PR #679, opened 2024-06-19
[PM-8863] Fido2 credential creation

fedemkr:
## 🎟️ Tracking

[PM-8863](https://bitwarden.atlassian.net/browse/PM-8863)

## 📔 Objective

Create Fido2 credentials using Autofill extension going through adding a customizable cipher.

_Note: Create Fido2 credential selecting a cipher from the list and creating a new pre-filled cipher automatically will be done in subsequent PRs. This also applies to connecting user verification with Fido2 credential creation._

## 📸 Screenshots

https://github.com/bitwarden/ios/assets/15682323/475f1ac1-72fc-47d9-b125-3ed158fe00fd

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-8863]: https://bitwarden.atlassian.net/browse/PM-8863?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenShared/Core/Autofill/Utilities/Fido2DebuggingUtils.swift
```diff
@@ -0,0 +1,68 @@
...
+        requestString.append("Options -> RK: \(request.options.rk) \n")
+        requestString.append("Options -> UV: ")
+        requestString.append(String(describing: request.options.uv))
+        requestString.append("\n")
+        let extensions = request.extensions?.description ?? "nil"
+        requestString.append("Extensions: \(extensions) \n")
+        return requestString
+    }
+
+    static func describe(result: MakeCredentialResult) -> String {
```
matt-livefront, line 54: ⛏️ Minor nitpick - Swift has a formal way of creating descriptions for a type, you can conform a type to `CustomDebugStringConvertible` and implement the `debugDescription` variable (or `CustomStringConvertible` and implement `description`). Would that be a better place for these?

BitwardenShared/Core/Autofill/Utilities/Fido2DebuggingUtilsTests.swift
```diff
@@ -0,0 +1,92 @@
...
+                name: "user"
+            ),
+            pubKeyCredParams: [PublicKeyCredentialParameters(ty: "public-key", alg: -7)],
+            excludeList: nil,
+            options: Options(rk: true, uv: .preferred),
+            extensions: nil
+        )
+
+        // swiftlint:disable:next line_length
+        let expectedResult = "ClientDataHash: 0101010101010101010101010101010101010101010101010101010101010101\nRP -> Id: someApp.com \nRP -> Name: nil \nUser -> Id: 0101010101010101010101010101010101010101010101010101010101010101\nUser -> Name: user \nUser -> DisplayName: userDisplay \nPubKeyCredParams: [BitwardenSdk.PublicKeyCredentialParameters(ty: \"public-key\", alg: -7)]\nExcludeList: nil \nOptions -> RK: true \nOptions -> UV: preferred\nExtensions: nil \n"
```
matt-livefront, line 41: 🎨 You could format this on multiple lines to avoid the long line:
```suggestion
        let expectedResult = """
            ClientDataHash: 0101010101010101010101010101010101010101010101010101010101010101
            RP -> Id: someApp.com
            RP -> Name: nil
            User -> Id: 0101010101010101010101010101010101010101010101010101010101010101
            User -> Name: user
            User -> DisplayName: userDisplay
            PubKeyCredParams: [BitwardenSdk.PublicKeyCredentialParameters(ty: "public-key", alg: -7)]
            ExcludeList: nil
            Options -> RK: true
            Options -> UV: preferred
            Extensions: nil
            """
```

BitwardenShared/UI/Vault/Vault/AutofillList/VaultAutofillListProcessor.swift
```diff
@@ -152,6 +163,76 @@ class VaultAutofillListProcessor: StateProcessor<
...
+    private func initFido2State() {
+        guard let fido2appExtensionDelegate = appExtensionDelegate as? Fido2AppExtensionDelegate else {
+            return
+        }
+
+        services.fido2UserInterfaceHelper.setupDelegate(fido2UserVerificationMediatorDelegate: self)
+
+        if let request = fido2appExtensionDelegate.getRequestForFido2Creation(),
+           let credentialIdentity = request.credentialIdentity as? ASPasskeyCredentialIdentity {
+            Task {
```
matt-livefront, line 176: 🎨 Since this function is called from an effect, I would make this function async rather than creating a task here.

You could also call this function from the `loadData` effect rather than create a separate effect.

BitwardenShared/UI/Vault/Vault/AutofillList/VaultAutofillListProcessor.swift
```diff
@@ -152,6 +163,76 @@ class VaultAutofillListProcessor: StateProcessor<
         }
     }
 
+    @available(iOSApplicationExtension 17.0, *)
+    private func initFido2State() {
+        guard let fido2appExtensionDelegate = appExtensionDelegate as? Fido2AppExtensionDelegate else {
+            return
+        }
+
+        services.fido2UserInterfaceHelper.setupDelegate(fido2UserVerificationMediatorDelegate: self)
```
matt-livefront, line 172: 🤔 Is the delegate that is passed in here only used for presenting alerts? Or does/will it have other uses? Instead of setting a delegate, I wonder if we could pass a closure that would be used for presenting alerts, such that you could pass `coordinator.showAlert`? I'm not sure how many methods that would change, but it was a thought I had.   Or maybe there's another way we could give it a way to present an alert (like have app processor set the app coordinator as the delegate).

BitwardenAutoFillExtension/Application/Support/Info.plist
```diff
@@ -85,8 +85,15 @@
...
-			<true/>
+			<key>ASCredentialProviderExtensionCapabilities</key>
+			<dict>
+				<key>ProvidesPasskeys</key>
+				<true/>
+				<key>ProvidesPasswords</key>
+				<true/>
+				<key>ShowsConfigurationUI</key>
+				<true/>
+			</dict>
```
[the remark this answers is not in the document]
fedemkr: Thanks for raising this, I'll double check but AFAIK we haven't had any reports on MAUI with this change.

BitwardenAutoFillExtension/CredentialProviderContext/ExtensionMode.swift
```diff
@@ -0,0 +1,30 @@
...
+
+    /// The extension is displaying a list of items in the vault that match a service identifier
+    /// and or passkey request parameters.
+    case autofillVaultList([ASCredentialServiceIdentifier], ASPasskeyCredentialRequestParameters)
+
+    /// The extension is being configured to set up autofill.
+    case configureAutofill
+
+    /// The extension is being configured to register a Fido2 credential.
+    case registerFido2Credential(ASPasskeyCredentialRequest)
```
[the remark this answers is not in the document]
fedemkr: This seems very reasonable and I like it way much better than my implementation and we can use the same approach for the iOS 18 changes. I'll give it a try, thanks!! 👏

BitwardenShared/UI/Vault/Vault/AutofillList/VaultAutofillListProcessor.swift
fedemkr: Oh I didn't want to create a `Task` without the need to do so. Like if this view is not called from a Fido2 flow then it would be unnecessary to async/await it. But yeah we save too little in comparison to keep up with the pattern followed elsewhere. 
I started putting this in `loadData` but was unsure if that was a good spot to wrap `initFido2State` in a `Task`, given that I don't want/need to await `initFido2State` to finish; just a fire and forget approach.

BitwardenShared/UI/Vault/Vault/AutofillList/VaultAutofillListProcessor.swift
fedemkr: It shows alerts and it's also used to perform `setupPin` navigation for user verification which is the last part of the flow if the other mechanisms aren't available on user verification preference `required`.

BitwardenShared/Core/Autofill/Utilities/ASAuthorizationPublicKeyCredentialUserVerificationPreference+Extensions.swift
```diff
@@ -0,0 +1,18 @@
+import AuthenticationServices
+import BitwardenSdk
+
+extension ASAuthorizationPublicKeyCredentialUserVerificationPreference {
+    /// Maps to the Bitwarden Sdk user verification preference object
+    func toSdkUserVerificationPreference() -> BitwardenSdk.Uv {
+        switch self {
+        case ASAuthorizationPublicKeyCredentialUserVerificationPreference.discouraged:
```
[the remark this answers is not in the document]
fedemkr: @matt-livefront I think it's not very straightforward here because `BitwardenSdk.Uv` is an enum and it doesn't have the `init(raw value:)` initializer. I'll do a static init if you agree. Would something like this work?
```swift
extension BitwardenSdk.Uv {
    static func from(preference: ASAuthorizationPublicKeyCredentialUserVerificationPreference) -> BitwardenSdk.Uv {
        switch preference {
        case ASAuthorizationPublicKeyCredentialUserVerificationPreference.discouraged:
            BitwardenSdk.Uv.discouraged
        case ASAuthorizationPublicKeyCredentialUserVerificationPreference.preferred:
            BitwardenSdk.Uv.preferred
        case ASAuthorizationPublicKeyCredentialUserVerificationPreference.required:
            BitwardenSdk.Uv.required
        default:
            BitwardenSdk.Uv.required
        }
    }
}
```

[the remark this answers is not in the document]
matt-livefront: You could still add an initializer to `BitwardenSdk.Uv`:
```
extension BitwardenSdk.Uv {
    init(preference: ASAuthorizationPublicKeyCredentialUserVerificationPreference) {
        switch preference {
        case .discouraged:
            self = .discouraged
        case .preferred:
            self = .preferred
        case .required:
            self = .required
        default:
            self = .required
        }
    }
}
```

BitwardenShared/UI/Vault/Vault/AutofillList/VaultAutofillListProcessor.swift
matt-livefront: `Task`s are lightweight so I wouldn't be too concerned about creating one unnecessarily. And since the task is created by the view when you do `.task { await store.perform(.initFido2) }`, you're already in a task, so by making `initFido2State` not be async, you end up creating a task to perform the effect and then another one for this work when it could be done all in a single task.

BitwardenAutoFillExtension/Application/Support/Info.plist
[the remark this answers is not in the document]
fedemkr: ⚠️  Tested on simulator iPhone SE (3rd) iOS 16.4 and you were right! We need to keep it or config doesn't appear on iOS < 17. Thanks!

BitwardenShared/Core/Autofill/Utilities/CredentialProviderContext/CredentialProviderContext.swift
```diff
@@ -0,0 +1,68 @@
...
+            return serviceIdentifiers
+        }
+
+        if case let .autofillFido2VaultList(serviceIdentifiers, _) = extensionMode {
+            return serviceIdentifiers
+        }
+        return []
+    }
+
+    /// Initilalizes the context
```
matt-livefront, line 63: ⛏️ 
```suggestion
    /// Initializes the context.
```

BitwardenShared/Core/Platform/Extensions/Array+Extensions.swift
```diff
@@ -0,0 +1,9 @@
+extension Array {
+    subscript(safeIndex index: Int) -> Element? {
```
matt-livefront, line 2: ⛏️ Could you add docs for this?

BitwardenShared/Core/Vault/Services/Fido2CredentialStoreService.swift
```diff
@@ -0,0 +1,72 @@
...
+               !ids.contains(fido2CredentialAutofillView.credentialId) {
+                continue
+            }
+
+            result.append(cipherView)
+        }
+        return result
+    }
+
+    func saveCredential(cred: BitwardenSdk.Cipher) async throws {
```
matt-livefront, line 56: ⛏️ Could you add docs for these?

BitwardenShared/Core/Vault/Services/Fido2CredentialStoreService.swift
```diff
@@ -0,0 +1,72 @@
...
+            try await cipherService.addCipherWithServer(cred)
+        } else {
+            try await cipherService.updateCipherWithServer(cred)
+        }
+    }
+}
+
+private extension Cipher {
+    /// Whether the cipher is active, is a login and has Fido2 credentials.
+    var isAciveWithFido2Credentials: Bool {
```
matt-livefront, line 67: ⛏️ 
```suggestion
    var isActiveWithFido2Credentials: Bool {
```

BitwardenShared/Core/Vault/Services/Fido2CredentialStoreServiceTests.swift
```diff
@@ -0,0 +1,309 @@
...
+        await assertAsyncThrows(error: BitwardenTestError.example) {
+            _ = try await subject.allCredentials()
+        }
+    }
+
+    /// `.findCredentials(ids:ripId:)` returns the login ciphers that are active, have Fido2 credentials
+    /// and match the `ripId` and the credential `ids` if any.
+    func test_findCredentials() async throws {
+        let expectedRpId = Fido2CredentialAutofillView.defaultRpId
+        let expectedCredentialId = Data((0 ..< 16).map { _ in 1 })
```
matt-livefront, line 98: 🎨 There's a `Data` initializer that should give you this same behavior. 
```suggestion
        let expectedCredentialId = Data(repeating: 1, count: 16)
```

BitwardenShared/Core/Vault/Services/Fido2CredentialStoreService.swift
matt-livefront, line 56: 🎨 I think instead of abbreviating "credential" here, you could make it an unnamed argument since it's repeating what's in the function name.
```suggestion
    func saveCredential(_ credential: BitwardenSdk.Cipher) async throws {
```

BitwardenShared/UI/Autofill/Application/Fido2AppExtensionDelegate.swift
```diff
@@ -0,0 +1,23 @@
+import AuthenticationServices
+
+/// A delegate that is used to handle actions and retrieve information from within an Autofill extension
+/// on Fido2 flows.
+public protocol Fido2AppExtensionDelegate: AppExtensionDelegate {
+    /// The mode in which the autofill extension is running.
+    var extensionMode: AutofillExtensionMode { get }
+
+    /// Completes the registration request iwth a Fido2 credential
```
matt-livefront, line 9: ⛏️ 
```suggestion
    /// Completes the registration request with a Fido2 credential
```

BitwardenShared/UI/Autofill/Utilities/Fido2UserInterfaceHelper.swift
```diff
@@ -0,0 +1,74 @@
...
+
+/// Default implemenation of `Fido2UserInterfaceHelper`.
+class DefaultFido2UserInterfaceHelper: Fido2UserInterfaceHelper {
+    /// Mediator which manages user verification on Fido2 flows.
+    private var fido2UserVerificationMediator: Fido2UserVerificationMediator
+
+    /// Continuation when picking a credential for creation.
+    var credentialForCreationContinuation: CheckedContinuation<CheckUserAndPickCredentialForCreationResult, Error>?
+
+    private(set) var fido2CredentialNewView: BitwardenSdk.Fido2CredentialNewView?
```
matt-livefront, line 27: ⛏️ Could you add docs for this?

fedemkr: The docs for this is already in the protocol, I don't see that being repeated in the implementation in other places. Or should I add it nonetheless because it's a property instead of a function?

matt-livefront: Ah, that's fine then we usually don't document protocol methods/variables.

matt-livefront (approved):
Looks great!

PR #682, opened 2024-06-20
BIT-2348: Add OTP key parsing for account name and issuer

matt-livefront:
## 🎟️ Tracking

[BIT-2348](https://livefront.atlassian.net/browse/BIT-2350)

## 📔 Objective

Adds support for parsing the account name and issuer from the OTP key.

This is needed as a precursor to searching for matching ciphers based on the account name or issuer from the key.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/BitwardenShared/Core/Vault/Services/TOTP/OTPAuthModel.swift
+++ b/BitwardenShared/Core/Vault/Services/TOTP/OTPAuthModel.swift
@@ -4,30 +4,58 @@ import Foundation
 ///
 /// This model includes the Base32-encoded key, the period, the number of digits, and the hashing algorithm.
 struct OTPAuthModel: Equatable {
+    // MARK: Properties
+
+    /// The name of the account for the key.
+    let accountName: String?
+
+    /// The hashing algorithm used for generating the OTP.
+    let algorithm: TOTPCryptoHashAlgorithm
+
+    /// The number of digits in the OTP.
+    let digits: Int
+
+    /// The provider or service the account is associated with.
+    let issuer: String?
+
     /// The Base32-encoded key used for generating the OTP.
     let keyB32: String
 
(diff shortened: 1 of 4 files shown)


BitwardenShared/Core/Vault/Services/TOTP/OTPAuthModel.swift
```diff
@@ -4,49 +4,101 @@ import Foundation
...
+                issuer = String(parts[0])
+                accountName = String(parts[1])
+            } else {
+                accountName = label
+            }
+        }
+
+        if issuer == nil {
+            issuer = queryItems.first { $0.name == "issuer" }?.value
+        }
```
KatherineInCode, line 88: 🤔 The OTP specs—particularly Google's—are maddeningly inconsistent and ambiguous about how to handle things like the issuer. It seemed to me like Google had regretted putting Issuer in the label, and therefore moved it to being another query parameter, but had to keep it around for backwards compatibility.

In [the Authenticator app](https://github.com/bitwarden/authenticator-ios/blob/main/AuthenticatorShared/Core/Vault/Services/TOTP/OTPAuthModel.swift#L101) I went the other way, treating the parameter as canonical and then stripping it out of the label if it was there to get the account name. Basically following this guidance from [Google's spec](https://github.com/google/google-authenticator/wiki/Key-Uri-Format):

> If the issuer parameter is absent, issuer information may be taken from the issuer prefix of the label.

I'm not sure if we want both apps to actually have the same implementation; in practice the two are I think functionally equivalent, it mostly comes down to how they handle incorrectly URIs that aren't to spec.

matt-livefront: Ah, that makes sense to me, I can swap those around.

PR #686, opened 2024-06-24
BIT-2414: Log organization events

KatherineInCode:
## 🎟️ Tracking

[BIT-2414](https://livefront.atlassian.net/browse/BIT-2414)

## 📔 Objective

This starts logging organizational events, particularly these:

- `clientViewed`
- `clientToggledPasswordVisible`
- `clientToggledHiddenFieldVisible`
- `clientToggledCardCodeVisible`
- `clientCopiedPassword`
- `clientCopiedHiddenField`
- `clientCopiedCardCode`
- `clientAutofilled`
- `clientToggledCardNumberVisible`

Since these were the ones logged in the MAUI app.

Event storage itself is in the `EventService`, which will be expanded in a Future PR to actually send the events to the server; this just stores them.

I was uncertain of where best to draw boundaries with wrapper functions, most notably because logging the events is `async throws`; I settled on a wrapper that does a default handling of the error (logging it) but not a wrapper that puts things in a `Task`. It seemed reasonable to me that we'd want to maintain the understanding that logging an event is async at the call site, but save the 4 lines of boilerplate to catch an error and log it. I'm open to discussion on this, however.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


BitwardenShared/UI/Vault/VaultItem/ViewItem/ViewItemProcessor.swift
```diff
@@ -443,6 +515,7 @@ private extension ViewItemProcessor {
                 }
                 newState.hasVerifiedMasterPassword = state.hasVerifiedMasterPassword
                 state = newState
+                try await services.eventService.collect(eventType: .cipherClientViewed, cipherId: itemId)
```
matt-livefront, line 518: 🤔 This is going to get called again any time the cipher is updated while on the view screen. Is that expected? Otherwise, maybe this could be done outside the for loop, unless we need to wait until the data is loaded.

BitwardenShared/UI/Vault/VaultItem/ViewItem/ViewItemProcessor.swift
```diff
@@ -199,10 +218,45 @@ private extension ViewItemProcessor {
...
+            Task {
+                await services.eventService.collect(
+                    eventType: .cipherClientCopiedCardCode,
+                    cipherId: cipherState.cipher.id,
+                    errorReporter: services.errorReporter
+                )
+            }
+        default:
+            break
+        }
```
matt-livefront, line 259: 🎨 Given that these calls are the same outside of the event type, maybe we could add a method that logs an event for the cipher in this processor? We could also add an extension to `CopyableField` that returns the event associated with copying that field?

BitwardenShared/UI/Vault/VaultItem/AddEditItem/AddEditItemProcessor.swift
```diff
@@ -311,8 +334,28 @@ final class AddEditItemProcessor: StateProcessor<// swiftlint:disable:this type_
...
             state.cardItemState.isCodeVisible = isVisible
+            if isVisible {
+                let cipherId = state.cipher.id
+                Task {
+                    await services.eventService.collect(
+                        eventType: .cipherClientToggledCardCodeVisible,
+                        cipherId: cipherId,
+                        errorReporter: services.errorReporter
+                    )
+                }
```
matt-livefront, line 345: 🎨 Similar to my other comment, I wonder if it makes sense to add a helper method in the processor that logs an event.

BitwardenShared/Core/Platform/Services/EventService.swift
```diff
@@ -0,0 +1,114 @@
...
+    func collect(
+        eventType: EventType,
+        cipherId: String? = nil,
+        uploadImmediately: Bool = false,
+        errorReporter: ErrorReporter
+    ) async {
+        do {
+            try await collect(eventType: eventType, cipherId: cipherId, uploadImmediately: uploadImmediately)
+        } catch {
+            errorReporter.log(error: error)
```
matt-livefront, line 39: > I was uncertain of where best to draw boundaries with wrapper functions, most notably because logging the events is async throws; I settled on a wrapper that does a default handling of the error (logging it) but not a wrapper that puts things in a Task. It seemed reasonable to me that we'd want to maintain the understanding that logging an event is async at the call site, but save the 4 lines of boilerplate to catch an error and log it. I'm open to discussion on this, however.

👍 I agree with your comments, and like the idea of just logging the error here instead of at every callsite.

❓ Is there a benefit of passing around the error reporter vs injecting it into `EventService`?

BitwardenShared/Core/Platform/Models/Domain/EventData.swift
```diff
@@ -0,0 +1,11 @@
+import Foundation
+
+// MARK: - EventData
+
+/// Domain model for tracked events.
+///
+public struct EventData: Codable, Equatable {
+    let type: EventType
+    let cipherId: String?
+    let date: Date
```
matt-livefront, line 10: 🎨 Can you add docs for these?

BitwardenShared/Core/Platform/Services/EventService.swift
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        let newEvent = EventData(
+            type: eventType,
+            cipherId: cipherId,
+            date: timeProvider.presentTime
+        )
+
+        // swiftlint:disable:next line_length
+        Logger.application.info("Event collected: \(eventType.rawValue) on cipher \(cipherId ?? "(none)", privacy: .public) at \(newEvent.date, privacy: .public))")
```
matt-livefront, line 111: 🎨 Using `String(describing: eventType)` vs `eventType.rawValue` will log the enum name, which was more helpful for me when trying to test this, but maybe there's a good reason for the raw value.

BitwardenShared/Core/Platform/Services/EventService.swift
```diff
@@ -0,0 +1,114 @@
...
+        try await stateService.setEvents([newEvent], userId: userId)
```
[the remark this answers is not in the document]
KatherineInCode: 🤦🏼‍♀️ It should append. That was just a miss on my part

BitwardenShared/UI/Vault/VaultItem/ViewItem/ViewItemProcessor.swift
KatherineInCode: It probably should only log if the data is loaded (since that's when the person sees it) but only doing so once would be preferred. I'll try a few things on this one.

BitwardenShared/Core/Platform/Services/EventService.swift
KatherineInCode: The main reason for injection is so that it works better with the protocol (rather than having one throwing and one non-throwing function that do the same thing). But if we're comfortable just always catching the error and logging it, I can simplify that a lot so as not to inject it on the call. I don't think we'd ever want to do anything different if there is an error collecting an event, now that I think about it. Definitely something that can be simplified.

BitwardenShared/Core/Platform/Services/EventService.swift
KatherineInCode: Mostly just habit; it'd be better to put the actual name in the logs, I agree.

BitwardenShared/UI/Vault/VaultItem/AddEditItem/AddEditItemProcessor.swift
KatherineInCode: With no longer having `errorReporter` at the call site, a helper method would basically just abstract out the `Task` aspect of it, and I'm not sure how much I like that 🤔

BitwardenShared/Core/Platform/Services/EventService.swift
```diff
@@ -0,0 +1,115 @@
...
+protocol EventService {
+    /// Save an event to disk for future upload. This does not propagate errors that might be
+    /// thrown in internal processing to avoid complexity at call sites, and instead just
+    /// logs the error.
+    ///
+    /// - Parameters:
+    ///   - eventType: The event to track.
+    ///   - cipherId: The ID of the relevant cipher for some events.
+    ///   - uploadImmediately: If `true` then immediately attempts an upload after saving.
+    func collect(eventType: EventType, cipherId: String?, uploadImmediately: Bool) async
```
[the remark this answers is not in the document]
KatherineInCode: That seems reasonable; I'll take it out (which should simplify some things a bit)

BitwardenShared/Core/Platform/Services/EventService.swift
```diff
@@ -0,0 +1,115 @@
...
+            let newEvent = EventData(
+                type: eventType,
+                cipherId: cipherId,
+                date: timeProvider.presentTime
+            )
+
+            events.append(newEvent)
+
+            // swiftlint:disable:next line_length
+            Logger.application.info("Event collected: \(String(describing: eventType)) on cipher \(cipherId ?? "(none)", privacy: .public) at \(newEvent.date, privacy: .public))")
```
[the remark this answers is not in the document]
KatherineInCode: In the past I've found it helpful in general for analytics-like events to be logged, though I could see putting it at the debug level in terms of release. I at the time thought it was analogous to something else in the codebase, but I can't find it now. I can just remove it.

BitwardenShared/Core/Platform/Services/EventServiceTests.swift
```diff
@@ -0,0 +1,137 @@
+import XCTest
+
+@testable import BitwardenShared
+
+class EventServiceTests: XCTestCase {
```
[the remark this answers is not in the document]
KatherineInCode: 🤦🏼‍♀️ I brought in that part and forgot to add tests. Good catch (pun intended)!

fedemkr (approved):
LGTM 🎉

KatherineInCode:
I was having a hard time getting Codecov to tell me what it thought was missing, but now it has that, so I'll see what I can do

fedemkr (approved):

PR #688, opened 2024-06-25
BIT-2366: Alert user to switch to existing account with matching email on login

matt-livefront:
## 🎟️ Tracking

[BIT-2366](https://livefront.atlassian.net/browse/BIT-2366)

## 📔 Objective

On login, if the user's entered email matches an existing account and the base URLs match, present an alert to switch to that account.

## 📸 Screenshots

https://github.com/bitwarden/ios/assets/126492398/8f506a9d-99ad-4c40-9027-a7ae21e179aa

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
+++ b/BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
@@ -473,13 +473,14 @@ extension DefaultAuthRepository: AuthRepository {
     }
 
     func existingAccountUserId(email: String) async -> String? {
-        guard let userId = await stateService.getUserId(email: email),
-              let baseUrl = try? await stateService.getEnvironmentUrls(userId: userId)?.base,
-              baseUrl == environmentService.baseURL
-        else {
-            return nil
+        let matchingUserIds = await stateService.getUserIds(email: email)
+        for userId in matchingUserIds {
+            guard let baseUrl = try? await stateService.getEnvironmentUrls(userId: userId)?.base,
+                  baseUrl == environmentService.baseURL
+            else { continue }
+            return userId
         }
-        return userId
+        return nil
     }
 
     func getAccount(for userId: String?) async throws -> Account {
--- a/BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
+++ b/BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
@@ -42,6 +42,14 @@ protocol AuthRepository: AnyObject {
     ///
     func deleteAccount(otp: String?, passwordText: String?) async throws
 
+    /// Returns the user ID of an existing account that is already logged in on the device matching
+    /// the specified email.
+    ///
+    /// - Parameter email: The email for the account to check.
+    /// - Returns: The user ID of the account that is already logged in on the device, or `nil` otherwise.
+    ///
+    func existingAccountUserId(email: String) async -> String?
+
     /// Gets the account for a user id.
     ///
     /// - Parameter userId: The user Id to be mapped to an account.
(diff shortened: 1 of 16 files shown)


BitwardenShared/Core/Auth/Repositories/AuthRepositoryTests.swift
```diff
@@ -247,6 +247,39 @@ class AuthRepositoryTests: BitwardenTestCase { // swiftlint:disable:this type_bo
...
+        stateService.userId = "1"
+
+        let userId = await subject.existingAccountUserId(email: "user@bitwarden.com")
+
+        XCTAssertEqual(userId, "1")
+    }
+
+    /// `existingAccountUserId(email:)` returns `nil` if there's an existing account with the same
+    /// email but the base URLs are different.
+    func test_existingAccountUserId_matchingAccountDifferentBaseUrl() async throws {
```
fedemkr, line 265: ⛏️ Could a test be added for when `await stateService.getEnvironmentUrls(userId: userId)` throws?

BitwardenShared/Core/Platform/Services/StateService.swift
```diff
@@ -1186,6 +1193,12 @@ actor DefaultStateService: StateService { // swiftlint:disable:this type_body_le
         try getAccount(userId: userId).profile.userDecryptionOptions?.hasMasterPassword ?? true
     }
 
+    func getUserId(email: String) async -> String? {
+        guard let state = appSettingsStore.state else { return nil }
+        let account = state.accounts.values.first { $0.profile.email == email }
```
fedemkr, line 1198: ⚠️ The user might have set up more than one account with the same email in different servers, so we should be checking all the accounts that match the email and not only the first one. Therefore I think this should be returning an array of possible user ids.

**Example**
Logged in Accounts:
- Account1:
  - Email: e1@example.com
  - UserId: 1
  - Base Url: US
- Account2:
  - Email: e2@example.com
  - UserId: 2
  - Base Url: US
- Account3:
  - Email: e1@example.com
  - UserId: 3
  - Base Url: EU
 
If we attempt to log in using `e1@example.com` in `EU` server then it should display the alert for the `Account3` but currently it won't happen because of only taking into consideration the first one.

fedemkr (changes requested):

BitwardenShared/Core/Platform/Services/StateService.swift
matt-livefront: Ah, good catch, I updated it to handle that scenario.

BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
```diff
@@ -464,6 +472,17 @@ extension DefaultAuthRepository: AuthRepository {
         await vaultTimeoutService.remove(userId: nil)
     }
 
+    func existingAccountUserId(email: String) async -> String? {
+        let matchingUserIds = await stateService.getUserIds(email: email)
+        for userId in matchingUserIds {
+            guard let baseUrl = try? await stateService.getEnvironmentUrls(userId: userId)?.base,
+                  baseUrl == environmentService.baseURL
+            else { continue }
+            return userId
```
fedemkr, line 481: ⛏️ I think that If we use `if` instead of `guard` we can remove the `else { continue }` and directly `return userId`. Or am I missing something that is not possible in the case?
```suggestion
            if let baseUrl = try? await stateService.getEnvironmentUrls(userId: userId)?.base,
                  baseUrl == environmentService.baseURL {
                return userId
            }
```

matt-livefront: Yeah, that works too.

fedemkr (approved):

PR #693, opened 2024-06-26
[PM-8965] Register for push notifications more aggressively

KatherineInCode:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-8965

## 📔 Objective

Previously, the app was only registering for push notifications on launch, which meant that it would only register for an active user at that time. This now registers more aggressively, by checking any time the user goes to the vault list. It will also register after a user has granted permission for push notifications. However, it will also only attempt to register once per day (defined as 86400 seconds), to match the MAUI app.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessor.swift
+++ b/BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessor.swift
@@ -256,9 +256,11 @@ extension VaultListProcessor {
         coordinator.showAlert(
             .pushNotificationsInformation { [services] in
                 do {
-                    _ = try await services.notificationService
+                    let authorized = try await services.notificationService
                         .requestAuthorization(options: [.alert, .sound, .badge])
-                    await self.registerForNotifications()
+                    if authorized {
+                        await self.registerForNotifications()
+                    }
                 } catch {
                     self.services.errorReporter.log(error: error)
                 }
--- a/BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessor.swift
+++ b/BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessor.swift
@@ -14,13 +14,15 @@ final class VaultListProcessor: StateProcessor<
 > {
     // MARK: Types
 
-    typealias Services = HasAuthRepository
+    typealias Services = HasApplication
+        & HasAuthRepository
         & HasAuthService
         & HasErrorReporter
         & HasNotificationService
         & HasPasteboardService
         & HasPolicyService
         & HasStateService
+        & HasTimeProvider
         & HasVaultRepository
 
     // MARK: Private Properties
@@ -61,7 +63,7 @@ final class VaultListProcessor: StateProcessor<
         switch effect {
         case .appeared:
             await refreshVault(isManualRefresh: false)
-            await requestNotificationPermissions()
+            await handleNotifications()
             await checkPendingLoginRequests()
             await checkPersonalOwnershipPolicy()
             if !isShowingNotificationPermissions {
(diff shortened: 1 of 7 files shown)


BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessor.swift
```diff
@@ -234,6 +258,7 @@ extension VaultListProcessor {
                 do {
                     _ = try await services.notificationService
                         .requestAuthorization(options: [.alert, .sound, .badge])
+                    await self.registerForNotifications()
```
fedemkr, line 261: 🤔 Shouldn't we check the result from `requestAuthorization` and only if it's `true` call `await self.registerForNotifications()`? [Docs](https://developer.apple.com/documentation/usernotifications/unusernotificationcenter/requestauthorization(options:completionhandler:))

BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessorTests.swift
```diff
@@ -129,6 +135,71 @@ class VaultListProcessorTests: BitwardenTestCase { // swiftlint:disable:this typ
...
+        XCTAssertFalse(application.registerForRemoteNotificationsCalled)
+        XCTAssertEqual(
+            stateService.notificationsLastRegistrationDates["1"],
+            timeProvider.presentTime.addingTimeInterval(-86399)
+        )
+    }
+
+    /// `perform(_:)` with `appeared` registers the device for notifications
+    /// if the device attempted registration more than one day (that is, 86400 seconds) ago.
+    func test_perform_appeared_notificationRegistration_moreThanADay() async throws {
```
fedemkr, line 180: 🎨 If I'm not mistaken this would be to exactly a day. Can another test be added for more than a day? i.e. 86401 so we have the three possible cases for `>=`.

BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessor.swift
KatherineInCode: Good call; I missed that

BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessorTests.swift
KatherineInCode: I was mostly trying to test the boundary conditions, but I can add another one explicitly for 86401

BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessor.swift
matt-livefront: You can register for notifications without yet having authorization (or if the user denied authorization). The benefit of that is if the user enables notifications in settings, they will start receiving notifications immediately without having to reopen the app.

BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessorTests.swift
fedemkr: Great thanks! Just in case my reasoning behind this is because on a given `x >= n` condition, if we don't add a test for `x > n` then all the tests will keep passing if the condition would change to `x == n`.

KatherineInCode: It makes sense, though the tests would still pass if the condition changed to `x == n || x == n+1`. Figuring out the optimal set of tests to cover as much as possible without creating so many tests that execution takes forever is a tricky thing.

fedemkr (approved):

PR #714, opened 2024-07-08
BIT-2362 - Maximum Access Count Field Cannot Have A Value Entered In Manually

phil-livefront:
## 🎟️ Tracking

[BIT-2362](https://livefront.atlassian.net/browse/BIT-2362)

## 📔 Objective

- Allow the user to edit of the maximum access count from a numerical keyboard instead of just using the stepper to increase/decrease the count.
- From figma: "the maximum access count is blank on default, then displays the count when edited. If the users sets it back to 0 after editing, it displays 0."
- Added a "Save" button to the numerical keyboard to control the focus of the TextField for keyboard dismissal.

## 📸 Screenshots

https://github.com/bitwarden/ios/assets/150719757/97e7eff4-19d8-4cdb-81e3-72e5e70378f8

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


BitwardenShared/UI/Tools/Send/SendItem/AddEditSendItem/AddEditSendItemView.swift
```diff
@@ -11,6 +11,12 @@ struct AddEditSendItemView: View { // swiftlint:disable:this type_body_length
     /// The `Store` for this view.
     @ObservedObject var store: Store<AddEditSendItemState, AddEditSendItemAction, AddEditSendItemEffect>
 
+    /// A state variable to track whether the TextField is focused
+    @FocusState private var isMaxAccessCountFocused: Bool
+
+    /// A state variable to store the text in the TextField.
+    @SwiftUI.State private var maximumAccessCountText: String = ""
```
fedemkr, line 18: 🤔 Can't `store. maximumAccessCountText` be used instead of duplicating the same here?

BitwardenShared/UI/Tools/Send/SendItem/AddEditSendItem/AddEditSendItemView.swift
```diff
@@ -489,6 +524,37 @@ struct AddEditSendItemView: View { // swiftlint:disable:this type_body_length
             .pickerStyle(SegmentedPickerStyle())
         }
     }
+
+    // MARK: Private Functions
```
fedemkr, line 528: 🤔 Can all or some of these functions be moved to the `AddEditSendItemProcessor` given some additional `AddEditSendItemAction`?

BitwardenShared/UI/Tools/Send/SendItem/AddEditSendItem/AddEditSendItemView.swift
phil-livefront: I thought so too but given the use case where a user can enter 0, we need the ability to show that "0" in the text field.  It needs to be blank by default tho so I was having trouble getting that exact logic to work without having two variables since the TextField itself can handle entering the text vs just showing whats currently in the state.  Do you have a better idea on how to accomplish that?

BitwardenShared/UI/Tools/Send/SendItem/AddEditSendItem/AddEditSendItemView.swift
phil-livefront: yeah, we can certainly do that.  Good idea!

BitwardenShared/UI/Tools/Send/SendItem/AddEditSendItem/AddEditSendItemView.swift
fedemkr: I guess you can use `store.binding` like in [here](https://github.com/bitwarden/ios/blob/main/BitwardenShared/UI/Auth/CreateAccount/CreateAccountView.swift#L78) and move all the logic to the processor, where you can use the binding property as a `string` and then have an `int` calculated property based on the string one.

phil-livefront: Yeah, I did try that initially but I couldn't figure out how to allow the user to enter in "0" and  also allow it be blank by default.  Since the count is 0 by default.  It seemed everything I tried that some part of the use case requirements would break.  I can keep messing with other ideas to get it to work but it seems like its one of those bizarre situations where not all of the logic can be abstracted from the View. 

```
the maximum access count is blank on default, then displays the count when edited. If the users sets it back to 0 after editing, it displays 0.
```

matt-livefront: If you make `maximumAccessCountText` in the state a variable rather than a computed property, you could default it to the empty string, but update it whenever `maximumAccessCount` changes. Then it would initially be empty, but once you change it to 0 or a number, it would be updated to reflect the count. Would that match the behavior you're looking for?

phil-livefront: yeah i can try that!  good call.

BitwardenShared/UI/Tools/Send/SendItem/AddEditSendItem/AddEditSendItemView.swift
```diff
@@ -53,9 +56,6 @@ struct AddEditSendItemView: View { // swiftlint:disable:this type_body_length
             title: store.state.mode.navigationTitle,
             titleDisplayMode: .inline
         )
-        .task {
-            await store.perform(.loadData)
```
phil-livefront, line 57: noticed this was being called twice.  I can add it back but wanted to call it out in case it was missed before. CC: @matt-livefront

matt-livefront (approved):
Looks good!

fedemkr (approved):

PR #735, opened 2024-07-15
PM-9094 -  TOTP not shown when viewing an item

phil-livefront:
## 🎟️ Tracking
[PM-9094](https://bitwarden.atlassian.net/browse/PM-9094)

## 📔 Objective

When creating a `TOTPKey` based on an authenticatorKey, if the key wasn't correctly set to `base32`, `otpAuthUri`, or `steamUri`, the `TOTPKey` was being set to `nil`. This was preventing the TOTP code from showing up on the `ViewItemDetailsView`. The specific bug I encountered was due to spaces within the key (ex. `"Test Key"`).

After speaking with Federico, we came to the conclusion that removing whitespaces will work best for this particular bug due to the prefixes needed for `steamUri` and `otpAuth`.  This is because if we filtered out any non base32 character then it would great the existing prefix logic.  

## 📸 Screenshots

https://github.com/user-attachments/assets/f60ff499-0d89-4707-8533-54d390ad9476

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-9094]: https://bitwarden.atlassian.net/browse/PM-9094?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenShared/Core/Vault/Services/TOTP/TOTPKey.swift
```diff
@@ -69,18 +75,20 @@ enum TOTPKey: Equatable, Sendable {
...
-            guard steamKey.uppercased().isBase32 else { return nil }
-            self = .steamUri(key: steamKey)
+            if steamKey.uppercased().isBase32 {
+                self = .steamUri(key: steamKey)
+            } else {
+                self = .unknown(key: key)
+            }
         } else {
-            return nil
+            self = .unknown(key: key)
```
matt-livefront, line 91: 🎨 Could you add a test that covers the unknown case with a key that has a space it in?

phil-livefront: I had added the below test but maybe i am misunderstanding what I missed here.  

```
func test_default_getTOTPConfiguration_unknown() throws {
        let keyWithSpaces = "key with spaces"
        let config = try DefaultTOTPService().getTOTPConfiguration(key: keyWithSpaces)
        XCTAssertEqual(config.totpKey, .unknown(key: keyWithSpaces))
    }
```

matt-livefront: Ah, I missed that. I was thinking it'd be good to add one to `TOTPCodeConfigTests` since that's testing the other cases.

BitwardenShared/UI/Vault/VaultItem/LoginTOTPState.swift
```diff
@@ -94,13 +87,7 @@ public enum LoginTOTPState: Equatable, Sendable {
...
         switch authKeyString {
         case let .some(string):
-            if let model = TOTPKeyModel(authenticatorKey: string) {
-                self = .key(model)
-            } else if !string.isEmpty {
-                self = .invalid(string)
-            } else {
-                self = .none
-            }
+            self = .key(TOTPKeyModel(authenticatorKey: string))
```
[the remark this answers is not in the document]
phil-livefront: yeah great catch!

BitwardenShared/Core/Vault/Services/TOTP/TOTPService.swift
```diff
@@ -18,10 +18,10 @@ struct DefaultTOTPService: TOTPService {
     /// - Throws: `TOTPServiceError.invalidKeyFormat` if the key format is invalid.
     /// - Returns: A `TOTPKeyModel` containing the configuration details.
     func getTOTPConfiguration(key: String?) throws -> TOTPKeyModel {
-        guard let key,
-              let config = TOTPKeyModel(authenticatorKey: key) else {
+        guard let key else {
             throw TOTPServiceError.invalidKeyFormat
```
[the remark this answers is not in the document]
phil-livefront: good catch!

BitwardenShared/Core/Vault/Services/TOTP/TOTPKey.swift
phil-livefront: np, i can do that!

KatherineInCode:
🤔 It's possible I'm missing something, but what's the value of having a separate `unknown` and `base32` case for the `TOTPKey`, if they're both effectively treated the same in terms of how they're used to generate codes?

phil-livefront:
> 🤔 It's possible I'm missing something, but what's the value of having separate `unknown` and `base32` cases for the `TOTPKey`, if they're both effectively treated the same in terms of how they're used to generate codes?

The bug for this was because if a key had a space inside of it, it failed the `base32` check which then set the code to `nil`.  I was speaking with @fedemkr who confirmed this matches to the logic used on the javascript client as well.  I didn't want to alter the logic too heavily so thats why I added in the `unknown` case.

KatherineInCode:
Consolidating the `unknown` and `base32` cases into one case that's just "a key" that allows spaces and non-base32 characters seems like a reasonable solution to me. I'm pretty sure the SDK just strips out non-base32 characters when generating the TOTP code for a key (at a time), though it'd be worth double-checking that.

BitwardenShared/Core/Vault/Services/TOTP/TOTPKey.swift
```diff
@@ -70,13 +70,14 @@ enum TOTPKey: Equatable, Sendable {
     ///
     /// - Parameter key: A string representing the TOTP key.
     init?(_ key: String) {
-        if key.uppercased().isBase32 {
-            self = .base32(key: key)
-        } else if key.hasOTPAuthPrefix,
-                  let otpAuthModel = OTPAuthModel(otpAuthKey: key) {
+        let trimmedKey = key.replacingOccurrences(of: " ", with: "")
```
KatherineInCode, line 73: 🤔  Is there a more robust way to delete whitespace characters? Would it make sense to also just trim out everything that's not valid base32? That might be a UX thing, to error to indicate the user typed an invalid character in case they mistyped, but it might also be nice to just handle things without erroring?

phil-livefront: @fedemkr and I were going back and forth on this.  If we trim out everything that isnt base32 than that messes up the symbol padding suffixes for steam and otp.  I can go with the original plan and combine base32/unknown cases but Federico thought it would be best to just remove whitespaces.  Happy to do whatever you guys think is best, just not trying to break anything :)

KatherineInCode: I think probably a bit more robust whitespace deletion would be sufficient, then. It would certainly improve things, and we can always circle back to consider how we handle the various sorts of TOTP keys we can get. In practice, I don't think people enter these very often, though, so it might not be worth the effort. Maybe something like `let trimmedKey = key.filter{ !$0.isWhitespace }`?

phil-livefront: okay makes sense to me!  i like it.

matt-livefront: If the SDK is removing any non-base32 characters, is it worth doing any filtering on our side?

KatherineInCode: At most we'd want to find Steam and `otp://` keys, but I tend to agree with that. It might be worth double-checking what the SDK does in that regard, but if we don't have to sanitize the user's input like that, I'd rather just leave it.

matt-livefront: Right, I agree. It looks like the SDK is: https://github.com/bitwarden/sdk/blob/27a71d899284da50e9a08ca87a0068f991dafe44/crates/bitwarden-vault/src/totp.rs#L218-L224

KatherineInCode: Then I think we should accept the user's input as-is and let the SDK handle it.

phil-livefront: i sync'd with David Severns on the android side of things and he confirmed they aren't doing any validation on the user input as well.

KatherineInCode (approved):

BitwardenShared/Core/Vault/Services/TOTP/OTPAuthModel.swift
```diff
@@ -19,7 +19,7 @@ public struct OTPAuthModel: Equatable, Hashable, Sendable {
     let issuer: String?
 
     /// The Base32-encoded key used for generating the OTP.
-    let keyB32: String
+    let key: String
```
matt-livefront, line 22: ⛏️ Can you remove `Base32` from the docs here?

matt-livefront (approved):
Looks good!

PR #739, opened 2024-07-16
BIT-2415: Send org events

KatherineInCode:
## 🎟️ Tracking

https://livefront.atlassian.net/browse/BIT-2415

## 📔 Objective

This adds a five-minute timer to the `AppProcessor` that regularly sends accumulated organization events (if there are any to send). As well, if the app is backgrounded, events are sent immediately and the timer is stopped, then restarted on foregrounding.

There were issues with testing `AppProcessor` particularly. As near as I can tell, because the `MockNotificationCenterService` uses a `CurrentValueSubject` for its mock, when the `Task` objects in `AppProcessor.init()` are created and subscribe, the current value (`Void`) is emitted, triggering the foreground and background blocks to run.

This means firstly that any tests that rely on e.g. `notificationCenterService.didEnterBackgroundSubject.send()` can have that line taken out and have the test still pass, because in this example `didEnterBackgroundSubject` is getting triggered on initialization. And secondly, that tests that rely on validating a series of events—namely, backgrounding then foregrounding the app—have nondeterministic results.

I attempted several solutions to this—adding a `dropFirst` to the subject, turning the subject into a `PassthroughSubject`, and various incarnations of `Task` manipulation in the tests—to no avail. At best I could get nondeterministic results.

As a result, I only have a test for the `init` currently, and not other scenarios. Given that this is an effective singleton object, I'm not especially concerned, but it would be nice to find a way to test the app background/foregrounding in the future.

The missing lines in the CodeCov report are, as near as I can tell, an issue with Xcode.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


BitwardenShared/Core/Platform/Services/API/EventAPIService.swift
```diff
@@ -1,5 +1,28 @@
...
 /// A protocol for an API service used to make event requests.
 ///
-protocol EventAPIService {}
+protocol EventAPIService {
+    /// Performs an API request to send an event to the backend.
+    ///
+    /// - Parameters:
+    ///   - body: The request model to send.
+    ///
+    func postEvents(_ events: [EventData]) async throws
```
matt-livefront, line 12: 🎨 
```suggestion
    ///   - events: The list of events to send.
    ///
    func postEvents(_ events: [EventData]) async throws
```

BitwardenShared/Core/Platform/Services/API/Requests/EventRequest.swift
```diff
@@ -0,0 +1,44 @@
...
+    var body: [EventRequestModel]? { requestBody.events }
+
+    /// The HTTP method of the request.
+    let method: HTTPMethod = .post
+
+    /// The URL path of this request.
+    let path = "/collect"
+
+    /// The actual body of the request.
+    let requestBody: EventRequestBody
```
matt-livefront, line 43: 🤔 I'm curious if the `EventRequestBody` is useful here, if we're sending the events as an array (as opposed to wrapped by the body struct). Could this just store the events as an array?

BitwardenShared/Core/Platform/Services/EventServiceTests.swift
```diff
@@ -6,7 +6,9 @@ class EventServiceTests: XCTestCase {
     // MARK: Properties
 
     var cipherService: MockCipherService!
+    var client: MockHTTPClient!
     var errorReporter: MockErrorReporter!
+    var eventAPIService: APIService!
```
matt-livefront, line 11: ⛏️ Could you `nil` these out in `tearDown()`?

BitwardenShared/UI/Platform/Application/AppProcessor.swift
```diff
@@ -254,6 +261,22 @@ public class AppProcessor {
...
+            Task {
+                await self.services.eventService.upload()
+            }
+        }
+        sendEventTimer?.tolerance = 10
+    }
+
+    /// Stops the timer for organization events
+    private func stopEventTimer() {
+        sendEventTimer?.fire()
```
fedemkr, line 277: 🤔 [In MAUI](https://github.com/bitwarden/mobile/blob/main/src/App/Platforms/iOS/AppDelegate.cs#L479) we're creating a [background task](https://developer.apple.com/documentation/uikit/uiapplication/1623031-beginbackgroundtask) when about to go to background so the events upload can have a bit more time before the app is suspended. Not sure if the scope changed here but wanted to raise this just in case.

BitwardenShared/UI/Platform/Application/AppProcessor.swift
```diff
@@ -254,6 +261,22 @@ public class AppProcessor {
             services.errorReporter.log(error: error)
         }
     }
+
+    /// Starts timer to send organization events regularly
+    private func startEventTimer() {
+        sendEventTimer = Timer.scheduledTimer(withTimeInterval: 5, repeats: true) { _ in
```
matt-livefront, line 267: ❓ This will fire every five seconds - should it be five minutes?

fedemkr, line 267: 🤔 [In MAUI](https://github.com/bitwarden/mobile/blob/main/src/App/Platforms/iOS/AppDelegate.cs#L467) this is configured to happen every 60seconds for iOS, 5min is for Android. Was this changed to unify the platforms?

BitwardenShared/UI/Platform/Application/AppProcessor.swift
fedemkr, line 1: 🤔 Maybe out of scope for this PR, but could this work in extensions? Like there is people who only use the Autofill extension on a daily basis, would this upload events be triggered in this scenario or would it need to wait until the main app is opened to sync events? I think it works but wanted to ask just in case.

BitwardenShared/UI/Platform/Application/AppProcessor.swift
matt-livefront, line 267: 🎨 We don't ever deallocate `AppProcessor` so this is pretty minor, but I think the timer should be referencing a weak self here to avoid a retain cycle since the processor keeps a reference to the timer.

fedemkr (commented):
Great work! About the two lines in the `EventRequest` missing tests by **Codecov** you can find an example [here](https://github.com/bitwarden/ios/blob/main/BitwardenShared/Core/Vault/Services/API/Cipher/Requests/AddCipherRequestTests.swift#L41-L48)

BitwardenShared/UI/Platform/Application/AppProcessor.swift
KatherineInCode: Ah, that makes sense! I'll make that change

BitwardenShared/UI/Platform/Application/AppProcessor.swift
KatherineInCode: The ticket I was going off of indicated that it was five minutes, so I was going off of that. The Slack discussions I've been able to dig up don't provide any more illumination for me.

BitwardenShared/UI/Platform/Application/AppProcessor.swift
KatherineInCode: I think? I'm not as familiar with what all the extension has access to or how its lifecycle would impact this.

BitwardenShared/Core/Platform/Services/API/EventAPIService.swift
KatherineInCode: Good catch!

BitwardenShared/Core/Platform/Services/API/Requests/EventRequest.swift
KatherineInCode: I admittedly went through several different variations of this while trying to get "sending an array to the server" to work, but I think there might have been some reason I couldn't? I'll double-check.

BitwardenShared/Core/Platform/Services/EventServiceTests.swift
KatherineInCode: Good catch!

BitwardenShared/UI/Platform/Application/AppProcessor.swift
KatherineInCode: 🤦🏼‍♀️ Yep

KatherineInCode: It's still good housekeeping to do that, yeah. Good catch!

fedemkr: Just asked in Slack to see what's the correct expected value.

KatherineInCode:
> Great work! About the two lines in the `EventRequest` missing tests by **Codecov** you can find an example [here](https://github.com/bitwarden/ios/blob/main/BitwardenShared/Core/Vault/Services/API/Cipher/Requests/AddCipherRequestTests.swift#L41-L48)

The thing is, I'm [doing that](https://github.com/bitwarden/ios/blob/83b12d1463c4f3033b2a3eacb9af456e479543e6/BitwardenShared/Core/Platform/Services/API/Requests/EventRequestTests.swift#L18) and it's not getting picked up.

BitwardenShared/UI/Platform/Application/AppProcessor.swift
```diff
@@ -254,6 +264,39 @@ public class AppProcessor {
...
+                if let backgroundTaskId = self?.backgroundTaskId {
+                    self?.services.application?.endBackgroundTask(backgroundTaskId)
+                    self?.backgroundTaskId = nil
+                }
+            }
+        )
+        sendEventTimer?.fire()
+        sendEventTimer?.invalidate()
+        if let taskId = backgroundTaskId {
+            services.application?.endBackgroundTask(taskId)
```
matt-livefront, line 296: 🤔 Since the upload API call is made asynchronously, is it possible/likely that the background task is ended before the upload completes? Would it be better to move this to after the request completes?

KatherineInCode: Ah, good point.

BitwardenShared/UI/Platform/Application/AppProcessor.swift
```diff
@@ -254,6 +264,43 @@ public class AppProcessor {
...
+        sendEventTimer?.tolerance = 10
+    }
+
+    /// Stops the timer for organization events
+    private func stopEventTimer() {
+        sendEventTimer?.fire()
+        sendEventTimer?.invalidate()
+    }
+
+    private func uploadEvents() async {
```
matt-livefront, line 284: ⛏️ Can you add docs for this?

matt-livefront (dismissed):
Looks great!

fedemkr (approved):
Looks good! The only caveat is whether the timing is correct. Approving so you don't need to keep merging main into it, we can always adjust the timing in a future PR.

PR #751, opened 2024-07-19
[PM-8360] Fido2 autofill with user interaction

fedemkr:
## 🎟️ Tracking

[PM-8360](https://bitwarden.atlassian.net/browse/PM-8360)

## 📔 Objective

Add flow for Fido2 autofill with user interaction having selected one credential from the OS bottom sheet.

## 📸 Screenshots

https://github.com/user-attachments/assets/158ba9e5-01ec-45ce-bde8-8c937f4ffe52

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-8360]: https://bitwarden.atlassian.net/browse/PM-8360?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenShared/Core/Autofill/Services/AutofillCredentialService.swift
```diff
@@ -225,6 +244,85 @@ extension DefaultAutofillCredentialService: AutofillCredentialService {
...
+    func provideFido2Credential( // swiftlint:disable:this function_body_length
+        for passkeyRequest: ASPasskeyCredentialRequest,
+        withUserInteraction: Bool,
+        fido2UserVerificationMediatorDelegate: Fido2UserVerificationMediatorDelegate
+    ) async throws -> ASPasskeyAssertionCredential {
+        guard let credentialIdentiy = passkeyRequest.credentialIdentity as? ASPasskeyCredentialIdentity else {
+            throw AppProcessorError.invalidOperation
+        }
+
+        let isLocked = try? await authRepository.isLocked()
```
matt-livefront, line 258: 🎨 Could we use `vaultTimeoutService.isLocked()` here instead of adding the auth repository? I think it's better if we don't have services depend on repositories.

BitwardenShared/Core/Autofill/Services/AutofillCredentialService.swift
```diff
@@ -225,6 +244,85 @@ extension DefaultAutofillCredentialService: AutofillCredentialService {
...
+        switch (vaultTimeout, isLocked) {
+        case (.never, true):
+            // If the user has enabled Never Lock, but the vault is locked,
+            // unlock the vault before continuing.
+            try await authRepository.unlockVaultWithNeverlockKey()
+        case (_, false):
+            break
+        default:
+            if !withUserInteraction {
+                throw Fido2Error.userInteractionRequired
```
matt-livefront, line 270: ❓ If you get to this point and the vault is locked, isn't this method ultimately going to fail whether or not user interaction is available? It seems like we could just throw an error here without needing the `withUserInteraction` flag?

BitwardenShared/UI/Platform/Application/AppProcessor.swift
```diff
@@ -328,77 +334,43 @@ extension AppProcessor: SyncServiceDelegate {
...
-        case (_, false):
-            break
-        default:
-            break
+extension AppProcessor: Fido2UserVerificationMediatorDelegate {
+    func onNeedsUserInteraction() async throws {
+        if let fido2AppExtensionDelegate = appExtensionDelegate as? Fido2AppExtensionDelegate,
+           !fido2AppExtensionDelegate.flowWithUserInteraction {
+            fido2AppExtensionDelegate.setUserInteractionRequired()
+            throw Fido2Error.userInteractionRequired
```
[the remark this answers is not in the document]
fedemkr: I thought about it but wasn't sure about mixing extension errors with errors in shared module. If you think it'd improve the code I'll do it no problem, don't have a strong argument in favor of it.

BitwardenShared/Core/Autofill/Services/AutofillCredentialService.swift
fedemkr: 🤔 That's a very good point. If I'm not mistaken there is no way this method gets called if the vault is locked with user interaction. So I think I could remove the `withUserInteraction` parameter. Nice catch!!!! 🎉

BitwardenShared/Core/Autofill/Services/AutofillCredentialService.swift
fedemkr: 🤔 But how would I replace `authRepository.unlockVaultWithNeverlockKey`?

BitwardenShared/UI/Autofill/Utilities/Fido2UserVerificationMediatorTests.swift
```diff
@@ -268,6 +286,39 @@ class Fido2UserVerificationMediatorTests: BitwardenTestCase {
                 credential: cipher
             )
         }
+        XCTAssertTrue(fido2UserVerificationMediatorDelegate.onNeedsUserInteractionCalled)
+    }
+
+    private func checkUser_throws_when_repromptNeedsUserInteraction(
```
KatherineInCode, line 292: 🎨 When there are internal functions that perform testing things and can fail a test on an assertion, I like taking advantage of the [file and line parameters of the XCTAssert varieties](https://developer.apple.com/documentation/xctest/1500669-xctassert), and in particular setting it up to use the file and line of the test that's calling the function, similar to what's discussed [here](https://www.iosdev.recipes/xctest/custom-assertions-file-and-line/) but frequently also with custom failure text. This means that when a test fails, the failure shows in Xcode in the test where it fails, so there's appropriate context to understand what's going on; this is particularly useful if there are several different parameters that cause a failure, because without it Xcode logs all of the errors inside of the helper function, which is not necessarily helpful for diagnosing what conditions led to the failure and why.

It does look like modern Xcode ameliorates this a bit, so it's probably a moot point, but I still think it would be a good pattern to consider adopting.

BitwardenShared/UI/Platform/Application/AppCoordinator.swift
```diff
@@ -389,4 +417,4 @@ extension AppCoordinator: VaultCoordinatorDelegate {
     func presentLoginRequest(_ loginRequest: LoginRequest) {
         showLoginRequest(loginRequest)
     }
-}
+} // swiftlint:disable:this file_length
```
KatherineInCode, line 420: ⛏️ Don't we disable at the tops of files and enable at the bottoms of files?

fedemkr: Not sure, I've seen this in a couple of places already, but I can change it if needed.

BitwardenShared/UI/Autofill/Utilities/Fido2UserVerificationMediatorTests.swift
fedemkr: Yeah, sounds good I'll add it 😄

BitwardenShared/UI/Autofill/Utilities/Fido2UserVerificationMediatorTests.swift
```diff
@@ -221,7 +237,9 @@ class Fido2UserVerificationMediatorTests: BitwardenTestCase {
     // MARK: Private
 
     private func checkUser_verified_when_reprompt_and_reprompt_verified(
-        _ userVerificationPreference: BitwardenSdk.Verification
+        _ userVerificationPreference: BitwardenSdk.Verification,
+        file: StaticString = #filePath,
```
KatherineInCode, line 241: 💡 TIL about `#filePath` and the changes happening to `#file`

KatherineInCode (approved):

BitwardenShared/UI/Platform/Application/AppProcessor.swift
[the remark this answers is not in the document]
matt-livefront: Ok, that's fair, I was thinking we wouldn't expect to use this outside of the extension, but I'm fine leaving it as-is too.

matt-livefront (approved):
Looks good!

PR #797, opened 2024-08-06
PM-10126 - Master Password or PIN Field Not Automatically Focused

phil-livefront:
## 🎟️ Tracking

[PM-10126](https://bitwarden.atlassian.net/browse/PM-10126)

## 📔 Objective

- Auto focus on the PIN or master password entry fields when unlocking the vault or logging in.

### Logging In 

https://github.com/user-attachments/assets/6ede3f6b-c2b7-4c3b-9975-64b61486de96

### Unlocking Vault
https://github.com/user-attachments/assets/4d7c88a9-7741-4369-964e-e6bcb6cd5318

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-10126]: https://bitwarden.atlassian.net/browse/PM-10126?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenShared/UI/Platform/Application/Views/BitwardenTextField.swift
```diff
@@ -140,11 +147,13 @@ struct BitwardenTextField<TrailingContent: View>: View {
         accessibilityIdentifier: String? = nil,
         passwordVisibilityAccessibilityId: String? = nil,
         canViewPassword: Bool = true,
+        isPasswordAutoFocused: Bool = true,
```
fedemkr, line 150: 🤔 By giving this `true` value as default doesn't it automatically focus the password field on the `AddEditLoginItemView`? Or what happens if the `AddEditCustomFieldsView` has several `hidden` custom fields?

phil-livefront: great call out, i actually shouldn't have done this.  I had it passing in the value to `true` but defaulting to false but at the last minute thought I could be clever.  Thanks for the catch!

BitwardenShared/UI/Auth/VaultUnlock/VaultUnlockState.swift
```diff
@@ -29,6 +29,9 @@ struct VaultUnlockState: Equatable {
     /// A flag indicating if the master password should be revealed or not.
     var isMasterPasswordRevealed = false
 
+    /// A flag indicating if the password/pin text field should be auto focused.
+    var isPasswordAutoFocused = true
+
```
fedemkr, line 34: ⛏️ Do we actually need this in the state? I mean unless I'm missing something it will always be `true`, so I guess we could just set the `true` value directly in the View. Maybe I'm breaking some pattern here, if so let me know. cc: @matt-livefront @KatherineInCode

phil-livefront: yeah, i went back and forth on this myself... I opted to follow the pattern but in this case it might not be worth it?  im good with whatever the team thinks is best here.

BitwardenShared/UI/Platform/Application/Views/BitwardenTextField.swift
```diff
@@ -32,6 +32,9 @@ struct BitwardenTextField<TrailingContent: View>: View {
     /// Whether a password in this text field is visible.
     let isPasswordVisible: Binding<Bool>?
 
+    /// If the keyboard should be presented immediately when the view appears.
+    let isPasswordAutoFocused: Bool
```
matt-livefront, line 36: 🤔 Is it possible to use the `focused` modifier to set focus on `BitwardenTextField` in the containing view without introducing changes to `BitwardenTextField`? I think we've done that elsewhere and that feels a little bit more SwiftUI-y.

BitwardenShared/UI/Auth/VaultUnlock/VaultUnlockState.swift
matt-livefront: If possible, I think we should use `@FocusState` for this. Ideally we would have focus in the state, but SwiftUI makes it difficult (`@FocusState` specifically). I think it can just be in the view if we don't have a reason to change it from the processor. But if we did need a static bool like this, I think keeping it in the view is ok.

fedemkr (approved):

BitwardenShared/UI/Auth/VaultUnlock/VaultUnlockView.swift
```diff
@@ -184,6 +186,7 @@ struct VaultUnlockView: View {
         }
     }
 
+
```
KatherineInCode, line 189: ⛏️ Warning

BitwardenShared/UI/Platform/Application/Views/BitwardenTextField.swift
phil-livefront: just for historical purposes, this wasn't working for the `pin` section because of how the views are laid out.  Switching back to the original implementation which worked for both.

matt-livefront (approved):

PR #804, opened 2024-08-07
PM-10639: Implement conditional logic for showing the intro carousel

matt-livefront:
## 🎟️ Tracking

[PM-10639](https://bitwarden.atlassian.net/browse/PM-10639)

## 📔 Objective

Adds the logic to determine when to show the intro carousel. If the feature is enabled, and there's no accounts on the device, the intro carousel should be shown on app launch. Once the user successfully creates an account or logs in, the carousel won't be shown on future app launches.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-10639]: https://bitwarden.atlassian.net/browse/PM-10639?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/UI/Auth/Extensions/AuthRouter+Redirects.swift
+++ b/BitwardenShared/UI/Auth/Extensions/AuthRouter+Redirects.swift
@@ -31,6 +31,12 @@ extension AuthRouter {
         if account.profile.forcePasswordResetReason != nil {
             return .updateMasterPassword
         } else {
+            let isCarouselEnabled: Bool = await services.configService.getFeatureFlag(.nativeCarouselFlow)
+            let introCarouselShown = await services.stateService.getIntroCarouselShown()
+            if isCarouselEnabled, !introCarouselShown {
+                await services.stateService.setIntroCarouselShown(true)
+            }
+
             return .complete
         }
     }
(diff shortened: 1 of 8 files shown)


BitwardenShared/UI/Auth/Extensions/AuthRouter+Redirects.swift
```diff
@@ -31,6 +31,12 @@ extension AuthRouter {
         if account.profile.forcePasswordResetReason != nil {
             return .updateMasterPassword
         } else {
+            let isCarouselEnabled: Bool = await services.configService.getFeatureFlag(.nativeCarouselFlow)
+            let introCarouselShown = await services.stateService.getIntroCarouselShown()
+            if isCarouselEnabled, !introCarouselShown {
+                await services.stateService.setIntroCarouselShown(true)
```
KatherineInCode, line 37: 🤔 If I'm understanding this correctly, we will also only show the carousel the first time the app is launched. So even if someone doesn't log in or create an account, they won't see it a second time. Just want to make sure this is intended behavior, as I feel the ticket is a bit ambiguous on this; and the PR summary indicates that it's only keyed to having accounts on the device, implying that it should pop up each time until a user successfully logs in or creates an account.

matt-livefront: >  we will also only show the carousel the first time the app is launched. So even if someone doesn't log in or create an account, they won't see it a second time

The `completeAuthRedirect()` method is only called once someone has successfully logged in or created an account right before the vault is shown. So if the app is launched and you see the carousel, but you don't log in or create an account, we wouldn't get here and then you would still see it on subsequent app launches until an account is added.

KatherineInCode: Makes sense, was just making sure!

KatherineInCode (approved):

matt-livefront:
I made one other change here to set the carousel shown flag on app startup if there's existing accounts, to prevent existing users from seeing the carousel. The logic in `completeAuthRedirect()` was meant to handle both existing accounts completing auth (vault unlock) and new logins / accounts. But for an existing user with never lock, that method is skipped, so the carousel shown flag wasn't being set.

fedemkr (approved):

KatherineInCode (approved):

PR #821, opened 2024-08-13
[PM-9362]  Check server's version before using Cipher Key Encryption

LRNcardozoWDF:
## 🎟️ Tracking

[PM-9362](https://bitwarden.atlassian.net/browse/PM-9362)

## 📔 Objective

Check server version and initialise flag `enableCipherKeyEncryption` accordingly. Additionally, improve how we update local config and load flags into the SDK.

### Notes

- Now the config is forced refresh every time a user logs in into the app
  - It's leveraging `AuthService -> getIdentityTokenResponse` that is called from all logins to centralize the call
- Added `preAuthServerConfig` to be able to have the server config saved if we don't have an active user id.
- On getting the config from server, if an error occurs while this is done, then if it's authenticated then it tries to update the local config with the `preAuthServerConfig`
- The `ClientService` now listens to new server config to update the corresponding SDK client for the proper user id.
- The `ConfigApiService` now calls the endpoint with auth data if the user is authenticated, otherwise it uses the unauthenticated service.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-9362]: https://bitwarden.atlassian.net/browse/PM-9362?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenShared/Core/Platform/Models/Domain/ServerConfig.swift
```diff
@@ -39,6 +39,28 @@ struct ServerConfig: Equatable, Codable {
...
+        for (serverV, minServerV) in zip(serverVersion, minServerVersion) {
+            if serverV < minServerV {
+                return false
+            } else if serverV > minServerV {
+                return true
+            }
+        }
+
+        return serverVersion.count >= minServerVersion.count
+    }
```
fedemkr, line 63: 🎨 Even though this is mostly good, I'd make some improvements:
- I'd create a `Version` struct that would take care of passing a `String`, cleaning it and splitting the components accordingly to use (defaulting the ones that don't have value because of wrong formatting or throwing an error depending on how you'd like to tackle it).
- In such struct I'd add the separators (`-` and `.`) as constants so they're not hardcoded on each use.
- I'd conform to `Comparable` and do the comparison logic there.
- So then you could simply do `Version("2024.3.1") > Version("2024.2.3")`, therefore it's easier to read and use.

🤔 Aside from the above, not sure why you're doing `return serverVersion.count >= minServerVersion.count`, here you're checking if both versions have the same number of components whereas IMO you should check that before zipping the versions and if they don't have the same number of components then you could fill the gaps with `-1` or just fail depending on how you'd like to tackle it. Therefore, at the end if it goes out of the `for` it means both versions are equal and you can return that they are.

BitwardenShared/Core/Platform/Services/ConfigService.swift
```diff
@@ -138,6 +145,17 @@ class DefaultConfigService: ConfigService {
         return localConfig
```
fedemkr, line 145: ⚠️ We need to also call loading flags here because as soon as the app starts if it's not forcing refresh, the `localConfig` is not `nil` and is not expired then it's not going to perform what's inside the `if` thus it won't load the flags which would be wrong. So we always need to update the flags with the latest version of the config, if there's one:
```suggestion
        if let localConfig {
            await loadFlags(localConfig)
        }
        return localConfig
```

fedemkr (changes requested):

BitwardenShared/Core/Platform/Models/Domain/ServerConfig.swift
KatherineInCode: 🎨 I would potentially call it `ServerVersion` and make sure it's also `Codable` but otherwise I agree with Fede that turning a version into its own object (that's a lot easier to write tests around vis-à-vis parsing server versions and handling errors) would be ideal. The `serverVersion` variable should then be of that type, which might require a migration and/or you expose a `Version` object that's a computed property off of the `serverVersion`.

Then at the calling site you can just say `config.serverVersion > Version("2024.2.3")` or whatever and it's taken care of; it'll also make writing tests in other places easier, I feel.

BitwardenShared/Core/Platform/Services/ConfigService.swift
```diff
@@ -138,6 +145,17 @@ class DefaultConfigService: ConfigService {
...
     }
 
+    func loadFlags(_ config: ServerConfig) async {
+        do {
+            try await clientService.platform().loadFlags(
+                [FeatureFlagsConstants.enableCipherKeyEncryption:
+                    config.isServerVersionAfter()]
```
KatherineInCode, line 152: 🎨 The typical format for arrays like this would be something like:
```
loadFlags([
    itemone,
    itemtwo,
])
```

BitwardenShared/Core/Platform/Services/ConfigService.swift
```diff
@@ -138,6 +145,17 @@ class DefaultConfigService: ConfigService {
...
+            )
+        } catch {
+            errorReporter.log(error: error)
```
KatherineInCode, line 155: ⛏️ Is there a way we could add a test for this?

BitwardenShared/Core/Platform/Services/ConfigServiceTests.swift
```diff
@@ -12,6 +12,7 @@ final class ConfigServiceTests: BitwardenTestCase {
     var stateService: MockStateService!
     var subject: DefaultConfigService!
     var timeProvider: MockTimeProvider!
+    var clientService: MockClientService!
```
KatherineInCode, line 15: ⛏️ Properties should be alphabetized

BitwardenShared/Core/Platform/Services/ConfigServiceTests.swift
```diff
@@ -20,6 +21,7 @@ final class ConfigServiceTests: BitwardenTestCase {
 
         client = MockHTTPClient()
         configApiService = APIService(client: client)
+        clientService = MockClientService()
```
KatherineInCode, line 24: ⛏️ Alphabetize

BitwardenShared/Core/Platform/Utilities/Constants.swift
```diff
@@ -79,6 +79,9 @@ enum Constants {
 
     /// The default file name when the file name cannot be determined.
     static let unknownFileName = "unknown_file_name"
+
+    /// The minimum server version required to have cipher key encryption on.
+    static let CipherKeyEncryptionMinServerVersion = "2024.2.0"
```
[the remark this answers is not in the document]
KatherineInCode: ⛏️ It should also be alphabetized

KatherineInCode (changes requested):

BitwardenShared/Core/Platform/Services/ConfigService.swift
```diff
@@ -166,6 +184,7 @@ class DefaultConfigService: ConfigService {
                 responseModel: configResponse
             )
             try? await stateService.setServerConfig(serverConfig)
+            await loadFlags(serverConfig)
```
LRNcardozoWDF, line 187: @fedemkr Do you think your [comment](https://github.com/bitwarden/ios/pull/821#discussion_r1715199771) is taken care of in this case?

fedemkr (commented):
Nice job, just some things I noticed that could be improved.

BitwardenShared/Core/Platform/Services/ConfigService.swift
```diff
@@ -157,6 +164,22 @@ class DefaultConfigService: ConfigService {
...
+    /// - Parameter config: Config to update the flags.
+    private func loadFlags(_ config: ServerConfig?) async {
+        do {
+            guard let config else {
+                return
+            }
+
+            try await clientService.platform().loadFlags([
+                FeatureFlagsConstants.enableCipherKeyEncryption: config.supportsCipherKeyEncryption(),
+            ])
```
matt-livefront, line 177: ⚠️ One thing I'm noticing is that this doesn't seem to get called when a new user signs into the app. I think the config is fetched for that user's environment prior to login, but since we don't have a user ID for the account yet to save that config, it's not saved to user defaults nor is it loaded here.

⚠️ And a related issue, since we rely on the active user ID when saving the config, I think there's potential for us to be saving a config to the wrong user account. For example, if you have an existing account signed in (userId=1) and you go to add an account in a different environment (userId=2), we load the config for that environment and then save it for the active user id (userId=1) since the new user isn't the active user yet.

@fedemkr Do you know offhand how MAUI handles this?

Neither of these were introduced here, so this work could be done separately but I'm not sure how it impacts this.

One potential idea: `ConfigService` could load the config for an environment and save it as a pre-auth config, similar to pre-auth URLs. And then when the user logs in, we save this pre-auth config to that user's config. We'd just need a way to indicate whether we're fetching the config for the active user or a potential new account.

fedemkr: We might have a similar issue on MAUI. The main difference is that on MAUI we're not saving a config per user but we maintain one globally which is fetched again when there's a change in user. So I think the native app does a much better job in this matter.
Yes, I like your idea of having a `pre-auth` config or something like that but I've just thought of another solution.

💡 @matt-livefront  What do you think about using the server url as key instead of a user id? We can use a hash of the server url if needed. With this we avoid having to worry about `pre-auth` or things like that. Additionally, we can reuse the same config amongst users of the same server.

matt-livefront: > 💡 @matt-livefront What do you think about using the api server url as key instead of a user id? We can use a hash of the server url if needed. With this we avoid having to worry about pre-auth or things like that. Additionally, we can reuse the same config amongst users of the same server.

Great idea, I like this approach a lot.

> As to this not being called when the user logs in, I'll make some changes for that to happen. Maybe call getting the config every time the active user changes 🤔

This makes sense to me!

BitwardenShared/Core/Platform/Services/ConfigService.swift
```diff
@@ -6,16 +7,21 @@ import OSLog
 /// A protocol for a `ConfigService` that manages the app's config.
 ///
 protocol ConfigService {
+    /// A publisher that publishes a new value when a new server config has been gotten.
```
andrebispo5, line 10: 🎨 "A publisher that updates with a new value when a new server configuration is received."
What do you think?

BitwardenShared/Core/Auth/Services/AuthServiceTests.swift
```diff
@@ -43,6 +45,7 @@ class AuthServiceTests: BitwardenTestCase { // swiftlint:disable:this type_body_
             appIdService: AppIdService(appSettingStore: appSettingsStore),
             authAPIService: authAPIService,
             clientService: clientService,
+            configService: configService,
```
KatherineInCode, line 48: 🎨 It feels strange to me to introduce the config service to the auth service, but have no tests around calling it.

BitwardenShared/Core/Platform/Services/ConfigService.swift
```diff
@@ -6,16 +7,21 @@ import OSLog
...
+    func configPublisher() async throws -> AsyncThrowingPublisher<AnyPublisher<MetaServerConfig?, Never>>
+
     /// Retrieves the current configuration. This will return the on-disk configuration if available,
     /// or will retrieve it from the server if not. It will also retrieve the configuration from
     /// the server if it is outdated or if the `forceRefresh` argument is `true`. Configurations
     /// retrieved from the server are saved to disk.
     ///
     /// - Parameters:
     ///   - forceRefresh: If true, forces refreshing the configuration from the server.
+    ///   - isPreAuth: If true, the call is coming before the user is auth or when adding a new account.
```
KatherineInCode, line 20: 🎨 "before the user is authenticated"?

BitwardenShared/Core/Platform/Services/ConfigService.swift
```diff
@@ -6,16 +7,21 @@ import OSLog
...
     /// - Returns: A server configuration if able.
     ///
-    func getConfig(forceRefresh: Bool) async -> ServerConfig?
+    @discardableResult
```
KatherineInCode, line 23: 👍🏻

BitwardenShared/Core/Platform/Utilities/ServerVersion.swift
```diff
@@ -0,0 +1,92 @@
...
+import OSLog
+
+// MARK: - ServerVersion
+
+/// Model that represents the server version.
+///
+struct ServerVersion: Comparable, Codable {
+    // MARK: - Constants
+
+    /// The separator between version and a suffix..
```
KatherineInCode, line 11: ⛏️ 
```suggestion
    /// The separator between version and a suffix.
```

BitwardenShared/Core/Platform/Utilities/ServerVersion.swift
```diff
@@ -0,0 +1,92 @@
...
+    private static let suffixSeparator: Character = "-"
+
+    /// The separator between version's components.
```
KatherineInCode, line 14: ⛏️ 
```suggestion
    /// The separator between a version's components.
```

BitwardenShared/Core/Platform/Utilities/ServerVersion.swift
```diff
@@ -0,0 +1,92 @@
...
+
+    // MARK: Parameters
+
+    /// A string representing the version
+    var version: String
+
+    // MARK: Init
+
+    /// Initializes a `SeverVersion` with the `version` if it's valid, otherwise returns `nil`.
+    /// - Parameter version: The string`version` to use.
```
KatherineInCode, line 25: 🎨 
```suggestion
    /// - Parameter version: The version string to parse.
```

BitwardenShared/Core/Platform/Utilities/ServerVersion.swift
```diff
@@ -0,0 +1,92 @@
...
+
+    /// Checks whether the `version` has the correct format.
+    /// - Parameter version: Version value to check.
+    /// - Returns: `true` if valid, `false` otherwise.
+    private static func isValid(version: String) -> Bool {
+        // Regular expression pattern for a version string: X.Y.Z-metadata
+        // where X, Y, Z are non-negative integers without leading zeros
+        // and X cannot be only zeros
+        // and an optional hyphen followed by alphanumeric metadata
+        let regexPattern = "^[1-9][0-9]{3}\\.[0-9]+\\.[0-9]+(-[a-zA-Z0-9]+)*$"
```
KatherineInCode, line 72: 🤔 This assumes that our major version number will always be a four-digit year, which is probably reasonable, but we could make the regex just `[1-9][0-9]*` in order to be slightly more flexible.

BitwardenShared/Core/Platform/Services/ConfigService.swift
```diff
@@ -112,62 +130,136 @@ class DefaultConfigService: ConfigService {
...
+
+            try? await setStateServerConfig(preAuthConfig, isPreAuth: false, userId: userId)
         }
     }
 }
+
+/// Helper object to send updated server config object with extra metadata
+/// like whether it comes from pre-auth and the user ID it belongs to.
+/// This is useful for getting the config on background and establishing which was the original context.
+struct MetaServerConfig {
```
KatherineInCode, line 258: 🤔 I'm not sure I love this name particularly. Maybe something like `ServerConfigInfo`? Not sure I like that either. It's probably fine, but I wouldn't mind finding a better name.

BitwardenShared/Core/Auth/Services/AuthServiceTests.swift
fedemkr: I still have several tests to add to the PR, that's why is not getting called.

BitwardenShared/Core/Platform/Services/ConfigService.swift
fedemkr: Yeah I struggled to find a name for it as well. Not a fan of `ServerConfigInfo`, what about `ServerConfigMessage`? As it gets inline with publishing a message and subscribing to a message sort of speak.

BitwardenShared/Core/Platform/Utilities/ServerVersion.swift
fedemkr: Yeah, makes sense 👍

BitwardenShared/Core/Platform/Services/ConfigService.swift
fedemkr: Just for the sake of the upcoming release, can we leave `MetaServerConfig`? We can create a tech-debt task to change the name if you wish so we don't block the PR for this.

KatherineInCode (approved):

matt-livefront (approved):

PR #825, opened 2024-08-13
PM-10836: Make PolicyService thread safe

shannon-livefront:
## 🎟️ Tracking

[PM-10836](https://bitwarden.atlassian.net/browse/PM-10836)

## 📔 Objective

Fix a crash that was reported through crashlytics. No reproduction steps that I know of.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-10836]: https://bitwarden.atlassian.net/browse/PM-10836?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/Core/Vault/Services/PolicyService.swift
+++ b/BitwardenShared/Core/Vault/Services/PolicyService.swift
@@ -51,7 +51,7 @@ protocol PolicyService: AnyObject {
 /// A default implementation of a `PolicyService` which manages syncing and updates to the user's
 /// policies.
 ///
-class DefaultPolicyService: PolicyService {
+actor DefaultPolicyService: PolicyService {
     // MARK: Properties
 
     /// The data store for managing the persisted policies for the user.
--- a/BitwardenShared/Core/Vault/Services/PolicyServiceTests.swift
+++ b/BitwardenShared/Core/Vault/Services/PolicyServiceTests.swift
@@ -391,6 +391,20 @@ class PolicyServiceTests: BitwardenTestCase { // swiftlint:disable:this type_bod
         XCTAssertFalse(onlyOrgApplies)
     }
 
+    /// `policyAppliesToUser()` called concurrently only makes a single request.
+    func test_policyAppliesToUser_calledConcurrently() async {
+        stateService.activeAccount = .fixture()
+        organizationService.fetchAllOrganizationsResult = .success([.fixture()])
+        policyDataStore.fetchPoliciesResult = .success([.fixture(type: .twoFactorAuthentication)])
+
+        async let concurrentTask1 = subject.policyAppliesToUser(.twoFactorAuthentication)
+        async let concurrentTask2 = subject.policyAppliesToUser(.twoFactorAuthentication)
+
+        _ = await (concurrentTask1, concurrentTask2)
+
+        XCTAssertEqual(policyDataStore.fetchPoliciesCount, 1)
+    }
+
     /// `policyAppliesToUser(_:)` returns whether the policy applies to the user when one
     /// organization has the policy enabled but not another.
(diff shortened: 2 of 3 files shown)


BitwardenShared/Core/Vault/Services/PolicyService.swift
```diff
@@ -51,7 +51,7 @@ protocol PolicyService: AnyObject {
 /// A default implementation of a `PolicyService` which manages syncing and updates to the user's
 /// policies.
 ///
-class DefaultPolicyService: PolicyService {
+actor DefaultPolicyService: PolicyService {
```
fedemkr, line 54: 🎨 Could a test be added accessing the dictionary from different threads concurrently proving that is now thread safe?

shannon-livefront: That's a great idea! Do you know if there's already an example of that somewhere in the app?

fedemkr: Oh sorry, I missed this earlier. See the tests now I think they're correct but were you able to confirm that they fail if you leave this as a class instead of an actor?

shannon-livefront: Yes, mostly. It was always a sporadic issue, so if I set this as a class instead of an actor, I have to run the test repeatedly before it fails

KatherineInCode: ❓ Are we going to have to change more classes into actors for managing thread safety with the Swift 6 concurrency model? Is a bunch of actors an idea solution?

shannon-livefront: That's probably a good idea, unless anyone knows of any drawbacks to using actors? I'm not entirely sure where to look in the app for which classes would need to be changed, though. Should we just go through and update every service?

KatherineInCode: My understanding is that actors will execute on non-main threads (aside from `@MainActor`-marked things) which could potentially produce unexpected behavior? There also seems to be [performance implications](https://www.hackingwithswift.com/quick-start/concurrency/what-is-actor-hopping-and-how-can-it-cause-problems) by having too many actors calling each other, but I'm not sure we're doing enough calls to hit that in a noticeable manner.

It might make more sense to [mark the particular things as isolated](https://www.massicotte.org/intro-to-isolation) rather than making the whole object into an actor? But there's a lot about the Swift 6 concurrency model that I still don't fully grok, so I'm not completely sure of the implications one way or another.

matt-livefront: Services should only need to be actors if they are managing shared state (like the policies dictionary here). Most services don't do this though; they mostly interact with other services rather than managing state of their own. StateService is already an actor and the database is accessed on a background context.

@KatherineInCode Those are good points; it's hard to say if we would run into performance implications without profiling, but my gut feeling as well is that we don't do enough of actor hopping that it would make a noticeable difference. I'm still figuring out isolation as well, but I guess I'm not following how that would help us here. Actor is going to isolate all methods/properties here; but most(all?) methods here end up relying on the policies dictionary so I'm not sure if we could make any methods be non-isolated.

KatherineInCode: That all makes sense, and I think we're good with just making this an actor for now. It's something for us to keep in mind over time, though, especially if we continue to run into these sorts of thread errors.

👍🏻

KatherineInCode (approved):

matt-livefront (approved):

PR #846, opened 2024-08-20
PM-10946: Handle key connector unlock for new user

matt-livefront:
## 🎟️ Tracking

[PM-10946](https://bitwarden.atlassian.net/browse/PM-10946)

## 📔 Objective

This handles vault unlock for a new user who needs to be converted to use Key Connector. If the user uses Key Connector, but doesn't yet have a user key, we'll generate a key, upload it to key connector and the API and then use that to unlock the user's vault.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


BitwardenShared/Core/Auth/Services/AuthService.swift
```diff
@@ -718,13 +718,13 @@ class DefaultAuthService: AuthService { // swiftlint:disable:this type_body_leng
     /// - Parameter response: The response received from the identity token request.
     /// - Returns: Whether the vault can be unlocked with a Key Connector key.
     ///
-    private func canUnlockWithKeyConnectorKey(_ response: IdentityTokenResponseModel) async throws -> Bool {
+    private func canUnlockWithKeyConnectorKey(_ response: IdentityTokenResponseModel) async throws -> (Bool, URL?) {
```
fedemkr, line 721: ⛏️ Update docs on what is returned.
🤔 Would it be simpler to just return the `URL?` and change a bit the name of the method? I mean `nil` would mean can't unlock with key connector and `not nil` would mean can unlock with key connector. However it may make it harder to follow on the code though, what do you think?

BitwardenShared/Core/Platform/Services/KeyConnectorService.swift
```diff
@@ -39,27 +49,53 @@ class DefaultKeyConnectorService {
...
+
         let account = try await stateService.getActiveAccount()
-        let keyConnectorUrlString = account.profile.userDecryptionOptions?.keyConnectorOption?.keyConnectorUrl
-        guard let keyConnectorUrlString,
-              let keyConnectorUrl = URL(string: keyConnectorUrlString) else {
-            throw KeyConnectorServiceError.missingKeyConnectorUrl
-        }
+        try await accountAPIService.setKeyConnectorKey(
+            SetKeyConnectorKeyRequestModel(
+                kdfConfig: account.kdf,
```
fedemkr, line 83: 🤔 In MAUI the KDF config is set from the `IdentityTokenResponse` not from the current `Account`. Not sure if it makes a difference though as it should be the same but wanted to raise it up just in case.

BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
```diff
@@ -748,12 +752,22 @@ extension DefaultAuthRepository: AuthRepository {
...
-        let encryptionKeys = try await stateService.getAccountEncryptionKeys(userId: account.profile.userId)
-        guard let encryptedUserKey = encryptionKeys.encryptedUserKey else { throw StateServiceError.noEncUserKey }
+        var encryptionKeys = try? await stateService.getAccountEncryptionKeys(userId: account.profile.userId)
+        if encryptionKeys?.encryptedUserKey == nil {
+            try await keyConnectorService.convertNewUserToKeyConnector(
+                keyConnectorUrl: keyConnectorURL,
+                orgIdentifier: orgIdentifier
+            )
+            encryptionKeys = try await stateService.getAccountEncryptionKeys(userId: account.profile.userId)
+        }
```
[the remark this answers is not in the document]
matt-livefront: The intention is that if the user doesn't have encryption keys, they are a new user who needs to be converted to use key connector. I'll change the logic to check for a specific "missing keys" error though, as opposed to any error. Think that makes sense?

BitwardenShared/Core/Auth/Services/AuthService.swift
matt-livefront: Yeah, I did like the name of that method, which is why I ended up the tuple. But it is simpler to return an optional URL. I ended up changing it to `func keyConnectorUrlForUnlock(_:) -> URL?`, which I think still reads ok.

BitwardenShared/Core/Platform/Services/KeyConnectorService.swift
matt-livefront: Good callout, in this case we create the `Account` from the `IdentityTokenResponse` right prior to calling this method, so the KDF in the account will be the same as in the response. That seems ok to me, but if you think we need to pass around the KDF from the response, I could change that.

fedemkr: Thanks for checking it, it's perfect as it is now.

fedemkr (approved):
LGTM 🎉

PR #901, opened 2024-09-05
PM-11697: Fix potential crash when creating a new account

matt-livefront:
## 🎟️ Tracking

[PM-11697](https://bitwarden.atlassian.net/browse/PM-11697)

## 📔 Objective

When creating a new account on a device that has at least one existing account, it's possible for the app to crash when checking the strength of the new account's password.

The root cause is that `DefaultClientService` maintains a cache of the SDK client as a dictionary, mapping a user ID to their SDK. It's possible for this dictionary to be modified from multiple threads. Making the service an actor fixes the concurrency issue.

I also noticed another issue here. When adding a new account, a user ID doesn't exist for the account yet, so we have some logic that returns a new SDK for these calls, and this SDK doesn't end up being cached. If there's an existing account on the device when adding a new account, that existing account is still considered the active user, so we end up using the existing account's SDK for some operations for the new account. To fix this, I added a `isPreAuth` property that can be used when getting a SDK client so we always return a new SDK instead of potentially using an SDK for an existing account.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-11697]: https://bitwarden.atlassian.net/browse/PM-11697?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
+++ b/BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
@@ -590,7 +590,8 @@ extension DefaultAuthRepository: AuthRepository {
     }
 
     func passwordStrength(email: String, password: String) async throws -> UInt8 {
-        try await clientService.auth().passwordStrength(password: password, email: email, additionalInputs: [])
+        try await clientService.auth(isPreAuth: true)
+            .passwordStrength(password: password, email: email, additionalInputs: [])
     }
 
     func sessionTimeoutAction(userId: String?) async throws -> SessionTimeoutAction {
--- a/BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
+++ b/BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
@@ -106,9 +106,11 @@ protocol AuthRepository: AnyObject {
     /// - Parameters:
     ///   - email: The user's email.
     ///   - password: The user's password.
+    ///   - isPreAuth: Whether the client is being used for a user prior to authentication (when
+    ///     the user's ID doesn't yet exist).
     /// - Returns: The password strength of the password.
     ///
-    func passwordStrength(email: String, password: String) async throws -> UInt8
+    func passwordStrength(email: String, password: String, isPreAuth: Bool) async throws -> UInt8
 
     /// Gets the profiles state for a user.
     ///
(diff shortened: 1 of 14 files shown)


BitwardenShared/Core/Platform/Services/ClientService.swift
```diff
@@ -199,28 +211,31 @@ class DefaultClientService: ClientService {
...
+
+        let userId = try await stateService.getAccountIdOrActiveId(userId: userId)
+
+        // If the user has a client, return it.
+        guard let client = userClientArray[userId] else {
+            // If not, create one, map it to the user, then return it.
+            let newClient = createAndMapClient(for: userId)
+            return newClient
+        }
+        return client
```
fedemkr, line 238: 🤔 Shouldn't we maintain the `catch StateServiceError.noAccounts, StateServiceError.noActiveAccount` logic in case this happens? Perhaps because of passing wrongly the `isPreAuth` flag.

BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
```diff
@@ -590,7 +590,8 @@ extension DefaultAuthRepository: AuthRepository {
     }
 
     func passwordStrength(email: String, password: String) async throws -> UInt8 {
-        try await clientService.auth().passwordStrength(password: password, email: email, additionalInputs: [])
+        try await clientService.auth(isPreAuth: true)
+            .passwordStrength(password: password, email: email, additionalInputs: [])
     }
```
fedemkr, line 595: 🤔 Is there a potential drawback if someone in the future uses this `passwordStrength` method while the vault is unlocked? I mean someone could just see the protocol definition and miss this implementation detail. Maybe just a comment on the protocol definition would be enough, if needed.

fedemkr:
Really nice catch, great job!!! 🎉 🎉

BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
matt-livefront: Maybe! I could expose it as a parameter here? I was thinking that `passwordStrength` probably isn't specific to the user, so maybe it wouldn't matter, but that could change.

BitwardenShared/Core/Platform/Services/ClientService.swift
matt-livefront: My thinking on removing it was that it would help us catch any potential uses where `isPreAuth` flag isn't set, but should be. Since if there's an active account and `isPreAuth` isn't set, it will just silently use the account's client.

But it would be safer to keep this, so I'll add it back and maybe see about logging an error.

fedemkr (approved):

PR #908, opened 2024-09-06
PM-11720 - TDE User Without MP Cannot Enable Autofill For Account

phil-livefront:
## 🎟️ Tracking

[PM-11720](https://bitwarden.atlassian.net/browse/PM-11720)
## 📔 Objective

- When a user doesn't have a MP set but has a TDE account, they should be brought to a SSO flow (if biometrics are enabled to authenticate). 

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-11720]: https://bitwarden.atlassian.net/browse/PM-11720?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
+++ b/BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
@@ -935,11 +935,9 @@ extension DefaultAuthRepository: AuthRepository {
             // No-op: nothing extra to do for decryptedKey.
             break
         case .deviceKey:
-            // No-op: nothing extra (for now).
-            break
+            try await configureBiometricUnlockIfRequired()
         case .keyConnector:
-            // No-op: nothing extra to do for Key Connector.
-            break
+            try await configureBiometricUnlockIfRequired()
         case let .password(password, _):
             let hashedPassword = try await authService.hashPassword(
                 password: password,
@@ -956,17 +954,7 @@ extension DefaultAuthRepository: AuthRepository {
                 try await stateService.setPinProtectedUserKeyToMemory(pinProtectedUserKey)
             }
 
-            // Re-enable biometrics, if required.
-            let biometricUnlockStatus = try? await biometricsRepository.getBiometricUnlockStatus()
-            switch biometricUnlockStatus {
-            case .available(_, true, false):
-                try await biometricsRepository.configureBiometricIntegrity()
-                try await biometricsRepository.setBiometricUnlockKey(
-                    authKey: clientService.crypto().getUserEncryptionKey()
-                )
-            default:
-                break
-            }
+            try await configureBiometricUnlockIfRequired()
         case .pin:
             // No-op: nothing extra to do for pin unlock.
             break
@@ -1030,5 +1018,18 @@ extension DefaultAuthRepository: AuthRepository {
     private func userIdOrActive(_ maybeId: String?) async throws -> String {
         if let maybeId { return maybeId }
         return try await stateService.getActiveAccountId()
+    }
+
+    private func configureBiometricUnlockIfRequired() async throws {
+        let biometricUnlockStatus = try? await biometricsRepository.getBiometricUnlockStatus()
+        switch biometricUnlockStatus {
+        case .available(_, true, false):
+            try await biometricsRepository.configureBiometricIntegrity()
+            try await biometricsRepository.setBiometricUnlockKey(
+                authKey: clientService.crypto().getUserEncryptionKey()
+            )
+        default:
+            break
+        }
     }
 }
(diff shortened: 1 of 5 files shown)


BitwardenShared/UI/Auth/Extensions/AuthRouter+Redirects.swift
```diff
@@ -354,13 +354,18 @@ extension AuthRouter {
...
-                return .vaultUnlock(
-                    activeAccount,
-                    animated: animated,
-                    attemptAutomaticBiometricUnlock: attemptAutomaticBiometricUnlock,
-                    didSwitchAccountAutomatically: didSwitchAccountAutomatically
-                )
+                let biometricUnlockStatus = try await services.biometricsRepository.getBiometricUnlockStatus()
+
+                if case let .available(_, enabled, validIntegrity) = biometricUnlockStatus, enabled, !validIntegrity {
+                    return .enterpriseSingleSignOn(email: activeAccount.profile.email)
```
matt-livefront, line 360: I think one other check we need to add here is to ensure the user doesn't have a master password, since otherwise they could use their MP to unlock the vault.
```
activeAccount.profile.userDecryptionOptions?.hasMasterPassword
```

matt-livefront (commented):
These changes look good! 

There may be one additional piece that we need though. After you use SSO to authenticate in the extension, if you open the extension again, I think we should prompt the user for Face ID at that point since they have authenticated and we could have saved the integrity state.

Here's where we set the integrity state if you unlock your vault with your MP:
https://github.com/bitwarden/ios/blob/c3f3a138c70fe7e3691c96259da98050b6a31873/BitwardenShared/Core/Auth/Repositories/AuthRepository.swift#L959-L966

I think we would want to do that if you use either a password, device key (TDE), or key connector in that method.

BitwardenShared/UI/Auth/Extensions/AuthRouter+Redirects.swift
```diff
@@ -354,13 +354,19 @@ extension AuthRouter {
...
+                if case .available(_, true, false) = biometricUnlockStatus, !hasMasterPassword {
+                    return .enterpriseSingleSignOn(email: activeAccount.profile.email)
+                } else {
+                    return .vaultUnlock(
+                        activeAccount,
+                        animated: animated,
+                        attemptAutomaticBiometricUnlock: attemptAutomaticBiometricUnlock,
+                        didSwitchAccountAutomatically: didSwitchAccountAutomatically
+                    )
+                }
```
fedemkr, line 369: 🎨 A small improvement I'd add is that we don't need to check for biometric status if `hasMasterPassword` is `true` so I'd rewrite this as:
```suggestion
                let hasMasterPassword = activeAccount.profile.userDecryptionOptions?.hasMasterPassword == true

                if !hasMasterPassword {
                    let biometricUnlockStatus = try await services.biometricsRepository.getBiometricUnlockStatus()
                    if case .available(_, true, false) = biometricUnlockStatus {
                        return .enterpriseSingleSignOn(email: activeAccount.profile.email)
                    }
                }

                return .vaultUnlock(
                    activeAccount,
                    animated: animated,
                    attemptAutomaticBiometricUnlock: attemptAutomaticBiometricUnlock,
                    didSwitchAccountAutomatically: didSwitchAccountAutomatically
                )
```

BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
```diff
@@ -1031,4 +1019,13 @@ extension DefaultAuthRepository: AuthRepository {
...
     }
+
+    private func configureBiometricUnlockIfRequired() async throws {
+        if case .available(_, true, false) = try? await biometricsRepository.getBiometricUnlockStatus() {
+            try await biometricsRepository.configureBiometricIntegrity()
+            try await biometricsRepository.setBiometricUnlockKey(
+                authKey: clientService.crypto().getUserEncryptionKey()
+            )
+        }
+    }
```
fedemkr, line 1030: ⛏️ Add docs.
❓ Would it be ok the unlock flows affected by this break if this throws? I'd guess it's fine but checking just in case.

phil-livefront: yeah I was thinking that if this fails for some reason then the calls wouldn't be executed and the user would have to provide another form of auth.  Instead of being given the optional to provide biometrics. Does that make sense?

matt-livefront (approved):
Looks good to me!

PR #913, opened 2024-09-10
[BITAU-137] Create AuthenticatorSyncKit SDK

brant-livefront:
## 🎟️ Tracking

[BITAU-137](https://livefront.atlassian.net/browse/BITAU-137)

## 📔 Objective

This PR adds the basic structure for sharing an `AuthenticatorSyncShared` Swift package from the main PM app repo. This allows us to share all of the code between the PM and Authenticator apps for building the sync feature, but doesn't require us to make a whole new git repo for managing it (it is simply shared from this repo). 

This PR is just about getting the basic structure set up correctly and getting feedback on this approach. However, there is the first piece of shared code `SharedKeychainRepository` which adds the ability to read/write keys in the shared keychain via the App Group. 

The next step after this PR is merged would then be a PR from the Authenticator repo to pull in this via Swift Package.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


Configs/AuthenticatorSyncShared.xcconfig
```diff
@@ -0,0 +1,4 @@
+#include "./Common.xcconfig"
+#include? "./Local.xcconfig"
+
+PRODUCT_BUNDLE_IDENTIFIER = $(ORGANIZATION_IDENTIFIER).AuthenticatorSyncShared
```
KatherineInCode, line 4: 🎨 All of our other bundle IDs are kebab-cased as near as I can tell; it probably makes sense for us to continue that

project.yml
```diff
@@ -55,12 +55,14 @@ schemes:
         - BitwardenAutoFillExtension
         - BitwardenShareExtension
         - BitwardenShared
+        - AuthenticatorSyncShared
```
KatherineInCode, line 58: 🎨 Can this be alphabetized?

Package.swift
```diff
@@ -0,0 +1,32 @@
...
+import PackageDescription
+
+let package = Package(
+    name: "BitwardenShared",
+    platforms: [
+        .iOS(.v15),
+    ],
+    products: [
+        .library(
+            name: "AuthenticatorSyncShared",
```
KatherineInCode, line 12: 🤔 I'm not super in love with this name, though I can't think of anything better. Just, if in the future we want to put more things in a shared library between the Password Manager and Authenticator apps, keeping it in the same package makes sense, but the name might not at that point. Probably moot, realistically, but something to think about

AuthenticatorSyncShared/Tests/SharedKeychainRepositoryTests.swift
```diff
@@ -0,0 +1,108 @@
+import CryptoKit
+import Foundation
+import XCTest
+
+@testable import AuthenticatorSyncShared
+
+final class SharedKeychainRepositoryTests: XCTestCase {
```
fedemkr, line 7: 🤔 Could we add a base class for `XCTestCase` and get some of [BitwardenTestCase](https://github.com/bitwarden/ios/blob/main/GlobalTestHelpers/Support/BitwardenTestCase.swift) methods into it? Like `assertAsyncThrows` which are really handy to use. Additionally, I'd like to have a base class for tests in case we need to do changes affecting the entire test suite.

Configs/AuthenticatorSyncShared.xcconfig
brant-livefront: Fixed in next commit. 👍

project.yml
brant-livefront: I assume you mean the whole file? (i.e. it should be the first target as well). I'll have it updated to that in the next commit

Package.swift
brant-livefront: 🤔 I see your point. We could possibly name it like the App Group and go with `BitwardenAuthenticatorShared`? That would simply imply where it's used (in both apps) rather than relating it just to sync.

Alternatively, we could keep this one scoped to just Sync (maybe even name it simply SyncShared) and set a pattern for future packages to be targeted to just one feature? i.e. The main package right now is BitwardenShared, but we are only sharing one product (`AuthenticatorSyncShared`). If you wanted to share Feature X in the future, you could make a FeatureXShared product/target and share via this same mechanism, rather than putting it also in the same product to share. That would add a bit more on the Authenticator side, however, as you'd need to explicitly pull in each product that you're interested in using.

AuthenticatorSyncShared/SharedKeychainRepository.swift
```diff
@@ -0,0 +1,155 @@
...
+        appSecAttrAccessGroupShared: String,
+        keychainService: AuthenticatorKeychainService
+    ) {
+        self.appSecAttrAccessGroupShared = appSecAttrAccessGroupShared
+        self.keychainService = keychainService
+    }
+
+    // MARK: Methods
+
+    /// Retrieve the value for the specifi item from the Keychain Service.
```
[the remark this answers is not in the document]
brant-livefront: 👍  will fix in next commit

AuthenticatorSyncShared/SharedKeychainRepository.swift
```diff
@@ -0,0 +1,155 @@
...
+                kSecClass: kSecClassGenericPassword,
+            ] as CFDictionary
+        )
+
+        if let resultDictionary = foundItem as? [String: Any],
+           let data = resultDictionary[kSecValueData as String] as? Data {
+            return data
+        } else {
+            throw AuthenticatorKeychainServiceError.keyNotFound(item)
+        }
```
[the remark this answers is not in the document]
brant-livefront: 👍  will fix in next commit

AuthenticatorSyncShared/Tests/SharedKeychainRepositoryTests.swift
brant-livefront: Sounds good. I was trying to make sure we didn't duplicate too much, but totally understand wanting to have one base class to work from 👍 `assertAsyncThrows` will definitely be handy to have. I'll add in the next commit.

project.yml
KatherineInCode: We're not the most consistent about alphabetization with the targets (notably, we put `BitwardenTests` after `Bitwarden` but before `BitwardenActionExtension`), but yeah, in general if we have a list of things that doesn't have an enforced order, by convention we like to alphabetize them.

Package.swift
KatherineInCode: Hm. I see merits to both approaches. If we did it by-feature, I'd suggest actually maybe `AuthenticationSyncShared`, so that it's more clearly "this is syncing authentication things".

But also having a broader one that makes it easier in the future for us to just move things from `BitwardenShared` to it in order to share between the two projects... (though personally, I would have called that `BitwardenShared` so I don't know 😂 )

@fedemkr Do you have a strong opinion on whether we want to name more broadly or do a per-feature naming, or something else entirely?

AuthenticatorSyncShared/Tests/SharedKeychainRepositoryTests.swift
brant-livefront: Added this in the latest commit. I took out the UI-related stuff (partially because we don't have UI for this in the package, but also because that had some BitwardenShared dependencies as well).

project.yml
brant-livefront: Updated in the latest commit. Let me know if we need to adjust anything else 👍

Package.swift
fedemkr: Naming is hard, can't think of a really good name for this 🤔 .

I think it depends on the scope (and possible future scope) of this. I would guess that what we're possibly going to share between the two apps would be **auth** data and syncing the **vault**. 
Additionally, I wouldn't put general "helpers" or controls shared between any potential apps as I would add these kind of stuff to an independent package we can reuse amongst all current/future apps without nothing specific of a particular one, e.g. we could also potentially use this future package on the Watch app.

Having said this I'm fine with the **particular naming** for now to denote what the package is about and helps not to include general stuff in there. Furthermore, we can revisit the naming while we include stuff if needed.
I thought about the next ones but not strongly in favor of any of these nor thinking they're better than the original one:
- `AuthenticatorSyncCore`
- `AuthenticatorSyncEngine`
- `AuthenticatorSyncManager`
- `AuthenticatorSyncKit`
- `AuthenticatorSyncFoundation`

AuthenticatorSyncShared/Tests/SharedKeychainRepositoryTests.swift
```diff
@@ -0,0 +1,108 @@
...
+        do {
+            _ = try await subject.getAuthenticatorKey()
+            XCTFail("Expected error `keyNotFound` was never thrown!")
+        } catch let caughtError as AuthenticatorKeychainServiceError {
+            XCTAssertEqual(caughtError, error)
+        } catch let caughtError {
+            XCTFail(
+                "The error caught (\(caughtError)) does not match the type of error provided (\(error))."
+            )
+        }
```
fedemkr, line 88: 🎨 `assertAsyncThrows` can be used instead of this to be more clean leveraging the new `AuthenticatorSyncSharedTestCase`.

Package.swift
KatherineInCode: I'm only a little cautious about naming here because changing the name of a package like this, when it's shared, could be mildly disruptive between the two projects.

I do actually like `Kit` as a suffix; that's something I've used before for this sort of thing, now that I think about it, and it's kind of how Apple likes it. `AuthenticationSyncKit`? `AuthenticatorSyncKit`? Something like that seems like it would be good.

fedemkr: Yeah, yesterday reviewed Apple's naming and `Kit` and `Foundation` are usually used as suffixes 😄 . Totally agree let's go with `AuthenticatorSyncKit` 😄 (I wouldn't go for `Authentication` given that if I recall correctly this would also allow the app to send ciphers from one place to the other)

KatherineInCode: Sounds good to me!

brant-livefront: Sounds good to me as well - I'll migrate to `AuthenticatorSyncKit` 👍

brant-livefront: @KatherineInCode @fedemkr OK, I think I've got all the names changed over. Latest commit should have everything running with the new name. But do be on the lookout for anything I might have missed.

AuthenticatorSyncShared/Tests/SharedKeychainRepositoryTests.swift
brant-livefront: This should be updated in the latest.

AuthenticatorSyncKit/SharedKeychainRepository.swift
```diff
@@ -0,0 +1,155 @@
...
+                kSecAttrAccessGroup: appSecAttrAccessGroupShared,
+                kSecAttrAccessible: kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly,
+                kSecAttrAccount: item.unformattedKey,
+                kSecClass: kSecClassGenericPassword,
+            ] as CFDictionary
+        )
+
+        guard let resultDictionary = foundItem as? [String: Any],
+              let data = resultDictionary[kSecValueData as String] as? Data else {
+            throw AuthenticatorKeychainServiceError.keyNotFound(item)
```
fedemkr, line 97: ⛏️ Could tests be added for when `foundItem` is not `[String: Any]` and another one for when `resultDictionary[kSecValueData as String]` is not of `Data` type?

fedemkr (commented):
Looks good, just a few tests missing.

AuthenticatorSyncKit/SharedKeychainRepository.swift
brant-livefront: Good catch 👍 Added those in the latest.

AuthenticatorSyncKit/Tests/SharedKeychainRepositoryTests.swift
```diff
@@ -0,0 +1,126 @@
...
+    override func tearDown() {
+        keychainService = nil
+        subject = nil
+    }
+
+    // MARK: Tests
+
+    /// Verify that `deleteAuthenticatorKey()` issues a delete with the correct search attributes specified.
+    ///
+    func testDeleteAuthenticatorKey() async throws {
```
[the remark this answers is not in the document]
brant-livefront: 👍 should be fixed in latest

fedemkr (approved):

victor-livefront (approved):

brant-livefront:
@fedemkr @victor-livefront Sorry, I realized right before merging that we should have a bundle version of `0.1.0` on the Swift package, rather than `1.0`. We've got more things to put in there before it's 1.0 and that matches what the Android side is doing as well.

This dismissed your approvals, so I'll need a re-approval before I merge.

victor-livefront (approved):

PR #923, opened 2024-09-12
PM-11491: Allow adding a new password autofill item when the website supports passkeys

matt-livefront:
## 🎟️ Tracking

[PM-11491](https://bitwarden.atlassian.net/browse/PM-11491)

## 📔 Objective

If a website supports passkeys and the user triggers the autofill extension, iOS seems to call `prepareCredentialList(for:requestParameters:)` instead of `prepareCredentialList(for:)`. We were using which method was called to determine whether or not the user should be able to add a new login item. Even though you can't add a new passkey in this flow, we should still allow adding a new login item. This brings back the add login buttons in this flow.

## 📸 Screenshots

| Before | After |
| -- | -- |
| <video src="https://github.com/user-attachments/assets/5fa8e751-45e9-467d-b8a5-cd08e1afc021" /> | <video src="https://github.com/user-attachments/assets/0da4f38b-e8af-4292-9571-b79cedfad145" /> |

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


BitwardenShared/UI/Vault/Vault/AutofillList/VaultAutofillListProcessor.swift
```diff
@@ -358,7 +358,6 @@ extension VaultAutofillListProcessor {
             }
         case let .autofillFido2VaultList(serviceIdentifiers, fido2RequestParameters):
             state.isAutofillingFido2List = true
-            state.emptyViewMessage = Localizations.noItemsToList
```
fedemkr, line 361: 🤔 Is this no longer necessary?

matt-livefront: Setting `emptyViewMessage`? I don't think so, it will default to `There are no items in your vault for this website/app. Tap to add one.`, which I think is what we would want to say in this scenario?

https://github.com/bitwarden/ios/blob/4de2663e2b7d9c598aacdb220d61ee95c3c914f4/BitwardenShared/UI/Vault/Vault/AutofillList/VaultAutofillListState.swift#L15

fedemkr: Oh you're right with the change in behavior of the add button, now this is no longer necessary. Thanks for checking it!

fedemkr (approved):

PR #926, opened 2024-09-13
[PM-11490] iOS 18 fix broken account switcher

fedemkr:
## 🎟️ Tracking

[PM-11490](https://bitwarden.atlassian.net/browse/PM-11490)

## 📔 Objective

iOS 18 -> Fix broken account switcher that was being shown behind the toolbar.

Removed `scrollOffsetChanged` and `scrollOffset` from the `ProfileSwitcherHandler`/`ProfileSwitcherState`/`ProfileSwitcherAction` as they're no longer necessary as the state is now handled by the `ProfileSwitcherView` -> Thanks @matt-livefront for this fix/improvement.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-11490]: https://bitwarden.atlassian.net/browse/PM-11490?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenShared/UI/Auth/ProfileSwitcher/ProfileSwitcherView.swift
```diff
@@ -67,7 +61,7 @@ struct ProfileSwitcherView: View {
     private var backgroundView: some View {
         VStack {
             Asset.Colors.backgroundPrimary.swiftUIColor
-                .frame(height: abs(min(store.state.scrollOffset.y, 0)))
+                .frame(height: store.state.isVisible ? 0.01 : 0)
```
phil-livefront, line 64: was this needed?

phil-livefront (dismissed):

BitwardenShared/UI/Auth/ProfileSwitcher/ProfileSwitcherView.swift
```diff
@@ -9,13 +9,7 @@ struct ProfileSwitcherView: View {
...
 
     var body: some View {
-        OffsetObservingScrollView(
-            axes: store.state.isVisible ? .vertical : [],
-            offset: .init(
-                get: { store.state.scrollOffset },
-                set: { store.send(.scrollOffsetChanged($0)) }
-            )
-        ) {
+        ScrollView {
```
matt-livefront, line 12: ❓ I thought @phil-livefront noticed that removing the axes ternary ended up with the same result? Curious if you tried that.

```
        OffsetObservingScrollView(
            axes: .vertical,
            offset: .init(
                get: { store.state.scrollOffset },
                set: { store.send(.scrollOffsetChanged($0)) }
            )
        ) {
```

BitwardenShared/UI/Auth/ProfileSwitcher/ProfileSwitcherView.swift
fedemkr: It doesn't seem to be necessary, at least on the iOS 18 device.

BitwardenShared/UI/Auth/ProfileSwitcher/ProfileSwitcherView.swift
fedemkr: Yep, that also works. But if we're not changing the `axes` and not going to use the `scrollOffset` then I'd just use a normal `ScrollView` for this as the `OffsetObservingScrollView` won't add any value to this.

matt-livefront: I guess I'm confused why we don't need the `scrollOffset` anymore?

matt-livefront: Maybe there's an issue with the scroll offset currently. But I think the intention of it was to hide the space above the scroll view if you scroll it beyond the top of the content.

| Current | Expected |
| -- | -- |
| <video src="https://github.com/user-attachments/assets/087b67fd-7134-4c2f-9b3f-4af8f72b2694" /> | <video src="https://github.com/user-attachments/assets/f041870a-9234-4826-9dcc-aadde6b226a6" /> |

I can fix this separately, but if it's not causing any issue maybe we should keep it in.

matt-livefront (dismissed):

fedemkr:
@matt-livefront had to remake the snapshots for the profile switcher because now it doesn't show the blank space behind the toolbar. I think it's ok and I've just overridden the old snapshots with the current behavior. Let me know if you think this could cause an issue somewhere but I can't think of a place where this could go wrong.

matt-livefront:
> @matt-livefront had to remake the snapshots for the profile switcher because now it doesn't show the blank space behind the toolbar. I think it's ok and I've just overridden the old snapshots with the current behavior. Let me know if you think this could cause an issue somewhere but I can't think of a place where this could go wrong.

Ah, I think that's ok. The gap at the top I think is from the safe area, and switching from the VStack to ZStack likely changed the background in the safe area. But since the profile switcher will always display below a navigation bar, I don't think it's an issue.

matt-livefront (approved):

matt-livefront (dismissed):

KatherineInCode (dismissed):

KatherineInCode (approved):

PR #937, opened 2024-09-16
[BITAU-134] [BITAU-121] Create Shared CoreData Store

brant-livefront:
## 🎟️ Tracking

[[BITAU-134](https://livefront.atlassian.net/browse/BITAU-134)]
[[BITAU-121](https://livefront.atlassian.net/browse/BITAU-121)] 

## 📔 Objective

This PR adds the CoreData store, model class, and helpers to AuthenticatorBridgeKit. This extends the shared code to include a way to share data between the two apps. The CoreData store is protected by the App Group, so that only members of the App Group are able to load data from it. There's a number of convenience methods on the AuthenticatorBridgeDataStore that anticipate a few use-cases where we'll be calling this in the future, with a few more to be added.

Currently, this does not do any encryption as part of the store. I've pulled that code out into a separate branch which I will put up into a PR shortly after this one ([PR for that branch](https://github.com/bitwarden/ios/pull/938)). I felt like it was a bit much to review in two pieces, but the intention is to have that run through the DataStore so that nothing is stored without first being encrypted.

I also have not added code to instantiate/use any of this yet to the PM app. I'm trying to do as much as we can in the background before adding it behind the feature flag.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


AuthenticatorBridgeKit/AuthenticatorBridgeDataStore.swift
fedemkr, line 1: 🤔 What do you think about implementing a similar approach as the main app? I.e. the `DataStore` is a class that defines general behavior and doesn't have specific data models in there, however provide protocols for these specific data models and then the `DataStore` can have extensions that conform to such protocols for usage.
Specifically I'm referring to:
`public func fetchAllForUserId(_ userId: String) async throws -> [AuthenticatorBridgeItemDataModel]`
and
`public func insertItems(_ items: [AuthenticatorBridgeItemDataModel], forUserId userId: String) async throws`
which are specific for `AuthenticatorBridgeItemDataModel`.

This approach is more flexible if we need to add additional data models in the library to provide a clean separation between such models.

AuthenticatorBridgeKit/AuthenticatorBridgeItemData.swift
```diff
@@ -0,0 +1,146 @@
+import CoreData
+import Foundation
+
+/// A data model for persisting authenticator items into the shared CoreData store.
+///
+public class AuthenticatorBridgeItemData: NSManagedObject {
```
fedemkr, line 6: 🤔 Maybe it's worth to add the [ManagedObject](https://github.com/bitwarden/ios/blob/main/BitwardenShared/Core/Platform/Models/Data/ManagedObject.swift) extension that we have on the main app. So when we create a share core library to use amongst swift apps we can quickly update/replace the code.

AuthenticatorBridgeKit/Tests/TestHelpers/Fixtures/AuthenticatorBridgeItemDataModel+Fixtures.swift
```diff
@@ -0,0 +1,34 @@
...
+@testable import AuthenticatorBridgeKit
+
+extension AuthenticatorBridgeItemDataModel {
+    static func fixture(
+        favorite: Bool = true,
+        id: String = UUID().uuidString,
+        name: String = "Name",
+        totpKey: String? = "TOTP Key",
+        username: String? = "Username"
+    ) -> AuthenticatorBridgeItemDataModel {
```
fedemkr, line 12: ⛏️ On fixtures I think it's better to rely on default values and then customize as needed or provide a particular `fixtureWhatever` func for customization. This helps when using this amongst different tests where it's more intuitive to know the value of a default parameter and the expect behavior of such. What do you think?
```suggestion
    static func fixture(
        favorite: Bool = false,
        id: String = UUID().uuidString,
        name: String = "Name",
        totpKey: String? = nil,
        username: String? = nil
    ) -> AuthenticatorBridgeItemDataModel {
```

brant-livefront: That sounds reasonable to me. 👍 I'll adjust this over in the next commit.

AuthenticatorBridgeKit/AuthenticatorBridgeItemData.swift
brant-livefront: In the POC, we just duplicated a couple of things (like ManagedObject, ManagedUserObject, etc). I had taken a slightly different tack of trying to minimize that and instead be more specific to just what we needed in the AuthenticatorBridgeKit.

But if we're OK with duplicating (for now, with an eye to refactoring to a shared library sometime in the future), I can move back to that. Then the DataStore would be much more generic (like the main app), and we'd add a wrapper (I liked your suggestion on the other PR of `AuthenticatorBridgeItemService`) to handle the convenience methods, encrypting, etc.

I'll refactor over to that here shortly.

AuthenticatorBridgeKit/AuthenticatorBridgeDataStore.swift
brant-livefront: See my comment below - I had pushed things up into the DataStore for a lot of this. But if we're ok with duplicating some of the CoreData extensions from the main app, I'll refactor this back to being more generic and add the convenience methods to a separate service.  👍

AuthenticatorBridgeKit/AuthenticatorBridgeItemData.swift
fedemkr: I guess that unless we know this data class is the only one we're going to use for now I'd go for the more general one. However I'd like to hear @KatherineInCode opinions as well.

KatherineInCode: It does look like `ManagedObject` is duplicated already between BWA and PM. It would be nice if we could just declare conformance of this to `ManagedObject` and use it from both 🤔 

Ultimately I'm okay with this duplication for the time being, but it does seem like a relatively easy target for sharing between the apps—though then everything that uses the `ManagedObject` in the PM will have to then import the Bridge library, which seems non-ideal from a dependency perspective.

Almost as though we actually need a *separate* library that's just "Core Data Components" 😂

AuthenticatorBridgeKit/AuthenticatorBridgeItemDataModel.swift
```diff
@@ -0,0 +1,39 @@
+import Foundation
+
+/// A struct for storing information about items that are shared between the Bitwarden and Authenticator apps.
+///
+public struct AuthenticatorBridgeItemDataModel: Codable, Equatable {
```
KatherineInCode, line 5: 🤔 I feel like there was some reason I didn't make the `AuthenticatorItemDataModel` to conform to `Equatable` in the Authenticator, though I don't remember why, exactly. Is there a particular reason we want it here?

(I'm also finding myself wondering, in light of the `ManagedObject` conversation, if we *actually* might want to eventually consolidate the `AuthenticatorItemDataModel` and `AuthenticatorBridgeItemDataModel` into one type that's just used by both apps; it just so happens the BWA has local storage of those objects as well 🤔 Not sure how that would play with Core Data, though. On the other hand...would it potentially be possible to just...use the `AuthenticatorItemDataModel` and move it into this library, to be shared code to begin with? And update BWA to use that instead?)

AuthenticatorBridgeKit/NSManagedObjectContext+Extension.swift
```diff
@@ -0,0 +1,52 @@
+import CoreData
+
+extension NSManagedObjectContext {
```
KatherineInCode, line 3: 🤔 With this, we'll have the same file in three different places, as near as I can tell, since there's copies of this already in PM and BWA. Does it make sense to add it again?

AuthenticatorBridgeKit/AuthenticatorBridgeItemData.swift
brant-livefront: I've refactored to use the more generic version and pulled in a handful of files. These are all duplicates of what is in the main PM app (and in some cases BWA as well, as Katherine points out), but it does give us the more generic version of all of this and a possible path to consolidating them later.

I'm OK with either approach, just let me know which one you prefer. But I'm going to push up my latest commits which will bring in the more generic version so you all can see what that looks like in practice. If we choose to go the more specific route, we can always roll back to some of the earlier versions.

AuthenticatorBridgeKit/NSManagedObjectContext+Extension.swift
brant-livefront: This one was the one piece I felt like there wasn't a good way around duplicating it. And as you mentioned above, I was hesitant to make everything depend on the Bridge library and move this exclusively in there. 🤔

AuthenticatorBridgeKit/AuthenticatorBridgeItemDataModel.swift
brant-livefront: 🤔 That's an interesting question. As far as CoreData, I'm not sure how it would work to have the local data store (writing to the app's container) and the shared data store (writing to the app's app group) using the exact same CoreData model. It's possible that's fine if we just reference everything properly. I could try doing a spike?

But also, that would mean that Authenticator immediately becomes wholly dependent on the Swift package for even its own local storage. The items are pretty much identical, but I wasn't sure if we want to jump all the way in on that.

KatherineInCode: I think it's fine currently, despite them being identical. Probably not worth digging in on that too much at the moment, but something to keep in mind in the future.

AuthenticatorBridgeKit/AuthenticatorBridgeItemData.swift
KatherineInCode: I think ultimately I prefer the generic version, even though that's a bit more chrome and duplication here. But I also think that makes it easier to identify the duplication (because what's duplicated is a clearly-named separate file rather than inline) so in the future if/when we want to consolidate we can.

fedemkr: Agree on the generic version.

KatherineInCode (approved):

AuthenticatorBridgeKit/CodableModelData.swift
```diff
@@ -0,0 +1,40 @@
...
+
+extension CodableModelData {
+    /// Encodes or decodes the model to/from the data instance.
+    var model: Model? {
+        get {
+            guard let modelData else { return nil }
+            do {
+                return try JSONDecoder().decode(Model.self, from: modelData)
+            } catch {
+                Logger.bridgeKit.error("Error decoding \(String(describing: Model.self)): \(error)")
```
fedemkr, line 24: 🤔 Should we log to crashlytics (if possible, i.e. use `ErrorReporter`) if an error happens here? Or directly not catch it and let the caller handle it? I mean this would be pretty hidden if it happens to an end user.  Maybe I'm being overcautious here but wanted to raise it just in case.

brant-livefront: In the main PM app's `CodableModelData` it just logs to Logger.application - so I was following that as a guide. I think that's probably the best choice though:
- This is coding/decoding the object that our code put in there (as opposed to say, something from a server). So it's pretty unlikely it would be of an incorrect format. And if it was, it's pretty unclear how we'd recover.
- Since these are in getter/setters it is a little awkward to throw the error out of this context. Instead, we're just returning `nil` and requiring that the property be Optional. So any error handling is likely going to be done where we unwrap that and use the value in the caller.

fedemkr: Makes sense thanks!

fedemkr (approved):

PR #938, opened 2024-09-16
[BITAU-122] [BITAU-133] Add encryption to/from shared the CoreData store

brant-livefront:
## 🎟️ Tracking

[[BITAU-122](https://livefront.atlassian.net/browse/BITAU-122)]
[[BITAU-133](https://livefront.atlassian.net/browse/BITAU-133)] 

## 📔 Objective

This PR extends [the previous CoreData PR](https://github.com/bitwarden/ios/pull/937) to add encryption to the process of storing and fetching data in the store. The cryptography service provides all of the encryption/decryption (with the key from SharedKeychainRepository). The ItemService uses the cryptoService to encrypt items before it stores any data into the Store, and then decrypts the items whenever it does a fetch. This allows clients to simply call the convenience methods on the ItemService and deal with the items as unencrypted, decoded objects.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


AuthenticatorBridgeKit/AuthenticatorBridgeDataStore.swift
fedemkr, line 1: 🤔 I'm not very fond of having the `DataStore` referencing services as IMO the store shouldn't be worrying whether the data is encrypted or not, it's just CRUD operations.
I think the encrypt/decrypt part should be in a service, e.g. `AuthenticatorBridgeItemService`, which calls this `AuthenticatorBridgeDataStore` to get the data and then performs the encryption/decryption operations.

brant-livefront: I agree. On the other PR I'm going to migrate the DataStore back to being generic. I'll move the convenience methods into a separate service - I like the naming of `AuthenticatorBridgeItemService` 👍

brant-livefront:
This PR has been refactored to adopt the latest from main, now that we merged in the changes to the DataStore and added the ItemService. It should be all set for re-review 👍

AuthenticatorBridgeKit/Tests/AuthenticatorBridgeItemDataTests.swift
```diff
@@ -16,19 +17,22 @@ final class AuthenticatorBridgeItemDataTests: AuthenticatorBridgeKitTestCase {
 
     override func setUp() {
         super.setUp()
+        cryptoService = MockSharedCryptographyService()
```
victor-livefront, line 20: In your existing tests, can you now check to see if the cryptography service decrypt and encrypt are called on the mock?

AuthenticatorBridgeKit/SharedCryptographyService.swift
```diff
@@ -0,0 +1,143 @@
...
+    /// returns the list with each member encrypted.
+    ///
+    /// - Parameter items: The decrypted array of items to be encrypted
+    /// - Returns: the array of items with their data encrypted
+    /// - Throws: AuthenticatorKeychainServiceError.keyNotFound if the Authenticator
+    ///     key is not in the shared repository.
+    ///
+    func encryptAuthenticatorItems(
+        _ items: [AuthenticatorBridgeItemDataModel]
+    ) async throws -> [AuthenticatorBridgeItemDataModel]
```
fedemkr, line 32: 🤔 Would it be worth keeping the same approach we have on the apps where the decrypted object name has `View` as a suffix? What do you think? cc: @KatherineInCode 
I personally would like to so we maintain the same approach everywhere but would like to hear your thoughts on this.
```swift
func decryptAuthenticatorItems(
    _ items: [AuthenticatorBridgeItemDataModel]
) async throws -> [AuthenticatorBridgeItemView]

func encryptAuthenticatorItems(
    _ items: [AuthenticatorBridgeItemView]
) async throws -> [AuthenticatorBridgeItemDataModel]
```

victor-livefront (commented):
Just a tiny comment on some possible extra testing.

AuthenticatorBridgeKit/Tests/AuthenticatorBridgeItemDataTests.swift
brant-livefront: Yeah, I definitely test that in the `AuthenticatorBridgeItemServiceTests` since the service has the responsibility. I didn't add that here since it's somewhat out of the scope of the `AuthenticatorBridgeItemData`. But since it's also running through the service to test the data object, I don't have an objection to adding that here as well.

AuthenticatorBridgeKit/SharedCryptographyService.swift
brant-livefront: I think of these as part of the syncing layer and the View portion as the responsibility of the calling apps - i.e. in the Authenticator app when we consume this, we're definitely going to be mapping these from models to views for actual display. And by not putting a View object here, we can map them directly into the View objects that already exist over there.

But I'm definitely open if you'd rather do that here. I do like the distinction of two different objects so we're not confused if this is the encrypted one or not.

KatherineInCode: With [the CryptographyService in BWA](https://github.com/bitwarden/authenticator-ios/blob/main/AuthenticatorShared/Core/Vault/Services/CryptographyService.swift#L42) I tried to follow the convention of "Encrypting goes from `thingView` to `thing` and decrypting goes from `thing` to `thingView`" that's essentially how the SDK handles it (despite my finding the use of "view" for something that isn't a View confusing).

Having the cryptography service be `AuthenticatorBridgeItemDataModel-> AuthenticatorBridgeItemDataModel` is in my opinion going to be very confusing and lead to problems, because now you have to keep track of whether you have an _encrypted_ object or _decrypted_ object. I'd much rather have two different types, so that when you look at code you always know what you're dealing with.

fedemkr: ℹ️ AFAIK here the term "View" is used as in databases where you have the plain table / data model and then a "View" which is a virtual "table" / object which is often used so it's ready to be used for higher layers without having to use the whole table/data model or having to perform additional logic which is taken care in the DB, amongst other reasons. So basically is a domain "View" not a UI view, but yes the mix is a bit confusing.

✅  I agree as well on using the suffix `View` here as is, in fact, a decrypted object which is easier to track and comprehend comparing it to the main apps code.

AuthenticatorBridgeKit/Tests/AuthenticatorBridgeItemDataTests.swift
brant-livefront: Added in the latest commit 👍

victor-livefront (approved):

AuthenticatorBridgeKit/AuthenticatorBridgeItemDataView.swift
```diff
@@ -0,0 +1,40 @@
...
+    ///   - totpKey: The TOTP key used to generate codes.
+    ///   - username: The username of the Bitwarden account that owns this iteam.
+    ///
+    public init(favorite: Bool, id: String, name: String, totpKey: String?, username: String?) {
+        self.favorite = favorite
+        self.id = id
+        self.name = name
+        self.totpKey = totpKey
+        self.username = username
+    }
```
fedemkr, line 39: ⛏️ I think we can remove this as Swift already gives us this initializer for free unless you're planning to do extra logic here in a future PR.
```suggestion
```

brant-livefront: That one needs to be there so that we can declare in as `public`. I took it off the Model version because we don't want callers (i.e. external to the AuthenticatorBridgeKit library) to init those. But these View ones will be used externally.

AuthenticatorBridgeKit/SharedCryptographyService.swift
KatherineInCode: Oh, yeah, "View" is definitely a term used in databases, and also has a meaning in server-side applications. It's just that "View" has a very particular sort of meaning in the iOS ecosystem (and its attempt at an MVC architecture) also, so the doubling-up was *very* confusing to me when I first got into the codebase.

It is, however, our convention. Thanks, computer science, for doubling up on terminology 😂

AuthenticatorBridgeKit/SharedCryptographyService.swift
```diff
@@ -0,0 +1,143 @@
...
+    ///   - plainText: The string to encrypt
+    ///   - key: The key to use to encrypt the string
+    /// - Returns: An encrypted string or `nil` if the string was nil
+    ///
+    private func encrypt(_ plainText: String?, withKey key: SymmetricKey) -> String? {
+        guard let plainText else {
+            return nil
+        }
+
+        let nonce = randomData(lengthInBytes: 12)
```
KatherineInCode, line 124: 🤔 "Nonce" is a problematic term due to its meaning in the UK, but unfortunately the cryptography community still uses it. It's probably moot here (the library function is called that, and we're *clearly* in cryptography land) so I mostly just wanted to sigh about it and how language is hard.

KatherineInCode (approved):
Looks reasonable to me

AuthenticatorBridgeKit/AuthenticatorBridgeItemDataView.swift
fedemkr: Ohhh missed the `public` modifier, thanks for clarifying 👍

fedemkr (approved):

PR #941, opened 2024-09-17
PM-11215 - Build Debug Menu

phil-livefront:
## 🎟️ Tracking
[PM-11215](https://bitwarden.atlassian.net/browse/PM-11215)

## 📔 Objective

- Build a debug menu that is only accessible from debug builds when you triple tap any screen with 3 fingers or a shake gesture
- Currently allows for feature flag toggling but this can be used for other configurations moving forward
- Allows for the ability to override a boolean feature flag value so you can easily test different flags without them needing to be remotely configured.  
- Also added in a refresh button to go back to the defaults
- For the demo video below, using the simulator only allows for 2 mutli-taps at a time but we have it set to 3 and made sure this worked on device

## 📸 Screenshots
Simulator Screenshot - iPhone 15 Pro - 2024-09-17 at 09 27 32

https://github.com/user-attachments/assets/25d5a785-913c-4490-a428-e8fda92c57e0

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-11215]: https://bitwarden.atlassian.net/browse/PM-11215?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


KatherineInCode:
🤔 Having the gesture for the debug menu be something that can't be done on the simulator is weird to me. Wouldn't it be better to have something that doesn't require a physical device? Or is the expectation that if we need the debug menu locally we'll go and change the gesture?

Also of note: Debug builds are *only* built for dev locally with Xcode. Any build in TestFlight is a Release build, so this will only be usable by devs and not QA.

BitwardenShared/Core/Platform/Services/Stores/AppSettingsStore.swift
```diff
@@ -131,6 +131,18 @@ protocol AppSettingsStore: AnyObject {
...
+    ///
+    /// This method fetches the value for a specified feature flag from the app's settings store.
+    /// The value is decoded into a specified type that conforms to `Codable`. If the flag does not exist
+    /// or cannot be decoded, the method returns `nil`.
+    ///
+    /// - Parameter name: The name of the feature flag to retrieve, represented as a `String`.
+    /// - Returns: The value of the feature flag as the specified type `T`, or `nil` if the flag does not exist
+    ///     or cannot be decoded.
+    ///
+    func featureFlag(name: String) -> Bool?
```
KatherineInCode, line 144: ⛏️ The docs indicate that it's a generic type T, but all we have here is Bool. This also means we don't have overrides for Int or String flags.

BitwardenShared/Core/Platform/Services/Stores/AppSettingsStore.swift
```diff
@@ -281,6 +293,19 @@ protocol AppSettingsStore: AnyObject {
...
+    /// This method updates or removes the value for a specified feature flag in the app's settings store.
+    /// If the `value` parameter is `nil`, the feature flag is removed from the store. Otherwise, the flag
+    /// is set to the provided boolean value.
+    ///
+    /// - Parameters:
+    ///   - name: The name of the feature flag to set or remove, represented as a `String`.
+    ///   - value: The boolean value to assign to the feature flag. If `nil`, the feature flag will be removed
+    ///    from the settings store.
+    ///
+    func setFeatureFlag(name: String, value: Bool?)
```
KatherineInCode, line 307: 🤔 If we're gating using these functions behind an `#if DEBUG`, wouldn't it make sense to also gate these functions and attendant code?

(Other discussion about whether we want to `#if DEBUG` aside)

BitwardenShared/Core/Platform/Services/Stores/AppSettingsStore.swift
phil-livefront: good catch, yeah for now I kept them as booleans but I can add capability for other types.  I just didn't want to hold this up for testing the onboarding flow but I can certainly tackle that.  Would another ticket be okay with you?

phil-livefront:
> 🤔 Having the gesture for the debug menu be something that can't be done on the simulator is weird to me. Wouldn't it be better to have something that doesn't require a physical device? Or is the expectation that if we need the debug menu locally we'll go and change the gesture?
> 
> Also of note: Debug builds are _only_ built for dev locally with Xcode. Any build in TestFlight is a Release build, so this will only be usable by devs and not QA.

Yeah I agree, the ticket stated to use the gesture so I went with that but we could add another vessel to open it as well. Do you have any that would be useful?  I've done a shake gesture before, which can be used in the simulator, or also triple tapping specific images, etc.

BitwardenShared/Core/Platform/Services/Stores/AppSettingsStore.swift
phil-livefront: yeah that a great call too, no need to have these functions available unless they are needed.

BitwardenShared/UI/Platform/DebugMenu/DebugMenuView.swift
```diff
@@ -0,0 +1,93 @@
...
+    var body: some View {
+        ZStack {
+            List {
+                Section {
+                    featureFlags
+                } header: {
+                    featureFlagSectionHeader
+                }
+            }
+        }
```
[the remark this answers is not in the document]
phil-livefront: ah yea, i actually had this originally and im not even sure why I added the ZStack tbh.  Will remove.

BitwardenShared/UI/Platform/DebugMenu/DebugMenuProcessor.swift
```diff
@@ -0,0 +1,90 @@
...
+    // MARK: Private Functions
+
+    /// Asynchronously fetches and updates feature flags for the debug menu.
+    /// If we are in a DEBUG build, we will try to use the value stored in User Defaults,
+    /// otherwise use the remote state or default to false.
+    private func fetchFlags() async {
+        let remoteFeatureFlags = await services.configService.getRemoteFeatureFlags()
+
+        state.featureFlags = FeatureFlag.allCases.map { feature in
+            let userDefaultValue = services.appSettingsStore.featureFlag(name: feature.rawValue)
```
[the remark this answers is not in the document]
phil-livefront: I think that is a great idea!

BitwardenShared/Core/Platform/Services/Stores/AppSettingsStore.swift
KatherineInCode: Future tickets for the additional types is reasonable to keep this a bit smaller and focused.

BitwardenShared/Core/Platform/Services/ConfigService.swift
```diff
@@ -187,6 +203,10 @@ class DefaultConfigService: ConfigService {
         return configuration?.featureStates[flag]?.stringValue ?? defaultValue
     }
 
+    func getRemoteFeatureFlags() async -> [FeatureFlag: AnyCodable] {
```
KatherineInCode, line 206: 🤔 Does it make sense to also gate this behind the `DEBUG_MENU` flag?

KatherineInCode:
> we could add another vessel to open it as well

I don't have any good ideas for it, unfortunately; the way apps I've worked on in the past handled debug screens and the like was through other means. The three-finger triple tap is probably reasonable enough, and if we need to access the debug menu in the simulator we could just manually update it to five two-finger taps or something like that. In practice, it's probably moot.

BitwardenShared/Core/Platform/Services/ConfigService.swift
phil-livefront: yeah, good call out.  That would make the most sense to me to avoid any issues in release builds

Bitwarden/Application/SceneDelegate.swift
```diff
@@ -121,4 +131,21 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
     private func showSplash() {
         splashWindow?.alpha = 1
     }
+
+    /// Handle the triple-tap gesture and launch the debug menu.
```
KatherineInCode, line 135: ⛏️ Can probably gate this behind `DEBUG_MENU` as well

Configs/BitwardenShared.xcconfig
```diff
@@ -1,4 +1,6 @@
 #include "./Common.xcconfig"
 #include? "./Local.xcconfig"
 
+BITWARDEN_FLAGS = DEBUG_MENU
```
KatherineInCode, line 4: 🤔  If we have it in Shared, it doesn't need to be in both Debug and Release, and potentially defeats the point of gating it?

Configs/Bitwarden-Release.xcconfig
```diff
@@ -5,6 +5,8 @@ CODE_SIGN_IDENTITY = Apple Development
 #include? "./Local.xcconfig"
 
 ASSETCATALOG_COMPILER_APPICON_NAME = $(APPICON_NAME)
+BITWARDEN_FLAGS = DEBUG_MENU
```
KatherineInCode, line 8: 🤔 Is the intention that this would be overridden with some other flags locally on some CI builds? How would we, ideally, know whether or not a build on CI should or should not have the debug menu at build time?

fedemkr (commented):
🤔 Are the changes needed in the CI/CD pipeline going to be in a different PR? So that we can modify `Configs/Base-Release.xcconfig` to introduce the flag for internal release builds.

phil-livefront:
> 🤔 Are the changes needed in the CI/CD pipeline going to be in a different PR? So that we can modify `Configs/Base-Release.xcconfig` to introduce the flag for internal release builds.

Right now the intention is to have builds from the Livefront CI/CD to update the builds with the flag so QA can begin testing the onboarding flow.  I wanted to have the structure in place when the Bitwarden CI/CD wants to implement this. Since beta builds go out to users we haven't come to a conclusion on how to handle that just yet but getting the builds with the menu will allow QA to test the new features. Does that make sense or am i being confusing?

KatherineInCode:
> 🤔 Are the changes needed in the CI/CD pipeline going to be in a different PR? So that we can modify `Configs/Base-Release.xcconfig` to introduce the flag for internal release builds.

We wouldn't want to modify that xcconfig file in the pipeline, though. We'd want to add something to `Local.xcconfig` after we set the variant, which [creates that file](https://github.com/bitwarden/ios/blob/main/Scripts/select_variant.sh#L42). We could also theoretically pass whether or not to do it at that point.

The real issue is that if we don't want this shipping to users, then we can't put it in any CI/CD pipeline builds unless we know for sure that particular build isn't going to be promoted, since the builds that "are uploaded to TestFlight" are the same ones we'd pull from to decide what to release.

In past apps I've worked on, we had a separate build variant (parallel to our prod variant) specifically for QA that had these sorts of debugging things turned on, because it's much easier to say "prod has debug menu off, qa has debug menu on". This solves the problem relatively nicely (though QA is then not necessarily actually testing the artifact that's going to be released to the public). However, because both our Prod variant and Beta variant go to users, we can't use them in this manner; we'd either have to stop releasing Beta to users or create a different variant that's only used internally by QA.

BitwardenShared/UI/Platform/DebugMenu/DebugMenuView.swift
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    /// The feature flags currently used in the app.
+    private var featureFlags: some View {
+        ForEach(store.state.featureFlags) { flag in
+            Toggle(
+                isOn: store.bindingAsync(
+                    get: { $0.isEnabled(for: flag.feature) },
+                    perform: { DebugMenuEffect.toggleFeatureFlag(flag.feature.rawValue, $0) }
```
matt-livefront, line 45: 🎨 Could this be simplified by getting the flag's `isEnabled` value from the flag itself?
```suggestion
            Toggle(
                isOn: store.bindingAsync(
                    get: { _ in flag.isEnabled },
                    perform: { DebugMenuEffect.toggleFeatureFlag(flag.feature.rawValue, $0) }
```

phil-livefront: nice!  much cleaner.

BitwardenShared/Core/Platform/Services/TestHelpers/MockConfigService.swift
```diff
@@ -8,9 +8,14 @@ class MockConfigService: ConfigService {
 
     var configMocker = InvocationMockerWithThrowingResult<(forceRefresh: Bool, isPreAuth: Bool), ServerConfig?>()
     var configSubject = CurrentValueSubject<BitwardenShared.MetaServerConfig?, Never>(nil)
+    var debugFeatureFlags = [DebugMenuFeatureFlag]()
+    var featureFlags = [FeatureFlag: AnyCodable]()
     var featureFlagsBool = [FeatureFlag: Bool]()
     var featureFlagsInt = [FeatureFlag: Int]()
     var featureFlagsString = [FeatureFlag: String]()
+    var getDebugFeatureFlagsCalled = false
```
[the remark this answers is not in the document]
phil-livefront: ah yeah, i gotta update the tests so this should be used.  good catch

BitwardenShared/UI/Platform/DebugMenu/DebugMenuState.swift
```diff
@@ -0,0 +1,15 @@
...
+/// The state used to present the `DebugMenuView`.
+///
+struct DebugMenuState: Equatable, Sendable {
+    /// The current feature flags supported.
+    var featureFlags: [DebugMenuFeatureFlag] = []
+
+    /// Computed property to get the isEnabled value for a given feature flag
+    func isEnabled(for feature: FeatureFlag) -> Bool {
+        featureFlags.first { $0.feature == feature }?.isEnabled ?? false
+    }
```
matt-livefront, line 14: 🎨 I think this isn't used anymore.

```suggestion
```

phil-livefront: good catch!

matt-livefront (approved):
Looks good!

PR #948, opened 2024-09-18
[PM-11894] update password generator policy

aj-rosado:
## 🎟️ Tracking
https://bitwarden.atlassian.net/browse/PM-11894

## 📔 Objective

Renaming property `defaultValue` to `overridePasswordType` as it is the name used on the other clients
Updating the behaviour so that it forces the user to use the `overridePasswordType` on the password generator

## 📸 Screenshots

https://github.com/user-attachments/assets/4f021f4d-fd13-45fb-be9e-f709cb54a117

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/BitwardenShared/Core/Tools/Models/Domain/PasswordGenerationOptions.swift
+++ b/BitwardenShared/Core/Tools/Models/Domain/PasswordGenerationOptions.swift
@@ -50,7 +50,7 @@ struct PasswordGenerationOptions: Codable, Equatable {
     var wordSeparator: String?
 
     /// Whether the password type should be enforced or not
-    var overridePasswordType: Bool = false
+    var overridePasswordType: Bool?
 }
 
 extension PasswordGenerationOptions {
--- a/BitwardenShared/Core/Tools/Models/Domain/PasswordGenerationOptions.swift
+++ b/BitwardenShared/Core/Tools/Models/Domain/PasswordGenerationOptions.swift
@@ -48,6 +48,9 @@ struct PasswordGenerationOptions: Codable, Equatable {
 
     /// The separator to put between words in the passphrase.
     var wordSeparator: String?
+
+    /// Whether the password type should be enforced or not
+    var overridePasswordType: Bool?
 }
 
 extension PasswordGenerationOptions {
--- a/BitwardenShared/Core/Tools/Models/Domain/PasswordGenerationOptions.swift
+++ b/BitwardenShared/Core/Tools/Models/Domain/PasswordGenerationOptions.swift
@@ -50,7 +50,7 @@ struct PasswordGenerationOptions: Codable, Equatable {
     var wordSeparator: String?
 
     /// Whether the password type should be enforced or not
-    var overridePasswordType: Bool?
+    var overridePasswordType: Bool = false
 }
 
 extension PasswordGenerationOptions {
--- a/BitwardenShared/Core/Vault/Models/Enum/PolicyOptionType.swift
+++ b/BitwardenShared/Core/Vault/Models/Enum/PolicyOptionType.swift
@@ -10,7 +10,7 @@ enum PolicyOptionType: String {
     case capitalize
 
     /// A policy option for the default type of the password generator.
-    case defaultType
+    case overridePasswordType
 
     /// A policy option for whether to include a number in a passphrase.
     case includeNumber
--- a/BitwardenShared/Core/Vault/Models/Enum/PolicyOptionType.swift
+++ b/BitwardenShared/Core/Vault/Models/Enum/PolicyOptionType.swift
@@ -9,7 +9,7 @@ enum PolicyOptionType: String {
     /// A policy option for whether to capitalize the passphrase words.
     case capitalize
 
-    /// A policy option for the default type of the password generator.
+    /// A policy option for the enforced type of the password generator.
     case overridePasswordType
 
     /// A policy option for whether to include a number in a passphrase.
--- a/BitwardenShared/Core/Vault/Services/PolicyService.swift
+++ b/BitwardenShared/Core/Vault/Services/PolicyService.swift
@@ -183,16 +183,14 @@ extension DefaultPolicyService {
         for policy in policies {
             if let overridePasswordTypeString = policy[.overridePasswordType]?.stringValue,
                let overridePasswordType = PasswordGeneratorType(rawValue: overridePasswordTypeString),
-               generatorType != .password,
-               options.overridePasswordType != true {
+               generatorType != .password {
                 // If there's multiple policies with different default types, the password type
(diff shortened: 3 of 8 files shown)


BitwardenShared/Core/Vault/Models/Enum/PolicyOptionType.swift
```diff
@@ -10,7 +10,7 @@ enum PolicyOptionType: String {
     case capitalize
 
     /// A policy option for the default type of the password generator.
-    case defaultType
+    case overridePasswordType
```
KatherineInCode, line 13: ⛏️ The documentation comment should probably be updated as well

BitwardenShared/Core/Vault/Services/PolicyService.swift
```diff
@@ -179,15 +179,19 @@ extension DefaultPolicyService {
...
+               let overridePasswordType = PasswordGeneratorType(rawValue: overridePasswordTypeString),
+               generatorType != .password,
+               options.overridePasswordType != true {
                 // If there's multiple policies with different default types, the password type
                 // should take priority. Use `generateType` as opposed to `options.type` to ignore
                 // the existing type in the options.
-                generatorType = defaultType
+                generatorType = overridePasswordType
             }
+            // if the stringValue is NOT emptu or null, overridePasswordType should be true
```
KatherineInCode, line 193: ⛏️ 
```suggestion
            // if the stringValue is NOT empty or nil, overridePasswordType should be true
```

BitwardenShared/Core/Tools/Models/Domain/PasswordGenerationOptions.swift
```diff
@@ -48,6 +48,9 @@ struct PasswordGenerationOptions: Codable, Equatable {
 
     /// The separator to put between words in the passphrase.
     var wordSeparator: String?
+
+    /// Whether the password type should be enforced or not
+    var overridePasswordType: Bool?
```
KatherineInCode, line 53: 🤔 Why is this boolean optional? What does a nil value represent as different from a false value?

BitwardenShared/Core/Vault/Services/PolicyService.swift
```diff
@@ -179,15 +179,19 @@ extension DefaultPolicyService {
...
+            options.overridePasswordType = !(policy[.overridePasswordType]?.stringValue?.isEmpty ?? true)
```
KatherineInCode, line 194: 🤔 Maybe I'm misunderstanding the logic, but doesn't putting this here mean that the `options.overridePasswordType != true` on line 187 will always be true for the first policy?

BitwardenShared/Core/Vault/Services/PolicyService.swift
david-livefront: Kotlin uses null not nil

BitwardenShared/Core/Vault/Models/Enum/PolicyOptionType.swift
aj-rosado: updated!

BitwardenShared/Core/Tools/Models/Domain/PasswordGenerationOptions.swift
aj-rosado: should never be nil actually, removed the optional

BitwardenShared/Core/Vault/Services/PolicyService.swift
aj-rosado: you are right I messed up here.
It would work for most of the scenario but would fail on some edge cases. Fixed it and changed tests in order to catch this.

BitwardenShared/Core/Vault/Services/PolicyService.swift
aj-rosado: this comment was removed

BitwardenShared/Core/Tools/Models/Domain/PasswordGenerationOptions.swift
aj-rosado: Actually, this is making the [test_decode_empty](https://github.com/bitwarden/ios/blob/235f340f670d02f279a2409ec9bf596f63f888a3/BitwardenShared/Core/Tools/Models/Domain/PasswordGenerationOptionsTests.swift#L54) on `PasswordGenerationOptionsTests` to fail.

I'm afraid it might throw an exception when calling the [decodePasswordOptions](https://github.com/bitwarden/ios/blob/235f340f670d02f279a2409ec9bf596f63f888a3/BitwardenShared/Core/Autofill/Utilities/ActionExtensionHelper.swift#L201) and there is no `overridePasswordType`.

BitwardenShared/UI/Tools/Generator/Generator/GeneratorProcessor.swift
```diff
@@ -348,9 +348,7 @@ final class GeneratorProcessor: StateProcessor<GeneratorState, GeneratorAction,
         var passwordOptions = state.passwordState.passwordGenerationOptions
         state.isPolicyInEffect = await (try? services.policyService
             .applyPasswordGenerationPolicy(options: &passwordOptions)) ?? false
-        // When updating the state, don't update the password generator type or it can override the
-        // user's current selection.
-        state.passwordState.update(with: passwordOptions, shouldUpdateGeneratorType: false)
+        state.passwordState.update(with: passwordOptions, shouldUpdateGeneratorType: true)
```
fedemkr, line 351: 🤔 Can't the scenario in the comment happen anymore? I mean isn't this line with `true` value already called in `loadGeneratorOptions()`? Is it necessary to also have it here?

aj-rosado: This new behaviour should enforce the user to use the `generatorType` from policies, so the scenario we were trying to avoid, now it is mandatory.

This was causing sync to not update the `generatorType`

fedemkr: Great makes sense thanks!

KatherineInCode (commented):
Seems reasonable so far, but we'll want to make a change so we don't have failing tests.

BitwardenShared/Core/Tools/Models/Domain/PasswordGenerationOptions.swift
matt-livefront: This model is persisted to UserDefaults to maintain the user's options between app runs. The optional bool allows decoding to succeed even if it doesn't exist in the previously persisted model.

fedemkr (approved):

PR #951, opened 2024-09-19
Fix issue where non-SSO users are forced into an SSO flow

KatherineInCode:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-12343

## 📔 Objective

Users who aren't SSO users are being forced into the SSO flow, particularly with the extension. This is likely because somehow their account in UserDefaults doesn't have user decryption options set. With this change, if we can't find those options, we assume the user has a master password, so we won't go into the SSO flow.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/BitwardenShared/UI/Auth/Extensions/AuthRouter+Redirects.swift
+++ b/BitwardenShared/UI/Auth/Extensions/AuthRouter+Redirects.swift
@@ -357,7 +357,7 @@ extension AuthRouter {
                     return .landingSoftLoggedOut(email: activeAccount.profile.email)
                 }
 
-                let hasMasterPassword = activeAccount.profile.userDecryptionOptions?.hasMasterPassword == true
+                let hasMasterPassword = activeAccount.profile.userDecryptionOptions?.hasMasterPassword ?? true
 
                 if !hasMasterPassword {
                     let biometricUnlockStatus = try await services.biometricsRepository.getBiometricUnlockStatus()
--- a/BitwardenShared/UI/Auth/Extensions/AuthRouter+Redirects.swift
+++ b/BitwardenShared/UI/Auth/Extensions/AuthRouter+Redirects.swift
@@ -328,7 +328,7 @@ extension AuthRouter {
     /// - Parameters:
     ///    - activeAccount: The active account.
     ///    - animated: If the suggested route can be animated, use this value.
-    ///    - shouldAttemptAutomaticBiometricUnlock: If the route uses automatic bioemtrics unlock,
+    ///    - shouldAttemptAutomaticBiometricUnlock: If the route uses automatic biometrics unlock,
     ///      this value enables or disables the feature.
     ///    - shouldAttemptAccountSwitch: Should the application automatically switch accounts for the user?
     /// - Returns: A suggested route for the active account with state pre-configured.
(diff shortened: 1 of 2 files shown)


BitwardenShared/UI/Auth/Extensions/AuthRouter+Redirects.swift
```diff
@@ -364,7 +364,7 @@ extension AuthRouter {
                     return .landingSoftLoggedOut(email: activeAccount.profile.email)
                 }
 
-                let hasMasterPassword = activeAccount.profile.userDecryptionOptions?.hasMasterPassword == true
+                let hasMasterPassword = activeAccount.profile.userDecryptionOptions?.hasMasterPassword ?? true
```
fedemkr, line 367: 🤔 Would this fall on a wrong behavior on a TDE user who's actually looking to go through this flow and somehow `userDecryptionOption` is `nil` as well? 
I mean AFAIK `userDecryptionOption` should never be `nil`; it was marked as optional only to provide support for old server versions if I'm not mistaken. So there must be something that is setting that to `nil` somehow. Maybe we can reach to server folks to confirm `userDecryptionOptions` is always sent in last 3 major versions of server both when logging in with whichever method the user chooses and also in the sync endpoints.

KatherineInCode: My current theory as to what happened is that somehow the `userDecryptionOptions` didn't properly carry forward through an app upgrade migration, and we don't fetch a new one until the user has signed in—but because of the SSO prompt, they can't sign in.

I dug through our code as thoroughly as I could, and if we ever set it to anything outside of what the server hands us or we pull off of disk, I couldn't find it. It might be worth checking with the server team, yeah.

Ultimately, though, I think there are significantly more MP users than SSO users, and so I'd rather default to the MP users and assume if we don't have that object then the user has a master massword—which is [what the state service does already](https://github.com/bitwarden/ios/blob/main/BitwardenShared/Core/Platform/Services/StateService.swift#L1383).

fedemkr: I'll approve but I believe we should check with server folks just to see if we need to make any additional changes or take into consideration that there may be occasions where `userDecryptionOptions` is sent with `nil` value.

fedemkr (approved):

BitwardenShared/UI/Auth/Extensions/AuthRouter+Redirects.swift
matt-livefront: I think the only time the app gets `userDecryptionOptions` is when the account is created or logged into. I don't think it's returned as part of sync or token refresh. So if it was `nil` when the account was created or last logged into (maybe possible for old accounts?), it would still be `nil` in user defaults.

matt-livefront (approved):

BitwardenShared/UI/Auth/Extensions/AuthRouter+Redirects.swift
trmartin4: @fedemkr I happened to see this since I was curious if TDE was involved in this issue.  The intent of the server code is that the `userDecryptionOptions` would always be populated.  It's initialized [here](https://github.com/bitwarden/server/blob/56600db011a0f3cb0b627bf02e8391195fa59be2/src/Identity/IdentityServer/UserDecryptionOptionsBuilder.cs#L25-L26) and populated with values depending on the user's account information.

The construct should be returned with all requests since TDE was introduced, back in `2023.8.2` if I remember correctly.

fedemkr: Great thanks for the info @trmartin4!!! That version is way more than 3 major server versions ago so we can be 99% sure, that's not causing the issue 😄

PR #969, opened 2024-09-24
[PM-10877] Use new version calculation in build workflow

vvolkgang:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-10877

## 📔 Objective

Implement the new version calculation in build.yml. While at it, added a step to output inputs.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


.github/workflows/build.yml
```diff
@@ -20,6 +20,13 @@ on:
       xcode-version:
         description: "Xcode Version Override - e.g. '15.2'"
         type: string
+      base_version_number:
+        description: "Base Version Number - Will be added to the calculated version number"
```
KatherineInCode, line 24: 🤔 Calculated version number, or calculated build number? (Something that somehow escaped my attention with the version workflow itself)

vvolkgang: Similar workflows in our repos have "version" or "version_name" already, I didn't want to break that pattern, and it got confusing really fast having version_name and build_number. Everything now starts with "version", followed by the specific property name, the equivalent to `version { name, number }`. 

Refrained from refactoring all of the other input names because we also need to align on `-` vs `_` use, thought that reviewing all of this at the same time would be a nightmare for you. Will go in it's own PR when everything else is sorted.

KatherineInCode: Hm, okay. So just to be clear, "version name" is what Xcode reports as "version", and "version number" is what Xcode reports as "build"? (`CFBundleShortVersionString` and `CFBundleVersion` respectively)

(Side note: I'm not fond of how Apple is so inconsistent with this. [agvtool[(https://developer.apple.com/library/archive/qa/qa1827/_index.html) calls them "marketing" and "internal" versions, and [the actual spec](https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleversion) indicates that `CFBundleVersion` should also be a three-integers-separated-by-periods field even though none of their tooling seems to support this?)

KatherineInCode (dismissed):

.github/workflows/build.yml
vvolkgang: That's it! Version.Name is the string we use to communicate Product versions (what agvtool calls "marketing"). TestFlight doesn't allow this but in other platforms we can use it with additional info for internal builds like branch names, e.g. `2024.9.2-feature1`. Version.Number is an int, an internal value that most other platforms use to compare builds. Apple happens to be the only one that allows this to be reset when Version.Name changes.

Mapping for mobile:

* Version.Name - iOS: `CFBundleShortVersionString` Android: `versionName`
* Version.Number - iOS: `CFBundleVersion` Android: `versionCode`

KatherineInCode (approved):

vvolkgang:
Successful test run after dealing with GitHub slow runners yesterday: https://github.com/bitwarden/ios/actions/runs/11134836990

.github/actions/dispatch-and-download/action.yml
```diff
@@ -27,8 +27,9 @@ runs:
   using: "composite"
   steps:
     - name: Log inputs to job summary
+      shell: bash
```
KatherineInCode, line 30: 🤔 It normally executes in `zsh` and that causes an issue with the steps? Or just being explicit to be safe and consistent with other steps?

KatherineInCode (approved):

.github/actions/dispatch-and-download/action.yml
vvolkgang: Actions require a shell to be specified either inside of it or by the caller. Went with defining this inside to reduce usage friction and ensure that we don't get weird errors due to different shells. 

No good reason for bash vs other options tho, just my (ignorant) default.

KatherineInCode: To be clear, I'm 100% fine with bash 😄 just didn't know it needed to be specified. Cool beans! 👍🏻

PR #977, opened 2024-09-26
[BITAU-148] [BITAU-154] Add Sync Service to the PM app

brant-livefront:
## 🎟️ Tracking

- [BITAU-148](https://livefront.atlassian.net/browse/BITAU-148) - Add Sync Service
- [BITAU-154](https://livefront.atlassian.net/browse/BITAU-154) - Monitor Cipher Updates and Write to Store

## 📔 Objective

Up until now, I've been building out the capabilities inside of `AuthenticatorBridgeKit`, but neither of the apps was actually using any of them. This PR adds the `AuthenticatorSyncService` which uses the pieces from `AuthenticatorBridgeKit` to listen for updates and write items to the shared core data store.  `AuthenticatorSyncService` was modeled on the work @victor-livefront did in the POC and the Watch connection done previously.

*Note: This code only handles things when the vault is unlocked and the app is in the foreground. There's a lot more edge cases that the POC explored, but I want to make sure we're all aligned on this approach and then I'll start adding pieces one by one to handle the edge cases. For example, there's nothing that cleans up the key/data when a user turns sync off., or handles running in the background, etc.*

There are two things that stop this from firing in the main app for now:
1. The `enableAuthenticatorSync` feature flag
   - The service checks this on `start()` and does nothing if the flag is off. It just simply returns without listening for anything.
2. The new Setting - `syncToAuthenticator` - must be turned on for an account to start listening to updates.

Both of these operate to make sure we don't release this feature to users until we're all ready. Even if the feature flag (which is not yet remote-enabled) were flipped to true, we don't yet have a way in the UI to turn on the sync setting.

If you want to override both of those so that you can see this in action, the easiest way is to override the `start()` method in the new `AuthenticatorSyncService`:
```Swift
     try await stateService.setSyncToAuthenticator(true)
     if await configService.getFeatureFlag(FeatureFlag.enableAuthenticatorSync,
                                           defaultValue: true) {
         subscribeToAppState()
     }

```
Then run the application, log in and unlock the vault and then move it to the background and back to the foreground. You should see the messages in the log indicating that everything is being picked up and sync'd.

### Notes on re-used code
I know we discussed this previously and our strategy has been to move things into `AuthenticatorBridgeKit` as duplicates so that we can clearly see what's being reused (and migrate to a shared library in the future). I just wanted to call out a few more files that I'm pulling over as part of this PR:
- FetchedResultsPublisher.swift
- Future+Extensions.swift
- Publisher+Async.swift
   - as well as it's tests:  PublisherAsyncTests.swift
  

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


brant-livefront:
Thanks to [Codecov](https://github.com/bitwarden/ios/pull/977#issuecomment-2377374497) I noticed that I forgot to add coverage for the new method on `SharedCryptographyService`. Latest commit fixes that 👍

AuthenticatorBridgeKit/SharedCryptographyService.swift
```diff
@@ -54,7 +69,26 @@ public class DefaultAuthenticatorCryptographyService: SharedCryptographyService
...
+
+            return AuthenticatorBridgeItemDataView(
+                favorite: item.favorite,
+                id: item.id,
+                name: (try? decrypt(item.name, withKey: symmetricKey)) ?? "",
+                totpKey: try? decrypt(item.totpKey, withKey: symmetricKey),
+                username: try? decrypt(item.username, withKey: symmetricKey)
+            )
+        }
+    }
```
fedemkr, line 89: 🤔 I think this function is not needed and also is duplicating decryption logic from the model. And the `decryptAuthenticatorItemModels` can be used instead with the proposed change in `sharedItemsPublisher`

AuthenticatorBridgeKit/AuthenticatorBridgeItemService.swift
```diff
@@ -124,4 +144,18 @@ public class DefaultAuthenticatorBridgeItemService: AuthenticatorBridgeItemServi
...
+        let fetchRequest = AuthenticatorBridgeItemData.fetchRequest()
+        fetchRequest.sortDescriptors = [NSSortDescriptor(keyPath: \AuthenticatorBridgeItemData.userId, ascending: true)]
+        return FetchedResultsPublisher(
+            context: dataStore.persistentContainer.viewContext,
+            request: fetchRequest
+        )
+        .asyncTryMap { items in
+            try await self.cryptoService.decryptAuthenticatorItemDatas(items)
+        }
+        .eraseToAnyPublisher()
```
fedemkr, line 159: 🤔 I think this can be improved to use `decryptAuthenticatorItemModels` by chaining the model mapping:
```suggestion
        .tryMap { dataItems in 
            return dataItems.compactMap { data in 
                guard let model = itemData.model else { 
                    return nil 
                }
                return model
            }
        }
        .asyncTryMap { itemModel in
            try await self.cryptoService.decryptAuthenticatorItemModels(items)
        }
        .eraseToAnyPublisher()
```

BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
```diff
@@ -0,0 +1,247 @@
+import AuthenticatorBridgeKit
+import BitwardenSdk
+import CryptoKit
+import Foundation
+import OSLog
+import UIKit
+
+// MARK: - AuthenticatorSyncService
+
+/// The service used to share TOTP codes to and from the Authenticator app..
```
fedemkr, line 10: ⛏️ 
```suggestion
/// The service used to share TOTP codes to and from the Authenticator app.
```

BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
```diff
@@ -0,0 +1,247 @@
...
+
+        return ciphersToUse.map { cipher in
+            AuthenticatorBridgeItemDataView(
+                favorite: false,
+                id: cipher.id ?? UUID().uuidString,
+                name: cipher.name,
+                totpKey: cipher.login?.totp,
+                username: username
+            )
+        }
```
fedemkr, line 154: 🎨 One can do the filtering before decrypting the ciphers to improve performance as we don't want all ciphers decrypted but only the ones who comply with the filter. This reduces the number of calls to the SDK and also the waiting time on decrypting unneeded ciphers.
Moreover, we can use the filter before decrypting because we don't need any decrypted value to perform the filter correctly:
```suggestion
    private func decryptTOTPs(_ ciphers: [Cipher],
                              userId: String) async throws -> [AuthenticatorBridgeItemDataView] {
        let totpCiphers = ciphers.filter { cipher in
            cipher.deletedDate == nil
                && cipher.type == .login
                && cipher.login?.totp != nil
        }
        let decryptedCiphers = try await totpCiphers.asyncMap { cipher in
            try await self.clientService.vault().ciphers().decrypt(cipher: cipher)
        }
        let account = try await stateService.getActiveAccount()
        let username = account.profile.name ?? account.profile.email

        return decryptedCiphers.map { cipher in
            AuthenticatorBridgeItemDataView(
                favorite: false,
                id: cipher.id ?? UUID().uuidString,
                name: cipher.name,
                totpKey: cipher.login?.totp,
                username: username
            )
        }
```

BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
```diff
@@ -0,0 +1,247 @@
...
+
+    /// This function handles the initial syncing with the Authenticator app as well as listening for updates
+    /// when the user adds new items. This is called when the sync is turned on.
+    ///
+    /// - Parameter userId: The userId of the user who has turned on sync.
+    ///
+    private func handleSyncOnForUserId(_ userId: String) async {
+        Logger.application.log("#### sync is on for userId: \(userId)")
+
+        if !vaultTimeoutService.isLocked(userId: userId) {
```
fedemkr, line 165: ⛏️ Could we use `guard` to reduce nesting?

BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
```diff
@@ -0,0 +1,247 @@
...
+        cipherPublisherTasks[userId]??.cancel()
+        cipherPublisherTasks[userId] = nil
+    }
+
+    /// Subscribe to NotificationCenter updates about if the app is in the foreground vs. background.
+    ///
+    private func subscribeToAppState() {
+        Task {
+            for await _ in notificationCenterService.willEnterForegroundPublisher() {
+                Logger.application.log("#### app entered foreground")
```
fedemkr, line 193: 🤔 Do we want the app to log this on PRD? I see its value on debugging but I wouldn't leave this for PRD.

AuthenticatorBridgeKit/SharedCryptographyService.swift
brant-livefront: 👍 Great suggestion. Fixed in latest.

AuthenticatorBridgeKit/AuthenticatorBridgeItemService.swift
brant-livefront: 👍 Great suggestion. I like this a lot better! I had been playing around with different ways of doing this, but for some reason didn't land on this one. Reduces duplication and cleans up the CryptoService's responsibilities 💯 

Adopted this in the latest.

BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
brant-livefront: 👍 Good note. Fixed in latest.

BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
brant-livefront: 👍 Fixed in latest.

BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
brant-livefront: I was debating pulling them out entirely once we are close to done. They are really useful for now in debugging stuff though so I've moved them over to `.debug` in the latest. 👍

AuthenticatorBridgeKit/AuthenticatorBridgeItemService.swift
```diff
@@ -124,4 +144,21 @@ public class DefaultAuthenticatorBridgeItemService: AuthenticatorBridgeItemServi
...
+        let fetchRequest = AuthenticatorBridgeItemData.fetchRequest()
+        fetchRequest.sortDescriptors = [NSSortDescriptor(keyPath: \AuthenticatorBridgeItemData.userId, ascending: true)]
+        return FetchedResultsPublisher(
+            context: dataStore.persistentContainer.viewContext,
+            request: fetchRequest
+        )
+        .tryMap { dataItems in
+            dataItems.compactMap(\.model)
+        }
+        .asyncTryMap { itemModela in
```
KatherineInCode, line 159: 🎨 
```suggestion
        .asyncTryMap { itemModel in
```

AuthenticatorBridgeKit/FetchedResultsPublisher.swift
```diff
@@ -0,0 +1,145 @@
...
+import CoreData
+
+// MARK: - FetchedResultsPublisher
+
+/// A Combine publisher that publishes the initial result set and any future data changes for a
+/// Core Data fetch request.
+///
+/// Adapted from https://gist.github.com/darrarski/28d2f5a28ef2c5669d199069c30d3d52
+///
+class FetchedResultsPublisher<ResultType>: Publisher where ResultType: NSFetchRequestResult {
```
KatherineInCode, line 11: 🤔 Does it make sense to have tests wrapping this, just in case?

AuthenticatorBridgeKit/Future+Extensions.swift
```diff
@@ -0,0 +1,21 @@
+import Combine
+
+extension Future {
+    /// Initialize a `Future` with an async throwing closure.
+    ///
+    /// - Parameter attemptToFulfill: A closure that the publisher invokes when it emits a value or
+    ///     an error occurs.
+    ///
+    convenience init(_ attemptToFulfill: @Sendable @escaping () async throws -> Output) where Failure == Error {
```
KatherineInCode, line 9: 🤔 Can we test this directly, instead of just covering it through other places as well? Would it make sense to do that?

BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
```diff
@@ -0,0 +1,248 @@
...
+        }
+    }
+
+    /// This function handles the initial syncing with the Authenticator app as well as listening for updates
+    /// when the user adds new items. This is called when the sync is turned on.
+    ///
+    /// - Parameter userId: The userId of the user who has turned on sync.
+    ///
+    private func handleSyncOnForUserId(_ userId: String) async {
+        Logger.application.debug("#### sync is on for userId: \(userId)")
```
KatherineInCode, line 162: 🎨  We probably don't want log statements getting merged in, especially debug ones

AuthenticatorBridgeKit/FetchedResultsPublisher.swift
brant-livefront: This was just a pickup form BitwardenShared and I didn't see any (direct) tests there. I can try adding some if we think it should be tested directly. It does seem to be tested as part of other tests however.

AuthenticatorBridgeKit/Future+Extensions.swift
brant-livefront: Same as above. I don't see any tests currently in BitwardenShared for this, but I'm open if we think it needs direct tests.

BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
brant-livefront: Should I just pull them all out? I'm open to that. They are nice in debugging this, but we're getting much closer to just seeing the actual results directly (and I can add them in as needed while debugging).

AuthenticatorBridgeKit/FetchedResultsPublisher.swift
KatherineInCode: We're probably good, just wanted to double-check.

AuthenticatorBridgeKit/Future+Extensions.swift
KatherineInCode: Same as above, we're probably good, just wanted to double-check.

AuthenticatorBridgeKit/AuthenticatorBridgeItemService.swift
brant-livefront: fixed in latest 👍

BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
fedemkr: 🤔 Which would be the downsides of having these debug logs in `main`? I mean end users won't see that, only debug builds will use that.
I often go by the rule: if the debug log will help debugging the feature at a later stage or because of some bug reported then I'd leave that and merge it but if it's something trivial or just to see where the flow goes I'd remove it from the commit.

brant-livefront: That's a reasonable line and I think these fall outside of it - they are useful for debugging for now, but I don't see them as being useful later when the whole system is up and running. I've pulled them all the way out in the latest commit. 👍

BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
```diff
@@ -0,0 +1,237 @@
...
+        syncSettingSubscriberTask = Task {
+            for await (userId, shouldSync) in await self.stateService.syncToAuthenticatorPublisher().values {
+                guard let userId else { continue }
+
+                if shouldSync {
+                    await handleSyncOnForUserId(userId)
+                } else {
+                    handleSyncOffForUserId(userId)
+                }
+            }
```
fedemkr, line 223: 🤔 I think we should take into consideration whether the vault is `locked` / `unlocked` to call the handle methods. I see this only happening inside the `handleSyncOn...` func but what would that work correctly in the next scenario?
1. The active user id is unlocked and should sync so `handleSyncOnForUserId` is called and it subscribes to cipher updates.
2. Then the active user id gets locked but as far as I see the cipher updates are not unsubscribed possibly introducing race conditions errors depending if the lock happens while decrypting totps or replacing the items.

IMO if the vault locks/logsout then it should cancel any running operations to avoid any unwanted behaviors.
What do you think?

BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
```diff
@@ -0,0 +1,237 @@
...
+    // MARK: Public Methods
+
+    public func start() {
+        Task {
+            if await configService.getFeatureFlag(FeatureFlag.enableAuthenticatorSync,
+                                                  defaultValue: false) {
+                subscribeToAppState()
+            }
+        }
+    }
```
fedemkr, line 106: 🤔 Should a check be added here to avoid `start` being called multiple times thus subscribing multiple times to  app state? Additionally, should this service be an `actor` to prevent multiple thread trying to call `start` concurrently?

BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
brant-livefront: My intention with this PR was just to do the first minimalist pass, but Vault unlocking and background stuff is definitely on my mind. 👍  I'm working on that in a separate branch, I just wanted to check in with this PR first on the overall architecture of adding this service, proving out the basic sync, etc. As such, I was just bombing out if the Vault was unlocked. For this PR, you had to have an unlocked vault when you turn sync on, or background/foreground the app with sync on and vault unlocked for anything to work.

My current thinking is the best way is to listen to two publishers: the sync setting AND the vault lock/unlock. If we get a vault lock/unlock or if the user turns on/off sync, we need to re-evaluate the sync in light of both and clean up accordingly (e.g. the point you mentioned about the vault locking _while_ we were decrypting ciphers is an excellent one). What do you think about that approach?

BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
brant-livefront: The original goal of adding a start() method was to get async out the `ServiceContainer` (because we have to `await` the configService call). So `ServiceContainer` could just call `start()` and move on and then the service would start up asynchronously in the background. Making it an actor means that it implicitly needs to become async and thus the `ServiceContainer` needs to wrap this in a `Task` to let it run without blocking.

However, now that you've mentioned it, I think that's probably the safest way to go. We're relying on the compiler instead of just our own code conventions of not calling it multiple times. And without making it an `actor`, I think it's hard to guarantee that our check for !started is doing what we think it's doing.

I've pushed up a commit that moves this over to an `actor` and adds a check to see if it's started 👍

BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
fedemkr: Oh excellent then and I agree that's a nice approach to take 😄 Thanks for clarifying.

fedemkr (approved):
Looks good 🎉

PR #972, opened 2024-09-25
PM-12565: Add settings badges for account setup steps

matt-livefront:
## 🎟️ Tracking

[PM-12563](https://bitwarden.atlassian.net/browse/PM-12563) - Add badge indicator to bottom navigation for missing autofill setup
[PM-12564](https://bitwarden.atlassian.net/browse/PM-12564) - Add badge indicator to bottom nav for missing unlock method setup
[PM-12565](https://bitwarden.atlassian.net/browse/PM-12565) - Implement badge count logic

## 📔 Objective

Adds a badge to the settings tab indicating how many account setup steps remain. These will be added if the user chooses to set up autofill or vault unlock at a later point from the onboarding flow. Once they enable autofill or a vault unlock method, the corresponding badge will be removed.

## 📸 Screenshots

Screenshot 2024-09-25 at 10 31 53 AM

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

[PM-12563]: https://bitwarden.atlassian.net/browse/PM-12563?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[...]

PR #988, opened 2024-09-30
PM-12565: Don't fail login if setting account setup progress fails

matt-livefront:
## 🎟️ Tracking

[PM-12565](https://bitwarden.atlassian.net/browse/PM-12565)

## 📔 Objective

This addresses a comment @fedemkr made here: https://github.com/bitwarden/ios/pull/972/files#r1777654377

If setting the user's account setup progress fails as part of login, we should log the error rather than throw which would cause the login to fail.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-12565]: https://bitwarden.atlassian.net/browse/PM-12565?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/Core/Platform/Services/StateService.swift
+++ b/BitwardenShared/Core/Platform/Services/StateService.swift
@@ -58,14 +58,14 @@ protocol StateService: AnyObject {
     /// - Parameter userId: The user ID associated with the autofill setup progress.
     /// - Returns: The user's autofill setup progress.
     ///
-    func getAccountSetupAutofill(userId: String?) async throws -> AccountSetupProgress?
+    func getAccountSetupAutofill(userId: String) async -> AccountSetupProgress?
 
     /// Gets the user's progress for setting up vault unlock.
     ///
     /// - Parameter userId: The user ID associated with the vault unlock setup progress.
     /// - Returns: The user's vault unlock setup progress.
     ///
-    func getAccountSetupVaultUnlock(userId: String?) async throws -> AccountSetupProgress?
+    func getAccountSetupVaultUnlock(userId: String) async -> AccountSetupProgress?
 
     /// Gets all accounts.
     ///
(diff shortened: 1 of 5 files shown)


BitwardenShared/Core/Platform/Services/StateService.swift
```diff
@@ -58,14 +58,14 @@ protocol StateService: AnyObject {
     /// - Parameter userId: The user ID associated with the autofill setup progress.
     /// - Returns: The user's autofill setup progress.
     ///
-    func getAccountSetupAutofill(userId: String?) async throws -> AccountSetupProgress?
+    func getAccountSetupAutofill(userId: String) async -> AccountSetupProgress?
```
matt-livefront, line 61: This is a slight change to the pattern that exists in `StateService`.

Our current pattern is that the protocol defines a throwing function which takes an optional `userId`. If the user ID parameter is `nil`, then we get the active user ID. And then we have an extension method which doesn't accept any parameters to essentially provide a default value of `nil` as the user ID to get the active user ID.

If we instead move the "default to active user ID" logic into the extension method, we could make the protocol method require a user ID and in most cases make it no throw. 

This won't affect most places where we call the extension method without a user ID. But in cases where we already have the user ID (some services/repositories methods), we could avoid throwing. If we like this pattern, I can update the rest of StateService to match.

fedemkr: Awesome! I really like the approach but I wouldn't perform the changes right now. We're working on a proposal to reduce most of the boilerplate in `StateService` and `AppSettingsStore`, so reducing the effort to maintain the code and tests. So I'll add this there and share the proposal with you as well to revise it if you agree.

fedemkr (approved):

BitwardenShared/Core/Platform/Services/StateService.swift
matt-livefront: Sounds good, thanks!

PR #995, opened 2024-10-02
PM-13104: Remove biometric integrity checks

matt-livefront:
## 🎟️ Tracking

[PM-13104](https://bitwarden.atlassian.net/browse/PM-13104)

## 📔 Objective

Removes biometric integrity checks. The downside of these checks is that they are unique per-process. So even if biometrics were enabled in the app, opening the autofill extension would require the user to enter their MP or PIN before biometrics would be enabled in that extension. The biometrics auth key already uses [biometryCurrentSet](https://developer.apple.com/documentation/security/secaccesscontrolcreateflags/biometrycurrentset) which will invalid the keychain item if a finger or face is added for Touch/Face ID. Without these checks, enabling biometrics in the app will cause it to work in any of the extensions immediately vs after the first successful unlock with MP or PIN.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-13104]: https://bitwarden.atlassian.net/browse/PM-13104?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenShared/UI/Auth/Extensions/AuthRouter+Redirects.swift
```diff
@@ -363,14 +361,6 @@ extension AuthRouter {
                     return .landingSoftLoggedOut(email: activeAccount.profile.email)
                 }
 
-                let hasMasterPassword = activeAccount.profile.userDecryptionOptions?.hasMasterPassword ?? true
-
-                if !hasMasterPassword {
-                    let biometricUnlockStatus = try await services.biometricsRepository.getBiometricUnlockStatus()
-                    if case .available(_, true, false) = biometricUnlockStatus {
-                        return .enterpriseSingleSignOn(email: activeAccount.profile.email)
```
KatherineInCode, line 371: 🤔 We're good with not navigating to the enterprise SSO screen here because if the biometrics fails (because the user changed their settings), it will drop to the login screen, which will give them an option to re-login with SSO?

BitwardenShared/Core/Platform/Services/MigrationService.swift
```diff
@@ -178,39 +171,38 @@ class DefaultMigrationService {
...
+        appGroupUserDefaults.removeObject(forKey: "bwPreferencesStorage:iOSExtensionBiometricIntegritySource")
+        appGroupUserDefaults.removeObject(forKey: "bwPreferencesStorage:iOSShareExtensionBiometricIntegritySource")
+    }
 
+    /// Performs migration 4.
+    ///
+    /// Notes:
+    ///   - Removes the integrity state values.
+    ///
+    private func performMigration4() async throws {
```
KatherineInCode, line 189: 🤔 Thinking through the situation where we might need to revert this change and re-add the integrity in the future, such as if there's some other implication that we're not realizing right now that means we need it back. This would mean that a user using the app with re-added integrity would just need to turn Face ID (etc.) off and then back on again with both the app and the extension(s) to get it working again, correct?

That's not great, but doesn't seem too particularly onerous, I just want to make sure there's nothing I'm missing.

KatherineInCode (commented):
Looks good to me, just a few things I want to double-check before approving.

BitwardenShared/Core/Platform/Services/MigrationService.swift
matt-livefront: Yeah, either turn it off and back on, or we could potentially get the integrity state value on the next successful Face ID scan and persist that.

BitwardenShared/UI/Auth/Extensions/AuthRouter+Redirects.swift
matt-livefront: Correct. That's a good point though, in the case of a biometrics failure (and if you don't have a MP or PIN), I think you would still remain on the vault unlock screen. We don't have logic that would log out you in that case. Should I add that?

fedemkr: Can't we just show the unlock with biometrics button in the screen when the user doesn't have MP and has biometrics configured? So they can try again and if they fail 5 times then it'd go through the logic of automatic logout because of reiterative failed attempts.

BitwardenShared/Core/Platform/Services/MigrationService.swift
KatherineInCode: Seems reasonable. Not necessarily super thrilling to tell all our users to do that, but at least doable.

BitwardenShared/UI/Auth/Extensions/AuthRouter+Redirects.swift
matt-livefront: > Can't we just show the unlock with biometrics button in the screen when the user doesn't have MP and has biometrics configured? So they can try again and if they fail 5 times then it'd go through the logic of automatic logout because of reiterative failed attempts.

Once biometrics fails because the key doesn't exist, we disable biometrics and remove the biometrics button. iOS also doesn't prompt you for Face ID in this case since the key has been removed. So even if we keep the biometrics button visible, you would tap it and it would look like nothing happened. Until you eventually get logged out, so we might as well just logout from the start.

matt-livefront: I added a change to log the user out if biometrics fails because the key doesn't exist, and the user doesn't have a MP or PIN. 4698b86592adddeca142e9e9b2b0a97a0794ab89

BitwardenShared/UI/Auth/VaultUnlock/VaultUnlockProcessor.swift
```diff
@@ -253,7 +253,17 @@ class VaultUnlockProcessor: StateProcessor<
...
+            let hasMasterPassword = try? await services.authRepository.hasMasterPassword()
+            let isPinEnabled = try? await services.authRepository.isPinUnlockAvailable()
+            if hasMasterPassword == false, isPinEnabled == false {
+                // If biometrics is enabled, but the auth key doesn't exist and the user doesn't
+                // have a master password or PIN, log the user out.
+                await logoutUser(userInitiated: false)
+            } else {
+                // Otherwise, refresh the data to remove biometrics as an unlock method.
+                await loadData()
+            }
```
fedemkr, line 266: 🤔 If `hasMasterPassword` is `nil` because of throwing shouldn't we logout the user as well? Given that either there are no accounts or there are no active accounts; so I guess on these cases it makes sense to logout as well.

KatherineInCode (dismissed):

KatherineInCode (approved):

fedemkr (approved):

PR #998, opened 2024-10-03
[BITAU-152] Handle Vault Lock/Unlock with Sync

brant-livefront:
## 🎟️ Tracking

[BITAU-152](https://livefront.atlassian.net/browse/BITAU-152)

## 📔 Objective

As I mentioned in [the previous PR](https://github.com/bitwarden/ios/pull/977#discussion_r1782863340), I needed to add explicit handling of vault lock/unlock status when performing sync. This PR adds subscribers for both the vault status as well as the sync setting. When either is updated (for any account), we examine if that account is ready for sync or not. To be ready for sync, the user must have opted in to sync _and_ they must have an unlocked vault. If either of those are missing, we dismantle any cipher listening and wait for another signal. I've also added a lot of tests to cover all the various scenarios with multiple accounts.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
```diff
@@ -102,9 +106,33 @@ actor DefaultAuthenticatorSyncService: NSObject, AuthenticatorSyncService {
...
-                                              defaultValue: false) {
-            subscribeToAppState()
+
+        guard await configService.getFeatureFlag(FeatureFlag.enableAuthenticatorSync,
+                                                 defaultValue: false) else {
+            return
+        }
+
+        syncSettingSubscriberTask = Task {
+            for await (userId, _) in await self.stateService.syncToAuthenticatorPublisher().values {
```
fedemkr, line 116: 🤔 Isn't `shouldSync` (2nd parameter) needed anymore?

BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
```diff
@@ -55,6 +55,10 @@ actor DefaultAuthenticatorSyncService: NSObject, AuthenticatorSyncService {
     /// a user opts-in to Authenticator sync.
     private var syncSettingSubscriberTask: Task<Void, Error>?
 
+    /// a Task that subscribes to the vault lock publisher for accounts. This allows us to take action once
+    /// a user unlocks their vault..
+    private var vaultUnlockSubscriberTask: Task<Void, Error>?
+
```
fedemkr, line 61: 🤔 Are these tasks going to be used for something else in a future PR? Because on the current code we could avoid having these variables.

BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
```diff
@@ -102,9 +106,33 @@ actor DefaultAuthenticatorSyncService: NSObject, AuthenticatorSyncService {
...
+            for await vaultStatus in await self.vaultTimeoutService.vaultLockStatusPublisher().values {
+                guard let vaultStatus else { continue }
+
+                do {
+                    try await determineSyncForUserId(vaultStatus.userId)
+                } catch {
+                    errorReporter.log(error: error)
+                }
+            }
         }
```
fedemkr, line 136: 🤔 Would it make sense to use just one `Task` with both publishers being combined using something like `Publishers.CombineLatest`?

BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
brant-livefront: It's not needed currently because we re-check that in `determineSyncForUserId(userId)`. The key piece of info we're getting from this subscription is that this user's sync setting has been _updated_ and therefore we need to re-evaluate. Since we re-evaluate based on _both_ vault and sync status, we don't need to pass this in here and we can just look it up.

That said, I suppose we could optionally pass this in to save ourselves that extra lookup. But it might make things a bit complicated in `determineSyncForUserId(userId)` (because you could be landing there with a vault update instead, where we would have to look this one up)

BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
brant-livefront: I don't think I have a clear vision for them in a future PR. You're suggesting just eliminating the `private var` reference to them and just declare them inline as Tasks without a reference? I think that should be fine. 🤔  I'll take a look at that.

BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
brant-livefront: I was thinking of going that route, but that puts the two into a tuple. The problem there is we could have updates for multiple different accounts, which then would be tupled together and we wouldn't be able to know which of the two is the actionable side of that. It also seems weird to have the syncing for `account1` tupled together with the vault status of `account2`.

If you can think of a way to handle this, I'd definitely be open. I'm less well-versed on combine. But the end result of both of these paths is just calling `determineSyncForUserId(userId)` and handling errors. So I'd rather have them all handled in one task if we can find a way.

brant-livefront: Perhaps using `merge` if we first translated them both to just publishing a `String?` of their username?

BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
fedemkr: Yep, indeed. If we're not going to handle cancellation or something else with this task reference then we could remove these private var references.

brant-livefront: Update to this in the latest commit. 👍

fedemkr (approved):
Looks good on my side! However I'd wait for @matt-livefront and @KatherineInCode to review as well in case I've missed something.

BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
```diff
@@ -102,9 +98,33 @@ actor DefaultAuthenticatorSyncService: NSObject, AuthenticatorSyncService {
...
+                    errorReporter.log(error: error)
+                }
+            }
+        }
+        Task {
+            for await vaultStatus in await self.vaultTimeoutService.vaultLockStatusPublisher().values {
+                guard let vaultStatus else { continue }
+
+                do {
+                    try await determineSyncForUserId(vaultStatus.userId)
```
matt-livefront, line 123: 🤔 I think `vaultLockStatusPublisher()` only publishes the vault lock status for the active account. Does that cause any problems if an inactive account's vault is locked when it times out?

- Unlock vault for account A
- Switch to account B
- Some time later account A's vault times out and locks
- Account A's vault updates in the background, trigging a sync. Decrypting these ciphers would then fail.

We might need a publisher for all accounts for this?

BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
```diff
@@ -154,43 +174,24 @@ actor DefaultAuthenticatorSyncService: NSObject, AuthenticatorSyncService {
...
     ///
-    /// - Parameter userId: The userId of the user who has turned on sync.
+    /// - Parameter userId: The userId of the user whose sync status is being determined.
     ///
-    private func handleSyncOnForUserId(_ userId: String) async {
-        guard !vaultTimeoutService.isLocked(userId: userId) else {
+    private func determineSyncForUserId(_ userId: String) async throws {
+        guard try await stateService.getSyncToAuthenticator(userId: userId),
+              !vaultTimeoutService.isLocked(userId: userId) else {
+            cipherPublisherTasks[userId]??.cancel()
```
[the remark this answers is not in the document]
brant-livefront: They are functionally equivalent (setting `nil` deletes the value and key). But now that you mention it, I think it might be better practice to call `removeValue(forKey:)`. I moved this over to that in the latest commit 👍

BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
brant-livefront: I adjusted this for now like we discussed on Friday - added a guard to check if that particular Vault got locked in the background where we wouldn't get the notification. I think this should handle the situation for now and then I can look at if we need the larger all-accounts publisher (I don't think we will)

matt-livefront (dismissed):
Looks good to me!

BitwardenShared/Core/Platform/Services/AuthenticatorSyncServiceTests.swift
```diff
@@ -455,6 +455,31 @@ final class AuthenticatorSyncServiceTests: BitwardenTestCase { // swiftlint:disa
...
 
+    /// The AuthService may not get notified about Vault locking if the user has switched accounts. Verify
+    /// that it checks the vault lock before beginning to decrypt and write new Ciphers received.
+    ///
+    @MainActor
+    func test_writeCiphers_vaultLocked() async throws {
+        setupInitialState()
+        await subject.start()
+        stateService.syncToAuthenticatorSubject.send(("1", true))
+        try await Task.sleep(nanoseconds: 10_000_000)
```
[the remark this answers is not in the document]
brant-livefront: Yeah, the sleeps that are in these tests are because there isn't an explicit thing to wait _for_. We need to give the service time to do things like subscribe to Ciphers or something else that happens out of the view of the tests. So we're waiting to avoid testing the wrong thing or false positives. 

In this case, not waiting can cause us to set the vault to locked before the `syncToAuthenticatorSubject.send` kicks off the `determineSyncForUserId` method. That would mean we look at the vault status and (correctly) don't set anything up. But in this case, we actually want to test that _after_ we've subscribed to cipher updates, we get the vault locked without a notification. So we have to wait until that subscription happens but there's not an external thing the tests can wait for to know it happens. I actually tested this one using code coverage and breakpoints to see that when I wrote it without any sleep, it never gets into the guard I'm trying to verify.

BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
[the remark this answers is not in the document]
brant-livefront: Oh, gotcha. Good catch. 👍 I added a commit with that.

matt-livefront (approved):

PR #1004, opened 2024-10-03
[BITAU-174] [BITAU-151] Remove shared items and key when a user disables sync

brant-livefront:
## 🎟️ Tracking

- [BITAU-151](https://livefront.atlassian.net/browse/BITAU-151) Delete the shared Authenticator key when sync is turned off for all accounts.
- [BITAU-174](https://livefront.atlassian.net/browse/BITAU-174) Delete all shared items when sync is turned off for an account.

## 📔 Objective

This PR cleans up the DataStore and keys when a user disables sync. When a user disables sync (after turning it on) for an account, we delete all of that account's items from the shared store. We then check to see if this was the last account to have sync turned on. If there are no other accounts with sync turned on, we then delete the main shared Authenticator key from the App Group keychain. If there are more accounts still with sync turned on, we leave the key in place.

This allows a user to turn off sync for one or all of their accounts and we clear everything up so that there's no left over data from having turned sync on. 

Note: This PR is based off of the branch for [this open PR](https://github.com/bitwarden/ios/pull/998). The other PR would need to be approved/merged first, or this will merge into that PR and wait for it to be approved/merge.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
```diff
@@ -174,22 +174,40 @@ actor DefaultAuthenticatorSyncService: NSObject, AuthenticatorSyncService {
...
+        for account in try await stateService.getAccounts() {
+            hasAccountWithSync = try await stateService.getSyncToAuthenticator(userId: account.profile.userId)
+            if hasAccountWithSync {
+                break
+            }
+        }
+        if !hasAccountWithSync {
+            try? sharedKeychainRepository.deleteAuthenticatorKey()
+        }
+    }
```
fedemkr, line 190: 🎨 You can simplify this a bit like this:
```suggestion
    private func deleteKeyIfSyncingIsOff() async throws {
        for account in try await stateService.getAccounts() {
            let hasAccountWithSync = try await stateService.getSyncToAuthenticator(userId: account.profile.userId)
            guard !hasAccountWithSync else {
                return
            }
        }

        try? sharedKeychainRepository.deleteAuthenticatorKey()
    }
```

BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
```diff
@@ -174,22 +174,40 @@ actor DefaultAuthenticatorSyncService: NSObject, AuthenticatorSyncService {
...
+    private func deleteKeyIfSyncingIsOff() async throws {
+        var hasAccountWithSync = false
+        for account in try await stateService.getAccounts() {
+            hasAccountWithSync = try await stateService.getSyncToAuthenticator(userId: account.profile.userId)
+            if hasAccountWithSync {
+                break
+            }
+        }
+        if !hasAccountWithSync {
+            try? sharedKeychainRepository.deleteAuthenticatorKey()
```
fedemkr, line 188: 🤔 Should the error be reported to Crashlytics if it happens?

BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
```diff
@@ -174,22 +174,37 @@ actor DefaultAuthenticatorSyncService: NSObject, AuthenticatorSyncService {
...
+    ///
+    private func deleteKeyIfSyncingIsOff() async throws {
+        for account in try await stateService.getAccounts() {
+            let hasAccountWithSync = try await stateService.getSyncToAuthenticator(userId: account.profile.userId)
+            guard !hasAccountWithSync else {
+                return
+            }
+        }
+        try sharedKeychainRepository.deleteAuthenticatorKey()
+    }
```
fedemkr, line 187: ⛏️ I might have missed them but aren't tests missing for when `getAccounts`, `getSyncToAuthenticator` and `deleteAuthenticatorKey` throw?

brant-livefront: 💯 Great note. I added tests for `getAccounts` and `deleteAuthenticatorKey` (which previously didn't throw, but was fixed). This also prompted me to add a couple more error handling checks for the errors in `determineSyncForUserId` func. 👍 Great example of "100% coverage is not always 100% covered" 😆  The latest commit has all the new tests.

I can't think of a reliable way to test the `getSyncToAuthenticator` - we access that same method in `determineSyncForUserId` so it will hit the error there _before_ it comes here (i.e. it has to first realize sync is turned off for an account, then it comes here to check if there's any accounts left syncing). So when I was trying to insert this to test an error here, it was instead failing first in `determineSyncForUserId`. However, I did add the test for that explicitly and this throws up to `determineSyncForUserId` which then throws back up to the initial Tasks. So we've covered all the places that could report this and I'm confident that it gets reported, but I can't think of a way to insert into the flow such that one call succeeds and the second one fails.

fedemkr: 🤔 The only way I can come up with is to have an array of return values and a counter, so you return the value of the index of the array based on the number of times the method has been called; so if you have (simplified) `[false, throw]` the first time it calls `getSyncToAuthenticator` it will return `false` and the second one will `throw`. Nonetheless not sure if it's a must to add such test behavior.

brant-livefront: Tests should be fixed with the merge of the latest stuff from the branch this was based off of.  👍 

That's a good idea for testing this 💯  I think it's probably not worth the added complexity of the tests (unless we find this issue coming up more often), but I'm up for taking a shot at it if you think it's worth it. 👍

fedemkr (dismissed):

brant-livefront:
@fedemkr This dismissed your approval because this was re-targeted to `main` once I merged the other PR in and I needed to fix some conflicts when this merged in.

fedemkr (approved):

PR #1010, opened 2024-10-04
PM-10267 PM-12988 - Fix saving master password generation and match swipe/close gesture ux

phil-livefront:
## 🎟️ Tracking

[PM-12988](https://bitwarden.atlassian.net/browse/PM-12988)
[PM-10267](https://bitwarden.atlassian.net/browse/PM-10267)

## 📔 Objective

- The close button and swipe-down gesture in the Create Account flow led to inconsistent navigation, landing users on different screens. This issue stemmed from how we were using dismissPresented in the AuthCoordinator. When navigating via an email verification deep link, it wasn't a problem because the stack was cleared before showing the view. However, during QA testing—where the email verification step was bypassed—this bug surfaced.
- The same root cause affected the "Save" button functionality, where users weren’t being returned to the CompleteRegistrationView. Since we were using presentedViewController, all current modals were dismissed, sending testers back to the Login screen.
- In this solution we are using the same `UINavigationController` used in the `MasterPasswordGuidanceView` and pushing the `MasterPasswordGeneratorView` on top of the existing stack instead of making a new one.  This allowed us to use the new implemented `dismissPresented` function which dismisses only the top VC as opposed to closing more than 1.
- I discussed the proposed solution with design, and they've confirmed it effectively resolves all the issues. 

## 📸 Screenshots

https://github.com/user-attachments/assets/1515484f-acda-4a64-aa72-ff05d730d7bb

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


BitwardenShared/UI/Auth/AuthCoordinator.swift
```diff
@@ -571,8 +571,13 @@ final class AuthCoordinator: NSObject, // swiftlint:disable:this type_body_lengt
         let store = Store(processor: processor)
         let view = MasterPasswordGeneratorView(store: store)
         let viewController = UIHostingController(rootView: view)
-        let navigationController = UINavigationController(rootViewController: viewController)
-        stackNavigator?.present(navigationController)
+
+        let topmostViewController = stackNavigator?.rootViewController?.topmostViewController()
+        topmostViewController?.navigationItem.backButtonTitle = Localizations.back
```
fedemkr, line 576: 🤔 I think for now this is fine but wanted to call out that I've had issues in the past where the `backButtonTitle` would  remain for the `topmostViewController` so if we use that controller to show a different screen will use that too. However, it doesn't seem to be the case here so I'll approve ✅

fedemkr (approved):

PR #1017, opened 2024-10-07
[BITAU-121] Add bitwardenAccountName to AuthenticatorBridge items

brant-livefront:
## 🎟️ Tracking

[BITAU-121](https://livefront.atlassian.net/browse/BITAU-121)

## 📔 Objective

As part of reviewing [this PR on the Authenticator repo](https://github.com/bitwarden/authenticator-ios/pull/163), we realized that it would clarify things to add an explicit `bitwardenAccountName` to the AuthenticatorBridge items. This PR adds that to the relevant objects, adds encryption for those fields in the `SharedCryptographyService`, and adds tests. Once this is merged, I'll revisit the above PR and pull in this new data point.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
```diff
@@ -160,16 +160,16 @@ actor DefaultAuthenticatorSyncService: NSObject, AuthenticatorSyncService {
         let decryptedCiphers = try await totpCiphers.asyncMap { cipher in
             try await self.clientService.vault(for: userId).ciphers().decrypt(cipher: cipher)
         }
-        let account = try await stateService.getActiveAccount()
-        let username = account.profile.email
+        let account = try? await stateService.getAccount(userId: userId)
 
         return decryptedCiphers.map { cipher in
             AuthenticatorBridgeItemDataView(
+                bitwardenAccountName: account?.profile.email,
```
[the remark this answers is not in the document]
brant-livefront: Currently, the comps for the Authenticator don't distinguish organizations or collections. It's just one big list under the account. This field is just to support the items we need to show over there. Maybe we should call it `bitwardenAccountEmail` since it will be the email that we're displaying?

In addition, your question here led me to clarify one other UX question I had and it sounds like we _do_ need to store the domain. I am going to add that as a column now.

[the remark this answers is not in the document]
brant-livefront: I renamed it to `bitwardenAccountEmail` for clarity and added `bitwardenAccountDomain` in the latest commit 👍

BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
```diff
@@ -160,16 +160,17 @@ actor DefaultAuthenticatorSyncService: NSObject, AuthenticatorSyncService {
         let decryptedCiphers = try await totpCiphers.asyncMap { cipher in
             try await self.clientService.vault(for: userId).ciphers().decrypt(cipher: cipher)
         }
-        let account = try await stateService.getActiveAccount()
-        let username = account.profile.email
+        let account = try? await stateService.getAccount(userId: userId)
 
         return decryptedCiphers.map { cipher in
             AuthenticatorBridgeItemDataView(
+                bitwardenAccountDomain: account?.settings.environmentUrls?.webVaultHost,
```
victor-livefront, line 167: Just want to double check with @fedemkr and @matt-livefront: This is the correct way to get the account domain, right? We found this being used in the profile selector, which is a similar use case for this.

victor-livefront (commented):
Looks good to me. Just that question to confirm the account domain.

victor-livefront (dismissed):

AuthenticatorBridgeKit/AuthenticatorBridgeItemDataModel.swift
```diff
@@ -6,6 +6,12 @@ import Foundation
 public struct AuthenticatorBridgeItemDataModel: Codable, Equatable {
     // MARK: Properties
 
+    /// The domain of the Bitwarden account that owns this item. (e.g. https://vault.bitwarden.com)
+    public let bitwardenAccountDomain: String?
+
+    /// The email of the Bitwarden account that owns this item.
+    public let bitwardenAccountEmail: String?
+
```
fedemkr, line 14: 🤔 IMO it shouldn't be necessary to have the `bitwarden` prefix on these new properties. Having only the `account` prefix should be enough to know the properties are from the Bitwarden account and not from the item being shared.
🤔  Another idea would be to have a different Authenticator bridge data model for account data so we don't mix account properties with item properties and also to only send that data once, these two properties will be the same for all items shared in the account. But this would require a larger change.
What's your thoughts on this? Also cc: @matt-livefront @KatherineInCode

matt-livefront: `accountDomain` and `accountEmail` seem clear enough to me too.

I think for simplicity, flattening the data model and including the domain and email with each one makes sense to me. The size of this data seems manageable and it'll be nice to not have to merge data from multiple tables for these shared items.

brant-livefront: Sounds good - I'll add a commit with `accountDomain` and `accountEmail` and drop the `bitwarden` prefix. 👍

fedemkr (approved):

PR #1028, opened 2024-10-09
PM-13066 - Add FAB on various screens

phil-livefront:
## 🎟️ Tracking

[PM-13066](https://bitwarden.atlassian.net/browse/PM-13066)

## 📔 Objective

- Implemented a floating action button to the bottom right of the `VaultListView`, `VaultGroupView`, `VaultAutofillListView`, `SendListView`, `VaultItemSelectionView`, and `FoldersView`.  
- Created a `CircleButtonStyle` and a `FloatingActionButton`.  
- Added extensions on `View` to create an `add` and `edit` floating action buttons which were used in the views above to prevent duplication. 
- Confirmed with design that the FAB should be shown during the loading and empty state.  I did however follow the pattern to hide the FAB when searching.
- I didn't remove the "+" from the nav bar because that will be done in this [ticket](https://bitwarden.atlassian.net/browse/PM-12663) and i wanted to keep these changes minimal. 
- The current `plus.pdf` is actually a little off centered.  I am going to be replacing it in the icons update ticket so I left it alone in this PR but you may notice the "+" looks a bit weird on the FAB and thats why. 
- Sorry about the large file change amount, its mostly snapshots though.

** Going to confirm with design about the empty states...  After seeing some snapshots with large font, it just doesn't make sense to have two buttons overlapping that do the same thing. Keeping as is for now since figma matches but might update in this Pr or a future one.

Update from design:  Currently working on a design update and approved keeping the FAB on the empty state as-is until coming to a decision. 

## 📸 Screenshots

see snapshots for all of the screenshot changes but heres an example.

Simulator Screenshot - iPhone 15 Pro - 2024-10-09 at 13 20 22

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-13066]: https://bitwarden.atlassian.net/browse/PM-13066?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenShared/UI/Vault/Vault/VaultList/VaultListView.swift
```diff
@@ -67,6 +70,36 @@ private struct SearchableVaultListView: View {
 
     // MARK: Private Properties
 
+    /// A floating action button positioned at the bottom-right corner of the screen.
+    @ViewBuilder private var newLoginFloatingActionButton: some View {
+        VStack {
+            Spacer()
+            HStack {
+                Spacer()
```
matt-livefront, line 78: 🎨 Instead of multiple stacks with spacers, I wonder if you could add it as an overlay to the vault's loading view? Then you wouldn't have to worry about hiding it during search and you could get the bottom trailing behavior without stacks and spacers.

```
    /// A view that displays either the my vault or empty vault interface.
    @ViewBuilder private var vault: some View {
        LoadingView(state: store.state.loadingState) { sections in
            if sections.isEmpty {
                emptyVault
            } else {
                vaultContents(with: sections)
            }
        }
        .overlay(alignment: .bottomTrailing) {
            newLoginFloatingActionButton
        }
    }
```

Also, this button can be used to create more things than logins so maybe `newItemFloatingActionButton` is better.

BitwardenShared/UI/Vault/Vault/VaultList/VaultListView.swift
```diff
@@ -67,6 +70,36 @@ private struct SearchableVaultListView: View {
...
+                                    .init(
+                                        color: Asset.Colors.buttonFilledForeground.swiftUIColor,
+                                        scaleWithFont: false,
+                                        width: 24,
+                                        height: 24
+                                    )
+                                )
+                        )
+                }
+                .accessibilityIdentifier("new-login-floating-action-button")
```
matt-livefront, line 97: ❓ Curious, I didn't see this in the ticket, where did it come from? Our other accessibility identifiers are all pascal case, so just wanted to confirm.

BitwardenShared/UI/Vault/Vault/VaultList/VaultListView.swift
```diff
@@ -67,6 +70,36 @@ private struct SearchableVaultListView: View {
...
+                        .frame(width: 50, height: 50)
+                        .overlay(
+                            Asset.Images.plus.swiftUIImage
+                                .imageStyle(
+                                    .init(
+                                        color: Asset.Colors.buttonFilledForeground.swiftUIColor,
+                                        scaleWithFont: false,
+                                        width: 24,
+                                        height: 24
+                                    )
```
matt-livefront, line 93: 🎨 This pressed state is slightly different than our other buttons. Should they match?

| Add an Item | FAB |
| --- | --- |
| Screenshot 2024-10-09 at 1 37 51 PM | Screenshot 2024-10-09 at 1 37 58 PM |

I think the FAB is also used on `VaultGroupView`, so maybe making this its own view makes sense.

BitwardenShared/UI/Vault/Vault/VaultList/VaultListView.swift
matt-livefront, line 97: 🎨 We should also add an accessibility label.

matt-livefront (commented):
If there's a bunch of items in the vault list, it seems like this overlaps the last item. Should we add additional padding to the bottom of the list to account for the FAB?

https://github.com/user-attachments/assets/b4456bc7-131b-4b73-9428-2202a2891ba6

The FAB looks like it's used in a few different places. Do you know if there are other tickets for that, or should we do that as part of this? I think this includes `VaultGroupView`, `VaultAutofillListView`, `SendListView`, `VaultItemSelectionView`, and `FoldersView` (anywhere `addToolbarItem` is used).

BitwardenShared/UI/Vault/Vault/VaultList/VaultListView.swift
phil-livefront: yeah great call, thanks!  ill update.

BitwardenShared/UI/Vault/Vault/VaultList/VaultListView.swift
phil-livefront: oh i was using this to get the button for testing... is that bad practice?

BitwardenShared/UI/Vault/Vault/VaultList/VaultListView.swift
phil-livefront: yeah they should match now that i am thinking about it.  Ill make the update.

BitwardenShared/UI/Vault/Vault/VaultList/VaultListView.swift
matt-livefront: Ah, the QA team also uses these for automated testing, but usually they specify the ID so the buttons match between platforms. This should probably match the add button's ID in the toolbar (`AddItemButton`) if I had to guess. 

Usually we use the accessibility label to get the button for testing, but the identifier works as well.

BitwardenShared/UI/Vault/Vault/VaultList/VaultListViewTests.swift
```diff
@@ -68,6 +68,14 @@ class VaultListViewTests: BitwardenTestCase {
         XCTAssertEqual(processor.dispatchedActions.last, .addItemPressed)
     }
 
+    /// Tapping the new login floating acrtion button dispatches the `.addItemPressed` action.`
```
KatherineInCode, line 71:
```suggestion
    /// Tapping the new login floating action button dispatches the `.addItemPressed` action.`
```

BitwardenShared/UI/Vault/Vault/VaultList/VaultListView.swift
KatherineInCode: It might be worth running this by QA to see what they prefer for the ID here

KatherineInCode (commented):
🤔 There are a couple of these screenshots that have the FAB that I'm wondering if they should, such as loading, or the empty state where there's an "Add Item" button front and center. Is this what we want, or will more conditional logic for this be in a future PR?

phil-livefront:
> 🤔 There are a couple of these screenshots that have the FAB that I'm wondering if they should, such as loading, or the empty state where there's an "Add Item" button front and center. Is this what we want, or will more conditional logic for this be in a future PR?

Im about to make a commit to fix a lot of what Matt brought up and add the FAB to the other screens as well.  I did confirm with design that the FAB should be on the loading screen and on the figma designs they show the FAB being on the empty state as well.

BitwardenShared/UI/Vault/Vault/VaultList/VaultListView.swift
phil-livefront: i removed the id and added a label, i was using it incorrectly and didn't realize QA was using them.  sorry about the confusion here!

BitwardenShared/UI/Tools/Send/Send/SendList/SendListView.swift
```diff
@@ -98,7 +103,9 @@ private struct MainSendListView: View {
                     )
                 }
             }
-            .padding(16)
+            .padding(.top, 16)
+            .padding(.horizontal, 16)
+            .padding(.bottom, FloatingActionButton.bottomOffsetPadding)
```
matt-livefront, line 108: 🎨 I noticed that sometimes you remove the original bottom padding and separate it out like this so as to not add extra bottom padding and sometimes you just add bottom padding for the FAB.

https://github.com/bitwarden/ios/blob/34a1278a4331ddb38f76ccb7cd5a9ca110d5396c/BitwardenShared/UI/Vault/Vault/VaultItemSelection/VaultItemSelectionView.swift#L189-L190

I think the scroll view modifier is going to also add extra bottom padding. We should be consistent if we can. If you removed the extra 16pts of padding in `FloatingActionButton.bottomOffsetPadding`, could we just add the additional padding for the FAB without having to specify the top and horizontal padding separately? Like what exists in `VaultItemSelectionView.swift`.

BitwardenShared/UI/Platform/Application/Views/FloatingActionButton.swift
```diff
@@ -0,0 +1,36 @@
...
+// MARK: - FloatingActionButton
+
+/// A view representing a floating action button for adding new items.
+///
+struct FloatingActionButton: View {
+    // MARK: Properties
+
+    /// The bottom content offset padding around the floating action button.
+    /// This is the button size plus the bottom/top padding (50 + 16 + 16).
+    static let bottomOffsetPadding: CGFloat = 82
```
matt-livefront, line 12: ❓ It looks like this is assuming the button is 50x50, but it looks like the image is 24x24 with 12 pt of padding = 48x48. Would it be better to give the button a fixed frame and not specify the padding?

BitwardenShared/UI/Platform/Application/Views/FloatingActionButton.swift
```diff
@@ -0,0 +1,46 @@
...
+// MARK: - FloatingActionButton
+
+/// A view representing a floating action button for adding new items.
+///
+struct FloatingActionButton: View {
+    // MARK: Properties
+
+    /// The bottom content offset padding around the floating action button.
+    /// This is the button size plus the bottom/top padding (50 + 16 + 16).
+    static let bottomOffsetPadding: CGFloat = 66
```
matt-livefront, line 12: 🎨 
```suggestion
    /// This is the button size plus the bottom/top padding (50 + 16).
    static let bottomOffsetPadding: CGFloat = 66
```

BitwardenShared/UI/Vault/Vault/VaultGroup/VaultGroupView.swift
```diff
@@ -107,6 +107,11 @@ struct VaultGroupView: View {
                 groupView(with: items)
             }
         }
+        .overlay(alignment: .bottomTrailing) {
+            addItemFloatingActionButton {
+                store.send(.addItemPressed)
```
matt-livefront, line 112: ⚠️ I think we sometimes don't show the add button in the toolbar, like if you're in the trash view. We should hide the FAB as well in that case. 
https://github.com/bitwarden/ios/blob/2990c57ea207b4c872097916f88b4ae92ad4cb3e/BitwardenShared/UI/Vault/Vault/VaultGroup/VaultGroupView.swift#L37-L38

BitwardenShared/UI/Platform/Application/Extensions/View.swift
```diff
@@ -125,4 +125,36 @@ extension View {
...
+    }
+
+    /// Returns a floating action button positioned at the bottom-right corner of the screen.
+    ///
+    /// - Parameter action: The action to perform when the button is tapped.
+    /// - Returns: A `FloatingActionButton` configured for adding an item.
+    ///
+    func editItemFloatingActionButton(action: @escaping () -> Void) -> some View {
+        floatingActionButton(image: Asset.Images.pencil.swiftUIImage, action: action)
+            .accessibilityLabel("EditItemFloatingActionButton")
```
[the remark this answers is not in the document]
matt-livefront: Yeah, that's fine, can you add a comment to that ticket?

matt-livefront (commented):
Just a few additional comments, but otherwise this looks great. There's a few lint warnings that should get resolved too.

BitwardenShared/UI/Platform/Application/Extensions/View.swift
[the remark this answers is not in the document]
phil-livefront: done!

matt-livefront (approved):

PR #1029, opened 2024-10-09
[PM-13193] Create Xcode version file

KatherineInCode:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-13193

## 📔 Objective

This creates a `.xcode-version` file per the [Fastlane proposal](https://github.com/fastlane/ci/blob/master/docs/xcode-version.md). There is a minor difference that format of the versions has to match how the [setup-xcode](https://github.com/maxim-lobanov/setup-xcode) expects it; the only place where this matters is with beta Xcode versions, which need to be e.g. `16.1-beta` rather than the proposal's `16.1b2`. Because Fastlane currently does not do the build, there should be no interference there.

As part of this, it made sense to add parameters to the `test.yml` workflow to allow overriding the Xcode version, the simulator name, and simulator iOS version respectively. Currently the defaults for the latter two are still hardcoded in the workflow, which is not ideal—future ideas could be for `.test-simulator-name` and `.test-simulator-version` files, though that could clutter things up. While storing all of these defaults in one file could also work, that breaks the `.xcode-version` file proposal earlier, which could impede adopting Fastlane for building—though we would still run into that problem with the override.

This lays the groundwork to more easily migrate to Xcode 16, and is part of the broader iOS 18 / Xcode 16 / Swift 6 effort.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


fedemkr (approved):
Really like this 🎉

PR #1032, opened 2024-10-10
[BITAU-141] [BITAU-142] [BITAU-143] [BITAU-150] Add Vault Unlock

brant-livefront:
## 🎟️ Tracking

[[BITAU-141](https://livefront.atlassian.net/browse/BITAU-141)] Add ability to read vault key from user's unlocked vault and save to keychain
[[BITAU-142](https://livefront.atlassian.net/browse/BITAU-142)] Unlock Vault using stored Authenticator Vault Key
[[BITAU-143](https://livefront.atlassian.net/browse/BITAU-143)] Delete Authenticator Vault Key on Turning Sync Off
[[BITAU-150](https://livefront.atlassian.net/browse/BITAU-150)] Create Keychain Item for Authenticator Vault Key

## 📔 Objective

This PR adds the ability to handle background and locked-vault syncs for the user. There's a few different parts:
- If the vault is locked when the user turns sync on, we don't take any action. This is the same as before.
- When the user has an unlocked vault and has turned on sync for the active user, just as before we kick off sync. However we now also store a copy of their vault key in the keychain as their `authenticatorVaultKey`.
- This allows us to unlock their vault when a Cipher update happens, but they have a locked vault. For instance: Unlock Account 1, switch to Account 2, account 1's vault timeout occurs and then you receive a Push for Account 1 (this is similar to the case Matt mentioned on the previous PR)
- Lastly, when they turn off sync, we clean everything up
   - Remove their `authenticatorVaultKey`
   - (From a previous PR) Remove their sync'd data and (if this is the last account) the shared key 

Note: There is one known issue with this PR (I'll handle it as part of [BITAU-153](https://livefront.atlassian.net/browse/BITAU-153)): Since the keys in the PM app are stored with an accessibility level of `kSecAttrAccessibleWhenUnlockedThisDeviceOnly` they are not available if the app is truly in the background (i.e. an update comes in via Push notification while the app is backgrounded and the device is locked). Thus we can't unlock the vault if the app is not in the foreground. I will need to add a refactor to move _just this one key_ to `kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly` which does allow background access.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
```diff
@@ -39,8 +42,8 @@ actor DefaultAuthenticatorSyncService: NSObject, AuthenticatorSyncService {
     /// The service used by the application to report non-fatal errors.
     private let errorReporter: ErrorReporter
 
-    /// Notification Center Service to subscribe to foreground/background events.
-    private let notificationCenterService: NotificationCenterService
```
KatherineInCode, line 43: 🤔 Is the notification center service getting removed because we don't actually need to know if the app is foregrounded/backgrounded?

brant-livefront: Yes. It was left over from a previous refactor (previously we were waiting for the foreground to take action).

BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
```diff
@@ -194,16 +216,18 @@ actor DefaultAuthenticatorSyncService: NSObject, AuthenticatorSyncService {
     /// - Parameter userId: The userId of the user whose sync status is being determined.
     ///
     private func determineSyncForUserId(_ userId: String) async throws {
-        if try await !stateService.getSyncToAuthenticator(userId: userId) {
+        guard try await stateService.getSyncToAuthenticator(userId: userId) else {
```
KatherineInCode, line 219: 🎨 In general I think `guard` is best for quick "fail out of flow" checks; this feels like it's doing a lot in the event of a failure, and I kind of think it would be better as an `if` block. Though I also appreciate the positive vs. negative aspects of it, and it's rather borderline because of that, and could go either way 🤔

KatherineInCode (approved):

BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
brant-livefront: In general, I agree -  I like to `guard` for more exceptional cases and get those out of the way first. 

While this isn't an exceptional case (when it launches it will be the most common case until people start turning it on), it did feel like a special case to handle first and foremost. And it means the rest of the method isn't sitting inside the `if/else` - which gets a little confusing because it needs to also check the vault status in another `if`.

I'm also slightly on the borderline as well, so I'm open to changing it. 🤔

BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
```diff
@@ -233,9 +257,14 @@ actor DefaultAuthenticatorSyncService: NSObject, AuthenticatorSyncService {
...
-        guard !vaultTimeoutService.isLocked(userId: userId) else { return }
-
+        let useKey = vaultTimeoutService.isLocked(userId: userId)
+        if useKey {
+            try await authRepository.unlockVaultWithAuthenticatorVaultKey(userId: userId)
+        }
         let items = try await decryptTOTPs(ciphers, userId: userId)
+        if useKey {
+            await vaultTimeoutService.lockVault(userId: userId)
+        }
```
fedemkr, line 267: 🤔 Hmmm TBH I wouldn't include an automatic unlock without the user explicitly knowing, I feel like this would at least need to be an **opt-in** feature so the user consents on having this behavior and we only automatically do this if the vault is currently unlocked. Could you check with Product about this?

fedemkr (changes requested):

BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
victor-livefront: When we originally put together a POC around this behavior, we presented options for warning the user about this (similar to the never lock) to Kyle Spearrin and Brian Gentry. Where we landed was that the toggle to turn on the feature was the opt in enough, and doesn't leave the entire PM permanently unlocked like never lock does.

BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
KatherineInCode: Having thought about it, I think it's fine. Yes, it's a little weird, but it *does* describe the situation in words pretty well, even if the "clean up" bit is a bit lengthy.

On the other hand, having `try` statements in the `else` block of a `guard` *also* feels a little weird.

I think for now we keep it as `guard`?

BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
```diff
@@ -759,6 +767,11 @@ extension DefaultAuthRepository: AuthRepository {
         )
     }
 
+    func unlockVaultWithAuthenticatorVaultKey(userId: String) async throws {
+        let authenticatorKey = try await keychainService.getAuthenticatorVaultKey(userId: userId) ?? ""
+        try await unlockVault(method: .decryptedKey(decryptedUserKey: authenticatorKey), hadUserInteraction: false)
```
fedemkr, line 772: 🤔 If `keychainService.getAuthenticatorVaultKey` is `nil` then unlocking will definitely fail so it feels like a `guard` should be added here and throw an error if that happens so we have a more specific error as to why this failed.

BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
fedemkr: @victor-livefront something else now that I re-read it is if `let items = try await decryptTOTPs(ciphers, userId: userId)` then the vault would remain unlocked. I think a `do .... catch` should be added here to lock the vault if the func throws and then re-throw the error for the caller.

victor-livefront: That's a good callout. Brant and I discussed threading here but overlooked the throw.

brant-livefront: @fedemkr @victor-livefront OK, I've pushed a commit that I think addresses all of these issues:
- I've moved the fetching of the account to this method. If that fails, it will throw immediately and get caught/logged  by the CipherPublisherTask's catch.
- I've changed the `keychainService.getAuthenticatorVaultKey` to return a non-optional `String`. It was always throwing `KeychainServiceError.keyNotFound` in the case of the key missing, so there's no reason for it to be optional
   - if this is thrown in the course of unlocking their vault, it will now be caught by the `catch` in `writeCiphers` and immediately logged.
- Likewise, if decrypting ciphers or writing them to the shared store fails, we're catching that explicitly and logging it right here in the method, so we should have insight into what's failing
- Lastly, outside of all of that, we check the `useKey` at the end of the method and make sure to lock the Vault back up even if any errors have occurred.

victor-livefront (dismissed):
Updated logic makes sense to me!

BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
```diff
@@ -233,9 +256,20 @@ actor DefaultAuthenticatorSyncService: NSObject, AuthenticatorSyncService {
...
+                try await authRepository.unlockVaultWithAuthenticatorVaultKey(userId: userId)
+            }
+            let items = try await decryptTOTPs(ciphers, account: account)
+            try await authBridgeItemService.replaceAllItems(with: items, forUserId: userId)
+        } catch {
+            errorReporter.log(error: error)
+        }
+        if useKey {
+            await vaultTimeoutService.lockVault(userId: userId)
+        }
```
fedemkr, line 273: ⛏️ Maybe I couldn't find it, but could a test be added when `let items = try await decryptTOTPs(ciphers, account: account)` throws? and when `try await authBridgeItemService.replaceAllItems(with: items, forUserId: userId)` throws ensuring the vault is locked afterwards?

brant-livefront: Added two more tests to explicitly check each of those 2 lines.  👍  Verifies that the vault gets properly re-locked even when we hit an error.

BitwardenShared/Core/Platform/Services/AuthenticatorSyncServiceTests.swift
```diff
@@ -602,9 +749,117 @@ final class AuthenticatorSyncServiceTests: BitwardenTestCase { // swiftlint:disa
...
+                    totp: "totp"
+                )
+            ),
+        ])
+
+        waitFor(!errorReporter.errors.isEmpty)
+        XCTAssertTrue(vaultTimeoutService.isLocked(userId: "1"))
+    }
+
+    /// Verify that `writeCiphers()` correctly catches and logs errors that occur in `replaceAllItems`. The user's vault is
```
fedemkr, line 815: ⛏️ Line length > 120.

brant-livefront: 🤦‍♂️ Fixed in latest.

fedemkr (approved):

PR #1045, opened 2024-10-14
[PM-13239] Add isPreAuth to ClientService.generator() calls

KatherineInCode:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-13239

## 📔 Objective

This fixes an error in the SSO flow that caused a crash on debug builds. This was happening because the SSO flow uses the Generator for TDE things before the user is authorized, but was not indicating that we were pre-auth at that time; in release builds this was just outputting an error log and continuing, but that crashes on debug builds.

This fix just adds the `isPreAuth` flag to the `generator()` function, and handles things appropriately upstream.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


BitwardenShared/Core/Platform/Services/ClientService.swift
```diff
@@ -97,8 +100,11 @@ extension ClientService {
 
     /// Returns a `ClientGeneratorsProtocol` for generator data tasks.
     ///
-    func generators() async throws -> ClientGeneratorsProtocol {
-        try await generators(for: nil)
+    /// - Parameter isPreAuth: Whether the client is being used for a user prior to authentication
+    ///     (when the user's ID doesn't yet exist). This primarily will happen in SSO flows.
+    ///
+    func generators(isPreAuth: Bool) async throws -> ClientGeneratorsProtocol {
```
matt-livefront, line 106: 🎨 I think we could default this to `false` here if we don't want to always specify it when not in pre-auth flows. That would match the `auth()` function, although I think we could also default that rather than specify an alternate function.

https://github.com/bitwarden/ios/blob/e8fc3661429e368d25fa06db43eba195d4be7a30/BitwardenShared/Core/Platform/Services/ClientService.swift#L74-L86

fedemkr: 🤔 I see pros and cons with any of the approaches because if we provide default values then someone could introduce a bug by copy/pasting a `generators()` call without parameter and not knowing they need to specify `isPreAuth: false` for that on pre auth flows. OTOH, it feels great not to specify the parameter as it would be `false` in the majority of use cases and it introduces less code changes.

🤔 Also I've been thinking on this for a while on trying not to pass the parameter at all and have some "context" where we can check whether the current flow is actually pre-auth or not so we reduce the chance of introducing bugs on these calls. I have some ideas but need to keep thinking on some cases, for now I see any of the two approaches working fine and we'd just need to keep an eye when calling it and on the PR reviews to be sure we're in the right flow.

KatherineInCode: @matt-livefront 
I missed that, and I agree—though I don't think you can provide defaults in a protocol. It makes sense for both `auth` and `generator` to match here, though.

@fedemkr 
I would definitely much rather some sort of broader context that we can check, rather than passing this around so much. I originally wrote it to just add that flag to *all* of the calls and it was a massive spiderweb of them.

Some of this is also motivated by the fact that in Debug modes, if we log an error we intentionally crash; the issue this is fixing wouldn't actually ever happen in a Release.

matt-livefront: @KatherineInCode This function is in the extension, which I think would allow adding a default value.

@fedemkr I think the context idea is a good one. In regard to default values, between crashlytics logging and crashing in debug builds these pretty easy to find if you miss a spot. And we don't have too many spots where we use the SDK prior to authentication, so I like the default value for those reasons.

KatherineInCode: @matt-livefront Oh! I see what you mean! Yes, I like that idea and can do something to that effect

matt-livefront (approved):

KatherineInCode:
@fedemkr @matt-livefront I updated which branch this is going against because this fixes an issue found with the release

matt-livefront:
> @fedemkr @matt-livefront I updated which branch this is going against because this fixes an issue found with the release

Would it be better to merge this into main and then have a separate PR to pull it into the release branch? That prevents a fix from ever being left out of main. I'm curious what issue this fixes with the release - I thought this should all work without the isPreAuth flag.

KatherineInCode:
@matt-livefront The idea with the RC branches is they'll be regularly merged into `main` as well, so that we don't miss that sort of thing (cf. [Mobile Release Branching Process Draft](https://bitwarden.atlassian.net/wiki/spaces/EN/pages/1227325457/Mobile+Release+Branching+Process+Draft))

The ticket in question is listed as a blocker for the release, and was found in regression testing the release, as I understand it; hence why the putting it in the release branch.

@differsthecat Does that all seem right to you?

matt-livefront (approved):

matt-livefront:
> @matt-livefront The idea with the RC branches is they'll be regularly merged into `main` as well, so that we don't miss that sort of thing (cf. [Mobile Release Branching Process Draft](https://bitwarden.atlassian.net/wiki/spaces/EN/pages/1227325457/Mobile+Release+Branching+Process+Draft))

Ah, ok.

> The ticket in question is listed as a blocker for the release, and was found in regression testing the release, as I understand it; hence why the putting it in the release branch.

When you said it was a blocker, I assumed this change was causing some larger issue. But crashing in debug builds is what I would expect, worth nothing that this wouldn't have crashed in production builds though.

KatherineInCode:
> worth nothing that this wouldn't have crashed in production builds though

I did in the PR objective, but I'll also add a note to the ticket about it.

fedemkr (approved):

PR #1052, opened 2024-10-17
[BITAU-155] Deep link to Settings Account Security screen

brant-livefront:
## 🎟️ Tracking

[BITAU-155](https://livefront.atlassian.net/browse/BITAU-155)

## 📔 Objective

This PR adds the ability to handle a deep link into the Settings -> Account Security screen: `bitwarden://settings/account_security`. It refactors the openURL method in AppProcessor slightly to clearly delineate `bitwarden://` URLs from `otpauth://` URLs. 

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


victor-livefront (approved):
Looks good. Next up are changes to the Auth app to send this Deep link path, right?

brant-livefront:
> Looks good. Next up are changes to the Auth app to send this Deep link path, right?

Yep 💯 I actually put a commit on the end of this PR: https://github.com/bitwarden/authenticator-ios/pull/170
That updates our `ExternalLinkConstants` to have this official link.  👍

BitwardenShared/UI/Platform/Application/AppProcessor.swift
```diff
@@ -115,22 +115,13 @@ public class AppProcessor {
     /// - Parameter url: The deep link URL to handle.
     ///
     public func openUrl(_ url: URL) async {
-        guard url.scheme?.isOtpAuthScheme == true else { return }
+        guard let scheme = url.scheme else { return }
 
-        guard let otpAuthModel = OTPAuthModel(otpAuthKey: url.absoluteString) else {
-            coordinator?.showAlert(.defaultAlert(title: Localizations.anErrorHasOccurred))
-            return
+        if scheme == "bitwarden" {
```
fedemkr, line 120: ⛏️ Can `"bitwarden"` scheme be moved to a constant?

BitwardenShared/UI/Platform/Application/AppProcessor.swift
```diff
@@ -380,6 +371,47 @@ extension AppProcessor {
...
     }
 
+    /// Handle a "bitwarden://" url.
+    ///
+    /// - Parameter url: The Bitwarden URL received by the app.
+    ///
+    private func handleBitwardenUrl(url: URL) async {
+        guard let components = URLComponents(url: url, resolvingAgainstBaseURL: true) else { return }
+
+        if components.host == "settings", components.path == "/account_security" {
```
fedemkr, line 381: 🎨 Could it be added to the `guard` instead?
```suggestion
        guard let components = URLComponents(url: url, resolvingAgainstBaseURL: true),
                   components.host == "settings", components.path == "/account_security" else { 
            return 
        }
```

BitwardenShared/UI/Platform/Application/AppProcessor.swift
```diff
@@ -380,6 +371,47 @@ extension AppProcessor {
...
+            let accountSecurityRoute = AppRoute.tab(.settings(.accountSecurity))
+            guard let userId = try? await services.stateService.getActiveAccountId(),
+                  !services.vaultTimeoutService.isLocked(userId: userId),
+                  await (try? services.vaultTimeoutService.hasPassedSessionTimeout(userId: userId)) == false
+            else {
+                await coordinator?.handleEvent(.setAuthCompletionRoute(accountSecurityRoute))
+                return
+            }
+            coordinator?.navigate(to: accountSecurityRoute)
```
fedemkr, line 390: 🎨 This is doing the same as in `handleOtpAuthUrl`, could the shared part be extracted to a method where we just pass the `AppRoute` and it takes care of the locked logic and navigation using such `AppRoute`?

BitwardenShared/UI/Platform/Application/AppProcessor.swift
brant-livefront: Pulled this into a constant in the latest. 👍

BitwardenShared/UI/Platform/Application/AppProcessor.swift
brant-livefront: The only reason I didn't use a `guard` in the first place is I am planning to add at least one more route in a future PR. So it was cleaner to split it into components and then plan on taking multiple paths. But it was so much cleaner for now as a `guard` I just switched it to a `guard` in the latest. I'll refactor as needed in the future PR where I need to. 👍

BitwardenShared/UI/Platform/Application/AppProcessor.swift
brant-livefront: Refactored in the latest. Now the `handle*Url` methods return `AppRoute?` and we just go through one path to handle whether we need to wait for unlock or not.  👍

BitwardenShared/UI/Platform/Application/AppProcessor.swift
```diff
@@ -115,22 +115,23 @@ public class AppProcessor {
...
-        guard let otpAuthModel = OTPAuthModel(otpAuthKey: url.absoluteString) else {
-            coordinator?.showAlert(.defaultAlert(title: Localizations.anErrorHasOccurred))
-            return
+        var route: AppRoute?
+        if scheme.isBitwardenAppScheme {
+            route = await handleBitwardenUrl(url: url)
+        } else if scheme.isOtpAuthScheme {
+            route = await handleOtpAuthUrl(url: url)
         }
+        guard let route else { return }
```
fedemkr, line 126: 🎨 IMO the func should be renamed given that they do not entirely handle the URL anymore but just get/create their `AppRoute`:
- `handleBitwardenUrl` -> `getBitwardenUrlRoute`
- `handleOtpAuthUrl` -> `getOtpAuthUrlRoute`

⛏️ Additionally, up to you but what would you think about moving the `scheme....` check inside the functions (returning `nil` if the scheme doesn't match) so this can be rewritten as:
```suggestion
        var route = await getBitwardenUrlRoute(url: url) ?? await getOtpAuthUrlRoute(url: url)
        guard let route else { return }
```
IMO this makes it easier to read and to scale if new url handling are needed without having a bunch of `if ... else if`

brant-livefront: I'm up for the name change and I'd be up for moving the scheme check as well. However, I don't think there's a way to make the cleaner nil-check work with concurrency and optimization. 

Doing this:
``` Swift
var route = await getBitwardenUrlRoute(url: url) ?? await getOtpAuthUrlRoute(url: url)
```
fails because `await` can't appear to the right of `??`. I tried moving it inside of a `guard`, but you end up with the same problem of it being an autoclosure that doesn't support concurrency.

This works, but you lose the optimization of not running subsequent calls if the first succeeds:
``` Swift
        let bitWardenRoute = await getBitwardenUrlRoute(url: url)
        let otpAuthRoute = await getOtpAuthUrlRoute(url: url)

        guard let route = bitWardenRoute ?? otpAuthRoute else { return }
```

This one gives you optimization, but I think scales the same or worse that the multiple `if else`s:
``` Swift
        var route = await getBitwardenUrlRoute(url: url)
        if route == nil {
            route = await getOtpAuthUrlRoute(url: url)
        }
        guard let route else { return }
```

Of those options, I think I lean towards the last one, or just going with the `if else` option. Thoughts?

fedemkr: Oh that's a pity that it's in fact an auto closure then it can't be awaited, thanks for checking that and raising it 😞. Yep IMO the last option is fine, the code is shorter and easy to scale.

brant-livefront: Updated to this in the latest. 👍  Agreed that it's a shame this doesn't work. I've been bitten by this same issue a couple of times and it _feels_ like it should work and it would make the code much nicer.

fedemkr (approved):

PR #1053, opened 2024-10-17
[PM-13015] Remember last opened view (View/Edit cipher)

fedemkr:
## 🎟️ Tracking

[PM-13015](https://bitwarden.atlassian.net/browse/PM-13015)

## 📔 Objective

This PR lays the grounds to remember the last opened view from the user when timeout lock happens.
So if user is automatically locked the last opened view with a minimum state will be saved and when the user unlocks, it will navigate to such view loading it.
This is currently constrained to:
- View Cipher
- Edit Cipher

And it's remembered for 5 minutes; after that unlocking will proceed to the default navigation.

## 📸 Screenshots

https://github.com/user-attachments/assets/e381675c-a87b-4e6b-ac48-cba0e6cb89e8

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-13015]: https://bitwarden.atlassian.net/browse/PM-13015?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenShared/UI/Platform/Application/AppProcessor.swift
```diff
@@ -380,6 +381,25 @@ extension AppProcessor {
...
+        guard appExtensionDelegate?.isInAppExtension != true else {
+            return
+        }
+        do {
+            guard let target = try await services.rehydrationHelper.getSavedRehydratableTarget() else {
+                return
+            }
+
+            await coordinator?.handleEvent(.setAuthCompletionRoute(target.appRoute))
+            try await services.rehydrationHelper.clearAppRehydrationState()
```
matt-livefront, line 397: ❓ Is this needed if you also have logic in the router to set the auth completion route?

BitwardenShared/UI/Auth/Extensions/AuthRouter+Redirects.swift
```diff
@@ -361,11 +370,14 @@ extension AuthRouter {
...
 
+                let rehydratableTarget = try await services.rehydrationHelper.getSavedRehydratableTarget()
+
                 return .vaultUnlock(
                     activeAccount,
                     animated: animated,
                     attemptAutomaticBiometricUnlock: attemptAutomaticBiometricUnlock,
-                    didSwitchAccountAutomatically: didSwitchAccountAutomatically
+                    didSwitchAccountAutomatically: didSwitchAccountAutomatically,
+                    rehydratableTarget: rehydratableTarget
```
matt-livefront, line 380: ⚠️ I think there might be a bug here. The `rehydratableTarget` is being set as the auth completion route when you navigate to the vault unlock screen, before the vault is unlocked. If you then switch accounts and unlock that vault, the app will try to rehydrate that route, for the wrong user.

It might be better of to do this in `AuthRouter.completeAuthRedirect()`. That occurs after the user's vault is unlocked, right before navigating to the vault. We could also pass the route to restore as part of the `.complete` `AuthRoute`. The `AuthCoordinator` then calls `delegate?.didCompleteAuth()`, which I think we could use to pass that route and then we wouldn't have to add a separate `setAuthCompletionRoute` delegate method?

https://github.com/bitwarden/ios/blob/e21b142444e0e631de70cbd395593f8f647b09e7/BitwardenShared/UI/Auth/Extensions/AuthRouter%2BRedirects.swift#L27-L39

BitwardenShared/UI/Platform/Application/AppProcessor.swift
fedemkr: Nop, just forgot to remove this part after I implemented the router logic; thanks for caching this!

BitwardenShared/UI/Vault/VaultItem/ViewItem/ViewItemProcessor.swift
```diff
@@ -76,6 +76,12 @@ final class ViewItemProcessor: StateProcessor<ViewItemState, ViewItemAction, Vie
         // When the view is dismissed, ensure any temporary files are deleted.
         services.vaultRepository.clearTemporaryDownloads()
     }
+    
+    // MARK: Public properties
+
+    var rehydrationState: RehydrationState? {
+        RehydrationState(target: .viewCipher(cipherId: itemId))
```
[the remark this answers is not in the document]
fedemkr: If we save the `AppRoute` directly here, then we need to make `AppRoute` conform to `Codable` as well as all its internal usage of other types like `AuthRoute` and many others; that's why I chose this approach as we don't need to worry of such conformance there and only worry about the specific targets we want to rehydrate.
I can definitely change everything to `Codable` if you prefer that approach though.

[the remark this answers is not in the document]
matt-livefront: Ah, I didn't consider that. Your reasoning makes sense! We can always reevaluate if needed, but creating the mapping is straightforward enough.

BitwardenShared/UI/Vault/VaultItem/VaultItemCoordinator.swift
```diff
@@ -402,6 +405,8 @@ class VaultItemCoordinator: NSObject, Coordinator, HasStackNavigator { // swiftl
             timeProvider: services.timeProvider
         )
         stackNavigator?.replace(view)
+
+        services.rehydrationHelper.addRehydratableTarget(processor)
```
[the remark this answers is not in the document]
fedemkr: Yep, I guess it can be moved directly into the processors and add the target on its `init`.

BitwardenShared/UI/Auth/Extensions/AuthRouter+Redirects.swift
fedemkr: Excellent point, thanks for raising this! And as further discussed, changed the approach so the logic is done in `completeAuthRedirect()` and added a new route `.completeWithRehydration(rehydratableTarget)` which passes the target using `delegate?.didCompleteAuth(rehydratableTarget)` to perform the correct navigation after auth completes.

phil-livefront (dismissed):
LGTM!  just need to resolve conflicts but approving nonetheless

BitwardenShared/UI/Auth/Extensions/AuthRouter+Redirects.swift
```diff
@@ -43,6 +42,14 @@ extension AuthRouter {
             }
         }
 
+        do {
+            if let rehydratableTarget = try await services.rehydrationHelper.getSavedRehydratableTarget() {
+                return .completeWithRehydration(rehydratableTarget)
```
matt-livefront, line 47: 🤔 Should this be behind the `if !isInAppExtension` check above? (Maybe the AppCoordinator changes prevent this route from being used in the extension, but might not hurt to also check it here?)

matt-livefront (approved):
This turned out nice! One minor comment but I'll leave that up to you.

BitwardenShared/UI/Auth/Extensions/AuthRouter+Redirects.swift
fedemkr: Yep totally agree, it's safer to do it in there so just moved it, thanks 😄

matt-livefront (approved):

PR #1060, opened 2024-10-19
[BITAU-156] Handle Deep Link From Authenticator to Add a New Item

brant-livefront:
## 🎟️ Tracking

[BITAU-156](https://livefront.atlassian.net/browse/BITAU-156)

related: [BITAU-179](https://livefront.atlassian.net/browse/BITAU-179)

## 📔 Objective

This PR adds a new ability to store a temporary item in the shared core data storage managed by AuthenticatorBridgeKit. This allows for the Authenticator to store a temporary item and call the PM app via deep link to handle the item. This functionality will be used in the Authenticator app to drive things like the option to "Move to BW", the option to save a QR code scan to BW, and the option to save a manual addition to BW.

This PR also refactors the deep link handling to support this new deep link. It will pull the item out of the temporary storage and present it in the Vault selection flow (i.e. how the PM app currently supports handling scanned otpauth:// URLs) so the user can save it permanently to the BW account. 

### Known Issue
In building this out, I found that the current Vault selection flow only supports `OTPAuthModel`. This is fine for its original purpose because that is what we'd get when scanning QR codes of `otpauth://` URLs. However, there are also `steamUri` and `standard` codes that could have been saved locally in the Authenticator app. 

The related ticket above ([BITAU-179](https://livefront.atlassian.net/browse/BITAU-179)) is for refactoring the Vault selection flow to handle `TOTPKeyModel` instead of `OTPAuthModel`. That would give us support for all 3 TOTP key types in the Vault selection.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
```diff
@@ -101,6 +109,13 @@ actor DefaultAuthenticatorSyncService: NSObject, AuthenticatorSyncService {
 
     // MARK: Public Methods
 
+    public func getTemporaryTotpItem() async -> AuthenticatorBridgeItemDataView? {
+        guard await configService.getFeatureFlag(FeatureFlag.enableAuthenticatorSync) else {
+            return nil
+        }
+        return try? await authBridgeItemService.fetchTemporaryItem()
```
fedemkr, line 116: 🎨  If an error is thrown we should report it to the logger/crashlytics so we know there's an error happening in the app.

BitwardenShared/Core/Platform/Utilities/IncomingLinkConstants.swift
```diff
@@ -0,0 +1,14 @@
+// MARK: - IncomingLinksConstants
+
+/// Links that are coming into our app via the "bitwarden://" deep link scheme.
+///
+enum IncomingLinkConstants {
```
fedemkr, line 5: 🎨 If this is only for `bitwarden://` I would rename this enum because a different scheme could also be an "incoming link". Something like:
```suggestion
enum BitwardenDeepLinkConstants {
```

BitwardenShared/UI/Platform/Application/AppProcessor.swift
```diff
@@ -383,14 +383,27 @@ extension AppProcessor {
...
+                                                     message: Localizations.pleaseTryAgain))
+                return nil
+            }
+
+            route = AppRoute.tab(.vault(.vaultItemSelection(otpAuthModel)))
+        default:
+            route = nil
+        }
+
+        return route
```
fedemkr, line 406: 🎨 No need for the extra `route` variable, you can simply return on each case:
```suggestion
        switch url.absoluteString {
        case IncomingLinkConstants.accountSecurity:
            return AppRoute.tab(.settings(.accountSecurity))
        case IncomingLinkConstants.authenticatorNewItem:
            guard let item = await services.authenticatorSyncService?.getTemporaryTotpItem(),
                  let totpKey = item.totpKey,
                  let otpAuthModel = OTPAuthModel(otpAuthKey: totpKey) else {
                coordinator?.showAlert(.defaultAlert(title: Localizations.somethingWentWrong,
                                                     message: Localizations.pleaseTryAgain))
                return nil
            }

            return AppRoute.tab(.vault(.vaultItemSelection(otpAuthModel)))
        default:
            return nil
        }
```

BitwardenShared/UI/Platform/Application/AppProcessor.swift
```diff
@@ -383,14 +383,27 @@ extension AppProcessor {
...
+        var route: AppRoute?
+        switch url.absoluteString {
+        case IncomingLinkConstants.accountSecurity:
+            route = AppRoute.tab(.settings(.accountSecurity))
+        case IncomingLinkConstants.authenticatorNewItem:
+            guard let item = await services.authenticatorSyncService?.getTemporaryTotpItem(),
+                  let totpKey = item.totpKey,
+                  let otpAuthModel = OTPAuthModel(otpAuthKey: totpKey) else {
+                coordinator?.showAlert(.defaultAlert(title: Localizations.somethingWentWrong,
+                                                     message: Localizations.pleaseTryAgain))
```
fedemkr, line 397: 🤔  Is it possible to have a more specific message to the user as to what went wrong? Would trying again actually fix this somehow at any moment for the same `totpKey`?

AuthenticatorBridgeKit/AuthenticatorBridgeItemService.swift
```diff
@@ -28,6 +35,20 @@ public protocol AuthenticatorBridgeItemService {
...
+    /// in, so we save it to a temporary account and retrieve it for processing in the PM app.
+    ///
+    /// The expectation is that only *one* temporary item will be stored at a time. Each time this method is called, it
+    /// will replace the one temporary item. When `fetchTemporaryItem()`is called, it will retrieve only
+    /// the last item stored.
+    ///
+    /// - Parameters:
+    ///   - item: The temporary `AuthenticatorBridgeItemDataModel` to be inserted into the store.
+    ///
+    func insertTemporaryItem(_ item: AuthenticatorBridgeItemDataView) async throws
```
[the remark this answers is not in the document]
brant-livefront: Right now, all of the proposed features are one-at-a-time. Overall, while we call this "sync" it's really more publishing PM app -> Authenticator one way. There's just these one-off places where we allow a code to be moved from Authenticator to PM.

Given that there's a bit of a process of where to store the info coming from Authenticator (i.e. the whole Vault selection process), it's likely to stay one at a time for the time being.

BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
brant-livefront: Updated in latest and added this to the test as well. 👍

BitwardenShared/Core/Platform/Utilities/IncomingLinkConstants.swift
brant-livefront: I was trying to differentiate against external links that might involve Bitwarden. But I think having deeplink in there (plus my doc comment) is probably explanatory enough. 👍  Adopted in latest commit.

BitwardenShared/UI/Platform/Application/AppProcessor.swift
brant-livefront: I like that 👍  Adopted in latest commit.

BitwardenShared/UI/Platform/Application/AppProcessor.swift
brant-livefront: I talked this over with Emily and she had the same question 😆 This is the error we landed on - while trying again won't likely solve the problem, this really shouldn't have happened. If you came here via the correct deeplink and the Authenticator app is storing it correctly in the store, it should be there when we fetch it.

The one exception to this is the known issue I listed above which would be addressed by [BITAU-179](https://livefront.atlassian.net/browse/BITAU-179)

AuthenticatorBridgeKit/AuthenticatorBridgeItemService.swift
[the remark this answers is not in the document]
fedemkr: Awesome thanks for the explanation! 👍

BitwardenShared/UI/Platform/Application/AppProcessor.swift
fedemkr: Alright, if this is only possible from the Authenticator app then it's highly unlikely to happen but if any other app can launch this URL I believe we should improve the messaging to state something about the OTP format being wrong or something like that.

brant-livefront: This particular url will only ever work with the Authenticator app because it depends on an item being added first to the temporary store (which only the Authenticator app can do). The deeplink serves as a way to launch the PM app and tell it to check for the new item (i.e. it's not providing any info via the deeplink like you might from a different app or the camera).

fedemkr: Indeed but I was thinking about the **non-happy path** where some app or something fires this URL and is opened by Bitwarden and given the temp item won't exist the user will see _"Something went wrong, please try again"_ with no idea about why that is shown. Again not 100% sure if any other app can fire this Deep link, can't remember right now if it's somehow configured to work only inside our AppGroup or something like that.

brant-livefront: @fedemkr I talked through this with Emily and she suggested two possible alternatives (keeping the headline of "Something went wrong" the same):
- "Unable to open the selected item. Please try again."
- "Unable to move the selected item to Bitwarden. Please try again."

We both lean towards the first one. Thoughts?

fedemkr: @brant-livefront talked it with Micah and he said if we could know the source application that fired the deep link. I've checked and there's a [sourceApplication](https://developer.apple.com/documentation/uikit/uiscene/openurloptions) in the url options of the scene `urlContexts` options which I think it's exactly what we want.
Also, mentioned the next thing:
> If it turns out that we can verify the source of the request as the authenticator app, use this copy:
“Unable to move the selected item from the Bitwarden Authenticator. Please try again.”
If we can’t, then I would use this copy:
“Unable to move the selected item. Please try again.”

brant-livefront: @fedemkr I've implemented this locally, but I'm not sure it's worth all the added complexity - we need to fetch this in SceneDelegate, pass down to AppProcessor.openUrl, pass through to getBitwardenRoute and then we can do the check:
``` Swift
                let message = source == BitwardenDeepLinkConstants.authenticatorBundleId ?
                    Localizations.unableToMoveTheSelectedItemFromTheBitwardenAuthenticatorPleaseTryAgain :
                    Localizations.unableToMoveTheSelectedItemPleaseTryAgain

                coordinator?.showAlert(.defaultAlert(title: Localizations.somethingWentWrong,
                                                     message: message))
```

But that just gives us a slightly more verbose error message when we verify the source of the link? Am I understanding that correctly? 

Emily and I both lean towards using the "Unable to move the selected item. Please try again.” copy in both cases. If you and Micah feel strongly though, I can push up a commit that adds the additional source check.

fedemkr: @brant-livefront Micah agrees to go with your suggestion on the generic message for both cases, so no need to validate the source. Thanks for all the feedback and sorry for all the back and forth with this.

fedemkr (approved):

PR #1065, opened 2024-10-22
[PM-10900] Fix master password policy bypassed when org user has 2FA

LRNcardozoWDF:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-10900

## 📔 Objective

Fix the flow so the update master password screen is presented to the user. Added the `MasterPasswordPolicyResponseModel` to the `IdentityTokenRequestError` so we can have the master password policies the org has active in order to check if user's master password complies with them.
In case it doesn't comply, the update master password screen will be presented to the user.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


BitwardenShared/Core/Auth/Services/AuthService.swift
```diff
@@ -207,6 +207,7 @@ protocol AuthService {
     func requirePasswordChange(
         email: String,
         masterPassword: String,
+        preAuth: Bool,
```
fedemkr, line 210: ⛏️ Add docs and could we maintain consistency with other places where we are using `isPreAuth` instead of `preAuth`?

BitwardenShared/Core/Auth/Services/AuthService.swift
```diff
@@ -573,17 +578,55 @@ class DefaultAuthService: AuthService { // swiftlint:disable:this type_body_leng
...
+                email: username,
+                captchaToken: captchaToken
+            )
+        } catch let error as IdentityTokenRequestError {
+            if case let .twoFactorRequired(_, _, masterPasswordPolicyResponseModel, _) = error {
+                try await checkMasterPasswordPolicies(masterPassword, true, username, masterPasswordPolicyResponseModel)
+            }
+            // Re-throw the error.
+            throw error
+        }
```
fedemkr, line 595: 🎨 🤔 We're currently handling `IdentityTokenRequestError` inside `getIdentityTokenResponse` and even more we're also handling `.twoFactorRequired(...)` in there as well so in order to maintain that handling as centralized as possible I think this should be moved there. If the issue was that this is only required for login with master password, then just do an if check that the `authenticationMethod` is `.password` and then you can handle your specific case all in the same place.

BitwardenShared/Core/Auth/Services/AuthService.swift
```diff
@@ -573,17 +578,55 @@ class DefaultAuthService: AuthService { // swiftlint:disable:this type_body_leng
...
+    ///
+    /// - Parameters:
+    ///  - masterPassword: The user's master password
+    ///  - preAuth: Whether this flow is before or after authentication
+    ///  - username: The user's email address.
+    ///  - masterPasswordPolicy: The master password policies that the org has active.
+    ///
+    private func checkMasterPasswordPolicies(
+        _ masterPassword: String,
+        _ preAuth: Bool,
```
fedemkr, line 624: ⛏️ Maintain the same naming as we use elsewhere => `isPreAuth`

BitwardenShared/Core/Auth/Services/AuthService.swift
```diff
@@ -594,18 +637,18 @@ class DefaultAuthService: AuthService { // swiftlint:disable:this type_body_leng
...
-                errorReporter.log(error: error)
+        if try await requirePasswordChange(
+            email: username,
+            masterPassword: masterPassword,
+            preAuth: preAuth,
+            policy: policy
+        ) {
+            if preAuth {
+                // Since this is pre authentication we use a local var to cache this info.
+                forcePasswordResetReason = .weakMasterPasswordOnLogin
```
fedemkr, line 648: 🤔 Given this is cached in the service and this is a Singleton, could it happen that the user enters with master password then this cache is set and in the 2FA view the user goes back and changes account and when logging in it applies this reset reason to a different account? In the StateService this is already taken care of because it's tied to the `userId`.

BitwardenShared/Core/Auth/Services/AuthService.swift
```diff
@@ -681,6 +724,12 @@ class DefaultAuthService: AuthService { // swiftlint:disable:this type_body_leng
         // Get the identity token to log in to Bitwarden.
         let response = try await getIdentityTokenResponse(email: email, request: twoFactorRequest)
 
+        // If it's assigned then we need to update the required reset password and remove the cache.
+        if forcePasswordResetReason != nil {
+            try await stateService.setForcePasswordResetReason(.weakMasterPasswordOnLogin)
+            forcePasswordResetReason = nil
```
fedemkr, line 730: 🤔 If the user chooses not to login with the current account, when is `forcePasswordResetReason` cleared?

fedemkr (changes requested):

BitwardenShared/Core/Auth/Services/AuthService.swift
```diff
@@ -612,6 +609,48 @@ class DefaultAuthService: AuthService { // swiftlint:disable:this type_body_leng
...
+    ///  - isPreAuth: Whether this flow is before or after authentication
+    ///  - username: The user's email address.
+    ///  - masterPasswordPolicy: The master password policies that the org has active.
+    ///
+    private func checkMasterPasswordPolicies(
+        _ masterPassword: String,
+        _ isPreAuth: Bool,
+        _ username: String,
+        _ masterPasswordPolicy: MasterPasswordPolicyResponseModel?
+    ) async throws {
```
fedemkr, line 625: 🎨 When multiple parameters are passed to a function we're rarely remove all parameter labels, so I would remove the `_` to be consistent with other places. Additionally, order parameters alphabetically to be consistent with other places as well.

BitwardenShared/Core/Auth/Services/AuthService.swift
```diff
@@ -612,6 +609,48 @@ class DefaultAuthService: AuthService { // swiftlint:disable:this type_body_leng
...
+            policy = MasterPasswordPolicyOptions(
+                minComplexity: model.minComplexity ?? 0,
+                minLength: model.minLength ?? 0,
+                requireUpper: model.requireUpper ?? false,
+                requireLower: model.requireLower ?? false,
+                requireNumbers: model.requireNumbers ?? false,
+                requireSpecial: model.requireSpecial ?? false,
+                enforceOnLogin: model.enforceOnLogin ?? false
+            )
+        }
```
fedemkr, line 637: 🎨 One thing to improve this a bit is to have an `init?(responseModel: MasterPasswordPolicyResponseModel)` in `MasterPasswordPolicyOptions+Extensions` which would take care of building the options or returning `nil`.
```swift
init?(responseModel: MasterPasswordPolicyResponseModel) {
    guard let responseModel else { return nil }
    self.init(
        minComplexity: responseModel.minComplexity ?? 0,
        minLength: responseModel.minLength ?? 0,
        requireUpper: responseModel.requireUpper ?? false,
        requireLower: responseModel.requireLower ?? false,
        requireNumbers: responseModel.requireNumbers ?? false,
        requireSpecial: responseModel.requireSpecial ?? false,
        enforceOnLogin: responseModel.enforceOnLogin ?? false
    )
}
```
So then you can simplify this to:
```swift
let policy = MasterPasswordPolicyOptions(responseModel: masterPasswordPolicy)
```

BitwardenShared/Core/Auth/Services/AuthService.swift
```diff
@@ -851,15 +898,17 @@ class DefaultAuthService: AuthService { // swiftlint:disable:this type_body_leng
...
                 // If the token request require two-factor authentication, cache the request so that
                 // the token information can be added once the user inputs the code.
                 twoFactorRequest = request
                 twoFactorRequest?.captchaToken = captchaBypassToken
 
                 // Form the resend email request in case the user needs to resend the verification code email.
                 var passwordHash: String?
-                if case let .password(_, password) = request?.authenticationMethod { passwordHash = password }
+                if case let .password(_, password) = request?.authenticationMethod { passwordHash = password
+                    try await checkMasterPasswordPolicies(masterPassword!, true, email, masterPasswordPolicyResponseModel)
```
fedemkr, line 910: ⚠️ We try never to use `!` as it can be dangerous if, for some reason `masterPassword` doesn't have a value, so add an `if let` check or add a `guard let` check which logs the error if that happens depending on whether you consider an error should be logged if `masterPassword` is `nil` and provide some description on such error so if we encounter that in Crashlytics we understand the context.

fedemkr (changes requested):
Looks better, found some things to improve. Also would love if @matt-livefront and/or @KatherineInCode reviews this in case I've missed something here given that it affects the login flow.

BitwardenShared/Core/Auth/Services/AuthService.swift
KatherineInCode: It's worth noting that on the whole the `_` convention was somewhat intended as a replacement for the old Objective-C `[object with:thing and:otherThing]` idea. Early Swift as I recall even turned this into `object(with:thing and:otherThing)`, which evolved into the `object(thing, and: otherThing)` that's more common today.

BitwardenShared/Core/Auth/Models/Response/IdentityTokenErrorModel.swift
```diff
@@ -44,13 +40,17 @@ struct IdentityTokenErrorModel: Codable {
...
+        case authenticator = "0"
         case email = "1"
         case duo = "2"
         case organizationDuo = "6"
         case yubikey = "3"
         case webAuthn = "7"
     }
 
+    /// Information for two factor authentication with Authenticator
+    let authenticator: Bool
```
fedemkr, line 52: ⛏️ In this case is whether it's enabled or not:
```suggestion
    /// Whether authenticator provider is enabled.
    let authenticator: Bool
```

BitwardenShared/Core/Auth/Services/API/Auth/Requests/IdentityTokenRequest.swift
```diff
@@ -80,9 +80,7 @@ struct IdentityTokenRequest: Request {
                 from: response.body
             ) else { return }
 
-            if let twoFactorProviders = errorModel.twoFactorProviders,
-               var twoFactorProvidersData = errorModel.twoFactorProvidersData {
-                twoFactorProvidersData.providersAvailable = twoFactorProviders
+            if let twoFactorProvidersData = errorModel.twoFactorProvidersData {
```
fedemkr, line 83: 🤔 I think we should also be checking `let errorModel.twoFactorProvidersData.providersAvailable` to actually check there are 2FA providers available in the response data. Because if the response sends an empty 2FA providers data, then the current code won't fail and it should.

BitwardenShared/Core/Auth/Services/AuthService.swift
```diff
@@ -904,11 +899,28 @@ class DefaultAuthService: AuthService { // swiftlint:disable:this type_body_leng
                 twoFactorRequest = request
                 twoFactorRequest?.captchaToken = captchaBypassToken
 
-                // Form the resend email request in case the user needs to resend the verification code email.
                 var passwordHash: String?
                 if case let .password(_, password) = request?.authenticationMethod { passwordHash = password
-                    try await checkMasterPasswordPolicies(masterPassword!, true, email, masterPasswordPolicyResponseModel)
+                    // Ensure masterPassword has a value before proceeding.
+                    guard let masterPassword else {
+                        // Otherwise, log the error.
```
fedemkr, line 906: ⛏️ No need for this comment as it's obvious.
```suggestion
```

fedemkr (commented):
Looks good, just some small things to adjust.
Also, check the Files annotations given that there are some from Codecov worth noting regarding missing tests; although I'll leave that up to you to which ones are important to add.

BitwardenShared/Core/Auth/Services/AuthService.swift
```diff
@@ -931,6 +930,12 @@ class DefaultAuthService: AuthService { // swiftlint:disable:this type_body_leng
                 // If this error was thrown, it also means any cached two-factor token is not valid.
                 await stateService.setTwoFactorToken(nil, email: email)
             }
+            if case .twoFactorProvidersNotConfigured = error {
+                errorReporter.log(error: BitwardenError.generalError(
+                    type: "Identity Token Validation Failed.",
+                    message: "Two Factor Providers data is nil."
```
fedemkr, line 936: ⛏️ 
```suggestion
                    message: "Two Factor Providers data is nil or empty."
```

BitwardenShared/UI/Auth/Login/LoginProcessor.swift
```diff
@@ -159,6 +159,8 @@ class LoginProcessor: StateProcessor<LoginState, LoginAction, LoginEffect> {
                 coordinator.navigate(
                     to: .twoFactor(state.username, .password(state.masterPassword), authMethodsData, nil)
                 )
+            case .twoFactorProvidersNotConfigured:
+                await handleErrorResponse(error)
```
fedemkr, line 163: ⚠️ Either you handle the error (and it's logging here) or in the service. Now it's going to log the error twice, once in the service and once here. If you want just remove the check in the service and leave this one since you have a specific error case it's going to be easy to pin point the origin of the error.

BitwardenShared/UI/Auth/Login/LoginWithDevice/LoginWithDeviceProcessor.swift
```diff
@@ -191,6 +191,8 @@ final class LoginWithDeviceProcessor: StateProcessor<
                     nil
                 }
                 coordinator.navigate(to: .twoFactor(state.email, unlockMethod, authMethodsData, nil))
+            case .twoFactorProvidersNotConfigured:
+                services.errorReporter.log(error: error)
```
fedemkr, line 195: ⚠️ Either you handle the error (and it's logging here) or in the service. Now it's going to log the error twice, once in the service and once here. If you want just remove the check in the service and leave this one since you have a specific error case it's going to be easy to pin point the origin of the error.

BitwardenShared/UI/Auth/Login/LoginProcessor.swift
LRNcardozoWDF: I chose to handle the error here and remove the check in the service.

BitwardenShared/UI/Auth/Login/LoginWithDevice/LoginWithDeviceProcessor.swift
LRNcardozoWDF: I chose to handle the error here and remove the check in the service.

fedemkr (approved):
LGTM, but just in case wait for a second approval from @matt-livefront or @KatherineInCode in case I've missed something given that this affects directly to the login flow.

matt-livefront (approved):

BitwardenShared/Core/Auth/Models/Response/MasterPasswordPolicyResponseModel.swift
```diff
@@ -1,6 +1,19 @@
...
+    /// Key names used for encoding and decoding.
+    enum CodingKeys: String, CodingKey {
+        case enforceOnLogin = "EnforceOnLogin"
+        case minComplexity = "MinComplexity"
+        case minLength = "MinLength"
+        case requireLower = "RequireLower"
+        case requireNumbers = "RequireNumbers"
+        case requireSpecial = "RequireSpecial"
+        case requireUpper = "RequireUpper"
+    }
```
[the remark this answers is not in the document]
fedemkr: @andrebispo5 I believe this is because `IdentityTokenErrorModel` doesn't conform to `JSONResponse`, thus it doesn't use the `pascalOrSnakeCaseDecoder`.
I think we could just change: 
`try? JSONDecoder().decode` 
for 
`try? JSONDecoder.pascalOrSnakeCaseDecoder.decode`
in here and that should fix the issue without the `CodingKeys`
https://github.com/bitwarden/ios/blob/1f97eabc83c4c8d61109036bf34af574d32f3c4e/BitwardenShared/Core/Auth/Services/API/Auth/Requests/IdentityTokenRequest.swift#L80-L84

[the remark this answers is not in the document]
fedemkr: do you agree @matt-livefront ?

PR #1070, opened 2024-10-23
[PM-11213] Update to Xcode 16.1 and iOS 18.1

KatherineInCode:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-11213

## 📔 Objective

This updates our workflows and code to use Xcode 16.1 for compilation and testing, and then to use an iPhone 15 Pro running iOS 18.1 for the tests.

This required making some changes to our ViewInspector tests, and they no longer will work on Xcode 15.4. Some workarounds had to be done to handle Xcode 16 [adding `AnyView` objects to the view hierarchy](https://indieweb.social/@curtclifton/113273571392595819). On the whole, I tried to create recursive or shadow methods that required minimal change to tests, however there were a few places where I couldn't, so had to add the `.implicitAnyView()` calls manually.

~~It might be reasonable to scale this back to just Xcode 16.1, but leave tests on an iPhone 15 Pro with iOS 18; the biggest difference is with iPhone type, because the [iPhone 16 Pro has more pixels than an iPhone 15 Pro](https://www.ios-resolution.com/), so the snapshots will be bigger.~~ We decided [in Slack](https://bitwarden.slack.com/archives/C05ENUBLMD5/p1730310680329679?thread_ts=1730304636.049669&cid=C05ENUBLMD5) to stick with the iPhone 15 Pro for now.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -56,8 +56,8 @@ body:
     attributes:
       label: Environment Details
       placeholder: |
-        - Device: [e.g. iPhone 15 Pro, iPad Air (5th Generation)]
-        - OS Version: [e.g. 17.4.1]
+        - Device: [e.g. iPhone 16 Pro, iPad Air (5th Generation)]
+        - OS Version: [e.g. 18.0.1]
```
fedemkr, line 60: ⛏️ This should be `18.1` as far as I can see in `.test-simulator-ios-version`

GlobalTestHelpers/Support/BitwardenTestCase.swift
```diff
@@ -9,10 +9,10 @@ open class BitwardenTestCase: XCTestCase {
 
     @MainActor
     override open class func setUp() {
-        if UIDevice.current.name != "iPhone 15 Pro" {
+        if UIDevice.current.name != "iPhone 16 Pro" {
```
fedemkr, line 12: 🤔 Could we add the OS version check here as well? So it's easier to spot someone is using the wrong simulator OS version to perform the tests.

fedemkr (commented):
Looks good! Just two small things.

.github/ISSUE_TEMPLATE/bug.yml
KatherineInCode: I intentionally made it different, because it's giving it as an example

GlobalTestHelpers/Support/BitwardenTestCase.swift
KatherineInCode: We certainly could, though given that people have more reason to change which version of iOS they're running tests again—and much less possibility for failures to happen as a result—I'm not sure it makes sense to block it like we want to do with the device, because that can pretty easily cause a cascade of failed tests. I'm inclined to just check device, and not also version.

.github/ISSUE_TEMPLATE/bug.yml
fedemkr: Ohhh makes sense!

GlobalTestHelpers/Support/BitwardenTestCase.swift
fedemkr: I'm not sure I understand why you say that can easily cause a cascade of failed tests if we're blocking the exact version in which the tests should run on. Furthermore, this is using `assertionFailure` so the first test to run will just throw `EXC_BREAKPOINT` which halts the execution and avoid running any others, marking this spot as the reason so you can easily see that it's because of either the simulator device or version used.

Additionally, some other reasons why we should add the OS version:
- Even though I haven't done a survey on this I would think people usually use the device + OS version indicated to run the snapshot tests so you can run the whole suite and check whether some test broke (at least that's what I usually do).
- If we don't force the OS version, people might take new snapshots in the wrong OS version which, unless they run the whole suite, they won't realize about it until CI runs and let them know the snapshots don't match with the ones taken on the CI's simulator; wasting CI process time and dev time making another round of taking snapshots and pushing again to the repo.
- Not forcing the version also might let you think there could be some snapshot tests wrong and may let you waste valuable time checking what's wrong between snapshots when in fact is just the OS version mismatch. Additionally, one may re-take the snapshot and then find out it's wrong when testing in CI.
- If someone needs to run the tests in another Device/OS version they can easily comment the line and that's it. We can even document this so people know it beforehand.

I'm thinking on changing a bit this when I have a bit of time so we can use Test Plans or something like that so we can force this device + OS version only in Snapshot tests but until that happens I believe we should force the OS version to save time.

What do you think?

KatherineInCode: You're right that it doesn't cascade because it just halts execution; it's mostly in my mind I think of that as "everything fails" 😅 though it's still really frustrating to hit when I've been intentionally working on some other device simulator in order to check how things look on that device, and then run tests and it just stops.

On the whole, especially with changing the screenshot threshold to 0.99, changing iOS version doesn't have an impact on the snapshot tests. For instance, the current tests pass on both an 18.0 and 18.1 simulator (but not 17.* only because of changes to the backgrounds on the `UITabBarControllerExtensions`.

Realistically, it's not that we need an iPhone 15 Pro; it's just any iPhone with a 3x scale. Especially now that I've made the `UIViewController` tests stick to the iPhone 13 screen size, the actual device should matter a lot less.

That said, if we still are specifying an exact device despite not needing to, I suppose it makes sense to also specify an iOS version despite not needing to, just to enforce it further. Though I do like the idea of using Test Plans or something along those lines to only do this check specifically on snapshot tests (and I almost find myself thinking it would be better to skip them rather than halt execution, but I'd have to think that through more before recommending it).

I'll add it in; if it gets too bothersome in the future we can always revisit.

fedemkr: Makes total sense and I didn't remember about the change in threshold and view controllers that's great, Thanks 🎉 !

fedemkr (approved):

GlobalTestHelpers/Support/BitwardenTestCase.swift
KatherineInCode: I only figured out the view controller thing this morning. I was so caught up on `device` yesterday I didn't think to look for `size`, but when I took a look this morning I was like "oh!" so made that change and verified that it now works better between different devices.

So ironically, we actually might want to be more restrictive to iOS version than device type, because that can have a larger impact on rendering.

PR #1072, opened 2024-10-23
PM-13424 - Show error when logging into an unofficial Bitwarden server 

phil-livefront:
## 🎟️ Tracking

[PM-13424](https://bitwarden.atlassian.net/browse/PM-13424)

## 📔 Objective
- Displays a specific error message when login fails against an unrecognized Bitwarden server, helping users identify the issue. This message indicates that the server may not be a genuine Bitwarden instance, suggesting they check with their provider or update their server URL.
- This follows the same login for the Android implementation ([PR](https://github.com/bitwarden/android/pull/4088)) where we are only catching login related errors for now.
- Slack [thread](https://livefront.slack.com/archives/C05ENUBLMD5/p1729008312476519?thread_ts=1728570245.277469&cid=C05ENUBLMD5) regarding login related actions being a good place to start.

### Implementation Details
After conversation in [Slack](https://bitwarden.slack.com/archives/C05ENUBLMD5/p1728570245277469), we decided to use the presence of `ConfigResponse.Server` property to determine if we are communicating with an official Bitwarden server or not.

- When `ConfigResponse.Server` is null we assume it is an official Bitwarden server
- When `ConfigResponse.Server` is not null we assume it is an unofficial Bitwarden server.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-13424]: https://bitwarden.atlassian.net/browse/PM-13424?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenShared/UI/Platform/Application/Support/Localizations/en.lproj/Localizable.strings
```diff
@@ -1049,3 +1049,4 @@
 "NoAccountFoundPleaseLogInAgainIfYouContinueToSeeThisError" = "No account found. Please log in again if you continue to see this error.";
 "ImportError" = "Import error";
 "NoLoginsWereImported" = "No logins were imported";
+"ThisIsNotARecognizedBitwardenServer" = "This is not a recognized Bitwarden server";
```
KatherineInCode, line 1052: 🤔 Should this end with a period? It also doesn't seem to be the same as we have on the Android side.

BitwardenShared/UI/Platform/Application/Extensions/Alert+NetworkingTests.swift
```diff
@@ -44,6 +44,21 @@ class AlertNetworkingTests: BitwardenTestCase {
         XCTAssertNil(action.handler)
     }
 
+    /// `.networkResponseError` builds an alert to display a server error message for an unofficial Bitwarden server..
```
KatherineInCode, line 47: ⛏️ 
```suggestion
    /// `.networkResponseError` builds an alert to display a server error message for an unofficial Bitwarden server.
```

BitwardenShared/UI/Auth/Login/LoginProcessor.swift
```diff
@@ -185,10 +184,26 @@ class LoginProcessor: StateProcessor<LoginState, LoginAction, LoginEffect> {
...
     }
+
+    /// Handles error responses by checking if the Bitwarden server is official or unofficial.
+    /// If the server is unofficial, it assigns a custom error; otherwise, it uses the provided error.
+    /// Displays an alert for the error and logs it using the error reporter.
+    ///
+    /// - Parameter errorResponse: The error received from the network response.
+    ///
+    private func handleErrorResponse(_ errorResponse: Error) async {
+        let error = if await services.configService.getConfig()?.isOfficialBitwardenServer() == false {
```
KatherineInCode, line 198: 🤔 My concern here is that we might eat errors not related to it being an unofficial server, presenting them as an unofficial server error. Or would we have handled other errors (such as "incorrect username/password" before this, rendering that concern moot?

BitwardenShared/UI/Platform/Application/Support/Localizations/en.lproj/Localizable.strings
phil-livefront: ah good catch! thanks!

BitwardenShared/UI/Auth/Login/LoginProcessor.swift
phil-livefront: this is a good point...  If the `server` property isn't `nil`, regardless of which error is given back from the server then we could display the new error.  Im wondering if we should check if its a `validationError(error:)` before checking the `server` value?

phil-livefront: Does this make sense?

```
private func handleErrorResponse(_ errorResponse: Error) async {
        if case let serverError as ServerError = errorResponse,
           case .validationError = serverError {
            coordinator.showAlert(.networkResponseError(serverError))
            services.errorReporter.log(error: serverError)
            return
        }

        // Check if the server is unofficial, otherwise use the original error
        let error = await services.configService.getConfig()?.isOfficialBitwardenServer() == false
            ? ServerError.unofficialBitwardenServerError
            : errorResponse

        // Show the appropriate alert and log the error
        coordinator.showAlert(.networkResponseError(error))
    }
```

phil-livefront: hmmm actually I think i need to think about this more...  We probably only want to show the unofficial server error is we cant deserialize an existing error?

phil-livefront: This ended up feeling safe but lemme know if you find any issues with this logic

```
   private func handleErrorResponse(_ errorResponse: Error) async {
        let error = if case let serverError as ServerError = errorResponse {
            serverError
        } else {
            await services.configService.getConfig()?.isOfficialBitwardenServer() == false
                ? ServerError.unofficialBitwardenServerError
                : errorResponse
        }

        coordinator.showAlert(.networkResponseError(error))
    }
```

KatherineInCode: The ticket itself indicates that it's only when we get an unknown error—so yeah, if we get a *better* error from the server, we should use that; this is to cover the situations we have where the older server version is sending something we can't properly parse in some way, or other things we haven't anticipated yet

BitwardenShared/UI/Auth/Login/LoginProcessor.swift
```diff
@@ -185,10 +184,28 @@ class LoginProcessor: StateProcessor<LoginState, LoginAction, LoginEffect> {
...
+        let error = if case let serverError as ServerError = errorResponse {
+            serverError
+        } else {
+            await services.configService.getConfig()?.isOfficialBitwardenServer() == false
+                ? ServerError.unofficialBitwardenServerError
+                : errorResponse
+        }
+
+        coordinator.showAlert(.networkResponseError(error))
+        services.errorReporter.log(error: error)
```
matt-livefront, line 207: 🤔 One other option here might be to pass `isOfficialBitwardenServer` into the `networkResponseError(_:)` method (e.g. `coordinator.showAlert(.networkResponseError(error, isOfficialBitwardenServer: isOfficialBitwardenServer)`) and let that function handle whether it should show the unofficial server error. Because it also handles things like URLErrors (e.g. request timed out, no internet), which we probably want to take priority.

phil-livefront: yeah thats a great idea!

KatherineInCode: I like that suggestion as well

BitwardenShared/UI/Platform/Application/Extensions/Alert+Networking.swift
```diff
@@ -35,6 +35,7 @@ extension Alert {
     ///
     static func networkResponseError(
         _ error: Error,
+        isOfficialBitwardenServer: Bool = true,
```
matt-livefront, line 38: ⛏️ Can you add this to the docs above?

BitwardenShared/UI/Auth/Login/LoginProcessor.swift
```diff
@@ -185,8 +184,29 @@ class LoginProcessor: StateProcessor<LoginState, LoginAction, LoginEffect> {
...
+    /// along with the error to the coordinator to display an appropriate alert on the main thread.
+    /// The error is also logged using the error reporter.
+    ///
+    /// - Parameter error: The error received from the network request.
+    ///
+    private func handleErrorResponse(_ error: Error) async {
+        let isOfficialBitwardenServer = await services.configService.getConfig()?.isOfficialBitwardenServer() ?? true
+        services.errorReporter.log(error: error)
+
+        await MainActor.run {
```
matt-livefront, line 203: 🤔 Is this needed if the processor is already `@MainActor`?

BitwardenShared/UI/Auth/Login/LoginProcessor.swift
```diff
@@ -185,8 +184,29 @@ class LoginProcessor: StateProcessor<LoginState, LoginAction, LoginEffect> {
...
+    /// Handles network error responses.
+    ///
+    /// Determines whether the Bitwarden server is official or unofficial and passes this information
+    /// along with the error to the coordinator to display an appropriate alert on the main thread.
+    /// The error is also logged using the error reporter.
+    ///
+    /// - Parameter error: The error received from the network request.
+    ///
+    private func handleErrorResponse(_ error: Error) async {
+        let isOfficialBitwardenServer = await services.configService.getConfig()?.isOfficialBitwardenServer() ?? true
```
matt-livefront, line 200: 🤔 This doesn't seem to be working for me. I think you might need to pass `isPreAuth: true` when getting the config.

BitwardenShared/UI/Auth/Login/LoginProcessor.swift
phil-livefront: oh yeah, 🧠 💨 on my end

BitwardenShared/UI/Auth/Login/LoginProcessor.swift
phil-livefront: hmmm yeah, i had that originally and thought i didnt need it so i removed it... but maybe i was in another weird cache situation.  good catch.

matt-livefront (approved):
Looks good!

BitwardenShared/UI/Auth/Landing/LandingProcessor.swift
```diff
@@ -224,9 +233,6 @@ extension LandingProcessor: RegionDelegate {
         guard !urls.isEmpty else { return }
         await services.environmentService.setPreAuthURLs(urls: urls)
         state.region = region
-        // After setting a new region, feature flags need to be reloaded
-        Task {
-            await loadFeatureFlag()
-        }
+        await refreshConfig()
```
fedemkr, line 236: ⚠️ @phil-livefront This introduces several issues on slow internet connection:
1. On [perform](https://github.com/bitwarden/ios/blob/main/BitwardenShared/UI/Auth/Landing/LandingProcessor.swift#L67) `.appeared` It won't call  `refreshProfileState` because it's going to await until `refreshConfig` finishes.
2. When setting a Self-host environment it won't close the Self-host settings view until `refreshConfig` finishes given that it's called in [func didSaveEnvironment(urls: EnvironmentUrlData) async](https://github.com/bitwarden/ios/blob/main/BitwardenShared/UI/Auth/Landing/LandingProcessor.swift#L208)

I'd go back to wrap the config refresh call in a `Task` so we don't have these issues and if we **really need** to wait for the config for some thing, we'd need to use some other approach, but given this is only for feature flags I'm not that concerned that is not loaded correctly on low internet connection as we have fallback flows for `emailVerificationFeatureFlag`.

PR #1080, opened 2024-10-25
[PM-14033] Configure test workflow simulator type and iOS version with config files

KatherineInCode:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-14033

## 📔 Objective

Similar to how we now have an `.xcode-version` file to denote which version of Xcode should be used when building/testing, this creates `.test-simulator-device-name` and `.test-simulator-ios-version` files to denote which device tests should be run against and which iOS version on that device, respectively.

I went back and forth on whether to make this two files like this, or one file with the full `name=[blah],OS=[blah]` string; I went with the former because it's more straightforward, and easier to potentially use in other workflows/tools in the future.

It would be nice if there were a way to use this in [BitwardenTestCase](https://github.com/bitwarden/ios/blob/main/GlobalTestHelpers/Support/BitwardenTestCase.swift#L12) as well, but I don't think that's readily possible without pre-build cleverness unfortunately.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


.github/workflows/test.yml
```diff
@@ -72,6 +72,8 @@ jobs:
       - name: Read default Xcode version
         run: |
           echo "DEFAULT_XCODE_VERSION=$(cat .xcode-version | tr -d '\n')" >> "$GITHUB_ENV"
+          echo "DEFAULT_SIMULATOR_NAME=$(cat .test-simulator-device-name | tr -d '\n')" >> "$GITHUB_ENV"
+          echo "DEFAULT_SIMULATOR_VERSION=$(cat .test-simulator-ios-version | tr -d '\n')" >> "$GITHUB_ENV"
```
fedemkr, line 76: 🤔 Maybe the name of the step should be updated given that it does more than just the Xcode version. What do you think?

fedemkr (commented):
🤔 Looks good! About the configuration being in different files I was thinking that most likely these files will be updated alongside when the Xcode version changes so IMO we could have everything on one file and use the `source` / `.` command to load if the file looks something like this:
```
XCODE_VERSION=15.4
SIMULATOR_NAME=iPhone 15 Pro
SIMULATOR_VERSION=18.0
```
So we just do `. /path/to/theconfigfile` and the variables will be automatically available and ready for use in the step to be added to `$GITHUB_ENV` without needing to do any `cat` or `sed` commands.
Thoughts?

.github/workflows/test.yml
KatherineInCode: 🤦🏼‍♀️ It is in the Xcode 16 branch, I just didn't bring it over here

KatherineInCode:
> 🤔 Looks good! About the configuration being in different files I was thinking that most likely these files will be updated alongside when the Xcode version changes so IMO we could have everything on one file and use the `source` / `.` command to load if the file looks something like this:

I considered that option, too, and went down a long rabbit hole of how to import environment variables from a file into GitHub actions.

The thing is, the `.xcode-version` file is [something of a standard](https://github.com/fastlane/ci/blob/master/docs/xcode-version.md), and used by other things in the ecosystem (particularly Fastlane). So there's benefit to having that as a standalone file itself, which is why ultimately I [did it that way](https://github.com/bitwarden/ios/pull/1029).

Ultimately it's probably six of one, half dozen of another. If I had a good reason for abandoning the `.xcode-version` standard, then yeah, I'd be inclined to put them all in one file.

fedemkr (approved):
Agree makes total sense, thanks for explaining @KatherineInCode ! 🎉

PR #1091, opened 2024-10-30
PM-14259: Fix never lock timeout value doesn't reset on logout

matt-livefront:
## 🎟️ Tracking

[PM-14259](https://bitwarden.atlassian.net/browse/PM-14259)

## 📔 Objective

If a user has the never timeout value selected and logs out, upon logging into that account again the app was showing their timeout still as never timeout, even though the key doesn't exist to unlock the vault. This fixes that so the timeout is reset to 15 minutes upon logging in again.

## 📸 Screenshots

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/0378182f-7e43-4e03-930c-217b778fb134" /> | <video src="https://github.com/user-attachments/assets/15771c7b-d769-4a94-ad24-67b8a07d1efb" /> |

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-14259]: https://bitwarden.atlassian.net/browse/PM-14259?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


phil-livefront (approved):
Looks good!

BitwardenShared/Core/Platform/Services/StateService.swift
```diff
@@ -1494,11 +1494,19 @@ actor DefaultStateService: StateService { // swiftlint:disable:this type_body_le
...
+            // If there isn't a stored value, it may be because MAUI stored `nil` for never timeout.
+            // So if the never lock key exists, set the timeout to never.
+            userAuthKey != nil ? .never : .fifteenMinutes
+        }
+        if timeoutValue == .never, userAuthKey == nil {
+            // If never lock but no key (possibly due to logging out), return the default timeout.
+            return .fifteenMinutes
         }
-        return SessionTimeoutValue(rawValue: rawValue)
+        return timeoutValue
```
fedemkr, line 1509: 🎨 🤔 I'll leave it up to you but I think this could be written in the next way to reduce cyclomatic complexity and be easier to folllow:
```swift
let userAuthKey = try? await keychainRepository.getUserAuthKeyValue(for: .neverLock(userId: userId))
guard let rawValue = appSettingsStore.vaultTimeout(userId: userId) else {
    // If there isn't a stored value, it may be because MAUI stored `nil` for never timeout.
    // So if the never lock key exists, set the timeout to never, otherwise to default.
    return userAuthKey != nil ? .never : .fifteenMinutes
}

let timeoutValue = SessionTimeoutValue(rawValue: rawValue)
if timeoutValue == .never, userAuthKey == nil {
    // If never lock but no key (possibly due to logging out), return the default timeout.
    return .fifteenMinutes
}
return timeoutValue
```
Given that `userAuthKey != nil ? .never : .fifteenMinutes` will cause the `if timeoutValue == .never, userAuthKey == nil` to never apply so we can return early in that case.
What do you think?

matt-livefront: I like that!

fedemkr (approved):

PR #1122, opened 2024-11-08
PM-14646: Fix JSON decoding errors

matt-livefront:
## 🎟️ Tracking

[PM-14646](https://bitwarden.atlassian.net/browse/PM-14646) - Fix JSON decoding errors for invalid CipherRepromptType, SecureNoteType, decimal policy timeout minutes
[PM-13105 ](https://bitwarden.atlassian.net/browse/PM-13105) - [iOS] Users getting "An Error has occurred" and can't access their vaults

## 📔 Objective

This fixes the following JSON decoding errors:

Invalid CipherRepromptType:
```
Cannot initialize CipherRepromptType from invalid Int value 65
```

Invalid SecureNoteType:
```
Cannot initialize SecureNoteType from invalid Int value 2
```

Unexpected decimal when decoding a vault timeout policy minutes property:
```
( "CodingKeys(stringValue: \"policies\", intValue: nil)", "_JSONKey(stringValue: \"Index 7\", intValue: 7)", "CodingKeys(stringValue: \"data\", intValue: nil)", "_JSONKey(stringValue: \"minutes\", intValue: nil)" )
Unable to decode AnyCodable value.
```

Fixes:

- For the two enum errors, I added a `@DefaultValue` property wrapper that will use a default value if the item is unable to be decoded, either because it's missing or invalid. Both of these enums have reasonable default values that we could use in the case of an invalid value.
- For the policy enum, I added support for `AnyCodable` to decode `Double` values. This is then used to convert the value to an `Int` if that's what we're expecting when we attempt to use the value.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-14646]: https://bitwarden.atlassian.net/browse/PM-14646?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/Core/Platform/Utilities/DefaultValue.swift
+++ b/BitwardenShared/Core/Platform/Utilities/DefaultValue.swift
@@ -35,14 +35,14 @@ extension DefaultValue: Decodable {
             if let intValue = try? container.decode(Int.self) {
                 Logger.application.warning(
                     """
-                    Cannot initialize \(T.self) from invalid Int value \(intValue), \
+                    Cannot initialize \(T.self) from invalid Int value \(intValue, privacy: .private), \
                     defaulting to \(String(describing: T.defaultValue)).
                     """
                 )
             } else if let stringValue = try? container.decode(String.self) {
                 Logger.application.warning(
                     """
-                    Cannot initialize \(T.self) from invalid String value \(stringValue), \
+                    Cannot initialize \(T.self) from invalid String value \(stringValue, privacy: .private), \
                     defaulting to \(String(describing: T.defaultValue))
                     """
                 )
--- a/BitwardenShared/Core/Platform/Utilities/DefaultValue.swift
+++ b/BitwardenShared/Core/Platform/Utilities/DefaultValue.swift
@@ -1,0 +1,99 @@
+import OSLog
+
+// MARK: - DefaultValueProvider
+
+/// A protocol for defining a default value for a `Decodable` type if an invalid or missing value
+/// is received.
+///
+protocol DefaultValueProvider: Decodable {
+    /// The default value to use if the value to decode is invalid or missing.
+    static var defaultValue: Self { get }
+}
+
+// MARK: - DefaultValue
+
+/// A property wrapper that will default the wrapped value to a default value if decoding fails.
+/// This is useful for decoding types which may not be present in the response or to prevent a
+/// decoding failure if an invalid value is received.
+///
+@propertyWrapper
+struct DefaultValue<T: DefaultValueProvider> {
+    // MARK: Properties
+
+    /// The wrapped value.
+    let wrappedValue: T
+}
+
+// MARK: - Decodable
+
+extension DefaultValue: Decodable {
+    init(from decoder: any Decoder) throws {
+        let container = try decoder.singleValueContainer()
+        do {
+            wrappedValue = try container.decode(T.self)
+        } catch {
+            if let intValue = try? container.decode(Int.self) {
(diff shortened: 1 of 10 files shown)


BitwardenShared/Core/Platform/Utilities/DefaultValue.swift
```diff
@@ -0,0 +1,99 @@
...
+extension DefaultValue: Decodable {
+    init(from decoder: any Decoder) throws {
+        let container = try decoder.singleValueContainer()
+        do {
+            wrappedValue = try container.decode(T.self)
+        } catch {
+            if let intValue = try? container.decode(Int.self) {
+                Logger.application.warning(
+                    """
+                    Cannot initialize \(T.self) from invalid Int value \(intValue), \
```
fedemkr, line 38: 🤔 I think we should be using [OSLogPrivacy private](https://developer.apple.com/documentation/os/oslogprivacy/3578096-private) instead of the default [auto](https://developer.apple.com/documentation/os/oslogprivacy/3578094-auto) here where the system decides whether to show or redact the `intValue`. Or at least choose which one to use depending on whether one is on debugging.

BitwardenShared/Core/Platform/Utilities/DefaultValue.swift
```diff
@@ -0,0 +1,99 @@
...
+                    defaulting to \(String(describing: T.defaultValue)).
+                    """
+                )
+            } else if let stringValue = try? container.decode(String.self) {
+                Logger.application.warning(
+                    """
+                    Cannot initialize \(T.self) from invalid String value \(stringValue), \
```
fedemkr, line 45: 🤔 I think we should be using [OSLogPrivacy private](https://developer.apple.com/documentation/os/oslogprivacy/3578096-private) instead of the default [auto](https://developer.apple.com/documentation/os/oslogprivacy/3578094-auto) here where the system decides whether to show or redact the `stringValue`. Or at least choose which one to use depending on whether one is on debugging.
IMO it's more important here than the int one as we should always hide string values from logs when we don't actually know if it's safe to potentially be displayed.

BitwardenShared/Core/Platform/Utilities/DefaultValue.swift
matt-livefront: Ok I will update it.

BitwardenShared/Core/Platform/Utilities/DefaultValue.swift
matt-livefront: My understanding is that strings are always hidden when interpolated but I can specify the privacy level.

KatherineInCode: That's my understanding as well, but I don't think there's any harm in explicitly specifying it, especially if someone is combing through the code later and doesn't know about the default being "hide".

KatherineInCode (approved):
👍🏻 I like it! Makes us more robust.

My only thought is if there might be other places we want to make something be `@DefaultValue`, though I can't name any offhand. Worst case is we apply it as we find it, I suppose.

matt-livefront:
> My only thought is if there might be other places we want to make something be @DefaultValue, though I can't name any offhand. Worst case is we apply it as we find it, I suppose.

I had that thought too, but I didn't want to introduce too many changes here. I can take another pass through separately and see if there's other places we could apply this (enums are the most susceptible to failures around this).

PR #1125, opened 2024-11-11
[BITAU-153] [BITAU-144] [BITAU-208] Enable Background Syncing When the Phone Is Locked

brant-livefront:
## 🎟️ Tracking

[BITAU-144](https://livefront.atlassian.net/browse/BITAU-144)
[BITAU-153](https://livefront.atlassian.net/browse/BITAU-153)

## 📔 Objective

There is currently an error in the way we receive Push Notifications when the device is completely locked. All of the keys in the app are currently stored with the protection level `kSecAttrAccessibleWhenUnlockedThisDeviceOnly`. This level of protection means that if we get a Push when the device is locked, none of the keys will be accessible. The first thing the NotificationService does is check `stateService.isAuthenticated()` which immediately tries to access the accessToken. This throws a security error, which is caught, logged, and fails silently. Therefore all background Push notifications will fail if the device is locked.

In addition, when we setup the Authenticator vault key to facilitate syncing to the Authenticator app, it was stored like all other keys with `kSecAttrAccessibleWhenUnlockedThisDeviceOnly`. This was a known issue that would break syncing in the background if the device is locked.

This PR attempts to fix both of these situations by allowing some keys to be accessed in the background even if the device is locked:
- It refactors the way we setup permissions to allow for optionally specifying the protection level associated with the `KeychainItem`.
- This allows us to specify `kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly` for the `accessToken` (which fixes Push) and the `authenticatorVaultKey` (which enables full background syncing to Authenticator).

With these changes in place, users can receive background pushes when the device is locked and the app will be able to process them. And those pushes can trigger the AuthenticatorSyncService to use it's key to unlock/lock the vault and perform the sync.

#### Existing users will already have accessToken with the old permissions:
- When this rolls out in the app, all users will have previously stored `accessToken`s with the old protection level.
- `setAccessToken` like all keychain operations will first _delete_ the existing item before _adding_ the new value.
- From my testing so far, the delete succeeds, allowing us to write the new `accessToken` with the new protection level. **But this should be thoroughly tested**.
- This also means that a user will have to have written a new accessToken before they can have push and sync working with the device locked. Until they get the new protection level set, it will behave as it does now and fail silently.
 
#### Testing Steps
Here's a quick rundown of what I did to test this case:
- Ran the app with the old setting of device unlocked to make sure my account was setup. Logged in, etc.
- Locked vault and device and sent a push. Fails due to the `accessToken`'s protection level
- Rebuild to device with the NEW accessToken protection.
- Locked vault and device and sent another push. Fails again (same error, due to the fact that the `accessToken` hasn't been re-written and therefore has the old protection level)
- Unlocked everything, breakpointed the call for pending logins and sent back a `401`. This causes the app to fetch and store a new `accessToken`
- Verified that the setAccessToken call was going through and didn't have any errors. This means it should now have the new protection level.
- Locked vault and Device, sent push, and everything completes. Saw the notification go through, saw the sync service unlock vault, sync keys, lock it back up, etc.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


victor-livefront (dismissed):
Code looks good to me. Here are the scenarios I tested with this change on a physical device:

Fresh install new version starting as logged out
* foregrounded/unlocked vault: ✅
* foregrounded/locked vault: ✅
* backgrounded/unlocked vault: ✅
* backgrounded/locked value: ✅
* device locked/unlocked vault: ✅
* device locked/locked vault: ✅

Installed old version and logged out:
* foregrounded/unlocked vault: ✅
* foregrounded/locked vault: ✅
* backgrounded/unlocked vault: ✅
* backgrounded/locked value: ✅
* device locked/unlocked vault: ✅
* device locked/locked vault: ❌ Error: BitwardenShared.KeychainServiceError.osStatusError(-25308)

Updated to new version:
* foregrounded/unlocked vault: ✅
* foregrounded/locked vault: ✅
* backgrounded/unlocked vault: ✅
* backgrounded/locked value: ✅
* device locked/unlocked vault: ✅
* device locked/locked vault: ❌ Error: BitwardenShared.KeychainServiceError.osStatusError(-25308) 

Auth Token refresh (updated my devices clock to 4 hours in the future, and foregrounded the app. Did not enter master password):
* foregrounded/unlocked vault: ✅
* foregrounded/locked vault: ✅
* backgrounded/unlocked vault: ✅
* backgrounded/locked value: ✅
* device locked/unlocked vault: ✅
* device locked/locked vault: ✅

I needed to comment the assertion failure in the `OSLogErrorReporter` to allow it to move forward for DEBUG, but this failure is based on other factors and is expected.

victor-livefront:
One thing I noticed when testing, which is only tangentially related to this since this PR is about keychain access, is that the authenticator sync service does not run if you start the app from cold and don't unlock it (but you have a valid auth token from a previous session). We may need to account for this scenario.

brant-livefront:
Per @victor-livefront's comment above, I added a ticket to track this. The last commit has a new test that accounts for this scenario and a fix for the issue.

Here's the scenario:
- User has enabled sync and synced at least once
- User exits the app completely and starts from scratch
- User gets a cipher update (i.e. via Push) while the app is in the foreground with the vault still locked

**Expected**: Sync should sync the new cipher update because it already has the vault key when sync was setup.
**Actual**: Sync service doesn’t get started because the check for the vault unlock failed.

victor-livefront (dismissed):
Updates look good. Scenario is fixed on my device now. Thanks!

BitwardenShared/Core/Auth/Services/KeychainRepository.swift
```diff
@@ -42,6 +42,21 @@ enum KeychainItem: Equatable {
...
+        switch self {
+        case .biometrics,
+             .deviceKey,
+             .neverLock,
+             .pendingAdminLoginRequest,
+             .refreshToken:
+            kSecAttrAccessibleWhenUnlockedThisDeviceOnly
+        case .accessToken,
+             .authenticatorVaultKey:
+            kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly
```
matt-livefront, line 56: ❓ Does it make sense to also do this for the `refreshToken`? If the access token is expired, we'll need to use the refresh token to fetch a new access token before proceeding.

BitwardenShared/Core/Auth/Services/KeychainService.swift
```diff
@@ -35,6 +48,12 @@ protocol KeychainService: AnyObject {
     func search(query: CFDictionary) throws -> AnyObject?
 }
 
+extension KeychainService {
+    func accessControl(for flags: SecAccessControlCreateFlags) throws -> SecAccessControl {
+        try accessControl(protection: kSecAttrAccessibleWhenUnlockedThisDeviceOnly, for: flags)
```
matt-livefront, line 53: 🤔 It seems like this is only used for one migration in `MigrationService`. Since we have the protection and flags defined on `KeychainItem` I feel like we should prefer using those over this default? Thoughts on removing this and updating the one instance to pass in the protection level?

fedemkr (commented):
Looks good. I'm not aware of any attacks on this protection level change but I would like @matt-livefront and @KatherineInCode to check it as well in case I'm missing something before approving.
Additionally, given that `accessToken` can now be used in background shouldn't we also allow `refreshToken` as well so when the `accessToken` is expired it can be refreshed? Otherwise sync will fail on background when trying to get the `refreshToken` on expired access token scenario given its protection level.

BitwardenShared/Core/Auth/Services/KeychainRepository.swift
brant-livefront: Sounds good. I was trying to keep as few things as possible from moving to the wider protection setting, but both you and Fede mentioned this and I think it makes sense to move refreshToken there as well. I'll put up a commit for that. 👍

BitwardenShared/Core/Auth/Services/KeychainService.swift
brant-livefront: When I first added this, I thought it might be nice to have a default that was strict and matched what we had before. But I think after I refactored the KeychainItem piece, we're down to just that one. I think it makes more sense to be explicit in that case and remove the extension. I'll make that change. 👍

brant-livefront:
@matt-livefront @fedemkr I've moved the `refreshToken` out to be accessible while the device is locked and I removed the protocol extension in favor of just explicitly adding the protection in `MigrationService`. 👍

matt-livefront:
> Looks good. I'm not aware of any attacks on this protection level change but I would like @matt-livefront and @KatherineInCode to check it as well in case I'm missing something before approving.

@fedemkr No concerns from me.

fedemkr (approved):

PR #1130, opened 2024-11-11
[PM-12991] Only show delete if user can manage any collections item is in

KatherineInCode:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-12991

## 📔 Objective

This changes the logic of whether or not to show the Delete button on the View Item and Edit Item screens. Previously it would always show if the user had Can Edit permissions; now it only shows if the user has Can Manage permissions. In the event that a cipher belongs to multiple collections, the ability to manage any of them gives the ability to delete (per [this comment in Jira](https://bitwarden.atlassian.net/browse/PM-14580?focusedCommentId=75962))

In the event that the server doesn't send the `manage` flag, we default to `!readonly`, as this matches past behavior as best we can.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


BitwardenShared/UI/Vault/VaultItem/CipherItemState.swift
```diff
@@ -107,6 +107,13 @@ struct CipherItemState: Equatable {
         self
     }
 
+    /// Whether or not this item can be deleted by the user.
+    var canBeDeleted: Bool {
+        guard !collectionIds.isEmpty else { return true }
+        let cipherCollections = collections.filter { collectionIds.contains($0.id ?? "") }
+        return cipherCollections.map(\.manage).contains(true)
```
fedemkr, line 114: 🎨 I think it's better to use a `contains` directly here so you don't need to traverse the whole `collections` sequence potentially three times (`filter` + `map` + `contains`) if `manage = true` is never valid or is the last one. Additionally, I feel it's simpler.
```suggestion
        return collections.contains { collection in
            collectionIds.contains(collection.id ?? "") && collection.manage
        }
```

BitwardenShared/UI/Vault/VaultItem/ViewItem/ViewItemProcessor.swift
```diff
@@ -533,6 +533,7 @@ private extension ViewItemProcessor {
 
                 if case var .data(itemState) = newState.loadingState {
                     itemState.loginState.totpState = totpState
+                    itemState.collections = try await services.vaultRepository.fetchCollections(includeReadOnly: true)
                     newState.loadingState = .data(itemState)
                 }
```
fedemkr, line 538: 🤔 If this throws, does the view state remain ok for the user? I believe it does but wanted to raise it up just in case something is missing given the `totpState` is being updated but not the `loadingState`.

BitwardenShared/UI/Vault/VaultItem/AddEditItem/AddEditItemState.swift
```diff
@@ -12,6 +12,8 @@ protocol AddEditItemState: Sendable {
     /// The card item state.
     var cardItemState: CardItemState { get set }
 
+    var canBeDeleted: Bool { get }
```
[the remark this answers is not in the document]
KatherineInCode: 🤦🏼‍♀️ Always something! Good catch!

BitwardenShared/UI/Vault/VaultItem/CipherItemState.swift
KatherineInCode: TIL `.contains` has a closure variation! Yeah, doing it that way is probably better.

BitwardenShared/UI/Vault/VaultItem/ViewItem/ViewItemProcessor.swift
KatherineInCode: The `totpState` won't get updated either, because it'll error out before the updated `itemState` is assigned within `newState` or `state` is updated with `newState`. There are a number of other cases that can error out here before the `state` is updated; the main difference is that the rest of the checks occur earlier in the `do` block, which come to think of it, probably makes sense to be consistent in that regard.

fedemkr (approved):

PR #1132, opened 2024-11-12
[BITAU-179] Refactor Vault Selection to Include Standard and SteamURI TOTP Types

brant-livefront:
## 🎟️ Tracking

[BITAU-179](https://livefront.atlassian.net/browse/BITAU-179)

## 📔 Objective

This PR refactors the Vault selection to handle `TOTPKeyModel` instead of `OTPAuthModel`. Previously, the vault selection was only used for incoming otpauth URLs from the camera's QR code scanning. This meant that it was built using `OTPAuthModel` to hold all of the TOTP details. When building the Authenticator sync feature, we wanted to allow users to move any of their codes from Authenticator to the PM app. But some of those codes could be Steam URI codes or standard codes, which wouldn't have parsed when handed to `OTPAuthModel`.

By refactoring the route, processor, state, etc to use `TOTPKeyModel` we allow for handling of any codes that could be sent from the Authenticator app.

### Testing notes
In testing this, I discovered that the way the Authenticator stores manually entered codes means that even Base32/Standard codes would have been represented using `otpauth://` style reference. So these codes probably worked before this refactor. However, it would still be necessary to support SteamURI codes. You can test this with the example: `steam://JBSWY3DPEHPK3PXP`.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/BitwardenShared/UI/Platform/Application/AppProcessor.swift
+++ b/BitwardenShared/UI/Platform/Application/AppProcessor.swift
@@ -407,14 +407,14 @@ extension AppProcessor {
             return AppRoute.tab(.settings(.accountSecurity))
         case BitwardenDeepLinkConstants.authenticatorNewItem:
             guard let item = await services.authenticatorSyncService?.getTemporaryTotpItem(),
-                  let totpKey = item.totpKey,
-                  let otpAuthModel = OTPAuthModel(otpAuthKey: totpKey) else {
+                  let totpKey = item.totpKey else {
                 coordinator?.showAlert(.defaultAlert(title: Localizations.somethingWentWrong,
                                                      message: Localizations.unableToMoveTheSelectedItemPleaseTryAgain))
                 return nil
             }
 
-            return AppRoute.tab(.vault(.vaultItemSelection(otpAuthModel)))
+            let totpKeyModel = TOTPKeyModel(authenticatorKey: totpKey)
+            return AppRoute.tab(.vault(.vaultItemSelection(totpKeyModel)))
         default:
             return nil
         }
@@ -428,12 +428,13 @@ extension AppProcessor {
     private func getOtpAuthUrlRoute(url: URL) async -> AppRoute? {
         guard let scheme = url.scheme, scheme.isOtpAuthScheme else { return nil }
 
-        guard let otpAuthModel = OTPAuthModel(otpAuthKey: url.absoluteString) else {
+        let totpKeyModel = TOTPKeyModel(authenticatorKey: url.absoluteString)
+        guard case .otpAuthUri = totpKeyModel.totpKey else {
             coordinator?.showAlert(.defaultAlert(title: Localizations.anErrorHasOccurred))
             return nil
         }
 
-        return AppRoute.tab(.vault(.vaultItemSelection(otpAuthModel)))
+        return AppRoute.tab(.vault(.vaultItemSelection(totpKeyModel)))
     }
 
     /// Starts timer to send organization events regularly
--- a/BitwardenShared/UI/Vault/Vault/VaultItemSelection/VaultItemSelectionState.swift
+++ b/BitwardenShared/UI/Vault/Vault/VaultItemSelection/VaultItemSelectionState.swift
@@ -11,9 +11,6 @@ struct VaultItemSelectionState: Equatable {
     /// The base url used to fetch icons.
     let iconBaseURL: URL?
 
-    /// The parsed OTP key used to find matching ciphers to add the key to.
-    let otpAuthModel: OTPAuthModel
-
     /// The user's current account profile state and alternative accounts.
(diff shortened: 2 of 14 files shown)


BitwardenShared/UI/Platform/Application/AppProcessor.swift
```diff
@@ -428,12 +428,13 @@ extension AppProcessor {
     private func getOtpAuthUrlRoute(url: URL) async -> AppRoute? {
         guard let scheme = url.scheme, scheme.isOtpAuthScheme else { return nil }
 
-        guard let otpAuthModel = OTPAuthModel(otpAuthKey: url.absoluteString) else {
+        let totpKeyModel = TOTPKeyModel(authenticatorKey: url.absoluteString)
+        guard case .otpAuthUri = totpKeyModel.totpKey else {
```
brant-livefront, line 432: @matt-livefront @fedemkr @KatherineInCode I'd love to hear your thoughts on this. I left it in to require it to be an OTP URL to keep the handling consistent. We have a few tests that expect a failure here when we send various incomplete `otpauth://` URLs.

However the TOTPKeyModel is much more forgiving and will simply treat these as other code types. So we could theoretically just remove those tests and we don't need to guard for the scheme or check to be sure it parsed to an `.otpAuthUri` type. Like I said, I left this in to keep it consistent, but I'm open to re-considering that.

BitwardenShared/UI/Vault/Vault/VaultItemSelection/VaultItemSelectionState.swift
```diff
@@ -42,6 +42,11 @@ struct VaultItemSelectionState: Equatable {
 
     /// The search term used to find ciphers that match the OTP key.
     var ciphersMatchingName: String? {
-        otpAuthModel.issuer ?? otpAuthModel.accountName
+        switch totpKeyModel.totpKey {
+        case let .otpAuthUri(otpAuthModel):
+            return otpAuthModel.issuer ?? otpAuthModel.accountName
+        case .standard, .steamUri:
+            return nil
```
fedemkr, line 49: 🤔 Is it ok that no vault items are streamed for standard and steam Uri? I would guess so but wanted to ask just in case.

fedemkr (approved):
Approving with a small question that I think it's fine.

BitwardenShared/UI/Platform/Application/AppProcessor.swift
matt-livefront: I think I'm in favor of keeping the `otpauth://` scheme check, but I like the idea of removing the `.otpAuthUri` case check and associated error. It seems better to let the user proceed rather than throwing an error if we can't parse to a `.otpAuthUri` type.

fedemkr: 🤔 I agree on keeping the scheme check. Now given the method `getOtpAuthUrlRoute` is explicitly for OTP auth URIs then IMO it makes sense to have the `.otpAuthUri` case check and associated error. Is there a scenario you can think of or @KatherineInCode (I know you've seen several special cases) where a `otpauth://` is passed and we cannot parse it but it's fine to let the user continue?

BitwardenShared/UI/Vault/Vault/VaultItemSelection/VaultItemSelectionState.swift
brant-livefront: I think that might be the only option. This string is just used to filter the list down to items we think match. But we might only have `steam://JBSWY3DPEHPK3PXP` or just `JBSWY3DPEHPK3PXP` in these cases, which makes it hard for us to guess as to what matches. 

But UI-wise, we already handle this because the older OTPAuth version could have the same problem (e.g. on a url that only passes secret with no issuer, etc). In fact, that's why this is an optional string. So you can always search for the item you want to add this one to and you'll still find it. 

One other note is that due to the way the Authenticator app saves these codes, I think it will actually generate an otpauth url style code for Standard codes manually entered. So let's say I put "Amazon" as the name and `JBSWY3DPEHPK3PXP`, when it passes it over to the PM app, it will likely hand that off in such a way that "Amazon" ends up as the issuer and is used here. 👍

BitwardenShared/UI/Platform/Application/AppProcessor.swift
fedemkr: I would definitely keep the check, we cannot allow TOTP creation without a key.

BitwardenShared/UI/Vault/Vault/VaultItemSelection/VaultItemSelectionState.swift
fedemkr: Roger makes sense, thanks!

PR #1139, opened 2024-11-13
[PM-14497] Implement OTP autofill from QuickType bar

fedemkr:
## 🎟️ Tracking

[PM-14497](https://bitwarden.atlassian.net/browse/PM-14497)

## 📔 Objective

Implement OTP autofill from QuickType bar with and without user interaction.

_Note: This feature is available from iOS 18._

## 📸 Screenshots

https://github.com/user-attachments/assets/fe85c35b-d762-4913-ad24-e5641b18a8cf

https://github.com/user-attachments/assets/a566ddc8-84a2-4418-bf81-3c3609652bd0

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-14497]: https://bitwarden.atlassian.net/browse/PM-14497?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenAutoFillExtension/CredentialProviderViewController.swift
```diff
@@ -90,6 +90,15 @@ class CredentialProviderViewController: ASCredentialProviderViewController {
...
         }
+
+        guard #available(iOSApplicationExtension 18.0, *) else {
+            return
+        }
+
+        if let otpRequest = credentialRequest as? ASOneTimeCodeCredentialRequest,
+           let otpIdentity = otpRequest.credentialIdentity as? ASOneTimeCodeCredentialIdentity {
+            provideOTPCredentialWithoutUserInteraction(for: otpIdentity)
+        }
```
matt-livefront, line 101: 🎨 Would it make sense to put this in the default case of the switch statement to keep everything in the switch? Also applies to `prepareInterfaceToProvideCredential`.

fedemkr: Not sure what would we gain given that we wouldn't leverage the case feature and the code will be even more nested. However if you have a strong opinion about it I can do the change no problem at all.
Or were you meaning to call a function from the `default` which has this code in it?

Sadly, dealing with `#available` is a bit messy and when dealing with multiple logic per OS version I'm not sure if I like having methods with suffixes like `myAwesomeMethodIOS17`, `myAwesomeMethodIOS18(...)`, so on; although maybe it could be a solution that is pretty self-explanatory.

matt-livefront: I think it initially reads like something that happens after the switch currently and could apply to any of the above cases, but it's really just another case that we're looking for based on `credentialRequest`. I agree the available checks aren't great. I'm fine keeping it as-is if you prefer that!

```
        default:
            if #available(iOSApplicationExtension 18.0, *),
               let otpRequest = credentialRequest as? ASOneTimeCodeCredentialRequest,
               let otpIdentity = otpRequest.credentialIdentity as? ASOneTimeCodeCredentialIdentity {
                provideOTPCredentialWithoutUserInteraction(for: otpIdentity)
            }
```

matt-livefront (approved):

matt-livefront (approved):
Nice!

matt-livefront (approved):

PR #1150, opened 2024-11-19
PM-14547: Remove soft logged out account

matt-livefront:
## 🎟️ Tracking

[PM-14547](https://bitwarden.atlassian.net/browse/PM-14547)

## 📔 Objective

When long-pressing on a logged out account in the profile switcher, the remove option should be displayed instead of log out.

## 📸 Screenshots

https://github.com/user-attachments/assets/2de7e812-90b2-427d-ac82-4fe2eb7911f8

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

[PM-14547]: https://bitwarden.atlassian.net/browse/PM-14547?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1155, opened 2024-11-21
PM-14585: Removed lock account option from Profile switcher if account has no master password

ezimet-livefront:
## 🎟️ Tracking

[PM-14585](https://bitwarden.atlassian.net/browse/PM-14585)
## 📔 Objective

- Added ```canBeLocked``` property for ```ProfileSwitcherItem```.
- Added check for ```canBeLocked```in ```AuthRepository```, An account can now be locked only if it has a master password, biometric authentication, or a PIN.
- Added unit tests.
## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-14585]: https://bitwarden.atlassian.net/browse/PM-14585?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
+++ b/BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
@@ -20,10 +20,19 @@ protocol AuthRepository: AnyObject {
     ///
     func allowBioMetricUnlock(_ enabled: Bool) async throws
 
-    /// Whether master password verification can be done for the active  user.
+    /// Whether a user account can be locked.
     ///
+    /// - Parameter userId: The user Id to be mapped to an account.
+    /// - Returns: `true` if the account can be locked, `false` otherwise.
+    ///
+    func canBeLocked(userId: String?) async -> Bool
+
+    /// Whether master password verification can be done for a userId.
+    ///
+    /// - Parameter userId:  The user Id to be mapped to an account.
     /// - Returns: `true` if one can verify master password, `false` otherwise.
-    func canVerifyMasterPassword() async throws -> Bool
+    ///
+    func canVerifyMasterPassword(userId: String?) async throws -> Bool
 
     /// Clears the pins stored on device and in memory.
     ///
@@ -72,11 +81,11 @@ protocol AuthRepository: AnyObject {
     ///
     func hasMasterPassword() async throws -> Bool
 
-    /// Whether pin unlock is available.
+    /// Whether pin unlock is available for a userId.
+    ///  - Parameter userId: The userId of the account.
+    ///  - Returns: Whether pin unlock is available.
     ///
-    /// - Returns: Whether pin unlock is available.
-    ///
-    func isPinUnlockAvailable() async throws -> Bool
+    func isPinUnlockAvailable(userId: String?) async throws -> Bool
 
     /// Checks the locked status of a user vault by user id.
     ///
@@ -273,6 +282,21 @@ protocol AuthRepository: AnyObject {
 }
 
 extension AuthRepository {
+    /// Whether active user account can be locked.
+    ///
+    /// - Returns: `true` if active user account can be locked, `false` otherwise.
+    ///
+    func canBeLocked() async -> Bool {
+        await canBeLocked(userId: nil)
+    }
+
+    /// Whether master password verification can be done for the active user.
+    ///
+    /// - Returns: `true` if active user has master password, `false` otherwise.
+    func canVerifyMasterPassword() async throws -> Bool {
+        try await canVerifyMasterPassword(userId: nil)
+    }
+
     /// Gets the account for the active user id.
     ///
     /// - Returns: The active user account.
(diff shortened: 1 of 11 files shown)


BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
```diff
@@ -960,8 +960,19 @@ extension DefaultAuthRepository: AuthRepository {
...
+        let hasMasterPassword: Bool = await (
+            try? stateService.getUserHasMasterPassword(userId: account.profile.userId)
+        ) ?? false
+        let isUnlockWithPinOn: Bool = await (
+            try? stateService.pinProtectedUserKey(userId: account.profile.userId)
+        ) != nil
+        let isUnlockWithBiometricOn: Bool = await (
+            try? biometricsRepository.getBiometricUnlockStatus().isEnabled
+        ) ?? false
+        let canBeLocked = hasMasterPassword || isUnlockWithPinOn || isUnlockWithBiometricOn
```
fedemkr, line 972: 🤔 Does it make sense to extract this logic into a func in this repository? So it's easier to be reused if needed elsewhere. It can be private for the time being if you want. What do you think?
🤔 Additionally, I see these functions already being used which is alike the logic you have here but without the user id:
- `isPinUnlockAvailable()`
- `canVerifyMasterPassword()`
Would it make sense to improve these functions passing `userId: String?` and having an extension method with no parameter that passes `userId: nil` so we reuse them in this logic as well?
I feel like by reusing these functions, then if the state service changes somehow it would be less code to maintain/update.

ezimet-livefront: I think that is good improvement, I am sure we would need this ```canBeLocked``` check somewhere else. I will add it.

BitwardenShared/UI/Auth/ProfileSwitcher/ProfileSwitcherItem.swift
```diff
@@ -5,6 +5,9 @@ import SwiftUI
 /// An object that defines account profile information relevant to account switching
 /// Part of `ProfileSwitcherState`.
 struct ProfileSwitcherItem: Equatable, Hashable {
+    /// Indicates if the account can be locked.
+    var canBeLocked: Bool
```
matt-livefront, line 9: 🤔 There is also an accessibility action in `ProfileSwitcherRow` which I think needs to be gated behind this.

https://github.com/bitwarden/ios/blob/bd6651bc1a222271549f6f2f9221eff9b3b9d799/BitwardenShared/UI/Auth/ProfileSwitcher/ProfileSwitcherRow.swift#L229-L234

BitwardenShared/UI/Auth/ProfileSwitcher/ProfileSwitcherRow.swift
```diff
@@ -213,7 +213,7 @@ struct ProfileSwitcherRow: View {
                         )
                     )
                 }
-                .onLongPressGesture(if: store.state.allowLockAndLogout) {
+                .onLongPressGesture(if: store.state.allowLock || store.state.allowLogout) {
```
matt-livefront, line 216: ⚠️ I think we need to be careful here. The purpose of `allowLockAndLogout` is to determine if the view supports showing the action sheet for the lock and logout actions. For example, you can't lock or logout from the autofill screen. So that should take priority over these new properties. But it's maybe also confusing to have `allowLockAndLogout`, `allowLock` and `allowLockout` whereas `allowLockAndLogout` is slightly different than the other two.

I have some similar changes [here](https://github.com/bitwarden/ios/pull/1150/files#diff-741bdaaa7f7c0db8a5fa1222f442067bf1284678be6ea8918e6698c98ffdbb52) to support remove account. Maybe it's easiest to stick with using the `canBeLocked` on `ProfileSwitcherItem`? For the lock accessibility action below, something like:

`if: store.state.allowLockAndLogout && profileSwitcherItem.isUnlocked && profileSwitcherItem.canBeLocked,`

matt-livefront (approved):

BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
```diff
@@ -466,8 +498,21 @@ extension DefaultAuthRepository: AuthRepository {
...
+        let hasMasterPassword: Bool = await (
+            try? canVerifyMasterPassword(userId: userId)
+        ) ?? false
+        let isUnlockWithPinOn: Bool = await (
+            try? isPinUnlockAvailable(userId: userId)
+        ) ?? false
+        let isUnlockWithBiometricOn: Bool = await (
+            try? biometricsRepository.getBiometricUnlockStatus().isEnabled
+        ) ?? false
+        return hasMasterPassword || isUnlockWithPinOn || isUnlockWithBiometricOn
```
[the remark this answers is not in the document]
ezimet-livefront: I will try to add this in my future PR.

PR #1172, opened 2024-11-26
[PM-15416] Change pull-to-refresh to not force a sync

KatherineInCode:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-15416

## 📔 Objective

This changes the behavior of pull-to-refresh to respect time gates and other business logic around when we should actually do a refresh. It does not change the behavior of other places we trigger a manual refresh.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessor.swift
+++ b/BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessor.swift
@@ -86,7 +86,7 @@ final class VaultListProcessor: StateProcessor<
         case .refreshAccountProfiles:
             await refreshProfileState()
         case .refreshVault:
-            await refreshVault(isManualRefresh: true)
+            await refreshVault(isManualRefresh: false)
         case let .search(text):
             state.searchResults = await searchVault(for: text)
         case .streamAccountSetupProgress:
--- a/BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessor.swift
+++ b/BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessor.swift
@@ -65,7 +65,7 @@ final class VaultListProcessor: StateProcessor<
     override func perform(_ effect: VaultListEffect) async {
         switch effect {
         case .appeared:
-            await refreshVault(isManualRefresh: false)
+            await refreshVault()
             await handleNotifications()
             await checkPendingLoginRequests()
             await checkPersonalOwnershipPolicy()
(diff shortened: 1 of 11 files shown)


BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessor.swift
```diff
@@ -86,7 +86,7 @@ final class VaultListProcessor: StateProcessor<
         case .refreshAccountProfiles:
             await refreshProfileState()
         case .refreshVault:
-            await refreshVault(isManualRefresh: true)
+            await refreshVault(isManualRefresh: false)
```
fedemkr, line 89: 🤔 `isManualRefresh` acts alongside with whether the user has `allowSyncOnRefresh` enabled or not. 
IMO if the user has `allowSyncOnRefresh` turned on then they would expect the vault to be synced (forced) when manually refreshing. Also, as I understand it that's what the [Bitwarden Help center says](https://bitwarden.com/help/vault-sync/#manual-sync):
> To manually sync your vault from a Bitwarden app:
Open the  Settings tab, tap the Other option, and tap Sync now.
Toggle the Enable sync on refresh option to allow your vault to be synced using a pulldown gesture on the  My Vault tab.

If we do this and let manually refreshing to have the timing logic then, e.g., if the user is part of an organization where some items were updated some minutes after the last sync and the user manually refreshes the view they won't see the updated items and they will think something is not working.
Furthermore, this would have a different behavior here than in `VaultGroup` and `Send`.

fedemkr (commented):
❓Sorry but I don't know much about the context on this, could more details be provided as to why this change is necessary?

BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessor.swift
fedemkr: Oh roger, now I've read the conversation it makes sense; thanks for the context!

fedemkr, line 89: 🤔 IMO we should change the name of the parameter `isManualRefresh` to something else as in this case would be contradictory given that the user is actually manually refreshing the vault. I would use `forceSync` as in the end is what internally is being treated as in the repositories.
Same applies to Send processor and Vault+Send repositories.

KatherineInCode: Good call.

Gerardv514:
What exactly is this PR doing? I have noticed on my ios app (2024.12) my vault is not syncing when I pull down gesture on the vault. If I go to settings/other and click Sync Now it syncs. I am running a self hosted and also have a BW cloud account, I can reproduce in both places.

PR #1181, opened 2024-12-03
[PM-12991] Don't show delete on edit screen if user can't delete

KatherineInCode:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-15419

## 📔 Objective

This is a defect on https://github.com/bitwarden/ios/pull/1130

There was an issue where we were showing the Delete button on the Edit Item screen even when a user didn't have Manage permissions for the collection that item was in. This was because when finding relevant collections, we were explicitly only getting ones that weren't readonly—which meant that when computing whether or not the user could delete the item, it would look like the item wasn't in any collections, which means it can be deleted. This updates the logic for getting those collections to include readonly collections.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/BitwardenShared/UI/Vault/VaultItem/AddEditItem/AddEditItemProcessor.swift
+++ b/BitwardenShared/UI/Vault/VaultItem/AddEditItem/AddEditItemProcessor.swift
@@ -261,7 +261,9 @@ final class AddEditItemProcessor: StateProcessor<// swiftlint:disable:this type_
             let ownershipOptions = try await services.vaultRepository
                 .fetchCipherOwnershipOptions(includePersonal: !isPersonalOwnershipDisabled)
 
-            state.collections = try await services.vaultRepository.fetchCollections(includeReadOnly: false)
+            // We need read-only collections so that we can include them in the state
+            // to correctly calculate if the item can be deleted
+            state.collections = try await services.vaultRepository.fetchCollections(includeReadOnly: true)
             // Filter out any collection IDs that aren't included in the fetched collections.
             state.collectionIds = state.collectionIds.filter { collectionId in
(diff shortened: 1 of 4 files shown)


BitwardenShared/UI/Vault/VaultItem/AddEditItem/AddEditItemProcessor.swift
```diff
@@ -261,7 +261,9 @@ final class AddEditItemProcessor: StateProcessor<// swiftlint:disable:this type_
             let ownershipOptions = try await services.vaultRepository
                 .fetchCipherOwnershipOptions(includePersonal: !isPersonalOwnershipDisabled)
 
-            state.collections = try await services.vaultRepository.fetchCollections(includeReadOnly: false)
+            // We need read-only collections so that we can include them in the state
+            // to correctly calculate if the item can be deleted
+            state.collections = try await services.vaultRepository.fetchCollections(includeReadOnly: true)
```
matt-livefront, line 266: 👍 The comment here is helpful context!

⚠️ I believe we were only fetching the non-read only collections because we don't want to show read-only collections as an option where you can assign the cipher to that collection in the ownership section. Do we need to add an additional filter for only showing ones that you have permissions for?

Screenshot 2024-12-04 at 10 02 45 AM

KatherineInCode: 🤦🏼‍♀️ I totally thought of this at some point this morning when I was making tea and then forgot when I got back to my computer. Yes, that is a thing that should be done. I'll make the fix.

fedemkr (approved):

PR #1202, opened 2024-12-11
PM-15435: Use TOTP manual entry in extension when camera isn't supported

matt-livefront:
## 🎟️ Tracking

[PM-15435](https://bitwarden.atlassian.net/browse/PM-15435)

## 📔 Objective

The camera isn't supported in iOS extensions. Previously, if the Set up TOTP button was tapped on the add vault item screen in the autofill extension, the camera feed would show as all black. This adds a check to redirect the user to the manual TOTP entry view instead of using the camera when in the extension.

## 📸 Screenshots

https://github.com/user-attachments/assets/dab53c1f-79de-4571-bc8a-787089589c62

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/BitwardenShared/UI/Vault/VaultItem/AddEditItem/AddEditItemProcessor.swift
+++ b/BitwardenShared/UI/Vault/VaultItem/AddEditItem/AddEditItemProcessor.swift
@@ -707,7 +707,9 @@ final class AddEditItemProcessor: StateProcessor<// swiftlint:disable:this type_
     /// Kicks off the TOTP setup flow.
     ///
     private func setupTotp() async {
-        guard services.cameraService.deviceSupportsCamera() else {
+        guard services.cameraService.deviceSupportsCamera(),
+              appExtensionDelegate?.isInAppExtension != true // Extensions don't allow camera access.
+        else {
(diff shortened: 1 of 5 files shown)


fedemkr (commented):
❓ Maybe this is a question more for Product team, but shouldn't we display a callout in there saying that the camera is not allowed in that context and that's why they can only enter the key manually? 
I would find it weird to be able to use the camera in the main app and not in there if nothing is explained to me of that difference.
Perhaps we can put the callout in the Help docs but IMO it's better to have these kind of things in-app.

matt-livefront:
> ❓ Maybe this is a question more for Product team, but shouldn't we display a callout in there saying that the camera is not allowed in that context and that's why they can only enter the key manually? I would find it weird to be able to use the camera in the main app and not in there if nothing is explained to me of that difference. Perhaps we can put the callout in the Help docs but IMO it's better to have these kind of things in-app.

@fedemkr Good question, I checked with product and it sounds like we don't want to make any additional UI changes at this time.

fedemkr (approved):

PR #1208, opened 2024-12-13
[PM-8216] Add warning to people who don't have two-factor authentication turned on

KatherineInCode:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-8216

## 📔 Objective

This is still a work in progress; I have it set up as a draft to help me track things as I tie up the remaining problems and do some additional refactoring. Some of these changes will be pulled into other PRs for simplicity.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


GlobalTestHelpers/MockAppModule.swift
```diff
@@ -148,6 +150,12 @@ class MockAppModule:
         tabCoordinator.asAnyCoordinator()
     }
 
+    func makeTwoFactorNoticeNoticeCoordinator(
```
andrebispo5, line 153: Was looking around and noticed this typo 😄

KatherineInCode: Good catch! Thanks!

BitwardenShared/UI/Auth/TwoFactorNotice/SetUpTwoFactor/SetUpTwoFactorView.swift
```diff
@@ -0,0 +1,114 @@
...
+            .buttonStyle(.secondary())
+
+            if store.state.allowDelay {
+                AsyncButton(Localizations.remindMeLater) {
+                    await store.perform(.remindMeLaterTapped)
+                }
+                .buttonStyle(.secondary())
+            }
+
+            Spacer()
```
matt-livefront, line 69: 🤔 Is this `Spacer()` necessary? It seems like this could be for pushing up the content, but the content in the scroll view should already be at the top? Or if it's just padding, I think the scroll view modifier also adds bottom padding.

BitwardenShared/UI/Auth/TwoFactorNotice/EmailAccess/EmailAccessView.swift
```diff
@@ -0,0 +1,100 @@
...
+                .padding(.horizontal, 12)
+
+            VStack(spacing: 12) {
+                AsyncButton(Localizations.continue) {
+                    await store.perform(.continueTapped)
+                }
+                .buttonStyle(.primary())
+            }
+            .dynamicTypeSize(...DynamicTypeSize.xxxLarge)
+            .padding(.horizontal, 16)
```
matt-livefront, line 48: 🤔 I think this might be adding double horizontal padding, since the `scrollView()` modifier also adds horizontal padding. `scrollView()` also adds the background color.

BitwardenShared/UI/Auth/TwoFactorNotice/TwoFactorNoticeCoordinator.swift
```diff
@@ -0,0 +1,98 @@
...
+        & HasAuthRepository
+        & HasEnvironmentService
+        & HasErrorReporter
+        & HasStateService
+        & HasTimeProvider
+
+    // MARK: Private Properties
+
+    /// The delegate for this coordinator, used to notify when the user logs out.
+    private weak var delegate: VaultCoordinatorDelegate?
```
matt-livefront, line 20: 🤔 Is this used?

BitwardenShared/UI/Auth/TwoFactorNotice/TwoFactorNoticeHelper.swift
```diff
@@ -0,0 +1,112 @@
...
+    case seen(Date)
+}
+
+// MARK: - TwoFactorNoticeHelper
+
+/// A protocol for a helper object to handle deciding whether or not to display
+/// the two-factor notice, and displaying it if so.
+///
+protocol TwoFactorNoticeHelper {
+    ///
```
matt-livefront, line 27: ❓ Missing docs?

BitwardenShared/UI/Auth/TwoFactorNotice/TwoFactorNoticeHelperTests.swift
```diff
@@ -0,0 +1,254 @@
...
+        subject = DefaultTwoFactorNoticeHelper(
+            coordinator: coordinator.asAnyCoordinator(),
+            services: services
+        )
+
+        // Because nearly all of the tests are "it doesn't show the notice if
+        // these conditions are true", it makes sense to set things up here to
+        // show the notice, and then in tests selectively set up the specific
+        // condition that causes it to not show. This hopefully makes the tests
+        // easier to read.
```
matt-livefront, line 50: 👍

BitwardenShared/UI/Platform/Application/Views/DynamicImageTextStackView.swift
```diff
@@ -0,0 +1,75 @@
...
+    init(
+        minHeight: CGFloat,
+        @ViewBuilder imageContent: @escaping () -> I,
+        @ViewBuilder textContent: @escaping () -> T
+    ) {
+        self.minHeight = minHeight
+        self.imageContent = imageContent()
+        self.textContent = textContent()
+    }
+
```
matt-livefront, line 74: ⛏️ 
```suggestion
```

BitwardenShared/UI/Auth/TwoFactorNotice/SetUpTwoFactor/SetUpTwoFactorProcessor.swift
```diff
@@ -0,0 +1,80 @@
...
+        case .remindMeLaterTapped:
+            await handleDismiss()
+        }
+    }
+
+    override func receive(_ action: SetUpTwoFactorAction) {
+        switch action {
+        case .changeAccountEmailTapped:
+            coordinator.showAlert(.changeEmailAlert {
+                self.state.url = self.services.environmentService.changeEmailURL
```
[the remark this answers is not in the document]
KatherineInCode: There's a discussion about this [on the Android ticket](https://bitwarden.atlassian.net/browse/PM-8217), but the end result is right now the user will need to kill and re-run the app to pick up on the change.

BitwardenShared/UI/Platform/Application/Support/Localizations/en.lproj/Localizable.strings
```diff
@@ -1059,8 +1059,17 @@
...
+"SetUpTwoStepLogin" = "Set up two-step login";
+"YouCanSetUpTwoStepLoginAsAnAlternative" = "You can set up two-step login as an alternative way to protect your account or change your email to one you can access.";
+"TurnOnTwoStepLogin" = "Turn on two-step login";
+"ChangeAccountEmail" = "Change account email";
 "NothingAvailableToAutofill" = "Nothing available to autofill";
 "FailedToGenerateVerificationCode" = "Failed to generate verification code";
 "FailedToAutofillItem" = "Failed to autofill item %1$@";
 "ExportingFailed" = "Exporting failed";
 "YouMayNeedToEnableDevicePasscodeOrBiometrics" = "You may need to enable device passcode or biometrics.";
+"TurnOnTwoStepLoginConfirmation" = "You can turn on two-step login on the bitwarden.com web vault. Do you want to visit the website now?";
```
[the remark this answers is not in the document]
KatherineInCode: 🤔 I was going off of the fact that the change email confirmation just says `bitwarden.com`. That's a good question, though.

BitwardenShared/UI/Auth/TwoFactorNotice/EmailAccess/EmailAccessView.swift
```diff
@@ -0,0 +1,100 @@
...
+                .accessibilityHidden(true)
+
+            Divider()
+
+            Toggle(Localizations.yesICanReliablyAccessMyEmail, isOn: store.binding(
+                get: \.canAccessEmail,
+                send: EmailAccessAction.canAccessEmailChanged
+            ))
+            .toggleStyle(.bitwarden)
+            .accessibilityIdentifier("ItemFavoriteToggle")
```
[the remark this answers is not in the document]
KatherineInCode: Hah, probably not. Thanks copy-paste!

BitwardenShared/UI/Auth/TwoFactorNotice/TwoFactorNoticeCoordinator.swift
KatherineInCode: Nope, copy-paste artifact :(

BitwardenShared/UI/Auth/TwoFactorNotice/SetUpTwoFactor/SetUpTwoFactorView.swift
KatherineInCode: At the time I had it in there, I needed it to get things pinned to the top, but I also changed around how this view was constructed since then. I'll double-check.

BitwardenShared/UI/Auth/TwoFactorNotice/EmailAccess/EmailAccessView.swift
```diff
@@ -0,0 +1,100 @@
...
+            } textContent: {
+                VStack(spacing: 16) {
+                    Text(Localizations.importantNotice)
+                        .styleGuide(.title, weight: .bold)
+
+                    Text(Localizations.bitwardenWillSendACodeToYourAccountEmail)
+                        .styleGuide(.title3)
+                }
+                .padding(.horizontal, 12)
+            }
```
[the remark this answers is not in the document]
KatherineInCode: I'd looked around for some other view that did this and apparently didn't search hard enough 😅 yes, it can be replaced that way. Though now I'm curious if we could also migrate the [IntroCarouselView](https://github.com/bitwarden/ios/blob/main/BitwardenShared/UI/Auth/IntroCarousel/IntroCarouselView.swift#L58) to also use this, possibly with additional parameters in some way for the spacing?

BitwardenShared/UI/Auth/TwoFactorNotice/EmailAccess/EmailAccessViewTests.swift
```diff
@@ -0,0 +1,55 @@
...
+
+    /// Tapping the continue button sends the `.continueTapped` effect
+    @MainActor
+    func test_continueButton_tap() throws {
+        let button = try subject.inspect().find(button: Localizations.continue)
+        try button.tap()
+
+        waitFor(!processor.effects.isEmpty)
+
+        XCTAssertEqual(processor.effects.last, .continueTapped)
```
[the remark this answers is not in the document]
KatherineInCode: Ah! Good to know!

BitwardenShared/UI/Auth/TwoFactorNotice/EmailAccess/EmailAccessView.swift
```diff
@@ -0,0 +1,100 @@
...
+        .frame(maxWidth: .infinity, maxHeight: .infinity)
+        .background(Asset.Colors.backgroundPrimary.swiftUIColor.ignoresSafeArea())
+        .foregroundStyle(Asset.Colors.textPrimary.swiftUIColor)
+        .multilineTextAlignment(.center)
+        .scrollView()
+    }
+
+    private var toggleCard: some View {
+        VStack(alignment: .leading, spacing: 8) {
+            Text(LocalizedStringKey(Localizations.doYouHaveReliableAccessToYourEmail("person\u{2060}@example.com")))
```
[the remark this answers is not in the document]
KatherineInCode: 🤦🏼‍♀️Yes, yes it should. Good catch!

BitwardenShared/UI/Auth/TwoFactorNotice/EmailAccess/EmailAccessView.swift
```diff
@@ -0,0 +1,100 @@
...
+                .styleGuide(.body)
+                .foregroundColor(Asset.Colors.textPrimary.swiftUIColor)
+                .accessibilityHidden(true)
```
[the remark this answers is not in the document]
KatherineInCode: 🤔 I'm not sure where that came from. Probably a copy-paste thing? But yeah, I don't think it should be hidden

BitwardenShared/UI/Platform/Application/Support/Localizations/en.lproj/Localizable.strings
```diff
@@ -1059,8 +1059,17 @@
 "CopyPrivateKey" = "Copy private key";
 "CopyFingerprint" = "Copy fingerprint";
 "SSHKeys" = "SSH keys";
+"ImportantNotice" = "Important notice";
+"BitwardenWillSendACodeToYourAccountEmail" = "Bitwarden will send a code to your account email to verify logins from new devices starting in February 2025.";
```
fedemkr, line 1063: ⛏️ I would add the suffix `DescriptionLong` to the key if you're not going to use the whole sentence as the key because there could be another value in the future that is only "Bitwarden will send a code to your account email" and if that happens we couldn't use the key as it would be already used by this one.

BitwardenShared/UI/Platform/Application/Support/Localizations/en.lproj/Localizable.strings
```diff
@@ -1059,8 +1059,17 @@
...
+"DoYouHaveReliableAccessToYourEmail" = "Do you have reliable access to your email, **%1$@**?";
+"YesICanReliablyAccessMyEmail" = "Yes, I can reliably access my email";
+"SetUpTwoStepLogin" = "Set up two-step login";
+"YouCanSetUpTwoStepLoginAsAnAlternative" = "You can set up two-step login as an alternative way to protect your account or change your email to one you can access.";
```
fedemkr, line 1067: ⛏️ Idem, I would add the suffix `DescriptionLong` here as well.

BitwardenShared/UI/Platform/Application/Support/Localizations/en.lproj/Localizable.strings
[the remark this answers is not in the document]
fedemkr: I guess this will change to not specify the host but to use "Bitwarden web vault" instead. Also I'd add `DescriptionLong` here as well to be consistent with other places.

BitwardenShared/UI/Auth/TwoFactorNotice/Alert+TwoFactorNotice.swift
```diff
@@ -0,0 +1,45 @@
...
+    /// in a browser.
+    ///
+    /// - Parameters:
+    ///   - action: The action taken if they select continue.
+    /// - Returns: An alert that asks if the user wants to navigate to the change email page
+    ///
+    static func changeEmailAlert(action: @escaping () -> Void) -> Alert {
+        Alert(
+            title: Localizations.changeAccountEmail,
+            message: Localizations.changeEmailConfirmation,
```
fedemkr, line 16: 🤔 Based on what was discussed internally I think this should be updated to not use the host url but the name of the product so to avoid the logic to see if it's US or EU host.

BitwardenShared/UI/Auth/TwoFactorNotice/EmailAccess/EmailAccessProcessor.swift
```diff
@@ -0,0 +1,76 @@
+import Combine
+import SwiftUI
+
+// MARK: - EmailAccessProcessor
+
+/// The processor used to manage state and handle actions for the new device notice screen.
```
fedemkr, line 6: ⛏️ Should this be like this?
```suggestion
/// The processor used to manage state and handle actions for the email access screen.
```

BitwardenShared/UI/Auth/TwoFactorNotice/EmailAccess/EmailAccessProcessor.swift
```diff
@@ -0,0 +1,76 @@
...
+            if state.canAccessEmail {
+                let displayState: TwoFactorNoticeDisplayState
+                displayState = state.allowDelay ? .canAccessEmail : .canAccessEmailPermanent
+                try await services.stateService.setTwoFactorNoticeDisplayState(state: displayState)
+                coordinator.navigate(to: .dismiss)
+            } else {
+                coordinator.navigate(to: .setUpTwoFactor(allowDelay: state.allowDelay))
+            }
+        } catch {
+            services.errorReporter.log(error: error)
```
fedemkr, line 73: 🤔 Shouldn't we display something to the user that states that something wrong happened? If not, the user may think the app got frozen or something like that if an error is thrown here but nothing is shown to the user.

BitwardenShared/UI/Auth/TwoFactorNotice/SetUpTwoFactor/SetUpTwoFactorViewTests.swift
```diff
@@ -0,0 +1,82 @@
...
+    /// Tapping the remind me later button sends `.remindMeLater`
+    @MainActor
+    func test_remindMeLater_tap() async throws {
+        let button = try subject.inspect().find(asyncButton: Localizations.remindMeLater)
+        try await button.tap()
+
+        XCTAssertEqual(processor.effects.last, .remindMeLaterTapped)
+    }
+
+    /// Tapping the turn on two-step login button sends `.changeAccountEmailTapped`
```
fedemkr, line 52:
```suggestion
    /// Tapping the turn on two-step login button sends `. turnOnTwoFactorTapped `
```

BitwardenShared/UI/Auth/TwoFactorNotice/TwoFactorNoticeCoordinator.swift
```diff
@@ -0,0 +1,95 @@
...
+        case let .setUpTwoFactor(allowDelay):
+            showSetUpTwoFactor(allowDelay)
+        }
+    }
+
+    func start() {}
+
+    // MARK: Private Methods
+
+    func showEmailAccess(_ allowDelay: Bool) {
```
fedemkr, line 66: ⛏️ Missing docs.

BitwardenShared/UI/Auth/TwoFactorNotice/TwoFactorNoticeCoordinator.swift
```diff
@@ -0,0 +1,95 @@
...
+            )
+        )
+        let store = Store(processor: processor)
+        let view = EmailAccessView(
+            store: store
+        )
+        stackNavigator?.replace(view)
+    }
+
+    func showSetUpTwoFactor(_ allowDelay: Bool) {
```
fedemkr, line 81: ⛏️ Missing docs.

BitwardenShared/UI/Auth/TwoFactorNotice/TwoFactorNoticeCoordinatorTests.swift
```diff
@@ -0,0 +1,77 @@
...
+        let action = try XCTUnwrap(stackNavigator.actions.last)
+        XCTAssertEqual(action.type, .pushed)
+        XCTAssertTrue(action.view is SetUpTwoFactorView)
+    }
+
+    /// `start()` does nothing
+    @MainActor
+    func test_start() throws {
+        subject.start()
+    }
```
fedemkr, line 76: ⛏️ You can check that no navigation is performed on start to test it does nothing:
```suggestion
    /// `start()` does nothing
    @MainActor
    func test_start() throws {
        subject.start()
        XCTAssertTrue(stackNavigator.actions.isEmpty)
    }
```

BitwardenShared/UI/Auth/TwoFactorNotice/TwoFactorNoticeHelper.swift
```diff
@@ -0,0 +1,113 @@
...
+            else { return }
+
+            let state = try await services.stateService.getTwoFactorNoticeDisplayState()
+            switch state {
+            case .canAccessEmail:
+                if permanent {
+                    coordinator.navigate(to: .twoFactorNotice(!permanent))
+                } else {
+                    return
+                }
```
fedemkr, line 99: ⛏️ IMO `guard` looks better in this case:
```suggestion
                guard permanent else {
                    return
                } 
                coordinator.navigate(to: .twoFactorNotice(false))
```
🤔 Shouldn't we pass `permanent` to `.twoFactorNotice`, i.e. `true`? Btw, given the current code the condition `!permanent` will always be `false` so that's why I changed it directly to the final value in the suggestion.

BitwardenShared/UI/Auth/TwoFactorNotice/TwoFactorNoticeHelper.swift
```diff
@@ -0,0 +1,113 @@
...
+            case .canAccessEmailPermanent:
+                return
+            case .hasNotSeen:
+                coordinator.navigate(to: .twoFactorNotice(!permanent))
+            case let .seen(date):
+                if services.timeProvider.timeSince(date) >= (86400 * 7) { // Seven days
```
fedemkr, line 105: 🎨 Extract `86400 * 7` into constant.

BitwardenShared/UI/Auth/TwoFactorNotice/TwoFactorNoticeHelperTests.swift
```diff
@@ -0,0 +1,254 @@
...
+        stateService.activeAccount = .fixture()
+        configService.featureFlagsBool[.newDeviceVerificationTemporaryDismiss] = true
+        configService.featureFlagsBool[.newDeviceVerificationPermanentDismiss] = true
+        environmentService.region = .unitedStates
+        stateService.doesActiveAccountHaveTwoFactorResult = .success(false)
+        policyService.policyAppliesToUserResult[.requireSSO] = false
+    }
+
+    override func tearDown() {
```
fedemkr, line 59: ⛏️ Tear down `timeProvider` as well.

BitwardenShared/UI/Auth/TwoFactorNotice/TwoFactorNoticeRoute.swift
```diff
@@ -0,0 +1,13 @@
+// MARK: - TwoFactorNoticeRoute
+
+/// A route to a specific screen in the No Two Factor notice.
+public enum TwoFactorNoticeRoute: Equatable, Hashable {
+    /// A route to dismiss the screen currently presented modally.
+    case dismiss
```
fedemkr, line 6: ⛏️ I would avoid specifying how the screen is presented so if then is not presented modally anymore we don't have to update the comment. Additionally, it's not like we have several `dismiss` cases for each type of presentation so we can simplify it.
```suggestion
    /// A route to dismiss the current screen.
    case dismiss
```

fedemkr (commented):
Great job, just some picks I saw.

BitwardenShared/Core/Platform/Services/StateServiceTests.swift
```diff
@@ -897,6 +905,27 @@ class StateServiceTests: BitwardenTestCase { // swiftlint:disable:this type_body
...
+    /// and throws an error if there is no current user.
+    func test_getTwoFactorNoticeDisplayState_noId() async throws {
+        appSettingsStore.setTwoFactorNoticeDisplayState(.canAccessEmail, userId: "1")
+
+        do {
+            try await _ = subject.getTwoFactorNoticeDisplayState()
+            XCTFail("subject.getTwoFactorNoticeDisplayState() should throw an error if there is no active account")
+        } catch {
+            XCTAssertEqual(error as? StateServiceError, StateServiceError.noActiveAccount)
+        }
```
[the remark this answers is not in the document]
KatherineInCode: Aha! I figured there was a function for it, just couldn't remember it!

BitwardenShared/UI/Auth/TwoFactorNotice/Alert+TwoFactorNotice.swift
KatherineInCode: There appears to be a few other places we explicitly name `bitwarden.com` as well, where it might make sense to not use the host URL. I kind of feel like that should be done in a separate PR, though?

BitwardenShared/UI/Auth/TwoFactorNotice/TwoFactorNoticeRoute.swift
KatherineInCode: I'll update it here, but there are several other instances of the comment elsewhere that should at some point be updated likewise 🤔  (I feel like that's a separate PR from this one, though)

BitwardenShared/UI/Auth/TwoFactorNotice/EmailAccess/EmailAccessProcessor.swift
KatherineInCode: The only way something could go wrong here is if it failed to save to the state service, which shouldn't actually happen, in practice, I think? After all, that only fails if there's no active account, and you can't get here without an active account. I'm not sure what sort of message we should display here if that fails, though—"Please restart the app and try again"?

BitwardenShared/UI/Auth/TwoFactorNotice/TwoFactorNoticeHelper.swift
KatherineInCode: `permanent` is the inverse of `allowDelay`, hence negating it. It's a weird naming thing because the flags are "temporary" (you can dismiss the notification) and "permanent" (you cannot dismiss the notification), and this is the point where we make the switch between the two

BitwardenShared/UI/Auth/TwoFactorNotice/EmailAccess/EmailAccessView.swift
```diff
@@ -41,10 +41,13 @@ struct EmailAccessView: View {
...
     private var toggleCard: some View {
         VStack(alignment: .leading, spacing: 8) {
-            Text(LocalizedStringKey(Localizations.doYouHaveReliableAccessToYourEmail("person\u{2060}@example.com")))
-                .styleGuide(.body)
-                .foregroundColor(Asset.Colors.textPrimary.swiftUIColor)
-                .accessibilityHidden(true)
+            Text(LocalizedStringKey(Localizations.doYouHaveReliableAccessToYourEmailDescriptionLong(
+                // Adding the Word Joiner character (U+2060) in the middle of the email address
+                // keeps the markdown rendering from making the email address a tappable link.
+                store.state.emailAddress.replacingOccurrences(of: "@", with: "\u{2060}@")
```
fedemkr, line 47: 🤔 Really cool! Could we have a helper / extension method to do the replacement so it's easier to reuse and to see that something already exists to fix it?

BitwardenShared/UI/Platform/Application/Support/Localizations/en.lproj/Localizable.strings
fedemkr: @KatherineInCode I meant `BitwardenWillSendACodeToYourAccountEmail` that needs to add `DescriptionLong`. I see you added it to `DoYouHaveReliableAccessToYourEmail` instead.

BitwardenShared/UI/Auth/TwoFactorNotice/EmailAccess/EmailAccessView.swift
KatherineInCode: Aha, good call!

BitwardenShared/UI/Platform/Application/Support/Localizations/en.lproj/Localizable.strings
KatherineInCode: 🤦🏼‍♀️ Yep, I can do that

BitwardenShared/UI/Auth/TwoFactorNotice/TwoFactorNoticeHelper.swift
```diff
@@ -0,0 +1,123 @@
...
+            // to always qualify.
+            let creationDate = profile.creationDate ?? Date(timeIntervalSince1970: 0)
+            let accountAge = services.timeProvider.timeSince(creationDate)
+            let hasTwoFactor = profile.twoFactorEnabled ?? false
+
+            guard accountAge > Constants.twoFactorNoticeMinimumAccountAgeInterval,
+                  !hasTwoFactor,
+                  services.environmentService.region != .selfHosted,
+                  await !services.policyService.policyAppliesToUser(.requireSSO)
+            else { return }
```
fedemkr, line 101: 🎨 IMO when we can short-circuit and return early we should to avoid wasting resources, like in here if the region is actually Self-Hosted then there is no need to get the active account and check all the other stuff. So I would re-order this as follows:
```suggestion
            guard services.environmentService.region != .selfHosted else {
                return
            }

            let profile = try await services.stateService.getActiveAccount().profile
            guard profile.twoFactorEnabled != true else {
                return
            }

            // If we don't have a creation date (possible for older accounts that
            // haven't synced recently, because the property is only being saved as of
            // this notice being implemented) then assume the account is old enough
            // to always qualify.
            let creationDate = profile.creationDate ?? Date(timeIntervalSince1970: 0)
            let accountAge = services.timeProvider.timeSince(creationDate)

            guard accountAge > Constants.twoFactorNoticeMinimumAccountAgeInterval,
                  await !services.policyService.policyAppliesToUser(.requireSSO)
            else { return }
```
What do you think?

BitwardenShared/Core/Platform/Utilities/Constants.swift
```diff
@@ -89,6 +89,14 @@ enum Constants {
...
     static let pbkdf2Iterations = 600_000
 
+    /// The minimum number of seconds an account must have existed for before
+    /// we display the two-factor notice.
+    static let twoFactorNoticeMinimumAccountAgeInterval: TimeInterval = 86400 * 7 // Seven days
+
+    /// The minimum number of seconds after a user delays the two-factor notice
+    /// before we display the notice again.
+    static let twoFactorNoticeDelayInterval: TimeInterval = 86400 * 7 // Seven days
+
```
fedemkr, line 99: ♻️ Maybe for another PR, but wanted to bring this up before I forget it that it would be pretty cool to have static methods from `TimeInterval` extensions so one can specify the time unit to create the interval, e.g. `TimeInterval.days(7)` or `TimeInterval.minutes(60)` and then it calculates the interval internally which would be much easier to use.

KatherineInCode: Agreed, though figuring out the best API surface for that is always the trick—especially as in a number of cases you actually want to go [through the Calendar object](https://developer.apple.com/documentation/foundation/nscalendar/1409577-date) to account for things like DST, and having too easy an API for "just add five days" very much can run afoul of those sorts of issues (I've had it happen before).

BitwardenShared/UI/Auth/TwoFactorNotice/TwoFactorNoticeHelper.swift
KatherineInCode: I was also trying to do some short-circuiting, in the order of "most likely to happen" to "least likely to happen", along with putting all the conditions in one place for easy reference. Otherwise, I trust the compiler to optimize the order of operations far better than I can (especially as, from what I understand, the Swift compiler will notice where the intermediate variable is actually used and reorder the steps for me to inline things and initialize things in order).

KatherineInCode: Haah, having thought about it, yeah, the order of these could be better in terms of taking advantage of that optimization. Splitting it out like this seems fine.

fedemkr (approved):

BitwardenShared/Core/Platform/Utilities/Constants.swift
fedemkr: Yeah makes sense. I mean if we were using iOS 16 as minimum we could use [Duration](https://developer.apple.com/documentation/Swift/Duration) object when we want high precision and avoid some of the issues you mention but well for now we need to keep using `TimeInterval`.
And totally agree that we don't need really high precision in these places but still I think refactoring to use some extension methods in another PR will be great 😄

PR #1210, opened 2024-12-13
[PM-8216] Add Change Email and Set Up Two-Factor URLs

KatherineInCode:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-8216

## 📔 Objective

This adds additional URLs to the Environment Service, Environment URLs, and Environment URL Data objects for 1) changing email (which is currently just the account screen) and 2) setting up two-factor authentication. These will be used by new screens in the future.

This also adds new tests to backfill some gaps in our test coverage.

This also does a bit of minor refactoring to unify casing structure inside each of these classes to follow the [Swift API Design Guidelines](https://www.swift.org/documentation/api-design-guidelines/), which suggests that "URL" should be capitalized inside of properties and function names. Ideally we would also change `EnvironmentUrls` and `EnvironmentUrlData` to `EnvironmentURLs` and `EnvironmentURLData` for this as well, but I wasn't sure if that was too big a change for this PR. If sentiment is towards ripping that bandaid off now, I can do that additionally.

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

PR #1212, opened 2024-12-16
[PM-8216] Update case of URL in various classes and variables

KatherineInCode:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-8216

## 📔 Objective

This follows on from [PR 1210](https://github.com/bitwarden/ios/pull/1210). Because the [Swift API Design Guidelines](https://www.swift.org/documentation/api-design-guidelines/) suggest that "URL" should be capitalized in classes and variables—and we have been inconsistent with that in the past—this attempts to go through and clean up some places where we use "Url" instead of "URL". There are plenty of other places that this does not capture—we can take care of those in a future pass.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


BitwardenShared/Core/Platform/Models/Domain/Account.swift
```diff
@@ -132,7 +132,7 @@ extension Account {
         // MARK: Properties
 
         /// The environment URLs for an account.
-        var environmentUrls: EnvironmentUrlData?
+        var environmentURLs: EnvironmentURLData?
```
matt-livefront, line 135: ❓ I think this is persisted to UserDefaults, is this going to cause decoding existing values to fail?

BitwardenShared/Core/Platform/Services/Stores/AppSettingsStore.swift
```diff
@@ -786,10 +786,10 @@ extension DefaultAppSettingsStore: AppSettingsStore {
...
             case let .pinProtectedUserKey(userId):
                 key = "pinKeyEncryptedUserKey_\(userId)"
-            case .preAuthEnvironmentUrls:
-                key = "preAuthEnvironmentUrls"
-            case let .accountCreationEnvironmentUrls(email):
-                key = "accountCreationEnvironmentUrls_\(email)"
+            case .preAuthEnvironmentURLs:
+                key = "preAuthEnvironmentURLs"
+            case let .accountCreationEnvironmentURLs(email):
+                key = "accountCreationEnvironmentURLs_\(email)"
```
matt-livefront, line 792: ⚠️ Changing this key will cause any existing values stored in UserDefaults with this key to become lost/orphaned without a migration. Is that ok in this scenario or do we need to consider how to handle this?

KatherineInCode: 🤦🏼‍♀️ Good catch. We shouldn't be making changes to the actual keys like this. I thought I caught all the cases we would do that, but I guess I missed one. I'll pull this one back.

BitwardenShared/Core/Platform/Models/Domain/Account.swift
KatherineInCode: Good question. I'll double-check that to make sure.

KatherineInCode: Having checked, it does cause decoding existing values to fail. Given a choice between either manually specifying a coding key (to maintain backwards compatibility), creating a migration, or just reverting the case change on the variable; I think the last makes the most sense. A coding key just feels weird, and a migration is too heavyweight, and the casing of these variables isn't so crucial that we can't just live with it being cased like it was.

matt-livefront (dismissed):

matt-livefront (approved):

PR #1231, opened 2024-12-30
[PM-16533] Rename files to indicate they belong to Password Manager

KatherineInCode:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-16533

## 📔 Objective

As part of the plan to merge the Authenticator repository into this one, we'll want to avoid filename conflicts. This moves various YML files out of the way by specifying that they're for Password Manager, so that the Authenticator ones will be able to sit next to them fine once the merge is done.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


.github/workflows/build.yml
```diff
@@ -355,8 +355,8 @@ jobs:
 
       - name: Update build version and number
         run: |
-          yq -i '.settings.MARKETING_VERSION = "${{ steps.version_info.outputs.version_name }}"' 'project.yml'
-          yq -i '.settings.CURRENT_PROJECT_VERSION = "${{ steps.version_info.outputs.version_number }}"' 'project.yml'
+          yq -i '.settings.MARKETING_VERSION = "${{ steps.version_info.outputs.version_name }}"' 'project_passwordmanager.yml'
+          yq -i '.settings.CURRENT_PROJECT_VERSION = "${{ steps.version_info.outputs.version_number }}"' 'project_passwordmanager.yml'
```
fedemkr, line 359: 🤔 As far as I can see we're using hyphen `-` instead of underscore `_` to separate words in the filenames for `yml` like `crowdin-pull.yml` or `github-release.yml`. Is there a reason to change to underscore for the new ones in this PR?

KatherineInCode: Both of those seem very reasonable to me; when I wrote the PM section of the [Action Plan](https://bitwarden.atlassian.net/wiki/spaces/MNAE/pages/1336868878/Mobile+Monorepos+Action+Plan) I was patterning off of what @vvolkgang had done for the BWA section, but I think hyphens and abbreviations is a good way to go. Álison, what do you think?

If we do this with hyphens and abbreviations, I can get the Action Plan updated to match.

vvolkgang: That just happened to be the original name of [.license_plist.yml](https://github.com/bitwarden/authenticator-ios/blob/main/.license_plist.yml). We can rename them all to use `-` while at it. 

Let's go with the shorter names. We'll scrap some / all instances of these later while merging files.

KatherineInCode: Awesome, I'll get the Action Plan updated accordingly, and then make the change here

Scripts/update_acknowledgements.sh
fedemkr, line 1: 🤔 Should this file also be renamed to use hyphen as well?

KatherineInCode: Hm. I was going off of all of the scripts (except `download-artifacts.sh`) using underscores. I'm not sure if updating all of them to also be hyphens would be in the scope of this change, though? If we want to do that?

(I think it's reasonable for us to do that at this point, to match everything else; it's more a question of whether this is the best place for that or not)

fedemkr: I'm fine to leave that for another PR if you feel it's out of the scope of this PR, no problem at all. Approving as is 👍

fedemkr (approved):

PR #1280, opened 2025-01-17
[PM-14014] Add toast if initial load is taking a while

KatherineInCode:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-14014

## 📔 Objective

This adds a toast that appears if the initial load of data takes more than five seconds; this toast will not disappear until the data load finishes. This also means updating the toast to be able to either automatically dismiss (as the previous functionality) or require the calling code to manually dismiss it (as for this situation).

## 📸 Screenshots

https://github.com/user-attachments/assets/7b797115-958d-4302-9ede-572cda4f4141

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessor.swift
```diff
@@ -226,6 +226,19 @@ extension VaultListProcessor {
     ///
     private func refreshVault() async {
         do {
+            let takingTimeTask = Task {
+                let delay: UInt64 = 5
+                if #available(iOS 16.0, *) {
+                    try await Task.sleep(for: .seconds(delay), tolerance: .seconds(1))
+                } else {
+                    try await Task.sleep(nanoseconds: delay * 1_000_000_000)
+                }
```
fedemkr, line 235: 🎨 Could this be extracted to an extension so we don't need to worry about the specific iOS version each time we need to call this?

BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessor.swift
```diff
@@ -226,6 +226,19 @@ extension VaultListProcessor {
...
+                self.state.toast = Toast(title: Localizations.thisIsTakingLongerThanExpected, mode: .manualDismiss)
+            }
+            defer {
+                state.toast = nil
+                takingTimeTask.cancel()
```
fedemkr, line 240: 🤔 From what I can see in the demo video the _toast_ is being shown a little bit longer after the data has been displayed to the user. So in order to have a smoother experience should this be also done directly before `state.loadingState = .data(sections)`? Thus the toast will go away just a moment before the data is displayed which is how I would imagine the flow.
What do you think?

BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessor.swift
KatherineInCode: Are you thinking something like a `Task.sleep(forSeconds:)` extension? 🤔 That could work

fedemkr: Indeed 😄 . Ideally I would generalize it to different units but I think it'd be much over engineered / YAGNI. So for now having the `forSeconds` extension will work just fine.

BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessor.swift
fedemkr: Just in case I wasn't saying to remove the defer block, just repeat it before the state is updated to improve the UI flow. So the defer will still be useful for catches. I get the point on DRY but in this case it's really tiny code and in the same method so I don't see it much as a code smell. However, I get also the point that this is an edge case so I'll leave it up to you as to whether make the change.

KatherineInCode: Ah, that makes sense, and that repetition shouldn't matter. Good call, I can do that (though I'll also probably include a comment to that effect)

fedemkr: Hmmm, I'll talk with the SDK team about this. AFAIK that method should only be called when dealing with Fido2 flows which is not the case for login.

fedemkr: ℹ️  I've checked this and it's because we're updating the `ASStore` when replacing all identities [here](https://github.com/bitwarden/ios/blob/main/BitwardenShared/Core/Autofill/Services/AutofillCredentialService.swift#L254) which ends up calling the `allCredentials` method (btw this additional sync is temporary and should not be necessary after some future changes).
However, in these cases of really slow connection yes there could be a race condition where in `needsSync` `lastSyncTime` doesn't get updated fast enough and two syncs end up being performed in parallel which is awful.
♻️ I like your temporal solution for now and IMO we should add a new task to deal with parallel syncs in another PR; which can be a little tricky considering the async nature of the calls.
What do you think?

fedemkr: We cannot change the sync there until some updates are done in the SDK/Server regarding passkeys data, I'll check with the team on the status of that.

fedemkr: Btw @KatherineInCode is the repetition of the code here not needed anymore with the changes you've made?

KatherineInCode: Correct; repeating it here wasn't doing anything useful (because it doesn't happen once the `ASStore` sync comes back, so the toast hangs out there regardless), but dismissing the toast when we load *any* data from a sync does.

fedemkr (approved):

PR #1292, opened 2025-01-21
[PM-11712] Fix keyboard scroll on cipher view so it doesn't overlap cursor

fedemkr:
## 🎟️ Tracking

[PM-11712](https://bitwarden.atlassian.net/browse/PM-11712)

## 📔 Objective

Fix keyboard scroll on cipher view so it doesn't overlap cursor on text fields like URI and Notes.

⚠️ This fix only applies to iOS 17+.

## 📸 Screenshots

https://github.com/user-attachments/assets/db1ab4ac-c713-491e-8bba-f6214a0fbd79

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-11712]: https://bitwarden.atlassian.net/browse/PM-11712?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenShared/UI/Vault/VaultItem/AddEditItem/AddEditItemView.swift
```diff
@@ -80,6 +83,11 @@ struct AddEditItemView: View {
                 .ignoresSafeArea()
         )
         .navigationBarTitleDisplayMode(.inline)
+        .apply { view in
```
KatherineInCode, line 86: 😕 I wish this could be compressed to just use [our .if() method](https://github.com/bitwarden/ios/blob/main/BitwardenShared/UI/Platform/Application/Extensions/View.swift#L85), but it looks like `#available` doesn't play nice there, alas.

BitwardenShared/UI/Platform/Application/Utilities/KeyboardResponder.swift
```diff
@@ -0,0 +1,42 @@
...
+    // MARK: Properties
+
+    /// Whether the keyboard is shown.
+    @Published var isShown: Bool = false
+
+    /// A publisher when the keyboard will show.
+    var keyboardWillShowNotification = NotificationCenter.default.publisher(
+        for: UIResponder.keyboardWillShowNotification
+    )
+    /// A publisher when the keyboard will hide.
```
KatherineInCode, line 16: ⛏️ Add a line between these?

Also I think the position should be switched for alphabetizing them, even though in practice they occur in show-hide sequence. Given that, and the limited functionality of this, I think leaving in that order is probably okay; I leave it to your judgement.

BitwardenShared/UI/Vault/VaultItem/AddEditItem/AddEditItemView.swift
fedemkr: Yeah, we could have special `.ifIOS17()`, `.ifIOS18()` methods which could simplify this a bit though 🤔

fedemkr: What do y'all think? @KatherineInCode @matt-livefront @ezimet-livefront ?

KatherineInCode: It looks like [we've discussed this before](https://github.com/bitwarden/ios/pull/655#issuecomment-2150897637), and I think I'm generally of the same opinion against those sorts of specific version-targeted things. However, I don't hold that particularly strongly, and if we have enough reason to do an `.ifIOS17()` then I'm fine with it (we're at two instances where it would be useful, at least; three is usually my threshold on such things).

fedemkr: Oh excellent, didn't recall that conversation thanks for pointing it out. I'm fine continuing with `apply` for now.

matt-livefront: I tend to agree with @KatherineInCode. I think I'd rather have us create a modifier that is similar to `contentMargins` and internally it would only apply the content margins modifier on iOS 17+ devices. This would keep the view itself clean and allow it to be reused if necessary. Here's one approach along these lines: https://davedelong.com/blog/2021/10/09/simplifying-backwards-compatibility-in-swift/ but we could also just add a `backportContentModifier()` or something else similarly named.

A similar example we have:
https://github.com/bitwarden/ios/blob/f53edf2bd6704d4f96d8c05c119abb6d654a847f/BitwardenShared/UI/Platform/Application/Extensions/View.swift#L15-L21

One word of caution on `.apply()` and similar modifiers. Because a modifier wraps the view and then returns the modified view, a view needs to be returned in all code paths. Because there's no else case to return `view`, it's going to return _nothing_ which will end up removing the wrapped content from the view on iOS devices prior to iOS 17.

This needs to instead be:
```
        .apply { view in
            if #available(iOS 17.0, *) {
                view.contentMargins(.bottom, keyboard.isShown ? 30.0 : 0.0, for: .scrollContent)
            } else {
                view
            }
        }
```

BitwardenShared/UI/Platform/Application/Utilities/KeyboardResponder.swift
```diff
@@ -0,0 +1,43 @@
...
+            false
+        }
+        .assign(to: \.isShown, on: self)
+        .store(in: &cancellableSet)
+
+        keyboardWillShowNotification.map { notification in
+            true
+        }
+        .assign(to: \.isShown, on: self)
+        .store(in: &cancellableSet)
```
matt-livefront, line 41: ⚠️ I think `assign(to:on:)` will end up creating a retain cycle on this object since you now have a reference to the publisher and the publisher has a reference back to the object.

Instead, I think you can assign the value directly to the publisher. In this case, Combine is managing the publisher so no need for cancellables either.

```suggestion
        keyboardWillHideNotification.map { _ in
            false
        }
        .assign(to: &$isShown)

        keyboardWillShowNotification.map { notification in
            true
        }
        .assign(to: &$isShown)
```

fedemkr: Oh nice catch! Thanks!

BitwardenShared/UI/Vault/VaultItem/AddEditItem/AddEditItemView.swift
fedemkr: I've made the changes and was checking on the blog post for backport modifiers and really like the idea. Should we address adding the `Backport<Content>` in another PR and add the other existing methods as well?

KatherineInCode: I really like the `Backport<Content>` idea as well, though I'm torn on whether to do it here (because we clearly need something like that) or in a separate PR.

It may just make sense to split the difference and create the `Backport<Content>` structure here and apply it to the keyboard things, and then to have a separate PR to update the rest of the code to use it.

matt-livefront: Views are structs, not classes so there's no reference counting or concerns around retain cycles. You'll just have a copy of the view in your case, so you should be good!

fedemkr: Ohhh you're right, often forget `View` is just a struct. Thanks !

matt-livefront (approved):
Nice 👍

PR #1300, opened 2025-01-24
[PM-13689] Prevent a verified user from deleting their account

LRNcardozoWDF:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-13689

## 📔 Objective

When the user selects the delete option in-app, they will see a message that tells the user they cannot delete their account and to contact their administrator.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


BitwardenShared/Core/Vault/Models/Response/ProfileOrganizationResponseModel.swift
```diff
@@ -41,6 +41,11 @@ struct ProfileOrganizationResponseModel: Codable, Equatable {
     /// Whether the profile organization uses policies.
     let usePolicies: Bool
 
+    /// Whether the user is managed by an organization.
+    /// A user is considered managed by an organization if their email domain
+    /// matches one of the verified domains of that organization, and the user is a member of it.
+    let userIsManagedByOrganization: Bool
```
fedemkr, line 47: ⚠️ If this may not be sent by the server, e.g. old server versions then this needs to be annotated with `@DefaultFalse` property wrapper or the decoding of the model could fail.
If that's the case add another API fixture where the property is not present to check that the decoding will succeed nonetheless.
```suggestion
    @DefaultFalse let userIsManagedByOrganization: Bool
```

BitwardenShared/UI/Platform/Application/Support/Localizations/en.lproj/Localizable.strings
```diff
@@ -1099,3 +1099,5 @@
 "TextToShare" = "Text to share";
 "SendDetails" = "Send details";
 "SendNameRequired" = "Send name (required)";
+"CannotDeleteAccount" = "Cannot delete account";
+"CannotDeleteAccountDescription" = "This action cannot be completed because your account is owned by an organization. Contact your organization administrator for additional details.";
```
fedemkr, line 1103: ⛏️ We usually use `DescriptionLong` as a suffix for these situations.
```suggestion
"CannotDeleteAccountDescriptionLong" = "This action cannot be completed because your account is owned by an organization. Contact your organization administrator for additional details.";
```

BitwardenShared/UI/Platform/Settings/Settings/AccountSecurity/DeleteAccount/DeleteAccountProcessor.swift
```diff
@@ -125,4 +127,15 @@ final class DeleteAccountProcessor: StateProcessor<DeleteAccountState, DeleteAcc
...
         }
     }
+
+    /// Load any initial data for the view.
+    private func loadData() async {
+        do {
+            state.shouldPreventUserFromDeletingAccount =
+                try await services.authRepository.isUserManagedByOrganization()
+        } catch {
+            coordinator.showAlert(.networkResponseError(error))
```
fedemkr, line 137: 🤔 If an error is thrown here then `state.shouldPreventUserFromDeletingAccount` will remain `false` and after the user taps ok then the user will be able to delete their account. Shouldn't we close the view after the user cancels the error alert?

BitwardenShared/UI/Platform/Settings/Settings/AccountSecurity/DeleteAccount/DeleteAccountView.swift
```diff
@@ -59,6 +27,85 @@ struct DeleteAccountView: View {
                 store.send(.dismiss)
             }
         }
+        .task {
+            await store.perform(.loadData)
+        }
+    }
+
+    /// This section is shown when the account can be deleted.
```
fedemkr, line 35: ⛏️ As a general rule you should comment what the property/method does or in this case creates/builds as to how the callers treat such method/property. I mean whether this is shown or not is not a concern of the actual computed property but one of the caller, so no need to comment it here because it may change in the future and one shouldn't need to update the comment here as well.
I would instead just comment:
```suggestion
    /// This is the section to delete an account.
```

BitwardenShared/UI/Platform/Settings/Settings/AccountSecurity/DeleteAccount/DeleteAccountView.swift
```diff
@@ -59,6 +27,85 @@ struct DeleteAccountView: View {
...
+                store.send(.dismiss)
+            } label: {
+                Text(Localizations.cancel)
+            }
+            .buttonStyle(.secondary(isDestructive: true))
+            .accessibilityIdentifier("CANCEL")
+        }
+    }
+
+    /// This section is shown when the account can't be deleted.
```
fedemkr, line 76: ⛏️ Same as in `deleteAccountSection`

BitwardenShared/UI/Platform/Settings/Settings/AccountSecurity/DeleteAccount/DeleteAccountView.swift
fedemkr, line 1: 🤔 @ifernandezdiaz I believe we should update the `accessibilityIdentifier` in here not to use all-caps, right? Like `CANCEL` to `CancelButton`. Confirming with you so we don't break any UI tests.

BitwardenShared/UI/Platform/Settings/Settings/AccountSecurity/DeleteAccount/DeleteAccountView.swift
```diff
@@ -59,6 +27,85 @@ struct DeleteAccountView: View {
...
+    @ViewBuilder var preventAccountDeletionSection: some View {
+        HStack(spacing: 12) {
+            Image(asset: Asset.Images.circleX16)
+                .foregroundColor(Color(asset: Asset.Colors.error))
```
fedemkr, line 80: 🤔 I would use `decorative` so SwiftUI ignores the image for accessibility purposes.
```suggestion
            Image(decorative: asset: Asset.Images.circleX16)
                .foregroundColor(Color(asset: Asset.Colors.error))
```

BitwardenShared/UI/Platform/Settings/Settings/AccountSecurity/DeleteAccount/DeleteAccountView.swift
```diff
@@ -59,6 +27,85 @@ struct DeleteAccountView: View {
...
+            } label: {
+                Text(Localizations.cancel)
+            }
+            .buttonStyle(.secondary(isDestructive: true))
+            .accessibilityIdentifier("CANCEL")
+        }
+    }
+
+    /// This section is shown when the account can't be deleted.
+    @ViewBuilder var preventAccountDeletionSection: some View {
```
fedemkr, line 77: 🎨 This whole section view is almost the same as `deleteAccountSection`. Please have just one computed-property for both of them so it can be reused and use the `State` binding to update the necessary changes in both of them.
There's no need to use the flag to show/hide any section if you do this way given that the content of the Texts will be updated depending on the state.

fedemkr (changes requested):
🤔 Is the "Close" button necessary? The user has the cancel button on the toolbar for the same action.

BitwardenShared/Core/Vault/Models/Domain/Organization.swift
```diff
@@ -36,6 +36,11 @@ public struct Organization: Equatable, Hashable, Sendable {
     /// Whether the profile's organization uses policies.
     let usePolicies: Bool
 
+    /// Whether the user is managed by an organization.
+    /// A user is considered managed by an organization if their email domain
+    /// matches one of the verified domains of that organization, and the user is a member of it.
+    let userIsManagedByOrganization: Bool
```
matt-livefront, line 42: ⚠️ Similar to @fedemkr's [comment](https://github.com/bitwarden/ios/pull/1300/files#r1932337147), I would also annotate this with `@DefaultFalse`. This model is populated from the database, so if the user's database doesn't yet have `userIsManagedByOrganization`, loading the organization would fail.

BitwardenShared/Core/Auth/Repositories/AuthRepositoryTests.swift
```diff
@@ -1099,6 +1099,64 @@ class AuthRepositoryTests: BitwardenTestCase { // swiftlint:disable:this type_bo
...
+        try await stateService.setActiveAccount(userId: "1")
+        organizationService.fetchAllOrganizationsResult = .success([])
+
+        let value = try await subject.isUserManagedByOrganization()
+        XCTAssertFalse(value)
+    }
+
+    /// `isUserManagedByOrganization` returns true if the user is managed by an organization.
+    @MainActor
+    func test_isUserManagedByOrganization_true_featureFlagON() async throws {
```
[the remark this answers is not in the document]
LRNcardozoWDF: Added the test [here](https://github.com/bitwarden/ios/pull/1300/files#diff-565a53fb69a68bc85cfd8100fb054d4d0e5dc4bef071f339a2cb6fdd77ffa1a3R1160-R1174).

BitwardenShared/UI/Platform/Settings/Settings/AccountSecurity/DeleteAccount/DeleteAccountView.swift
ifernandezdiaz: Hey @fedemkr! Don't worry; I'll update that ID on our side!
Can we use `"CancelDeletionButton"` as the ID? (to link the id string with the delete account view)

BitwardenShared/UI/Platform/Settings/Settings/AccountSecurity/DeleteAccount/DeleteAccountState.swift
```diff
@@ -2,4 +2,41 @@
...
+    }
+
+    /// A detailed description of the view.
+    var longDescription: String {
+        return if shouldPreventUserFromDeletingAccount {
+            Localizations.cannotDeleteAccountDescriptionLong
+        } else {
+            Localizations.deleteAccountExplanation
+        }
+    }
```
fedemkr, line 24: ⛏️ I'd change this to `title` and `description` as that's how it looks on the screen I think.

BitwardenShared/UI/Platform/Settings/Settings/AccountSecurity/DeleteAccount/DeleteAccountState.swift
```diff
@@ -2,4 +2,41 @@
...
+    var mainIcon: ImageAsset {
+        return if shouldPreventUserFromDeletingAccount {
+            Asset.Images.circleX16
+        } else {
+            Asset.Images.warning24
+        }
+    }
+
+    /// Whether the form to delete the account is showed.
+    var shouldPreventUserFromDeletingAccount = false
```
fedemkr, line 36: ⛏️ I'd change the docs description because if this is `true` the form is not showed so there isn't an exact correlation with the behavior. Maybe something like:
```suggestion
    /// Whether the user should be blocked from deleting their account.
    var shouldPreventUserFromDeletingAccount = false
```

BitwardenShared/UI/Platform/Settings/Settings/AccountSecurity/DeleteAccount/DeleteAccountView.swift
fedemkr: Oh ok, then yes we should use your recommendation Nacho cc: @LRNcardozoWDF could you make the change?

fedemkr (approved):

fedemkr (approved):

PR #1313, opened 2025-01-30
[PM-17710] Fix NotificationCenterServiceTests race condition failure

fedemkr:
## 🎟️ Tracking

[PM-17710](https://bitwarden.atlassian.net/browse/PM-17710)

## 📔 Objective

Fix NotificationCenterServiceTests race condition failure that caused tests to execute until timeout or fail randomly with the new `maximum-test-execution-time-allowance` setting.

This was caused when `notificationCenter.post(...)` was executed before the `iterator.next()` started. Therefore the notification value was missed from the publisher and the test never ended.

In order to fix this without adding an explicit `Task.sleep(...)` I've moved the `for await in publisher` to the setup where a flag is set when the value is received and then I call the `notificationCenter.post(...)` inside the `try await waitForAsync { ... }` so if the first time the post is missed then it will loop again in it `waitForAsync` and post again until the publisher is ready, receives the notification value and updates the flag.

I've run each test 100 times locally and with the fix all passed whereas before in around 30-40 times runs it was failing.
Plus all runs of each test are under 0,5 seconds so we're inside the `maximum-test-execution-time-allowance` of 1 second.
So hopefully we won't encounter these tests failing in CI anymore.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-17710]: https://bitwarden.atlassian.net/browse/PM-17710?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/Core/Platform/Services/NotificationCenterServiceTests.swift
+++ b/BitwardenShared/Core/Platform/Services/NotificationCenterServiceTests.swift
@@ -7,15 +7,39 @@ final class NotificationCenterServiceTests: BitwardenTestCase {
 
     var notificationCenter: NotificationCenter!
     var subject: DefaultNotificationCenterService!
+    var didEnterBackgroundTask: Task<Void, Never>!
+    var didEnterBackgroundPublished: Bool = false
+    var willEnterForegroundTask: Task<Void, Never>!
+    var willEnterBackgroundPublished: Bool = false
 
     // MARK: Setup & Teardown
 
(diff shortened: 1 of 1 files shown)


KatherineInCode (approved):
👍🏻 Good catch, and I like this solution!

PR #1317, opened 2025-01-31
[PM-17914] Show connect to watch toggle only on iPhone

infinitepower18:
## 📔 Objective

The connect to watch toggle is now only shown if the app is run on iPhone. There is no use showing the toggle on iPad and iPod touch as those devices can't be paired to an Apple Watch.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


BitwardenShared/UI/Platform/Settings/Settings/Other/OtherSettingsState.swift
```diff
@@ -19,4 +21,11 @@ struct OtherSettingsState {
 
     /// A toast message to show in the view.
     var toast: Toast?
+
+    // MARK: Computed Properties
+
+    /// Whether the connect to watch toggle should be shown.
+    var shouldShowConnectToWatchToggle: Bool {
+        UIDevice.current.model == "iPhone"
+    }
```
fedemkr, line 30: 🎨 Even though this is fine and achieves the goal it's not easy to mock and test. So in order to improve this and be consistent to how we implement this kind of stuff I'd make this a normal property:
```suggestion
    /// Whether the connect to watch toggle should be shown.
    var shouldShowConnectToWatchToggle: Bool = false
```
Then I'd inject in the `OtherSettingsProcessor` the `SystemDevice` object (this can be done by adding `& HasSystemDevice` to the `Services` typealias in there). And then in `loadInitialValues()` I would update the state with that:
```swift
state.shouldShowConnectToWatchToggle = services.systemDevice.isIPhone
```
As you can see I've used a property `isIPhone` which should be added to the `SystemDevice` extension so the logic can be reused and have everything concentrated in one place thus it's more maintainable:
```swift
var isIPhone: Bool {
    model == "iPhone"
}
```
🎨  And then add the tests for all these new logic which is easier now that the `SystemDevice` protocol is being used so in the tests you can use `MockSystemDevice` to change the `model` and check each behavior.

fedemkr (commented):
Hi there, great catch! Thank you for your contribution and here I leave you some improvements to make in order to be consistent on how we implement this kind of logic in the app.

fedemkr (commented):
🎉  Great thanks for the improvements, just one last minor thing.

BitwardenShared/UI/Platform/Settings/Settings/Other/OtherSettingsProcessor.swift
```diff
@@ -72,6 +72,7 @@ final class OtherSettingsProcessor: StateProcessor<OtherSettingsState, OtherSett
             state.clearClipboardValue = services.settingsRepository.clearClipboardValue
             state.isAllowSyncOnRefreshToggleOn = try await services.settingsRepository.getAllowSyncOnRefresh()
             state.isConnectToWatchToggleOn = try await services.settingsRepository.getConnectToWatch()
+            state.shouldShowConnectToWatchToggle = services.systemDevice.isIPhone
```
fedemkr, line 75: 🎨 Add a test for this line in `OtherSettingsProcessorTests -> test_perform_loadInitialValues_success`

BitwardenShared/Core/Platform/Utilities/SystemDevice.swift
```diff
@@ -14,6 +14,9 @@ protocol SystemDevice {
 
     /// The version of the operating system on the device, e.g. "17.0".
     var systemVersion: String { get }
+
+    /// Whether the device is an iPhone.
+    var isIPhone: Bool { get }
```
fedemkr, line 19: 🎨 Tests are failing because `MockSystemDevice` doesn't have the conformance to this new property, could you add this getter property to it?

fedemkr (approved):

fedemkr:
Hi @infinitepower18 this PR has been approved and moved to QA for testing. Thank you so much for your contribution!

fedemkr:
Hi @infinitepower18 , QA tested this and even though it works on most cases the toggle is still shown when the device is `iPod`. So I was thinking we could just check `WCSession.default.isSupported` which would be a more general solution for any device given that it will actually check if the device supports a watch session instead of changing which device you're on.
Would you like to do the change? Otherwise I can take of it. Thanks again for your contribution!

infinitepower18:
Hi @fedemkr, do you mind if you do the change? Thanks

fedemkr:
Hi @infinitepower18 sure thing I'll fix that and again thank you so much for your contribution! 😄

PR #1329, opened 2025-02-05
[PM-8714] Key displayed error on user delete on verification error

KatherineInCode:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-8714

## 📔 Objective

This changes what we display when a user attempts to delete their account but cannot for some reason. Previously, we treated all failures as an incorrect password or OTP; however, another reason is that the user is the sole owner of an organization, which was still seeing an invalid password message. This adjusts to look at the validation error sent by the server and if it is one that we recognize, provide a localized error message tailored to the issue.

We cannot use the server error message because it is not localized (at least, as near as I can tell); and the only way we have to identify the reason is through string compare. Further discussion is in the ticket.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


BitwardenShared/UI/Platform/Application/Support/Localizations/en.lproj/Localizable.strings
```diff
@@ -1112,3 +1112,4 @@
 "CheckPasswordForDataBreaches" = "Check password for data breaches";
 "PrivateNote" = "Private note";
 "XItems" = "%1$@ items";
+"CannotDeleteThisUserBecauseItIsTheSoleOwnerOfAtLeastOneOrganization." = "Cannot delete this user because it is the sole owner of at least one organization. Please delete these organizations or upgrade another user.";
```
fedemkr, line 1115: ⛏️ I wouldn't use symbols like dot (`.`) in the localizable keys and to be consistent with other long strings I'd add `DescriptionLong` as a suffix of the key maybe something shorter like `CannotDeleteUserSoleOwnerDescriptionLong`. Thoughts?

BitwardenShared/UI/Platform/Settings/Settings/AccountSecurity/DeleteAccount/DeleteAccountProcessor.swift
```diff
@@ -72,14 +72,33 @@ final class DeleteAccountProcessor: StateProcessor<DeleteAccountState, DeleteAcc
...
+                    .defaultAlert(
+                        title: Localizations.cannotDeleteThisUserBecauseItIsTheSoleOwnerOfAtLeastOneOrganization
+                    )
+                )
+            default:
+                coordinator.showAlert(
+                    .defaultAlert(
+                        title: errorModel.singleMessage()
+                    )
+                )
```
fedemkr, line 100: 🎨 IMO if it's not a known error then we should log it in the error reporter and maybe show the alert of the generic `catch` below given that the error message from the server will not be localized and could be too technical to be displayed to end users.

BitwardenShared/UI/Platform/Application/Support/Localizations/en.lproj/Localizable.strings
KatherineInCode: 🤦🏼‍♀️ I missed the period in there, good catch.

I went back and forth on the key name, ultimately going for just the first sentence, but that one makes sense, also. We're not the most consistent historically on naming 😅

BitwardenShared/UI/Platform/Settings/Settings/AccountSecurity/DeleteAccount/DeleteAccountProcessor.swift
KatherineInCode: Hm, that does make sense—but it also means that we will have to explicitly know about any new validation error that could occur and add logic for it, rather than at the least displaying the server's (English) text 🤔 See also discussion in the ticket where Wes indicated:
> I think for now using the server string instead is fine if it gives a more accurate error state to the user.
I'll ask him about it again there

KatherineInCode: Wes confirms your intuition: I'll update to use the generic error

BitwardenShared/UI/Platform/Application/Support/Localizations/en.lproj/Localizable.strings
fedemkr: Yep, there are several strings that should be updated but I've tried to maintain consistency with the localizable files for new strings and enforce the usage of the suffix to avoid some problems we've had in the past.
Additionally, I chose `DescriptionLong` instead of `LongDescription` as suffix to be less probable for a string to have such suffix actually in the value given that normally in English the adjective goes before; so inverting it on purpose decreases such chance. Nonetheless, I'd like to know what you think about it.

KatherineInCode: Hah, I didn't even think about that. But also, that makes sense from a categorization perspective, so that "Description - Long" and "Description - Short" end up together more logically (as `DescriptionLong` and `DescriptionShort`), etc. etc.

BitwardenShared/UI/Platform/Settings/Settings/AccountSecurity/DeleteAccount/DeleteAccountProcessor.swift
```diff
@@ -72,14 +72,34 @@ final class DeleteAccountProcessor: StateProcessor<DeleteAccountState, DeleteAcc
...
-                .defaultAlert(
-                    title: otp != nil
-                        ? Localizations.invalidVerificationCode
-                        : Localizations.invalidMasterPassword
+        } catch let ServerError.error(errorModel) {
+            // The only way we know what the actual error was is by looking at the validation errors.
+            // We have to compare by string for lack of any other way of identifying the errors.
+            // This allows us to localize the error message from the server as appropriately as we can.
+            switch errorModel.singleMessage() {
+            case "User verification failed.":
```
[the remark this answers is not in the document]
fedemkr: I agree, also we could transform these error strings into a string-based enum where we track non-2xx errors that are currently being handled by the app by these kind of comparisons. This would make it easier to adapt server side with error codes or whatever strategy they decide to implement and also with maintainability here as all the string errors will be in one place to modify.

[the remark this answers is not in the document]
KatherineInCode: I definitely went back and forth on putting this in the processor, for just that reason; I just didn't have a better idea of where to put it. I think since this is a one-off it's probably still okay here, but if we get to three times doing this finding that better place is probably worth it.

fedemkr (commented):
LGTM 🎉 , although it seems to have some conflicts.

KatherineInCode:
> LGTM 🎉 , although it seems to have some conflicts.

Conflicts should be resolved now. Gotta love the strings file 😂

PR #1372, opened 2025-02-21
PM-18127: Added authenticator app icon for 2FA screen

ezimet-livefront:
## 📔 Objective

- Added icon for authenticator app method.
## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


BitwardenShared/UI/Auth/Login/TwoFactorAuth/TwoFactorAuthView.swift
```diff
@@ -120,11 +118,20 @@ struct TwoFactorAuthView: View {
...
+                .resizable()
+                .frame(width: 124, height: 124)
+        case .authenticatorApp:
+            Image(decorative: Asset.Images.Illustrations.authenticatorApp)
+                .resizable()
+                .frame(width: 124, height: 124)
+        default:
+            EmptyView()
+        }
     }
```
fedemkr, line 135: 🤔 I think it'd be nice to have a computed property in the state that just returns an optional image to show, then you can change this to return an `EmptyView` if the computed property is `nil` and bind the property to the image `decorative` if there's a value in there. Therefore if more images are added per authMethod you only need to change the state instead of creating one `Image` view per `authMethod`.
```suggestion
    /// The image representing the selected authentication method.
    @ViewBuilder private var authMethodImage: some View {
        if let authMethodImage = store.state.authMethodImage {
            Image(decorative: authMethodImage)
                .resizable()
                .frame(width: 124, height: 124)
        } else {
            EmptyView()
        }
    }
```
🤔 Additionally, I was wondering if it'd be good to replace this and the `detailText` with `PageHeaderView` as I think @KatherineInCode suggested a while back. Although, we'd need to make several things optional in the custom control and also it doesn't have the `detailImageAsset` so maybe it's not a good idea to replace it. Thoughts?

ezimet-livefront: Yeah, I don't think it would be good idea to change ```PageHeaderView``` for this case.
I think computed property is good approach here.

fedemkr (approved):

PR #1383, opened 2025-02-25
[PM-18486] Fix register Fido2 credential excluded credential found flow

fedemkr:
## 🎟️ Tracking

[PM-18486](https://bitwarden.atlassian.net/browse/PM-18486)

## 📔 Objective

Fix on Fido2 credential creation excluded credential found flow, that when a user edits such excluded credential found and removes the passkey from it. Then on going back to the list screen the previous excluded credential shouldn't be shown and the normal Fido2 credential creation flow list should be shown.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-18486]: https://bitwarden.atlassian.net/browse/PM-18486?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/UI/Vault/Vault/AutofillList/VaultAutofillListProcessor.swift
+++ b/BitwardenShared/UI/Vault/Vault/AutofillList/VaultAutofillListProcessor.swift
@@ -406,6 +406,11 @@ class VaultAutofillListProcessor: StateProcessor<// swiftlint:disable:this type_
                 }
                 guard let cipher else { continue }
 
+                guard cipher.hasFido2Credentials else {
+                    state.excludedCredentialFound = false
+                    break
+                }
+
                 let vaultListSection = try await services.vaultRepository.createAutofillListExcludedCredentialSection(
                     from: cipher
                 )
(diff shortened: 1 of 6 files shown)


BitwardenShared/UI/Vault/Vault/AutofillList/VaultAutofillListProcessor.swift
```diff
@@ -486,7 +491,10 @@ extension VaultAutofillListProcessor: Fido2UserInterfaceHelperDelegate {
 
     func informExcludedCredentialFound(cipherView: CipherView) async {
         state.excludedCredentialFound = true
-        await updateExcludedCredentialSection(from: cipherView)
+        Task { [weak self] in
+            guard let self else { return }
+            await updateExcludedCredentialSection(from: cipherView)
```
matt-livefront, line 496: 🤔 I think it'd be better to not spin up a separate task for this. The problem with this is that the task is created, but nothing is going to stop it when the user navigates away from the view. Since it listens to a publisher, I think it will just keep on listening after the view is dismissed.

An alternative to this would be in the view, create a task that performs an action when the view appears or if `state.excludedCredentialFound` changes:

```
.task(id: store.state.excludedCredentialFound) { // Could replace the existing streamAutofillItems task.
    await store.perform(.streamAutofillItems)
}
```

I think we'd likely need to add logic in the processor to call `updateExcludedCredentialSection()` or `streamAutofillItems()`, but that seems like a better approach to me. Then when `excludedCredentialFound` changes, the existing task will be cancelled and the list will be re-populated. Do you think that could work?

fedemkr: Oh, you're totally right; forgot about task cancellation on navigation. Fixed 😄 Although no extra logic needed for the stream autofill items as it's done automatically when the view appears again.

matt-livefront (approved):

PR #1405, opened 2025-03-04
[PM-13428] Remove ownership option

LRNcardozoWDF:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-13428

## 📔 Objective

Removed ownership option when the user doesn't belong to an organization since they can't select anything else other than themselves.
Disabled the ownership picker when editing an item that belongs to an organization.

## 📸 Screenshots

- User doesn't belong to an organization, so the ownership picker doesn't show.

- User belongs to an organization, so the ownership picker is available when adding a new item. 

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


BitwardenShared/UI/Vault/VaultItem/AddEditItem/AddEditItemProcessor.swift
```diff
@@ -284,6 +284,7 @@ final class AddEditItemProcessor: StateProcessor<// swiftlint:disable:this type_
                 state.collections.contains(where: { $0.id == collectionId })
             }
 
+            state.hasOrganizations = try await services.vaultRepository.hasOrganizations()
```
matt-livefront, line 287: 🤔 Do we need to consider if any of the organizations are disabled or if the user isn't confirmed in the organization? For example `VaultRepository.fetchCipherOwnershipOptions()` has the following checks:

https://github.com/bitwarden/ios/blob/dc5f0864d77feec59f72e7da20a37082f346f1aa/BitwardenShared/Core/Vault/Repositories/VaultRepository.swift#L1012-L1015

I wonder if it would be easier to check `state.ownershipOptions.contains { !$0.isPersonal }` rather than do another fetch from the database?

BitwardenShared/UI/Vault/VaultItem/AddEditItem/AddEditItemView.swift
```diff
@@ -166,7 +166,7 @@ struct AddEditItemView: View {
                 )
                 .accessibilityIdentifier("FolderPicker")
 
-                if store.state.configuration.isAdding, let owner = store.state.owner {
+                if let owner = store.state.owner, store.state.hasOrganizations {
```
matt-livefront, line 169: ❓ I'm seeing that if I edit an item in my personal vault, I can change the owner. Is that expected, or should the ownership field be disabled? (e.g. should the disabled modifier not have the `!owner.isPersonal` check?). Because I thought you couldn't change the other while editing without also adding a separate API call to move the item.

And similarly, if I edit an item in an organization, should I be able to toggle the collection toggles on this screen? I think that also requires a separate API call which is why you needed to go to the edit collections screen.

LRNcardozoWDF: If you belong to an organization and the personal vault isn't disabled it will be possible to change the owner. And yes, we should be able to toggle the collection toggles. 
However, as you said both cases require a separate API call, thank you for feedback.
I'm going to remove these changes as it's more complex than I originally thought and falls outside the scope of the bug.

matt-livefront (dismissed):

BitwardenShared/UI/Vault/VaultItem/AddEditItem/AddEditItemState.swift
```diff
@@ -45,6 +45,9 @@ protocol AddEditItemState: Sendable {
     /// The state for guided tour view.
     var guidedTourViewState: GuidedTourViewState { get set }
 
+    /// Whether the user belongs to any organization.
+    var hasOrganizations: Bool { get set }
+
```
fedemkr, line 50: 🎨 From what I can see this depends on the `cipher` and the `ownershipOptions`. Couldn't this be just readonly and calculate the value based off these other properties? Correct me if I'm wrong but something like:
```suggestion
    /// Whether the user belongs to any organization.
    var hasOrganizations: Bool {
        cipher.organizationId != nil || ownershipOptions.contains { !$0.isPersonal }
    }
```
This would make the change much simpler and you avoid having to update the variable from outside the state.

fedemkr (approved):

BitwardenShared/UI/Vault/VaultItem/CipherItemState.swift
```diff
@@ -140,6 +137,10 @@ struct CipherItemState: Equatable { // swiftlint:disable:this type_body_length
         self
     }
 
+    var hasOrganizations: Bool {
+        cipher.organizationId != nil || ownershipOptions.contains { !$0.isPersonal }
+    }
```
fedemkr, line 142: ⛏️ Add tests.

fedemkr (changes requested):

BitwardenShared/UI/Vault/VaultItem/CipherItemState.swift
LRNcardozoWDF: Nice catch, ty! Tests added [c27b6cb](https://github.com/bitwarden/ios/pull/1405/commits/c27b6cbc156e110efb5f3a35b36b41cd076ca0af).

fedemkr: ⛏️ Nice! Could you add an additional test when `organizationId` is `nil` and `state.ownershipOptions` is empty? It would be the intermediate state while the view is fetching the cipher options.

BitwardenShared/UI/Vault/VaultItem/AddEditItem/AddEditItemView.swift
```diff
@@ -169,7 +169,7 @@ struct AddEditItemView: View {
                 )
                 .accessibilityIdentifier("FolderPicker")
 
-                if store.state.configuration.isAdding, let owner = store.state.owner {
+                if store.state.configuration.isAdding, let owner = store.state.owner, store.state.hasOrganizations {
```
fedemkr: ⛏️ It's really minor but could you swap places between the new check and `let owner..`?
```suggestion
                if store.state.configuration.isAdding, store.state.hasOrganizations, let owner = store.state.owner {
```
I believe in this way it's better for short-circuiting because if the state has no organizations then there's no need to have the new variable reference for `owner`.

BitwardenShared/UI/Vault/VaultItem/CipherItemStateTests.swift
```diff
@@ -211,6 +211,41 @@ class CipherItemStateTests: BitwardenTestCase { // swiftlint:disable:this type_b
...
+    func test_hasOrganizations_whenOwnershipIncludesNonPersonal_returnsTrue() throws {
+        let cipher = CipherView.fixture()
+        var state = try XCTUnwrap(CipherItemState(existing: cipher, hasPremium: true))
+        state.ownershipOptions = [CipherOwner.organization(id: "org123", name: "Organization")]
+
+        XCTAssertTrue(state.hasOrganizations)
+    }
+
+    /// `hasOrganizations` is false when ownership options are empty and organizationId is nil.
+    func test_hasOrganizations_whenCipherBelongsToPersonal_returnsSmth() throws {
```
fedemkr: ⛏️ Did you mean `..._returnsFalse` instead of `..._returnsSmth`?

fedemkr (approved):

PR #1430, opened 2025-03-11
[PM-16916] Custom fields need label to be saved

LRNcardozoWDF:
## 🎟️ Tracking

[PM-16916](https://bitwarden.atlassian.net/browse/PM-16916)

## 📔 Objective

Disable OK button if there's no input.

## 📸 Screenshots

- Custom field name is empty so the OK button is disabled.

- Custom field name is filled so the OK button is enabled.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-16916]: https://bitwarden.atlassian.net/browse/PM-16916?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenShared/UI/Platform/Application/Utilities/Alert/Alert/Alert.swift
```diff
@@ -26,6 +26,9 @@ public class Alert {
     /// The title of the message that is displayed at the top of the alert.
     let title: String?
 
+    /// The current alert shown.
+    private weak var currentAlertController: UIAlertController?
```
fedemkr, line 30: 🤔 Even though it's `weak` I'm not very fond of having this reference here in order to apply this logic.
Could you check if you could add a kind of inversion of control in the `AlertTextField -> textChanged` so it calls a function passed to the `AlertTextField` which ends up updating a reference of the alert action?

fedemkr: @KatherineInCode @matt-livefront what do you think?

matt-livefront: I agree. I'm also kind of surprised that this works because I'm not sure what's keeping a reference to the `Alert` to keep it from being deallocated. If Fede's suggestion doesn't work out, we already have an alert controller subclass [here](https://github.com/bitwarden/ios/blob/05d2c8a13559510b5959978a71bcc101bc916392/BitwardenShared/UI/Platform/Application/Utilities/Alert/Alert/Alert.swift#L197), could we use that house some of this logic?

BitwardenShared/UI/Platform/Application/Utilities/Alert/Alert/AlertAction.swift
```diff
@@ -28,10 +31,12 @@ public class AlertAction {
     ///
     public init(
         title: String,
+        shouldEnableAction: (([AlertTextField]) -> Bool)? = nil,
```
KatherineInCode, line 34: 🤔 Despite our desire to alphabetize parameters, I still somewhat prefer parameters that take closures to be piled at the end to potentially take advantage of trailing closure syntax.

BitwardenShared/UI/Platform/Application/Utilities/Alert/Alert/Alert.swift
KatherineInCode: I also agree that we should try to do inversion of control here if possible.

BitwardenShared/UI/Platform/Application/Utilities/Alert/Alert/Alert.swift
```diff
@@ -99,39 +99,68 @@ public class Alert {
...
                 textField.addTarget(
                     alertTextField,
                     action: #selector(AlertTextField.textChanged(in:)),
                     for: .editingChanged
                 )
             }
+
+            if updateActionsIfNeeded {
+                alertTextField.onTextChanged = { [weak self, weak alertController] in
+                    guard let self, let alert = alertController else { return }
```
fedemkr, line 127: 🎨 I think you can directly use `alertController` without assigning to `alert` here:
```suggestion
                    guard let self, let alertController else { return }
```

BitwardenShared/UI/Platform/Application/Utilities/Alert/Alert/Alert.swift
```diff
@@ -99,39 +99,68 @@ public class Alert {
...
+
                 textField.addTarget(
                     alertTextField,
                     action: #selector(AlertTextField.textChanged(in:)),
                     for: .editingChanged
                 )
             }
+
+            if updateActionsIfNeeded {
+                alertTextField.onTextChanged = { [weak self, weak alertController] in
```
fedemkr, line 126: 🤔 I'm not sure if this could introduce a retain cycle and if so we could set it to `nil` when dismissing the alert. What do you think @matt-livefront @KatherineInCode ?

KatherineInCode: The only thing I see that I think could introduce a retain cycle is the `alertActions.enumerated()` I think. Everything else is either weakified or declared in the closure. And I don't think it's escaping?

Could be worth double-checking by triggering it a few times and looking in the memory graph?

matt-livefront: The weak references here will prevent this from causing a retain cycle, so I think this is alright. I verified in the app that the alert controller and text field instances are being deallocated after the alert is dismissed.

fedemkr (approved):

KatherineInCode (approved):

PR #1438, opened 2025-03-17
[PM-18223] Add share error details button to error alerts without a message

matt-livefront:
## 🎟️ Tracking

[PM-18381](https://bitwarden.atlassian.net/browse/PM-18381) - Add Feature Flag for Error Reporting
[PM-18223](https://bitwarden.atlassian.net/browse/PM-18223) - Implement 'Share error details' Button in Error Dialogs

## 📔 Objective

This adds the mobile error reporting feature flag and implements the foundational pieces to support showing the 'Share error details' button in error alerts.

The 'Share error details' button will show for any default alerts which don't already show a message. Because 'Share error details' will need access to some services and the coordinator to show a share sheet, I've changed how displaying default error alerts works.

Before: `coordinator.showAlert(.networkResponseError(error))`
After: `await coordinator.showErrorAlert(error: error, services: services)`

This was the best compromise to keep this to a one-liner in a processor when presenting this alert, while supporting this new functionality. I tried writing an extension on `Coordinator` that would allow access to `services` without needing to pass those in, but ultimately couldn't get it to work with the protocol definition that we have. I think we would instead need a `Coordinator` base class, which we could explore in the future if needed.

To keep this PR small and focused on the setup pieces, I only updated `VaultListProcessor` to use this new pattern and will follow up with another PR updating the rest of the app.

> [!NOTE]
> This introduces a few SwiftLint warnings, which will be resolved in the next PR:
> >Backward matching of the unlabeled trailing closure is deprecated; label the argument with 'tryAgain' to suppress this warning.

## 📸 Screenshots

| Before | After |
|--------|--------|
| Screenshot 2025-03-17 at 10 41 52 AM | Screenshot 2025-03-17 at 10 42 06 AM | 

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-18381]: https://bitwarden.atlassian.net/browse/PM-18381?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PM-18223]: https://bitwarden.atlassian.net/browse/PM-18223?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/UI/Platform/Application/Utilities/Coordinator.swift
+++ b/BitwardenShared/UI/Platform/Application/Utilities/Coordinator.swift
@@ -6,6 +6,7 @@ protocol Coordinator<Route, Event>: AnyObject {
     // MARK: Types
 
     typealias ErrorAlertServices = HasConfigService & HasErrorReporter
+    typealias ErrorAlertTryAgain = (() async -> Void)?
 
     associatedtype Event
     associatedtype Route
@@ -52,8 +53,9 @@ protocol Coordinator<Route, Event>: AnyObject {
     ///   - error: The error that occurred, used to customize the details of the alert.
     ///   - services: A group of services used to gather additional information if the error details
     ///     are shared.
+    ///   - tryAgain: An optional closure allowing the user to retry whatever triggered the error.
     ///
-    func showErrorAlert(error: Error, services: ErrorAlertServices) async
+    func showErrorAlert(error: Error, services: ErrorAlertServices, tryAgain: (() async -> Void)?) async
 
     /// Shows the loading overlay view.
     ///
@@ -151,6 +153,17 @@ extension Coordinator {
     func showAlert(_ alert: Alert) {
         showAlert(alert, onDismissed: nil)
     }
+
+    /// Shows an alert for an error that occurred.
+    ///
+    /// - Parameters:
+    ///   - error: The error that occurred, used to customize the details of the alert.
+    ///   - services: A group of services used to gather additional information if the error details
+    ///     are shared.
+    ///
+    func showErrorAlert(error: Error, services: ErrorAlertServices) async {
+        await showErrorAlert(error: error, services: services, tryAgain: nil)
+    }
 }
 
 extension Coordinator where Self.Event == Void {
@@ -184,14 +197,15 @@ extension Coordinator where Self: HasNavigator {
     ///   - error: The error that occurred, used to customize the details of the alert.
     ///   - services: A group of services used to gather additional information if the error details
     ///     are shared.
+    ///   - tryAgain: An optional closure allowing the user to retry whatever triggered the error.
     ///
-    func showErrorAlert(error: Error, services: ErrorAlertServices) async {
+    func showErrorAlert(error: Error, services: ErrorAlertServices, tryAgain: (() async -> Void)?) async {
         let alert = if await services.configService.getFeatureFlag(.mobileErrorReporting) {
             Alert.networkResponseError(error, shareErrorDetails: {
                 // TODO: PM-18224 Show share sheet to export error details
-            })
+            }, tryAgain: tryAgain)
         } else {
-            Alert.networkResponseError(error)
+            Alert.networkResponseError(error, tryAgain: tryAgain)
         }
         showAlert(alert)
     }
--- a/BitwardenShared/UI/Platform/Application/Utilities/Coordinator.swift
+++ b/BitwardenShared/UI/Platform/Application/Utilities/Coordinator.swift
@@ -2,8 +2,10 @@ import Foundation
 
 /// A protocol for an object that performs navigation via routes.
 @MainActor
-public protocol Coordinator<Route, Event>: AnyObject {
+protocol Coordinator<Route, Event>: AnyObject {
     // MARK: Types
+
+    typealias ErrorAlertServices = HasConfigService & HasErrorReporter
 
     associatedtype Event
     associatedtype Route
@@ -43,6 +45,15 @@ public protocol Coordinator<Route, Event>: AnyObject {
     ///   - onDismissed: An optional closure that is called when the alert is dismissed.
     ///
     func showAlert(_ alert: Alert, onDismissed: (() -> Void)?)
+
+    /// Shows an alert for an error that occurred.
+    ///
+    /// - Parameters:
+    ///   - error: The error that occurred, used to customize the details of the alert.
+    ///   - services: A group of services used to gather additional information if the error details
+    ///     are shared.
+    ///
+    func showErrorAlert(error: Error, services: ErrorAlertServices) async
 
     /// Shows the loading overlay view.
     ///
@@ -114,7 +125,7 @@ protocol HasRouter<Event, Route> {
 
 // MARK: Extensions
 
-public extension Coordinator {
+extension Coordinator {
     /// Handles events that may require asynchronous management.
     ///
     /// - Parameter event: The event for which the coordinator handle.
@@ -143,7 +154,7 @@ public extension Coordinator {
 }
 
 extension Coordinator where Self.Event == Void {
-    /// Provide a default No-Op when a coodrinator does not use events.
+    /// Provide a default No-Op when a coordinator does not use events.
     ///
     func handleEvent(_ event: Void, context: AnyObject?) async {
         // No-Op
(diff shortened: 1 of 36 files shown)


BitwardenShared/UI/Platform/Application/Utilities/Coordinator.swift
```diff
@@ -2,9 +2,12 @@ import Foundation
 
 /// A protocol for an object that performs navigation via routes.
 @MainActor
-public protocol Coordinator<Route, Event>: AnyObject {
+protocol Coordinator<Route, Event>: AnyObject {
     // MARK: Types
 
+    typealias ErrorAlertServices = HasConfigService & HasErrorReporter
+    typealias ErrorAlertTryAgain = (() async -> Void)?
```
fedemkr, line 9: 🤔 Why is this `typealias` needed? Couldn't the closure be used in-situ or you wanted to be flexible to be changed in all places in the future just by changing it here?

BitwardenShared/UI/Platform/Application/Utilities/Coordinator.swift
```diff
@@ -2,9 +2,12 @@ import Foundation
 
 /// A protocol for an object that performs navigation via routes.
 @MainActor
-public protocol Coordinator<Route, Event>: AnyObject {
+protocol Coordinator<Route, Event>: AnyObject {
     // MARK: Types
 
+    typealias ErrorAlertServices = HasConfigService & HasErrorReporter
```
fedemkr, line 8: 🤔 Is `errorReporter` service going to be used in a future PR? Because I'm not seeing that used here.

fedemkr (commented):
🤔 Even though this approach makes sense, I'm not very fond of passing the services through a function to the `Coordinator`, although I know it's tricky given that it's a protocol.
Do we only need the services for the feature flag check? Or are you planning to use more services in the `Coordinator -> showErrorAlert` function? Or well anywhere in the `Coordinator`?
🤔 I was thinking about having some sort of proxy or composed/wrapped Coordinator so you can add any additional logic without disrupting the original implementation. Therefore you can pass any coordinator to be wrapped in the composed one to add this feature flag logic change.

BitwardenShared/UI/Platform/Application/Utilities/Coordinator.swift
matt-livefront: My plan was to put the logic of building the error details to share within that service. So once you tap 'Share error details' it would assemble the error details and app info which could then be shared.

BitwardenShared/UI/Platform/Application/Utilities/Coordinator.swift
matt-livefront: Not strictly necessary, but it was used in enough places where it seemed nice to have. At one point it wasn't async and I made it async and that triggered a trickle down of errors so I liked the convenience of just defining it once. I'm not particularly tied to it if you prefer to define it in place though.

matt-livefront:
> 🤔 Even though this approach makes sense, I'm not very fond of passing the services through a function to the `Coordinator`, although I know it's tricky given that it's a protocol. Do we only need the services for the feature flag check? Or are you planning to use more services in the `Coordinator -> showErrorAlert` function? Or well anywhere in the `Coordinator`? 

We need the services for the feature flag check and possibly some other service that builds the error details + app info that is ultimately shared.

We also have an `isOfficialBitwardenServer` parameter in `Alert.networkResponseError()` to determine if the server is an official one or not, but because you have to manually pass this in from the processor, we've only implemented that on a single screen. But this is another case where we could replace that with a shared solution if we had access to services when building the alert.

For the flight recording feature, we talked about possibly logging breadcrumbs. The easiest place to put this could be in the coordinator, or a coordinator base class? For this reason, I kind of had in my head that we might need to create a coordinator base class (or similar solution) which has access to the coordinator's services. And if we did that we could stop passing the services around here.

I have similar concerns around passing services, but putting this logic in the coordinator seemed like the best spot and it feels like we might be headed in a direction that we could get the services directly from the coordinator and avoid this passing.

> 🤔 I was thinking about having some sort of proxy or composed/wrapped Coordinator so you can add any additional logic without disrupting the original implementation. Therefore you can pass any coordinator to be wrapped in the composed one to add this feature flag logic change.

This kind of sounds like what `AnyCoordinator`. Is that kind of what you had in mind?

matt-livefront:
@fedemkr 6d260df406a674970e06b5edc30691184484d1e9 updates this to use the services from the coordinator, let me know what you think of that.

BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessor.swift
```diff
@@ -281,7 +281,7 @@ extension VaultListProcessor {
                     )
                 }
             } else {
-                await coordinator.showErrorAlert(error: error, services: services)
+                await coordinator.showErrorAlert(error: error)
             }
             services.errorReporter.log(error: error)
```
fedemkr, line 286: 🤔 Now that I see this, I think one think to keep in mind would be to almost always log to the error reporter first before showing any error alert; so in here `services.errorReporter.log(error: error)` should be at the beginning of the `catch` instead of at the end. Therefore if the user/OS kills the app while the alert is being shown the error gets reported nonetheless to Crashlytics.
If I'm thinking this correctly if the above happens now we're going to miss the error report because the catch clause will await for the error alert to finish in order to log in the `ErrorReporter`.

fedemkr (commented):
🤔 I think this is is better than the previous solution @matt-livefront and then as you mentioned before we can explore having a base one but I think this lays the grounds for that.
The only other thing I was thinking is: if a coordinator doesn't conform to the new protocol then it wouldn't display the new alert but it would have to have a custom implementation for `showErrorAlert`, right? So I believe it's a "safe" way to know we're missing that protocol in a new Coordinator.
What do you think @KatherineInCode ?

fedemkr:
Also @matt-livefront sorry I didn't answer the previous comment earlier, thanks so much for the additional detailed information 😄 It has been really helpful to understand everything.
Regarding the wrapper, yes I was thinking something alike the `AnyCoordinator` that actually could take an `AnyCoordinator` as a parameter to add any additional logic in top of that. However, maybe your last commit solution is better than that and requires less changes in structure.

matt-livefront:
> The only other thing I was thinking is: if a coordinator doesn't conform to the new protocol then it wouldn't display the new alert but it would have to have a custom implementation for `showErrorAlert`, right? So I believe it's a "safe" way to know we're missing that protocol in a new Coordinator.

@fedemkr Exactly, you'll get a build error that the coordinator doesn't conform to the `Coordinator` protocol. And then conforming to `HasErrorAlertServices` should be easier than implementing a custom `showErrorAlert` implementation.

KatherineInCode:
@fedemkr @matt-livefront I do think exploring what a base Coordinator might look like may actually be the best route to go. I'm a little concerned about us making Coordinator things internal as well, because ultimately we'll want to move the `Coordinator` protocol or base class into `BitwardenKit` so both PM and BWA can have access to it.

BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessor.swift
matt-livefront: I can move up the error reporter call. Even though `showErrorAlert` is async though, it will continue on after the coordinator calls present on the alert controller as we don't continue waiting for the alert to be dismissed. So probably minimal risk to the app being killed before the log call.

KatherineInCode:
> depending on where you draw the line for what goes in `BitwardenKit` vs not

Which is something we're still hashing out a bit, and will be able to a bit more once I get back to finishing up the first wave of monorepo work.

To a certain extent, I think we might want to pull a lot of those services into `BitwardenKit` when we're all said and done—especially obviously-common things like error alerting, logging, and such. It is something we can also figure out when we get there, and that includes the public/internal status of various protocols/structs/classes.

I do think exploring a base class might be a good idea. I'm usually about composition over inheritance, but this seems like one of the few places doing inheritance might be better.

fedemkr:
In my opinion when we move these classes to the `BitwardenKit` we'd need to think about it so not to add too much dependency as you were saying. Like the error or config service protocols should only have the minimum that the `BitwardenKit` needs and not everything else that may be needed for the actual app. I think that goes beyond the scope of the PR so perhaps we could do the refactor in a different one when we have more clarity about what to share; I would think this whole feature will be there but we can always reiterate on this.

fedemkr:
I believe we can approve as is or would you like any other changes @KatherineInCode done directly in this PR to prepare for the broader refactor/preparation for the Kit?

KatherineInCode:
I agree that the move to `BitwardenKit` is outside the scope of this PR, and I'd rather save any real preparations until a moment closer to the last responsible one, because at that point we'll have a much better idea of what we'll need to do.

I have no objections to moving forward as-is.

fedemkr (approved):

PR #1443, opened 2025-03-18
[PM-19045] feat: Implement Lock App Intents

fedemkr:
## 🎟️ Tracking

[PM-19045](https://bitwarden.atlassian.net/browse/PM-19045)

## 📔 Objective

Implemented **Lock all accounts** App Intent and its Shortcuts configuration.
This also adds `AppShortcutsLocalizations -> AppShortcuts.strings` that needs to localize Shortcuts phrases.

Also introduced some helpers to execute pending action on the main app after an `AppIntent` is executed and an `AppContextHelper` that can be injected wherever we want to get the current app context so we know if the main app is executing the code or the extension or an `AppIntent` which IMO is easier than passing around the `AppContext` everywhere.

### Notes

Some additional information is that the `AppIntent` executes the action and at the same time it can leave a pending action to be executed when the user goes back to the main app. This serves the purpose to re-navigate the user to the appropriate screen after the intent gets executed, specially when the app is still in background and the user goes back at it. Like e.g. the user has `Never` as session timeout, puts the app to background, the lock `AppIntent` is executed from a Shortcut and the user goes back to the app then the user needs to be redirected to the vault unlock screen.
This is also necessary given that depending on the `AppIntent` it will possibly run on another process.

### Lock shortcuts
Lock shortcuts

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-19045]: https://bitwarden.atlassian.net/browse/PM-19045?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
```diff
@@ -472,6 +479,7 @@ class DefaultAuthRepository {
     ///
     init(
         accountAPIService: AccountAPIService,
+        appContextHelper: AppContextHelper,
```
ezimet-livefront, line 482: 🔨 missing doc for init.

BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
```diff
@@ -106,6 +106,10 @@ protocol AuthRepository: AnyObject {
     ///
     func isUserManagedByOrganization() async throws -> Bool
 
+    /// Locks all vaults and clears decrypted data from memory
+    /// - Parameter isManuallyLocking: Whether the user is manually locking the account.
```
ezimet-livefront, line 110:
```suggestion
    /// Locks all vaults and clears decrypted data from memory.
    /// - Parameter isManuallyLocking: Whether the user is manually locking the account.
    ///
```

BitwardenShared/UI/Platform/Application/AppProcessor.swift
```diff
@@ -675,4 +715,27 @@ extension AppProcessor: Fido2UserInterfaceHelperDelegate {
...
+        case .lock, .lockAll:
+            guard let account = data as? Account else {
+                return
+            }
+            await coordinator?.handleEvent(
+                .accountBecameActive(
+                    account,
+                    attemptAutomaticBiometricUnlock: true,
+                    didSwitchAccountAutomatically: false
+                )
```
ezimet-livefront, line 735: ❓why do we want to pass ```true```  for ```attemptAutomaticBiometricUnlock:```after user successfully locked the vault/vaults?

Bitwarden/Application/Support/AppShortcutsLocalizations/en.lproj/AppShortcuts.strings
```diff
@@ -0,0 +1,2 @@
+"Lock current ${applicationName} account" = "Lock current ${applicationName} account";
+"Lock all ${applicationName} accounts" = "Lock all ${applicationName} accounts";
```
matt-livefront, line 2: ❓ I'm curious about these strings. Does this format work? I guess I would have expected it to be something like:
```
"Lock all %1$@ accounts" = "Lock current %1$@ account";
```

But I noticed you also pulled the `BitwardenShared` localizations into the `Bitwarden` target and are using that for some strings. Do we need both? Or should we use this for any strings within the `Bitwarden` target?

matt-livefront (commented):
🤔 What happens if the app is foregrounded when the intent is invoked, like via Siri? It seems like the account doesn't lock immediately; you need to background and foreground it again. Is there a hook into performing that action immediately?

Bitwarden/Application/Support/AppShortcutsLocalizations/en.lproj/AppShortcuts.strings
fedemkr: I struggled to find this but according to [here](https://sowenjub.me/writes/localizing-app-shortcuts-with-app-intents/) this is the right format for phrases localizations on App Shortcuts.
I put the file in this location because it's only needed for the AppIntents / Shortcuts so I didn't want it to be shared but yes they need to be in another localization file called `AppShortcuts.strings`.

Bitwarden/Application/AppIntents/LockAllAccountsIntent.swift
```diff
@@ -0,0 +1,35 @@
...
+    @MainActor
+    func perform() async throws -> some IntentResult & ProvidesDialog {
+        let errorReporter = ServiceContainer.createDefaultErrorReporter()
+        do {
+            let services = ServiceContainer(
+                appContext: .appIntent(.lockAll),
+                errorReporter: errorReporter
+            )
+            let appProcessor = AppProcessor(appModule: DefaultAppModule(services: services), services: services)
+            let appIntentMediator = appProcessor.getAppIntentMediator()
```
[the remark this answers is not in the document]
fedemkr: I forgot to refactor this as this is how I started but I think that depending on the `AppIntent` it may be needed but not for these ones given that they don't need to open the app. I'll check that and see if I can refactor this or perhaps I can do that in a different PR for all the intents given that all the PRs are correlated with the same approach so it'd be better to tackle all the movement once for all. Would that be ok?

BitwardenShared/UI/Platform/Application/AppProcessor.swift
```diff
@@ -326,6 +338,15 @@ public class AppProcessor {
         services.errorReporter.log(error: error)
     }
 
+    /// Gets the `AppIntentMediator` to handle `AppIntent` actions.
+    public func getAppIntentMediator() -> AppIntentMediator {
+        appIntentMediator
+            ?? DefaultAppIntentMediator(
+                authRepository: services.authRepository,
+                configService: services.configService
+            )
```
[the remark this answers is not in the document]
fedemkr: Yeah I think that makes sense and we can ignore the `AppProcessor` and module creation for `AppIntent`s that do not require opening the app. I was trying to make a more general approach for all of them but yes I think it makes sense to move this outward.
May I do the refactor change on another PR so I can tackle all the refactor at once for all intents?

fedemkr:
@matt-livefront Forgot to add notes about the case the app is in Foreground while the AppIntent gets executed. AFAIK there's no direct hook we can use as this action would happen in another process although I'm not completely sure if that's the case when the app is actually in foreground. I would need to dig deeper to see how could we achieve this.
If I were to implement this right now I'd add some sort of publisher from something on disk we can use from both sides. However I wanted to have this PoC working to show it and then we can refine it. I'll create a ticket for this so I don't forget, thanks!

Bitwarden/Application/Support/AppShortcutsLocalizations/en.lproj/AppShortcuts.strings
matt-livefront: Ah, interesting!

BitwardenShared/UI/Platform/Application/AppProcessor.swift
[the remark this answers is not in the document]
matt-livefront: Yep, sounds good!

Bitwarden/Application/Support/AppShortcutsLocalizations/en.lproj/AppShortcuts.strings
KatherineInCode: 🤔 We'll want to make sure this gets into the Crowdin configuration as well, right?

fedemkr: Indeed, here the [task](https://bitwarden.atlassian.net/browse/PM-19508)

BitwardenShared/UI/Platform/Application/TestHelpers/MockAppIntentMediator.swift
```diff
@@ -0,0 +1,15 @@
+@testable import BitwardenShared
+
+class MockAppIntentMediator: AppIntentMediator {
+    var canRunAppIntentsResult = false
+    var lockAllUsersCalled = false
+    var lockCurrentUserCalled = false
```
KatherineInCode, line 6: ⛏️ This variable is unused, it looks like?

Bitwarden/Application/Services/ErrorReporter/ErrorReporterFactory.swift
```diff
@@ -0,0 +1,15 @@
+import BitwardenShared
+
+/// A factory to create `ErrorReporter` instances.
+enum ErrorReporterFactory {
+    // MARK: Static Functions
+
+    /// Creates the default error reporter.
+    public static func createDefaultErrorReporter() -> ErrorReporter {
```
KatherineInCode, line 8: 🎨 The [Swift API Design Guidelines](https://www.swift.org/documentation/api-design-guidelines/) indicate that factory methods should start with `make`.

KatherineInCode (approved):

PR #1448, opened 2025-03-21
PM-17634 - Update When The App Checks If Authenticator Sync is Enabled

phil-livefront:
## 🎟️ Tracking

[PM-17634](https://bitwarden.atlassian.net/browse/PM-17634)

## 📔 Objective

Previously, when launching the app on a fresh install, we checked the feature flags for bitwarden.com before attempting to start the AuthenticatorSyncService. If the feature flag was initially disabled, the sync service would never start. This led to an issue where users logging into a self-hosted server—which has the feature flag enabled (`enable-pm-bwa-sync`)—would not see sync events during that session, even if logging into a server that has the flag on. This caused the service to only start syncing after a full app restart which then had a fresh value for the feature flag.

Added the check to be inside of `determineSyncForUserId(userId:)` which is called from both Tasks within the `start` function. This change ensures that the sync service always starts but dynamically guards data emission based on the feature flag’s state. This allows the service to respond immediately when the flag is enabled, without requiring an app restart.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-17634]: https://bitwarden.atlassian.net/browse/PM-17634?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/Core/Platform/Services/AuthenticatorSyncServiceTests.swift
+++ b/BitwardenShared/Core/Platform/Services/AuthenticatorSyncServiceTests.swift
@@ -802,6 +802,27 @@ final class AuthenticatorSyncServiceTests: BitwardenTestCase { // swiftlint:disa
         XCTAssertEqual(item.username, "masked@example.com")
     }
 
+    /// When the feature flag is off, determineSyncForUserId should return early and do nothing.
+    /// 
+    @MainActor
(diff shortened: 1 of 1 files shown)


matt-livefront (approved):
Looks good!

PR #1469, opened 2025-03-27
[BWA-155] Move Constants to BitwardenKit

KatherineInCode:
## 🎟️ Tracking

## 📔 Objective

To help lay more groundwork for moving the `ConfigService` into `BitwardenKit`, this brings `Constants` into `BitwardenKit`, so they can be used throughout both apps. Per discussion below, this doesn't bring all of them, but instead creates a `Constants` file with common constants that the apps can extend to add their own as necessary.

This does not bring in `ExternalLinksConstants` or `BitwardenDeepLinkConstants`.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


fedemkr (commented):
🤔 I know this is in Draft but IMO we shouldn't move all `Constants` into the shared kit. There are several of them that are not going to be used by the BWA app. 
I think we should move the ones that are actually used in both places but leave a `Constants` file per app for specific ones.
Thoughts?

KatherineInCode:
> 🤔 I know this is in Draft but IMO we shouldn't move all `Constants` into the shared kit. There are several of them that are not going to be used by the BWA app. I think we should move the ones that are actually used in both places but leave a `Constants` file per app for specific ones. Thoughts?

I don't think that's particularly unreasonable, though identifying which ones need to be where could be tricky, too 🤔 Especially since what necessitated this were constants that are needed by code currently in both `BitwardenKit` and `BitwardenShared`—maybe it makes more sense to just have the base `Constants` enum be in `BitwardenKit` and then each app can do extensions as it needs? 🤔 That would avoid having to update call sites.

matt-livefront (approved):

fedemkr:
🤔 Not sure I follow which are the benefits of having an extension of `Constants` in each app over the other solution of one constant enum in Kit, and one enum on each shared folder of each app, therefore we wouldn't need to `import BitwardenKit` when the constant is local to the app.
Additionally, about "...identifying which ones need to be where could be tricky, too..." which would be the tricky part? Isn't it just:
- If constant is present in only one app, then it goes to the app's constants file
- If constant is present in both apps or in one app and the Kit, then it goes to the Kit's constants file

However, if you both think the current one is still the best solution I'm fine with that as well.

PR #1474, opened 2025-03-31
[PM-18091] Update cipher delete & restore permissions

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-18091

## 📔 Objective

Use new cipher permissions properties to show or hide the delete and restore buttons on the UI.

## 📸 Screenshots

https://github.com/user-attachments/assets/1598475b-303a-4f49-8914-b02fb75ea440

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/BitwardenShared/Core/Platform/Models/Enum/FeatureFlag.swift
+++ b/BitwardenShared/Core/Platform/Models/Enum/FeatureFlag.swift
@@ -65,6 +65,9 @@ enum FeatureFlag: String, CaseIterable, Codable {
     /// A feature flag for the refactor on the SSO details endpoint.
     case refactorSsoDetailsEndpoint = "pm-12337-refactor-sso-details-endpoint"
 
+    /// A feature flag for the use of new cipher permission properties.
+    case restrictItemDeletion = "pm-15493-restrict-item-deletion-to-can-manage-permission"
+
     // MARK: Test Flags
 
     /// A test feature flag that isn't remotely configured and has no initial value.
@@ -140,6 +143,7 @@ enum FeatureFlag: String, CaseIterable, Codable {
              .nativeCarouselFlow,
              .nativeCreateAccountFlow,
              .refactorSsoDetailsEndpoint,
+             .restrictItemDeletion,
              .testRemoteFeatureFlag,
              .testRemoteInitialBoolFlag,
              .testRemoteInitialIntFlag,
--- a/BitwardenShared/Core/Platform/Models/Enum/FeatureFlag.swift
+++ b/BitwardenShared/Core/Platform/Models/Enum/FeatureFlag.swift
@@ -66,7 +66,7 @@ enum FeatureFlag: String, CaseIterable, Codable {
     case refactorSsoDetailsEndpoint = "pm-12337-refactor-sso-details-endpoint"
 
     /// A feature flag for the use of new cipher permission properties.
-    case restrictItemDeletion = "pm-15493-restrict-item-deletion-to-can-manage-permission"
+    case restrictCipherItemDeletion = "pm-15493-restrict-item-deletion-to-can-manage-permission"
 
     // MARK: Test Flags
 
@@ -143,7 +143,7 @@ enum FeatureFlag: String, CaseIterable, Codable {
              .nativeCarouselFlow,
              .nativeCreateAccountFlow,
              .refactorSsoDetailsEndpoint,
-             .restrictItemDeletion,
+             .restrictCipherItemDeletion,
              .testRemoteFeatureFlag,
              .testRemoteInitialBoolFlag,
              .testRemoteInitialIntFlag,
--- a/BitwardenShared/UI/Vault/VaultItem/AddEditItem/AddEditItemProcessor.swift
+++ b/BitwardenShared/UI/Vault/VaultItem/AddEditItem/AddEditItemProcessor.swift
@@ -127,6 +127,7 @@ final class AddEditItemProcessor: StateProcessor<// swiftlint:disable:this type_
     override func perform(_ effect: AddEditItemEffect) async {
         switch effect {
         case .appeared:
+            await loadRestrictItemDeletionFlag()
             await showPasswordAutofillAlertIfNeeded()
             await checkIfUserHasMasterPassword()
             await checkLearnNewLoginActionCardEligibility()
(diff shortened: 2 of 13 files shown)


BitwardenShared/UI/Vault/VaultItem/AddEditItem/AddEditItemProcessor.swift
```diff
@@ -680,6 +681,14 @@ final class AddEditItemProcessor: StateProcessor<// swiftlint:disable:this type_
         }
     }
 
+    /// Load the restrict item deletion flag from the config service to state.
+    ///
+    func loadRestrictItemDeletionFlag() async {
+        state.restrictItemDeletionFlagEnabled = await services.configService.getFeatureFlag(
+            FeatureFlag.restrictItemDeletion
```
fedemkr, line 688: ⛏️ I think there's no need to indicate the enum, it should be implicit.
```suggestion
            .restrictItemDeletion
```

BitwardenShared/Core/Platform/Models/Enum/FeatureFlag.swift
```diff
@@ -65,6 +65,9 @@ enum FeatureFlag: String, CaseIterable, Codable {
     /// A feature flag for the refactor on the SSO details endpoint.
     case refactorSsoDetailsEndpoint = "pm-12337-refactor-sso-details-endpoint"
 
+    /// A feature flag for the use of new cipher permission properties.
+    case restrictItemDeletion = "pm-15493-restrict-item-deletion-to-can-manage-permission"
+
```
fedemkr, line 70: ⛏️ 🤔 Does this affect any item of the app, like a send or an attachment? Or does this only affect deletion of ciphers? If the latter I'd update the case name to something more explicit like `restrictCipherItemDeletion`. I would guess you wanted to match the case name with the flag string value but if we can be more explicit the better so we don't actually need to read the flag comment here to see what's being used for.
However, if you have reasons to keep using the current one please let me know.

BitwardenShared/UI/Vault/VaultItem/CipherItemState.swift
```diff
@@ -140,6 +143,26 @@ struct CipherItemState: Equatable {
...
 
+    /// Whether or not this item can be deleted by the user.
+    /// New permission model from PM-18091
+    var canBeDeletedPermission: Bool {
+        // backwards compatibility for old server versions
+        guard let cipherPermissions = cipher.permissions else {
+            return canBeDeleted
+        }
+        return cipherPermissions.delete
+    }
```
fedemkr, line 154: 🎨 I believe you don't need the extra property and you can also save making the checks in the views if you combine this property with `canBeDeleted`, i.e.: 
```swift
    /// Whether or not this item can be deleted by the user.
    var canBeDeleted: Bool {
        // New permission model from PM-18091
        if restrictItemDeletionFlagEnabled, let cipherPermissions = cipher.permissions {
            return cipherPermissions.delete
        }

        guard !collectionIds.isEmpty else { return true }
        return collections.contains { collection in
            guard let id = collection.id else { return false }
            return collection.manage && collectionIds.contains(id)
        }
    }
```

BitwardenShared/UI/Vault/VaultItem/ViewItem/ViewItemProcessor.swift
```diff
@@ -521,7 +520,9 @@ private extension ViewItemProcessor {
                 let hasPremium = await (try? services.vaultRepository.doesActiveAccountHavePremium()) ?? false
                 let hasMasterPassword = try await services.stateService.getUserHasMasterPassword()
                 let collections = try await services.vaultRepository.fetchCollections(includeReadOnly: true)
-
+                let restrictItemDeletionFlagEnabled: Bool = await services.configService.getFeatureFlag(
+                    FeatureFlag.restrictItemDeletion
+                )
```
fedemkr, line 525: ⛏️ I believe you can skip the enum name reference:
```suggestion
                let restrictItemDeletionFlagEnabled: Bool = await services.configService.getFeatureFlag(
                    .restrictItemDeletion
                )
```

BitwardenShared/UI/Vault/VaultItem/CipherItemState.swift
```diff
@@ -140,6 +143,26 @@ struct CipherItemState: Equatable {
...
+    /// Whether or not this item can be restored by the user.
+    /// New permission model from PM-18091
+    var canBeRestoredPermission: Bool {
+        // backwards compatibility for old server versions
+        guard let cipherPermissions = cipher.permissions else {
+            return isSoftDeleted
+        }
+        return cipherPermissions.restore && isSoftDeleted
+    }
+
```
fedemkr, line 165: 🎨 You can also refactor this a bit so you have less logic in the view and it's easier to remove the feature flag in the future. Additionally, it's safer to use this property as it's safe-guarded in-situ with the flag so the consumer doesn't need to know the existence of the flag:
```suggestion
    /// Whether or not this item can be restored by the user.
    /// New permission model from PM-18091
    var canBeRestoredPermission: Bool {
        // backwards compatibility for old server versions
        guard restrictItemDeletionFlagEnabled, let cipherPermissions = cipher.permissions else {
            return isSoftDeleted
        }
        return cipherPermissions.restore && isSoftDeleted
    }

```

fedemkr (approved):

BitwardenShared/Core/Platform/Models/Enum/FeatureFlag.swift
andrebispo5: Yes this is a property only for ciphers, I'll update the naming to better reflect the scope of action.

BitwardenShared/UI/Vault/VaultItem/CipherItemState.swift
andrebispo5: This one I can't add here and need to leave it in the view, because otherwise in order to maintain state, I would have to pass the Flag value to the init of the `CipherItemState` and I wan't to avoid that.

fedemkr: Alright no problem

fedemkr (approved):

PR #1479, opened 2025-04-03
[PM-19546] Add enable flight recorder screen

matt-livefront:
## 🎟️ Tracking

[PM-19546](https://bitwarden.atlassian.net/browse/PM-19546)

## 📔 Objective

Adds the enable flight recorder screen, which provides additional context on the feature and allows changing the logging duration. Enabling the flight recorder doesn't persist at this time. 

## 📸 Screenshots

https://github.com/user-attachments/assets/6c1c1215-a350-4d5d-9e16-786467eabbf5

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


BitwardenShared/Core/Platform/Models/Enum/LoggingDuration.swift
```diff
@@ -0,0 +1,24 @@
+/// An enum that represents how long to enable the flight recorder.
+///
+enum LoggingDuration: CaseIterable, Menuable {
```
fedemkr, line 3: 🤔 What do you think about changing this name to `FlightRecorderDuration` or `FlightRecorderLoggingDuration`? So we can make a clear distinction with other forms of logging like crashlytics errors.

BitwardenShared/UI/Platform/Application/Support/Localizations/en.lproj/Localizable.strings
```diff
@@ -1153,3 +1153,11 @@
 "VerifyYourIdentity"="Verify your identity";
 "ShareErrorDetails" = "Share error details";
 "FlightRecorder" = "Flight recorder";
+"EnableFlightRecorder" = "Enable flight recorder";
+"ExperiencingAnIssue" = "Experiencing an issue?";
+"EnableTemporaryLoggingDescriptionLong" = "Enable temporary logging to collect and inspect logs locally. When enabled, application states and network calls may be logged—never sensitive vault information.";
```
fedemkr, line 1158: ⛏️ Should there be a space between `logged-never`?

KatherineInCode: 🤔 It should be an em-dash (—), I would think; that just doesn't show well in monospaced fonts

BitwardenShared/Core/Platform/Models/Enum/LoggingDuration.swift
KatherineInCode: 🤔 I think I'd rather keep it as just `LoggingDuration` so it can be used in those other places, and change the references to flight records to more note it as an example of where it could be used. Though I guess other forms of logging wouldn't be on timeframes, so maybe it's better not to generalize?

fedemkr: 🤔 I would not generalize something that we're probably not going to use elsewhere. IMO it's better to have specific names for this stuff so it's easy to pinpoint where it's being used (or may be being used) and then generalize if we have new logging/structure that needs this enum. Therefore we don't anticipate for something that may never come and if it does, well it's just renaming it.
However, I'd like to hear your thoughts as well if you still want to use the generalized/current version.

KatherineInCode: Ah, that's fair. Specifying the exact thing this is a logging duration for makes sense to me, then.

matt-livefront: I think it makes sense to be specific in this case, I can update it.

KatherineInCode (approved):

fedemkr (approved):

PR #1505, opened 2025-04-16
[PM-19577] Add flight recorder log creation

matt-livefront:
## 🎟️ Tracking

[PM-19577](https://bitwarden.atlassian.net/browse/PM-19577)

## 📔 Objective

This adds the logic to create the log file when the flight recorder is enabled along with persisting the flight recorder log metadata to UserDefaults.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-19577]: https://bitwarden.atlassian.net/browse/PM-19577?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/Core/Platform/Services/FlightRecorder.swift
+++ b/BitwardenShared/Core/Platform/Services/FlightRecorder.swift
@@ -26,9 +26,18 @@ protocol FlightRecorder: Sendable {
 
     /// Appends a message to the active log, if logging is currently enabled.
     ///
-    /// - Parameter message: The message to append to the active log.
+    /// - Parameters:
+    ///   - message: The message to append to the active log.
+    ///   - file: The file that called the log method.
+    ///   - line: The line number in the file that called the log method.
     ///
-    func log(_ message: String) async
+    func log(_ message: String, file: String, line: UInt) async
+}
+
+extension FlightRecorder {
+    func log(_ message: String, file: String = #file, line: UInt = #line) async {
+        await log(message, file: file, line: line)
+    }
 }
 
 // MARK: - DefaultFlightRecorder
@@ -177,15 +186,16 @@ extension DefaultFlightRecorder: FlightRecorder {
         return activeLogSubject.map { $0 != nil }.eraseToAnyPublisher()
     }
 
-    func log(_ message: String) async {
+    func log(_ message: String, file: String, line: UInt) async {
         guard let log = await fetchActiveLog() else { return }
         do {
             let timestampedMessage = "\(dateFormatter.string(from: timeProvider.presentTime)): \(message)\n"
             try await append(message: timestampedMessage, to: log)
         } catch {
+            let fileName = URL(fileURLWithPath: file).lastPathComponent
             errorReporter.log(error: BitwardenError.generalError(
                 type: "Flight Recorder Log Error",
-                message: "Unable to write message to log: \(message)",
+                message: "\(fileName):\(line) Unable to write message to log: \(message)",
                 error: error
             ))
         }
--- a/BitwardenShared/Core/Platform/Services/FlightRecorder.swift
+++ b/BitwardenShared/Core/Platform/Services/FlightRecorder.swift
@@ -1,4 +1,6 @@
+import BitwardenKit
 @preconcurrency import Combine
+import Foundation
 
 // MARK: - FlightRecorder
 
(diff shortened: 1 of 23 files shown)


BitwardenShared/UI/Platform/Settings/Settings/About/EnableFlightRecorder/EnableFlightRecorderProcessor.swift
```diff
@@ -64,7 +64,12 @@ final class EnableFlightRecorderProcessor: StateProcessor<
...
     private func saveAndEnableFlightRecorder() async {
-        await services.flightRecorder.enableFlightRecorder(duration: state.loggingDuration)
-        coordinator.navigate(to: .dismiss)
+        do {
+            try await services.flightRecorder.enableFlightRecorder(duration: state.loggingDuration)
+            coordinator.navigate(to: .dismiss)
+        } catch {
+            services.errorReporter.log(error: error)
+            await coordinator.showErrorAlert(error: error)
+        }
```
fedemkr, line 73: 🤔 I know now only creating the log file could throw here. However, considering that `enableFlightRecorder` could change and actually create a log file and then throw. Shouldn't we disable the flight recorder on the catch if that happens?

BitwardenShared/Core/Platform/Services/FlightRecorder.swift
```diff
@@ -45,27 +69,125 @@ final class DefaultFlightRecorder {
...
+            let timestampedMessage = "\(dateFormatter.string(from: timeProvider.presentTime)): \(message)\n"
+            try await append(message: timestampedMessage, to: log)
+        } catch {
+            errorReporter.log(error: BitwardenError.generalError(
+                type: "Flight Recorder Log Error",
+                message: "Unable to write message to log: \(message)",
+                error: error
+            ))
+        }
     }
```
fedemkr, line 192: 🤔 Should we pass the calling function or file line to this log function so if an error is thrown we can add it as a context of the message in order to check in which context this is throwing? I imagine we could have a report on this where it could be working on most places but perhaps it doesn't when called from a particular place, like an iOS Extension method or something like that.
What do you think?

BitwardenShared/UI/Platform/Settings/Settings/About/EnableFlightRecorder/EnableFlightRecorderProcessor.swift
matt-livefront: I feel like that should be the responsibility of `FlightRecorder`. I did set it up where it attempts to create the file first, and then only if that succeeds will it save the metadata to user defaults and enable the flight recorder. So I don't think you could get into a scenario where it would fail but still enable the flight recorder.

BitwardenShared/Core/Platform/Services/FlightRecorder.swift
matt-livefront: Good idea, I can add that information.

BitwardenShared/Core/Platform/Services/FlightRecorder.swift
```diff
@@ -45,27 +69,125 @@ final class DefaultFlightRecorder {
...
+    /// Appends a log message to the specified log.
+    ///
+    /// - Parameters:
+    ///   - message: The message to append to the log.
+    ///   - log: The log metadata for determining the log file to append the message to.
+    ///
+    private func append(message: String, to log: FlightRecorderData.LogMetadata) async throws {
+        let url = try fileURL(for: log)
+        try fileManager.append(Data(message.utf8), to: url)
+    }
```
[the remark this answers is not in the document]
matt-livefront: Ah, the extension is a good point. I may have to move the log files to access them in the extension. I think the filesystem is only locked prior to the first unlock after booting, but I'll try testing that.

BitwardenShared/UI/Platform/Settings/Settings/About/EnableFlightRecorder/EnableFlightRecorderProcessor.swift
fedemkr: Awesome, thanks for checking!

BitwardenShared/Core/Platform/Services/FlightRecorder.swift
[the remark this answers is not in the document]
matt-livefront: I moved the shared directory into the app group. As far as the phone locking, I wasn't able to see any problems with this while testing. I tried logging after a delay when the app went to the background while also locking my device and the log is still written to.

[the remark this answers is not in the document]
fedemkr: Awesome thanks Matt! 🎉

fedemkr (approved):

PR #1506, opened 2025-04-16
[BWA-155] Copy ConfigService to AuthenticatorShared

KatherineInCode:
## 🎟️ Tracking

## 📔 Objective

This copies the `ConfigService` from PM wholesale into BWA, overwriting the non-network-aware version that was previously there. This will allow remote feature flag enablement in BWA.

Having two copies like this is not a good long-term solution, however; this is a stopgap to enable functionality. However, to move actual `ConfigService` into `BitwardenKit` will require a bit more work around `StateService` and how we manage the `FeatureFlag` enum, which can all come in a later PR.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


AuthenticatorShared/Core/Platform/Services/ConfigService.swift
```diff
@@ -244,4 +296,50 @@ class DefaultConfigService: ConfigService {
...
+        // while this is loading.
+        let userId = try? await stateService.getActiveAccountId()
+
+        do {
+            let configResponse = try await configApiService.getConfig()
+            let serverConfig = ServerConfig(
+                date: timeProvider.presentTime,
+                responseModel: configResponse
+            )
+            try? await setStateServerConfig(serverConfig, isPreAuth: isPreAuth, userId: userId)
```
vvolkgang, line 314: ❓ When I looked at this I wasn't sure what would be ideal for `setStateServerConfig`, should we set `isPreAuth` to `true` for state to be saved using `setPreAuthServerConfig`? That would match PMs approach but I'm not sure if it goes against other parts of Authenticator and if they expect state to be tied to `userId == "local"`  .

KatherineInCode: If we do that, it should be set in whatever *calls* `ConfigService`, rather than inside of `ConfigService` itself, IMO. That way we keep the `ConfigService` the same between BWA and PM so that when we can drop it down into `BitwardenKit` we don't have to resolve differences in behavior.

I was planning on dealing with the "actually calling `ConfigService` and making sure that's set up right" in the next PR, though.

That said, I think it's probably reasonable for us to treat it as being in a pre-auth situation, and in that PR I can make sure we do that.

vvolkgang: Yep agreed; only looked at this yesterday because the debug menu doesn't set that, if we go with that direction as part of this PR we should probably update how the debug menu calls getConfig: 

https://github.com/bitwarden/ios/blob/beb392c97771541f1e9c96dcb2e529c2b7854e84/AuthenticatorShared/Core/Platform/Services/ConfigService.swift#L235

KatherineInCode: @vvolkgang Ah, good point. Do you think we should do that as part of this PR, or the next one that actually uses `ConfigService` everywhere?

@fedemkr Hm, I'll poke at that a bit and see what happens

PR #1524, opened 2025-04-23
[PM-18414] CI restructuring #2 - Consolidate /Configs folder and .xcconfig files

vvolkgang:
## 🎟️ Tracking

PM-18414 

## 📔 Objective

1. Moves Authenticator `.xcconfig` files to `/Configs`
2. Keeps the ability to independent local overrides per app with the following naming pattern: `Local-bwpm.xcconfig` for Password Manager and `Local-bwa.xcconfig` for Authenticator.

These changes will later be used by the new workflows, splitting them into their own PR. 

🟢 CI test run - https://github.com/bitwarden/ios/actions/runs/14626821608

---

📌 Possible further enhancement - We have a lot of overlap between debug / release files, I believe we can reduce some of it by using conditional configs:

* https://developer.apple.com/documentation/xcode/adding-a-build-configuration-file-to-your-project/#Apply-a-setting-conditionally-to-a-platform-or-architecture
* https://pewpewthespells.com/blog/xcconfig_guide.html#CondVarConfig

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


Scripts-bwa/select_variant.sh
```diff
@@ -21,8 +21,8 @@ variant=$1
 
 echo "🧱 Setting build variant to ${bold}${variant}${normal}."
 
-local_xcconfig_file="Configs-bwa/Local.xcconfig"
-export_options_file="Configs-bwa/export_options.plist"
+local_xcconfig_file="Configs/Local-bwa.xcconfig"
+export_options_file="Configs/export_options.plist"
```
vvolkgang, line 25: @KatherineInCode Should we also expect app specific plists here (e.g. `export_options-bwa.plist`)?

.gitignore
```diff
@@ -73,6 +73,8 @@ fastlane/README.md
 
 # Local xcconfig
 Configs/Local.xcconfig
+Configs/Local-bwpm.xcconfig
```
KatherineInCode, line 76: 🤔 I wonder if we could just make the rule `Configs/Local*.xcconfig` and cover all our future bases

Scripts-bwa/select_variant.sh
KatherineInCode: Hm, thinking about it, I think so? They have rather different plists. Though at some point the `select_variant` scripts could possibly be combined.

KatherineInCode: Actually, cancel that. This is only run on a build, and so they won't step on each other, so I think we're good if there's just the one plist file they both potentially write to. But each job will have its own separate copy of it, so there shouldn't be an issue, I think?

KatherineInCode (approved):
Looks good, assuming we don't have an issue with the plists.

Scripts-bwa/select_variant.sh
vvolkgang: Yep, that's not an issue from a CI build perspective but I wasn't sure if there was some local building angle I was missing. I'll leave it for now, easy change if we figure something out.

KatherineInCode (approved):

.gitignore
vvolkgang: _double facepalm_ good catch, updated!

PR #1528, opened 2025-04-25
chore/SRE-583 Deprecate usage of Auth-Email Header

bitwarden-charlie:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/SRE-583

## 📔 Objective

Deprecate usage of Auth-Email Header

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/BitwardenShared/Core/Auth/Services/API/Auth/Requests/IdentityTokenRequest.swift
+++ b/BitwardenShared/Core/Auth/Services/API/Auth/Requests/IdentityTokenRequest.swift
@@ -51,10 +51,8 @@ struct IdentityTokenRequest: Request {
 
     /// HTTP headers to be sent in the request.
     var headers: [String: String] {
-        guard case let .password(email, _) = requestModel.authenticationMethod else {
-            return [:]
-        }
-        return ["Auth-Email": Data(email.utf8).base64EncodedString().urlEncoded()]
+        // Auth-Email header is deprecated and no longer needed
+        return [:]
     }
 
     /// The HTTP method for this request.
--- a/BitwardenShared/Core/Auth/Services/API/Auth/Requests/IdentityTokenRequestTests.swift
+++ b/BitwardenShared/Core/Auth/Services/API/Auth/Requests/IdentityTokenRequestTests.swift
@@ -73,9 +73,9 @@ class IdentityTokenRequestTests: BitwardenTestCase {
         XCTAssertTrue(subjectAuthorizationCode.headers.isEmpty)
     }
 
-    /// `headers` returns the headers needed for a password request.
+    /// `headers` returns empty headers for a password request as Auth-Email is deprecated.
     func test_headers_password() {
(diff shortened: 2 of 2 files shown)


BitwardenShared/Core/Auth/Services/API/Auth/Requests/IdentityTokenRequest.swift
```diff
@@ -51,10 +51,8 @@ struct IdentityTokenRequest: Request {
 
     /// HTTP headers to be sent in the request.
     var headers: [String: String] {
-        guard case let .password(email, _) = requestModel.authenticationMethod else {
-            return [:]
-        }
-        return ["Auth-Email": Data(email.utf8).base64EncodedString().urlEncoded()]
+        // Auth-Email header is deprecated and no longer needed
```
matt-livefront, line 54: Is this change live in any environment? I tested this in dev, QA, and prod but always get a 400 error from the `/identity/connect/token` endpoint when attempting to log in.

```
POST https://vault.usdev.bitwarden.pw/identity/connect/token 400
Response:
{
    "error": "invalid_grant",
    "error_description": "Auth-Email header invalid."
}
```

fedemkr (commented):
🤔Additionally to what Matt commented, should this be behind a feature flag so we can better control when to disable the code? Not sure when this was deprecated on server, but there may be people using old self-host server versions and they will need the header to perform a good request.

sneakernuts:
> 🤔Additionally to what Matt commented, should this be behind a feature flag so we can better control when to disable the code? Not sure when this was deprecated on server, but there may be people using old self-host server versions and they will need the header to perform a good request.

This is a good call out. Server-side deprecation was done here https://github.com/bitwarden/server/pull/5709

We are past the support window according to https://bitwarden.com/help/bitwarden-software-release-support/

BitwardenShared/Core/Auth/Services/API/Auth/Requests/IdentityTokenRequest.swift
sneakernuts: @matt-livefront the change should be live in all environments server-side.

vvolkgang:
Added the server PR to the description. @fedemkr are you ok with the changes?

BitwardenShared/Core/Auth/Services/API/Auth/Requests/IdentityTokenRequest.swift
fedemkr: I think we can just remove the `headers` property of the `IdentityTokenRequest` then, as the default implementation of the property is the same as the one here. I don't think it's necessary to leave the comment either about `Auth-Email` being deprecated as there's no custom logic here for that and won't be needed anymore as far as I understand.

sneakernuts: @fedemkr thank you. Done!

PR #1550, opened 2025-05-05
[BWA-155] refactor: Convert FeatureFlag to struct in BitwardenKit

KatherineInCode:
## 🎟️ Tracking

## 📔 Objective

In order to continue the migration of `ConfigService` into `BitwardenKit`, this moves `FeatureFlag` into `BitwardenKit`. However, to accomplish this, it has been turned into a struct which PM and BWA extend with static members. This ends up being fairly straightforward, and avoids significant changes at the call sites.

I've also gotten rid of the static `debugFeatureFlags` variable, and instead changed those to be passed in. This change removes test feature flags from the apps themselves (since now they can exist in a mocks-only extension), which means a separate variable to clearly filter out test flags is unnecessary. As well, getting rid of the static means that each app can provide its own list of flags that should be in the debug menu (currently set to `allCases`). It's not the most elegant solution, but will continue to work as more things move into `BitwardenKit`.

I did try going down a path of making it a protocol, but that quickly got too complicated in terms of handling `any FeatureFlag` and trying to make `ConfigService` generic against a `FeatureFlag` protocol. The struct solution ends up being much cleaner and more straightforward.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


AuthenticatorShared/Core/Platform/Models/Enum/FeatureFlag.swift
```diff
@@ -5,82 +5,16 @@ import Foundation
 
 /// An enum to represent a feature flag sent by the server
 ///
-enum FeatureFlag: String, CaseIterable, Codable {
+extension FeatureFlag: @retroactive CaseIterable {
     // MARK: Feature Flags
 
-    /// A feature flag that determines whether or not the password manager sync capability is enabled.
-    case enablePasswordManagerSync = "enable-pm-bwa-sync"
+    static let enablePasswordManagerSync = FeatureFlag(
```
matt-livefront, line 11: ⛏️ Should we keep this documentation?

KatherineInCode: Yes, we should. Good catch!

matt-livefront (approved):

PR #1636, opened 2025-06-03
[PM-21681] Build out Shared Keychain functionality

KatherineInCode:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-21681

## 📔 Objective

This moves the classes for managing the Shared Keychain between PM and BWA into `BitwardenKit` from `AuthenticatorBridgeKit`, along with building out the set of things that it can store. This will enable building a shared timeout service in `BitwardenKit` that both apps can use.

The new functionality is not currently in use, but I wanted to have a PR for the shared-keychain architecture changes to encapsulate discussion and changes there, so that the shared timeout service PR would have less noise.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


BitwardenKit/Core/Platform/Services/SharedKeychain/SharedKeychainRepository.swift
```diff
@@ -0,0 +1,131 @@
+import Foundation
+
+public enum SharedTimeoutApplication: String {
+    case authenticator = "bwa"
+    case passwordManager = "pm"
+}
```
fedemkr: 🤔 If possible `BitwardenKit` shouldn't need to know anything about the apps that is using the Kit. IMO these strings should be configurable like using a property abstraction and compute it on each app. Another possibility is to have an "initialization" step of the Kit telling anything you need about the current app.

BitwardenKit/Core/Platform/Services/SharedKeychain/SharedKeychainRepository.swift
```diff
@@ -0,0 +1,131 @@
...
+        userId: String
+    ) async throws
+}
+
+public class DefaultSharedKeychainRepository: SharedKeychainRepository {
+    let storage: SharedKeychainStorage
+
+    public init(storage: SharedKeychainStorage) {
+        self.storage = storage
+    }
```
fedemkr: ⛏️ Add docs

BitwardenKit/Core/Platform/Services/SharedKeychain/SharedKeychainStorage.swift
```diff
@@ -4,47 +4,60 @@ import Foundation
...
-    /// Stores the access token for a user in the keychain.
+    /// Deletes the value in the keychain for the given item.
     ///
-    /// - Parameter value: The authenticator key to store.
+    /// - Parameters:
+    ///   - value: The value (Data) to be stored into the keychain
+    ///   - item: The item for which to store the value in the keychain.
     ///
-    func setAuthenticatorKey(_ value: Data) async throws
+    func deleteValue(for item: SharedKeychainItem) async throws
```
fedemkr: ⛏️ Fix functions order.

BitwardenKit/Core/Platform/Services/SharedKeychain/SharedKeychainRepository.swift
KatherineInCode: The apps are going to need to know about each other by way of BitwardenKit, because the whole thing this is driving towards is that PM will put data in the shared keychain about when the user will be automatically logged out, so that BWA can read it and also do the relevant parts of the logout (viz. deleting data in the shared storage)

KatherineInCode: 🤔 Though I do see your point, and wonder, then, if user ID is sufficient for figuring out the logout

...I think I have a question or two for Product I'll put on the ticket

matt-livefront: Does it make sense to keep this shared code for PM <-> BWA sync in `AuthenticatorBridgeKit` and keep `BitwardenKit` fairly app generic?

KatherineInCode: That's a good point, also—if this is primarily data that's PM -> BWA, then it probably makes sense to streamline/optimize around that and keep it in ABK, rather than the more generic solution I was pursuing. YAGNI and all that.

fedemkr: Yes, I agree. IMO BitwardenKit should be app agnostic and if we need something to be aware of both apps it should be in ABK.

KatherineInCode: Yep, I can move all this into ABK (where the pre-existing part was, anyways), and will also shift to a "this is when the login expires" model instead of storing each of the individual bits for the calculation

AuthenticatorBridgeKit/SharedKeychain/Mocks/MockSharedKeychainRepository.swift
```diff
@@ -0,0 +1,51 @@
...
+    public var authenticatorKey: Data?
+    public var errorToThrow: Error?
+    public var accountAutoLogoutTime = [String: Date]()
+
+    public init() {}
+
+    public func generateKeyData() -> Data {
+        let key = SymmetricKey(size: .bits256)
+        return key.withUnsafeBytes { Data(Array($0)) }
+    }
```
fedemkr: 🤔 Maybe I'm missing it but is this being used somewhere?

AuthenticatorBridgeKit/SharedKeychain/SharedKeychainRepository.swift
```diff
@@ -0,0 +1,106 @@
+import BitwardenKit
+import Foundation
+
+public enum SharedTimeoutApplication: String {
+    case authenticator = "bwa"
+    case passwordManager = "pm"
+}
```
fedemkr: 🤔 Is this being used somewhere? Or will this be used in a future PR?
⛏️ Also, if keeping this add docs.

fedemkr (commented):
👍 Besides the comments, I think it looks good.

AuthenticatorBridgeKit/SharedKeychain/Mocks/MockSharedKeychainRepository.swift
KatherineInCode: Nope, I just missed removing it

KatherineInCode: Yes, a couple of the tests use it to generate mock data. I can add a comment to that effect

fedemkr (approved):

PR #1644, opened 2025-06-05
[PM-22358] Refactor App Intents in order to fix several issues

fedemkr:
## 🎟️ Tracking

[PM-19400](https://bitwarden.atlassian.net/browse/PM-19400)
[PM-22358](https://bitwarden.atlassian.net/browse/PM-22358)
[PM-22360](https://bitwarden.atlassian.net/browse/PM-22360)
[PM-22367](https://bitwarden.atlassian.net/browse/PM-22367)

## 📔 Objective

Refactor App Intents in order to fix several issues:
- Refactor `ServiceContainer` so it allows to be used as a Singleton, i.e. sharing the instance between the App Intents and the main App.
- Removed app intent from the app context given they share the `ServiceContainer`.
- Improved handling of pending app intent actions so they are executed as well when the app is in foreground and improves stability when app comes from background.
- Fixed logout all accounts to work when app is in background where the SDK client was being initialized twice because of not sharing `ServiceContainer`; also fixed and improved stability of logging out all users.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-19400]: https://bitwarden.atlassian.net/browse/PM-19400?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PM-22358]: https://bitwarden.atlassian.net/browse/PM-22358?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PM-22360]: https://bitwarden.atlassian.net/browse/PM-22360?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PM-22367]: https://bitwarden.atlassian.net/browse/PM-22367?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenShared/Core/Platform/Services/ServiceContainer.swift
```diff
@@ -859,6 +864,36 @@ public class ServiceContainer: Services { // swiftlint:disable:this type_body_le
...
+    ///   - application: The application instance.
+    ///   - errorReporter: The closure to get a service used by the application to report non-fatal errors.
+    ///   - nfcReaderService: The closure to get a service used by the application to read NFC tags.
+    /// - Returns: Singleton `ServiceContainer`.
+    @MainActor
+    public static func shared(
+        appContext: AppContext = .mainApp,
+        application: Application? = nil,
+        errorReporter: () -> ErrorReporter,
+        nfcReaderService: () -> NFCReaderService? = { nil }
```
matt-livefront, line 882: ❓ Curious, why pass these as closures? Just to avoid duplicate initializations if the singleton is already created?

BitwardenShared/UI/Platform/Application/AppRoute.swift
```diff
@@ -47,6 +47,9 @@ public enum AppEvent: Equatable {
     ///
     case didLogout(userId: String, userInitiated: Bool)
 
+    /// Whent he user logs all their accounts.
```
matt-livefront: ⛏️ 
```suggestion
    /// When the user logs out all their accounts.
```

BitwardenShared/UI/Platform/Application/AppCoordinator.swift
```diff
@@ -92,6 +92,8 @@ class AppCoordinator: Coordinator, HasRootNavigator {
             )
         case let .didLogout(userId, userInitiated):
             await handleAuthEvent(.didLogout(userId: userId, userInitiated: userInitiated))
+        case .didLogOutAll:
+            await handleAuthEvent(.action(.logout(userId: nil, userInitiated: true)))
```
matt-livefront: 🎨 Minor feedback on this, I think the `handleAuthEvent(.action(.logout))` will attempt to log out of the active account (any of the `.action` events perform that action before redirecting). Since the app intent has already performed the logout, you could bypass that and use the `.didLogout` event directly.
```suggestion
        case .didLogOutAll:
            await handleAuthEvent(.didLogout(userId: nil, userInitiated: userInitiated))
```

However, this is identical to the `.didLogout` case above, so maybe we could use that instead of adding `.didLogOutAll` in this case?

BitwardenShared/Core/Platform/Services/ServiceContainer.swift
fedemkr: Yes, because these two services need to be created from the caller as BitwardenShared doesn't have access to them. So as you say in order not to duplicate them I pass the closures.
I don't like it that much so if you have a better way to handle this I'd love to hear it.
Also I thought other solutions but were much more complex so I went for this one which is simpler.

BitwardenShared/UI/Platform/Application/AppCoordinator.swift
fedemkr: @matt-livefront I'm using the other one because `.didLogout` can't pass a `nil` user ID and I wasn't that sure I wanted to enable that to be passed as optional. Although it could make sense to change it and use that directly since the check in `AuthRouter -> didLogoutRedirect` is to get the active account and I could add the existence of a `userId` parameter or else return to `.landing`. What do you think?

matt-livefront: Ah, I didn't realize that was the case. I think it'd be ok to make it optional as you suggested since we use that as a pattern elsewhere. Or add a `didLogoutAll` `AuthEvent` that just redirects to `.landing`.

BitwardenShared/Core/Platform/Services/ServiceContainer.swift
matt-livefront: Ok, makes sense. Yeah, this is kind of tricky, I'm not sure I have any better ideas either.

BitwardenShared/UI/Platform/Application/AppCoordinator.swift
fedemkr: Alright yeah makes sense, I'll change it, thanks for the feedback!.

matt-livefront (approved):

PR #1659, opened 2025-06-11
[PM-21681] Handle automatic timeout logout in BWA

KatherineInCode:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-21681

## 📔 Objective

Currently, automatic logout logic—including purging TOTP items shared from PM to BWA—is only being handled by Password Manager. This uses the additional keychain item introduced in [PR 1636](https://github.com/bitwarden/ios/pull/1636) to track the automatic logout timing between both apps, and if BWA determines that the user should be automatically logged out, deletes the shared TOTP items.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


AuthenticatorBridgeKit/AuthenticatorBridgeItemService.swift
```diff
@@ -210,4 +218,25 @@ public class DefaultAuthenticatorBridgeItemService: AuthenticatorBridgeItemServi
...
+
+        let results = try dataStore.persistentContainer.viewContext.fetch(fetchRequest)
+        let userIds = results.compactMap { ($0 as? [String: Any])?["userId"] as? String }
+
+        try await userIds.asyncForEach { userId in
+            if try await sharedTimeoutService.hasPassedTimeout(userId: userId) {
+                try await deleteAllForUserId(userId)
+            }
+        }
+    }
```
fedemkr, line 241: 🤔 Maybe this could be treated as Premature Optimization but for large data sets, e.g. in large organizations, I don't think this will have a good performance and it's something that directly affects time to load all the items (first time). Also it feels to me like an overkill to do this fetch for maximum 5 distinct user IDs.

♻️  Could we just have an array of user ID saved somewhere that we can fetch directly without traversing all `AuthenticatorBridgeItemData` entities? If it's a lot of effort a task can be created to improve this on a future PR.

🤔 Also, if one `userId` throws should it stop deleting for all the other users or should it keep going and try doing the process on other users?

BitwardenShared/Core/Vault/Services/VaultTimeoutService.swift
```diff
@@ -202,4 +261,32 @@ class DefaultVaultTimeoutService: VaultTimeoutService {
             .removeDuplicates()
             .eraseToAnyPublisher()
     }
+
+    /// Updates the shared timeout value in the SharedTimeoutService, so that BWA can log users out
+    /// on timeout. In the event that the user should not be automatically logged out after a time,
+    /// it will clear the timeout value.
+    private func updateSharedTimeout(lastActiveTime: Date?, timeoutValue: SessionTimeoutValue, userId: String?) async throws {
+        if let userId {
```
fedemkr: ⛏️ If possible use `guard` to reduce nesting.

AuthenticatorBridgeKit/SharedTimeoutServiceTests.swift
```diff
@@ -0,0 +1,86 @@
+import AuthenticatorBridgeKit
+import AuthenticatorBridgeKitMocks
+import BitwardenKit
+import BitwardenKitMocks
+import XCTest
+
+final class SharedTimeoutServiceTests: BitwardenTestCase {
```
fedemkr, line 8: ⛏️ Add tests for when all the functions throw on the different scenarios.

AuthenticatorBridgeKit/Tests/AuthenticatorBridgeItemServiceTests.swift
```diff
@@ -340,4 +342,40 @@ final class AuthenticatorBridgeItemServiceTests: AuthenticatorBridgeKitTestCase
...
+                    results.append(value)
+                }
+            )
+        defer { publisher.cancel() }
+
+        // Verify items are removed for "userId"
+        let deletedFetchResult = try await subject.fetchAllForUserId("userId")
+
+        XCTAssertNotNil(deletedFetchResult)
+        XCTAssertEqual(deletedFetchResult.count, 0)
```
fedemkr: ⛏️ I'd rename the variable as at first time it confused me as I thought it was going to have the number of items that were deleted by the operation and it's the current results for `userId`.

AuthenticatorBridgeKit/Tests/AuthenticatorBridgeItemServiceTests.swift
```diff
@@ -340,4 +342,40 @@ final class AuthenticatorBridgeItemServiceTests: AuthenticatorBridgeKitTestCase
         XCTAssertEqual(results[0], initialItems)
         XCTAssertEqual(results[1], replacedItems)
     }
+
+    /// The shared items publisher deletes items if the user is timed out.
+    ///
+    func test_sharedItemsPublisher_deletesItemsOnTimeout() async throws {
```
fedemkr, line 351: ⛏️ Add tests for all the scenarios where this can throw.

AuthenticatorBridgeKit/Tests/AuthenticatorBridgeItemDataTests.swift
```diff
@@ -15,6 +14,7 @@ final class AuthenticatorBridgeItemDataTests: AuthenticatorBridgeKitTestCase {
     var dataStore: AuthenticatorBridgeDataStore!
     var errorReporter: ErrorReporter!
     var itemService: AuthenticatorBridgeItemService!
+    var sharedTimeoutService: MockSharedTimeoutService!
```
fedemkr: ⛏️ This needs to be tear down. Moreover it seems there are other properties needing tear down.
However, I think for this specific service you could just add the mock inline as with `MockSharedKeychainRepository` as you're not using the property anywhere else other than the `setUp`.

BitwardenShared/Core/Auth/Repositories/AuthRepositoryTests.swift
fedemkr, line 1: ⛏️ I think tests are missing for when the modified functions throw.

BitwardenShared/Core/Vault/Services/TestHelpers/MockVaultTimeoutService.swift
```diff
@@ -45,30 +48,45 @@ class MockVaultTimeoutService: VaultTimeoutService {
         vaultTimeout[account.profile.userId] = value
     }
 
+    func isPinUnlockAvailable(userId: String?) async throws -> Bool {
+        if let userId {
+            try pinUnlockAvailabilityResult.get()[userId] ?? false
+        } else {
+            throw StateServiceError.noActiveAccount
+        }
+    }
```
fedemkr: ⛏️ Prefer using `guard`.

BitwardenShared/Core/Vault/Services/VaultTimeoutServiceTests.swift
```diff
@@ -7,12 +9,14 @@ import XCTest
 @testable import BitwardenShared
 
 @MainActor
-final class VaultTimeoutServiceTests: BitwardenTestCase {
+final class VaultTimeoutServiceTests: BitwardenTestCase { // swiftlint:disable:this type_body_length
```
fedemkr, line 12: ⛏️ Add tests for when all modified functions throw in their different scenarios.

AuthenticatorBridgeKit/AuthenticatorBridgeItemService.swift
KatherineInCode: I can look into having that list otherwise available, but I'm not sure how performant or not the query would be with large items, since it's effectively doing a `SELECT UNIQUE userId` 🤔 but avoiding the query altogether sounds like the way to go, yeah.

Currently if one userId has an issue it just halts entirely, but also means that it bails before it even pulls all the shared items for display, so that information shouldn't get to the person using the app. Doing otherwise would I think mean changing this to just log the error if it encounters it, and continue on?  Is that what we want it to do?

fedemkr: ♻️  The performance improvement no rush on doing so, and maybe we can just run a few performance tests in the future  before doing any changes. So I think I'm fine with that for the moment.

🤔  As for the behavior on how it should work for multiple users: if a user has 3 accounts, all 3 have timed out and the only the first one throws because of something I would continue to the other ones to clear items from the database as early as possible. 
We can just save a temp variable saying it failed to logout some account and throw an error after the loop so no shared items are shown.

♻️  Maybe for a future improvement but IMO the best solution for this would be to do the operation for all accounts and save the user IDs that threw error in some part of the process where then we can use these user IDs to filter out the shared data displayed as they may need to be logged out, but displaying all the other non-logged out accounts  data. However, I would need to re-check the code as I don't remember at the top of my head how multiple accounts items are shown in the UI and in which conditions they are shown / if it only shows like an active user or something like that.

AuthenticatorBridgeKit/SharedTimeoutService.swift
```diff
@@ -0,0 +1,81 @@
...
+        forUserId userId: String,
+        lastActiveDate: Date?,
+        timeoutLength: SessionTimeoutValue
+    ) async throws {
+        guard let lastActiveDate else {
+            try await clearTimeout(forUserId: userId)
+            return
+        }
+
+        let timeout = lastActiveDate.addingTimeInterval(TimeInterval(timeoutLength.seconds))
```
matt-livefront, line 77: 🌱 This matches our existing logic, so future improvement idea: seeing this again, should we be using `Calendar` to perform this calculation to take into account any time changes (daylight savings time, etc)?

AuthenticatorBridgeKit/AuthenticatorBridgeItemService.swift
KatherineInCode: Hm. At the moment, if one of them fails, _none_ of them will show. Which, as any error that would crop up is probably because of a CoreData issue, is probably pretty reasonable? If the database is corrupted or something, the later fetch won't work for the others as well.

matt-livefront: This feels like it would be pretty performant, but I like the idea of having a separate storage for the list of users. We duplicate the user's domain and email data AuthenticatorBridgeItemDataModel for each item we sync, so moving this to a separate table/storage would be nice.

AuthenticatorBridgeKit/AuthenticatorBridgeItemData.swift
```diff
@@ -26,7 +26,7 @@ public class AuthenticatorBridgeItemData: NSManagedObject, CodableModelData {
     ///   - context: The managed object context to insert the initialized item
     ///   - userId: The ID of the user who owns the item
     ///   - authenticatorItem: the `AuthenticatorBridgeItemDataModel` used to create the item
-    convenience init(
+    public convenience init(
```
matt-livefront: ❓ Does this need to be public? It seems like this would just be used in the module.

AuthenticatorBridgeKit/SharedTimeoutService.swift
KatherineInCode: Isn't `Date` under the hood just a count of seconds since midnight 1/1/70 UTC? Adding 15 minutes' worth of seconds onto that should be immune to DST changes, I'd think.

AuthenticatorBridgeKit/AuthenticatorBridgeItemData.swift
KatherineInCode: I can double check. There were a few places where I had to make things public because of unexpected places where things crossed module boundaries.

AuthenticatorBridgeKit/AuthenticatorBridgeItemService.swift
KatherineInCode: A separate storage for that would be nice, I agree, though I think outside of the scope of this PR. A good enhancement for the future, though, and this can be streamlined at that time.

AuthenticatorBridgeKit/SharedTimeoutService.swift
matt-livefront: Ah, ok, I thought using `addingTimeInterval` didn't take into account DST. But maybe we're safe because we're using a fixed number of seconds as opposed to a whole day. Seems to work though! Time calculations are confusing 🙃.

KatherineInCode: Yup. If we wanted to do something like a timeout of "one day", then we'd want to move to a `Calendar` because that's when you start getting into the weird things of needing to adjust for an hour across some day boundaries. Though I'd argue if we have a timeout of "twenty-four hours", we wouldn't, if that distinction makes sense?

AuthenticatorBridgeKit/AuthenticatorBridgeItemService.swift
```diff
@@ -210,4 +218,25 @@ public class DefaultAuthenticatorBridgeItemService: AuthenticatorBridgeItemServi
...
+        fetchRequest.propertiesToFetch = ["userId"]
+        fetchRequest.returnsDistinctResults = true
+        fetchRequest.resultType = .dictionaryResultType
+
+        let results = try dataStore.persistentContainer.viewContext.fetch(fetchRequest)
+        let userIds = results.compactMap { ($0 as? [String: Any])?["userId"] as? String }
+
+        try await userIds.asyncForEach { userId in
+            if try await sharedTimeoutService.hasPassedTimeout(userId: userId) {
+                try await deleteAllForUserId(userId)
```
fedemkr, line 238: ⛏️ I believe there's a test missing for when this function throws.

AuthenticatorBridgeKit/Mocks/MockAuthenticatorBridgeItemService.swift
```diff
@@ -0,0 +1,59 @@
...
+    public var sharedItemsSubject = CurrentValueSubject<[AuthenticatorBridgeItemDataView], Error>([])
+    public var storedItems: [String: [AuthenticatorBridgeItemDataView]] = [:]
+    public var syncOn = false
+    public var tempItem: AuthenticatorBridgeItemDataView?
+
+    public init() {}
+
+    public func checkForLogout() async throws {
+
+    }
```
fedemkr, line 16: ⛏️ Can this be removed? I see that's private in the service.

AuthenticatorBridgeKit/SharedTimeoutService.swift
```diff
@@ -0,0 +1,81 @@
...
+        return timeProvider.presentTime >= autoLogoutTime
+    }
+
+    public func updateTimeout(
+        forUserId userId: String,
+        lastActiveDate: Date?,
+        timeoutLength: SessionTimeoutValue
+    ) async throws {
+        guard let lastActiveDate else {
+            try await clearTimeout(forUserId: userId)
```
fedemkr, line 73: ⛏️ I think there's a test missing for when this line throws.

BitwardenShared/Core/Vault/Services/VaultTimeoutService.swift
```diff
@@ -171,12 +200,43 @@ class DefaultVaultTimeoutService: VaultTimeoutService {
         }
     }
 
+    func sessionTimeoutAction(userId: String?) async throws -> SessionTimeoutAction {
+        let hasMasterPassword = try await stateService.getUserHasMasterPassword(userId: userId)
+        let timeoutAction = try await stateService.getTimeoutAction(userId: userId)
+        guard hasMasterPassword else {
+            let isBiometricsEnabled = try await biometricsRepository.getBiometricUnlockStatus().isEnabled
+            let isPinEnabled = try await isPinUnlockAvailable(userId: userId)
```
fedemkr, line 208: ⛏️ There's only one test for throwing behavior based on `stateService.getUserHasMasterPassword`, but there are 3 tests missing. 
When:
- `try await stateService.getTimeoutAction(...)` throw
- `try await biometricsRepository.getBiometricUnlockStatus().isEnabled` throw
- `try await isPinUnlockAvailable(userId: userId)` throw

BitwardenShared/Core/Vault/Services/VaultTimeoutService.swift
```diff
@@ -171,12 +200,43 @@ class DefaultVaultTimeoutService: VaultTimeoutService {
...
     func setLastActiveTime(userId: String) async throws {
-        try await stateService.setLastActiveTime(timeProvider.presentTime, userId: userId)
+        let now = timeProvider.presentTime
+        try await stateService.setLastActiveTime(now, userId: userId)
+        let vaultTimeout = try await sessionTimeoutValue(userId: userId)
+        try await updateSharedTimeout(
+            lastActiveTime: now,
+            timeoutValue: vaultTimeout,
+            userId: userId
+        )
```
fedemkr, line 228: ⛏️ Tests are missing for when `try await sessionTimeoutValue(...)` throws and also for all the scenarios where `try await updateSharedTimeout(...)` can throw.

BitwardenShared/Core/Vault/Services/VaultTimeoutService.swift
```diff
@@ -171,12 +200,43 @@ class DefaultVaultTimeoutService: VaultTimeoutService {
...
 
     func setVaultTimeout(value: SessionTimeoutValue, userId: String?) async throws {
         try await stateService.setVaultTimeout(value: value, userId: userId)
+        guard let userId else { return }
+        let lastActiveTime = try await stateService.getLastActiveTime(userId: userId)
+        try await updateSharedTimeout(
+            lastActiveTime: lastActiveTime,
+            timeoutValue: value,
+            userId: userId
+        )
```
fedemkr, line 239: ⛏️ Tests are missing for all the scenarios where `try await updateSharedTimeout(...)` can throw.

fedemkr (approved):
Looks good, although there are several tests missing yet.
Approving as the tests can be added in a future PR and we let QA start testing this.

KatherineInCode:
> Looks good, although there are several tests missing yet. 

I don't in general think it's valuable to write a test for each and every possible called function that can throw, especially when the behavior is always "just throw the error up the stack instead of handling it". It feels to me like it constrains the implementation without specifying additional behavior, in addition to producing a lot of noise and clutter in the test file—both of which make it more difficult to refactor or otherwise edit later. Hence why I just went with the behavior-testing "does this not handle errors itself" tests.

PR #1660, opened 2025-06-11
[PM-10285] Update master password reprompt to protect item view

matt-livefront:
## 🎟️ Tracking

[PM-10285](https://bitwarden.atlassian.net/browse/PM-10285)

## 📔 Objective

Previously, the master password reprompt was triggered within the View Item view when performing certain actions. This protects the full View Item view by moving the prompt to display before navigating to the View Item view.

This handles the master password reprompt for the following flows:

- Vault list: view item
- Vault group list: view item
- Rehydrating the view item view after a timeout
- Autofill

## 📸 Screenshots

https://github.com/user-attachments/assets/0d455ec5-28a1-4c70-8cef-514e6fc3536e

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/BitwardenShared/UI/Vault/Extensions/Alert+Vault.swift
+++ b/BitwardenShared/UI/Vault/Extensions/Alert+Vault.swift
@@ -182,33 +182,30 @@ extension Alert {
     ///   - canCopyTotp: Whether the user can copy the TOTP code (because they have premium or the
     ///     organization uses TOTP).
     ///   - cipherView: The cipher view to show.
-    ///   - hasMasterPassword: Whether the user has a master password.
     ///   - id: The id of the item.
     ///   - showEdit: Whether to show the edit option (should be `false` for items in the trash).
     ///   - action: The action to perform after selecting an option.
     ///
     /// - Returns: An alert presenting the user with options to select an attachment type.
     @MainActor
-    static func moreOptions( // swiftlint:disable:this function_body_length function_parameter_count
+    static func moreOptions( // swiftlint:disable:this function_body_length
         canCopyTotp: Bool,
         cipherView: CipherView,
-        hasMasterPassword: Bool,
         id: String,
         showEdit: Bool,
         action: @escaping (_ action: MoreOptionsAction) async -> Void
     ) -> Alert {
         // All the cipher types have the option to view the cipher.
         var alertActions = [
-            AlertAction(title: Localizations.view, style: .default) { _, _ in await action(.view(id: id)) },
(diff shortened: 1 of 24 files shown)


BitwardenShared/UI/Vault/Extensions/Alert+Vault.swift
```diff
@@ -255,18 +252,15 @@ extension Alert {
                     await action(.copy(
                         toast: Localizations.password,
                         value: password,
-                        requiresMasterPasswordReprompt: cipherView.reprompt == .password && hasMasterPassword,
+                        requiresMasterPasswordReprompt: true,
```
matt-livefront, line 255: ℹ️ Slight change here. Previously, the `requiresMasterPasswordReprompt` value was being passed along for each action that may need master password reprompt. With the `MasterPasswordRepromptHelper` this check doesn't need to be done up-front and is handled internally in the helper. So instead of passing along this value, I moved the MP reprompt check into `VaultItemMoreOptionsHelper` for actions that may need reprompting, which helped to simplify some logic. For the copy action, `requiresMasterPasswordReprompt` is now indicating whether the MP reprompt check is needed for copying that value, since some things like copy username don't require the reprompt.

fedemkr (commented):
🤔 When I did some of the work for Autofill I added `UserVerificationHelper` which has several methods centralizing user verification, for this specific case: [verifyMasterPassword()](https://github.com/bitwarden/ios/blob/main/BitwardenShared/UI/Auth/Utilities/UserVerificationHelper.swift#L121). And my idea was to slowly start adapting other places to use it.

Therefore I'd use `UserVerificationHelper` in `MasterPasswordRepromptHelper` to reduce duplicated code. The caveat though is that it uses a delegate instead of the coordinator directly which I think that it's more flexible for the caller.

What do you think?

matt-livefront:
> 👍 Looks good overall, but I have some thoughts for `MasterPasswordRepromptHelper`.
> 
> 🤔 When I did some of the work for Autofill I added `UserVerificationHelper` which has several methods centralizing user verification, for this specific case: [verifyMasterPassword()](https://github.com/bitwarden/ios/blob/main/BitwardenShared/UI/Auth/Utilities/UserVerificationHelper.swift#L121). And my idea was to slowly start adapting other places to use it.
> 
> Therefore I'd use `UserVerificationHelper` in `MasterPasswordRepromptHelper` to reduce duplicated code. The caveat though is that it uses a delegate instead of the coordinator directly which I think that it's more flexible for the caller.
> 
> What do you think?

Ah, I can give that a try!

Issue #1216, opened 2024-12-17
Ui gets laggy


speedy1804:
### Steps To Reproduce

1. Open a password
2. Press and **hold** your finger on the URL
3. Slide up and and down with that finger **while pressing** the URL
4. —> Ui gets laggy as seen in the Attachments

### Expected Result

It is to be expected that the Ui continues to move smoothly and that the animation does not contain any jerks.

### Actual Result

The Ui has much performance issues.

### Screenshots or Videos

https://github.com/user-attachments/assets/9e2fed57-9826-4238-9572-e09b6a4b8063
(This is device1)

https://github.com/user-attachments/assets/b7a881eb-fed9-44d3-8203-9d9dd4e75c19
(This is device2)

### Additional Context

I already reported this issue here: https://github.com/bitwarden/ios/issues/1187 I was told it was not reproducible so I tested it on my other device (iPad Pro 11" M2) and it has the same issue.

### Build Version

Both (iPadOS and iOS) are on their latest build: 2024.12.0(1740)

### What server are you connecting to?

US

### Environment Details

- Device1: iPhone 14 running iOS 18.2
- Device2: iPad Pro 11" with M2 (so Gen 4 I believe) running iPadOS 18.2

### Issue Tracking Info

- [X] I understand that work is tracked outside of Github. A PR will be linked to this issue should one be opened to address it, but Bitwarden doesn't use fields like "assigned", "milestone", or "project" to track progress.

cbbit:
Hi there,

I am unable to reproduce this issue, it has been escalated for further investigation. If you have more information that can help us, please add it below. 

Thanks!

speedy1804:
can you tell/show me how you are trying to reproduce it? As I'm using iOS I unfortunately can't display screen touches within the screen recording - I think that would make it easier to understand what I mean

Gerardv514:
I was able to reproduce this too on ios 2024.12. I tapped and held onto uri and saw lag. Also tried username and same. I think I then swiped down regularly or backed out and went to try to reproduce again, but now not happening. I then forced closed the app and tried again, still not happening.

maxmini1:
Hi there, I was able to reproduce this too on ios (18.2.1) - 2024.12.0 @cbbit see here:

https://github.com/user-attachments/assets/43610cba-22cc-40b8-adf3-673fc1a02c6d


https://github.com/user-attachments/assets/5fd74e86-e24b-4433-8f1a-aefc42a1acfa

I hope this helps.

[...]

PR #1677, opened 2025-06-19
[PM-18285]  Vault repository refactor to improve performance

fedemkr:
## 🎟️ Tracking

[PM-18285](https://bitwarden.atlassian.net/browse/PM-18285)

## 📔 Objective

Refactor vault list publishers to increase the CPU and memory performance when building the vault list sections.
This is mainly done by changing how we process the ciphers and filter them. Mainly decrypting and handling the filtering in batches instead of all at the same time to reduce the effort on large vaults.

The interaction now would be:

- `VaultRepository` asks the `VaultListDirectorStrategyFactory` for a specific strategy based on the filters.
- `VaultListDirectorStrategy` requests the `VaultListDataPreparator ` to prepare the vault data to use based on ciphers, folders, collections and filters.
- `VaultListDataPreparator` prepares vault list data by iteratively decrypting batches of ciphers and filtering them, adding only the minimum useful information to the `VaultListPreparedData` builder which is then returned to the strategy.
- `VaultListDirectorStrategy` uses a `VaultListSectionsBuilder` from a factory to build the list sections and then they're returned to the repository.

With this approach I'm focusing mainly on improving performance and also on moving a lot of logic away from the `VaultRepository` and place it in smaller objects to ease centralization and also to reduce the large number of tests we have for this. This also makes maintaining easier as it's really easy to add/modify/delete sections and or the metadata.

> [!NOTE]  
> Added `SortDescriptorWrapper<T>` in order to wrap the [SortDescriptor](https://developer.apple.com/documentation/foundation/sortdescriptor) as there are some initializers that require iOS 17+, specially for strings so we can provide the backport version of it to be used in older OS.

**High level flow to build sections:**

Build sections high level diagram

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-18285]: https://bitwarden.atlassian.net/browse/PM-18285?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


speedy1804:
may this pr fix #1216 ?

fedemkr:
Hi @speedy1804 , no, this will not fix that issue. This PR focuses on CPU / Memory performance on vault list items loading.

BitwardenShared/Core/Vault/Helpers/VaultListDirectorStrategyFactory.swift
```diff
@@ -0,0 +1,42 @@
+import BitwardenKit
+
+protocol VaultListDirectorStrategyFactory {
+    func make(filter: VaultListFilter) -> VaultListDirectorStrategy
+}
+
+struct DefaultVaultListDirectorStrategyFactory: VaultListDirectorStrategyFactory {
+    let vaultListVBuilderFactory: VaultListBuilderFactory
```
KatherineInCode: ⛏️ 
```suggestion
    let vaultListBuilderFactory: VaultListBuilderFactory
```

BitwardenShared/Core/Vault/Helpers/VaultListDirectorStrategyFactory.swift
```diff
@@ -0,0 +1,42 @@
...
+    let cipherService: CipherService
+    let clientService: ClientService
+    let collectionService: CollectionService
+    let errorReporter: ErrorReporter
+    let folderService: FolderService
+    let stateService: StateService
+    let vaultListDataArranger: VaultListDataArranger
+
+    func make(filter: VaultListFilter) -> VaultListDirectorStrategy {
```
KatherineInCode: ⛏️ The name of this function should probably be `makeStrategy`

BitwardenShared/Core/Vault/Helpers/VaultListDirectors/VaultListDirectorStrategy.swift
```diff
@@ -0,0 +1,37 @@
...
+    /// - Returns: Sections to be displayed to the user.
+    func build(
+        filter: VaultListFilter
+    ) async throws -> AsyncThrowingPublisher<AnyPublisher<[VaultListSection], Error>>
+}
+
+// MARK: - VualtListBuilderMetadata
+
+/// Metadata helper object to hold temporary data the builder can then use to build the list sections.
+struct VualtListBuilderMetadata {
```
KatherineInCode: ⛏️ 
```suggestion
struct VaultListBuilderMetadata {
```

BitwardenShared/Core/Vault/Helpers/VaultListDirectors/VaultListDirectorStrategy.swift
```diff
@@ -0,0 +1,37 @@
+import BitwardenKit
+import BitwardenSdk
+import Combine
+
+// MARK: - VaultListDirectorStrategy
+
+protocol VaultListDirectorStrategy {
+    /// Builds the vault list sections.
+    /// - Parameters:
+    ///   - filter: Fitler to be used to build the sections.
```
KatherineInCode: ⛏️ 
```suggestion
    ///   - filter: Filter to be used to build the sections.
```

BitwardenShared/Core/Vault/Repositories/FastVaultRepository.swift
```diff
@@ -0,0 +1,62 @@
...
+    override func vaultListPublisher(
+        filter: VaultListFilter
+    ) async throws -> AsyncThrowingPublisher<AnyPublisher<[VaultListSection], any Error>> {
+        try await vaultListDirectorStrategyFactory
+            .make(filter: filter)
+            .build(filter: filter)
+    }
+
+    override func vaultListPublisher(
+        group: VaultListGroup,
```
KatherineInCode: 🤔 It looks like we're moving the group into the filter, so do we still need these separate functions like this?

BitwardenShared/Core/Vault/Helpers/VaultListBuilder.swift
```diff
@@ -0,0 +1,283 @@
...
+import BitwardenSdk
+import Combine
+import Foundation
+import OSLog
+
+// MARK: - VaultListBuilderFactory
+
+/// A factory protocol to make vault list builders.
+protocol VaultListBuilderFactory {
+    func make() -> VaultListBuilder
```
KatherineInCode: ⛏️ This should probably be `makeVaultListBuilder()` or `makeBuilder()`

BitwardenShared/Core/Vault/Helpers/VaultListBuilder.swift
```diff
@@ -0,0 +1,283 @@
...
+    func make() -> VaultListBuilder {
+        DefaultVaultListBuilder(clientService: clientService, errorReporter: errorReporter)
+    }
+}
+
+// MARK: - VaultListBuilder
+
+/// A protocol for a vault list builder which helps build items and sections for the vault lists.
+protocol VaultListBuilder {
+    func addTrashSection(
```
KatherineInCode: ⛏️ I think it's probably more Swifty for these to be `withTrashSection()` etc.

🤔 Does the tempdata need to be `inout` for all of these? Do we ever modify it in implementation?

🤔 Do these need to return a `some VaultListBuilder`, or am I getting my existential types with protocols mixed up on where they're needed?

BitwardenShared/Core/Vault/Helpers/VaultListBuilder.swift
```diff
@@ -0,0 +1,283 @@
...
+                id: "Types.SSHKeys",
+                itemType: .group(.sshKey, tempData.countPerCipherType[.sshKey, default: 0])
+            ),
+        ]
+
+        sections.append(VaultListSection(id: "Types", items: types, name: Localizations.types))
+        return self
+    }
+
+    func build() -> [VaultListSection] {
```
KatherineInCode: ⛏️ I think this would be more Swifty as `builtSections`, and arguably can just be the property `sections`

BitwardenShared/Core/Vault/Helpers/VaultListDataArranger.swift
```diff
@@ -0,0 +1,130 @@
+
+import BitwardenKit
+import BitwardenSdk
+
+protocol VaultListDataArranger {
```
KatherineInCode: ⛏️ Calling this a data arranger but having its functions be `arrangeMetadata` is a bit confusing to me. This also feels more like a metadata factory/initializer rather an arranger per se.

KatherineInCode (commented):
On the whole, I think is a plausible optimization for the `VaultRepository`, though I found it difficult to follow at times. I think some of that can be ameliorated with tests, more doc comments on things, and by naming things more in line with Swift conventions. I do find myself wondering if it's almost over-complicated, and if there might not be a way to do this in a more functional (as opposed to procedural) way. There's also possibly some nuance to it I'm not fully grasping as of yet.

BitwardenShared/Core/Vault/Repositories/FastVaultRepository.swift
fedemkr: Yes, that's something I'm planning to do as part of this PR.

BitwardenShared/Core/Vault/Helpers/VaultListBuilder.swift
fedemkr: I thought about the `with...` but I meant the name to have the sense that's added at the end of the sections. I often take `with...` like adding a part of something but not necessarily at the end. Also thought about `append...` but as far as I saw that's the wording when using in the context of low level types functions like in arrays.

I use `inout` because it's a struct and wanted to be passed by reference and not use COW behavior which would use more memory given the inner types are also structs.

No actual need, I just like fluent code and it's very often seen in builders so you can just use:
```swift
let foo = builder.addFirstThing()
    .addSecondThing()
    .addThirdThing()
    .build()
```

BitwardenShared/Core/Vault/Helpers/VaultListBuilder.swift
fedemkr: What I've seen so far on Build patterns is to just name the final build method `build()` given that the class/struct should give enough context on what it's building. Having said that I'll change the name of the object from `VaultListBuilder` to `VaultListSectionsBuilder` so it's more clear what this is about.

BitwardenShared/Core/Vault/Helpers/VaultListBuilder.swift
fedemkr: Even though I see the point, I see that a bit repetitive, I mean if it's a `VaultListBuilderFactory` then it makes `VaultListBuilder` objects and it's the only function in there so it's not like we have different things to make.

BitwardenShared/Core/Vault/Helpers/VaultListDirectorStrategyFactory.swift
fedemkr: Same as in the other one, given that this is a `VaultListDirectorStrategyFactory` then it's an object that makes/creates `VaultListDirectorStrategy` objects and there's only one method here so I find a bit repetitive to add `Strategy` suffix in there. If we were to add more method in here then I'd agree but in the current state not sure which is the value on adding the suffix. 
Furthermore, we usually name variables as the full object name so this would be:
`let foo = vaultListDirectorStrategyFactory.make(filter: myAwesomeFilter)`
which IMO is pretty self explanatory that `foo` will be a `VaultListDirectorStrategy` based on the factory pattern.
However, if you still feel like Swift is strict on this kind of naming I can make it more verbose no problem.

BitwardenShared/Core/Vault/Helpers/VaultListDataArranger.swift
fedemkr: Yeah, I struggled a lot to name this thing and not to use any other suffixes we're already using in the app that could cause confusion 😞 .
I'll see if I can improve the naming with the objects implicated on this. I went for arranger because it "arranges" (filter + saves in cache) cipher data into a more high level object with what's only needed for the sections.
Not sure if I like "Factory" much as IMO it doesn't follow much the factory pattern. I feel like this object makes more processing than an "Initializer" but I also think "Arranger" is a bit odd.
Do you have any other suffixes in mind?

KatherineInCode: Best suffixes I could think of were `Factory` or `Initializer`, but yeah. Naming things is hard 😓

BitwardenShared/Core/Vault/Helpers/VaultListBuilder.swift
KatherineInCode: The main thing is that [per Swift API guidelines](https://www.swift.org/documentation/api-design-guidelines/), a function called `build()` should have a side effect, because it's an imperative verb. My general rule of thumb is that any function that returns a thing should be named to know what the thing that's returned is; so it's weird to me that this returns something, rather than modifying in place.

BitwardenShared/Core/Vault/Helpers/VaultListBuilder.swift
KatherineInCode: I generally take `with...` functions on a struct to be ones that return the struct as-is with the one thing changed, thought digging around I'm not quite sure where I picked up that habit 🤔 

Would COW semantics matter if it's only read, though? The memory usage should remain the same so long as it's not mutated, I'd think?

I don't know that I've seen the builder pattern all that often in Swift code, to be honest; I think codebases I've been in have always solved the problem in a somewhat different way—though if `build()` is the pattern convention for that function, then I guess it's fine, it just doesn't feel like a very Swift-y name to me.

BitwardenShared/Core/Vault/Helpers/VaultListDirectorStrategyFactory.swift
KatherineInCode: I see the point on verbosity, and I think some of where the API design guidelines are coming from is assuming the factory's variable name will just be `factory` or something to that effect. It's fine.

BitwardenShared/Core/Vault/Helpers/VaultListArrangedDataBuilder.swift
```diff
@@ -0,0 +1,295 @@
...
+            return self
+        case let .collection(id, _, _):
+            guard cipher.collectionIds.contains(id) else { return self }
+        case let .folder(id, _):
+            guard cipher.folderId == id else { return self }
+        case .noFolder:
+            guard cipher.folderId == nil else { return self }
+        case .trash:
+            // this case is handled at the beginning of the function.
+            return self
```
matt-livefront: 🤔 The guard for handling the trash group looks very similar to the other cases here (it checks a condition `group == .trash` and then adds the `VaultListItem`). Any reason not to do that in this switch with the rest of the groups?

BitwardenShared/Core/Vault/Helpers/VaultListArrangedDataBuilder.swift
```diff
@@ -0,0 +1,295 @@
...
+    /// Adds a folder item to the arranged data.
+    func addFolderItem(
+        cipher: CipherListView,
+        filter: VaultListFilter,
+        folders: [Folder]
+    ) -> VaultListArrangedDataBuilder
+    /// Adds an item for a specific group to the arranged data.
+    func addItem(forGroup group: VaultListGroup, with cipher: CipherListView) async -> VaultListArrangedDataBuilder
+    /// Adds a no folder item to the arranged data.
+    func addNoFolderItem(cipher: CipherListView) -> VaultListArrangedDataBuilder
```
matt-livefront: 🤔 I'm not quite following the usage behind these methods. I think my confusion stems from the fact that there are methods for adding actual vault items (folders, collections, ciphers) and then there are separate methods for adding ciphers to the metadata for certain optional groups. 

As a consumer of this API, I think I would wonder if I should call `addFavoriteItem()` even if the item isn't a favorite (I would think no because it's not a favorite, but I think I should based on the implementation). And similarly for the other methods.

I'm guessing the reasoning behind the separate methods is an optimization? For example, we don't need to build the favorites list if you're in a group view?

I think it'd be clearer IMO to just have the three methods for adding the actual data (folders, collections, and ciphers). And then for the `addCipher()` method, maybe we could pass an `OptionSet` parameter indicating if we want to build out the data items for the favorites, folders, and group items? Alternatively, instead of passing this data to each `addCipher()` call, maybe we could initialize the builder with the set of options?

BitwardenShared/Core/Vault/Helpers/VaultListArrangedDataBuilder.swift
```diff
@@ -0,0 +1,295 @@
...
+    /// Builds the arranged data.
+    func build() -> VualtListBuilderMetadata
+    /// Increments the cipher type count in the arranged data.
+    func incrementCipherTypeCount(cipher: CipherListView) -> VaultListArrangedDataBuilder
+    /// Increments the cipher deleted count in the arranged data.
+    func incrementCipherDeletedCount() -> VaultListArrangedDataBuilder
+    /// Increments the collection count in the arranged data.
+    func incrementCollectionCount(cipher: CipherListView) -> VaultListArrangedDataBuilder
+    /// Increments the TOTP count in the arranged data.
+    func incrementTOTPCount(cipher: CipherListView) async -> VaultListArrangedDataBuilder
```
matt-livefront: 🤔 Is it necessary to manually call these / compute the counts as the data is being added? Could the counts be computed at the end when the result is built? Manually incrementing the counts seems like it would be more error prone than having some logic that calculates the counts once all of the data is known, but maybe I'm missing something.

BitwardenShared/Core/Vault/Helpers/VaultListArrangedDataBuilder.swift
```diff
@@ -0,0 +1,295 @@
...
+        if cipher.favorite,
+           let favoriteListItem = VaultListItem(cipherListView: cipher) {
+            arrangedData.favorites.append(favoriteListItem)
+        }
+        return self
+    }
+
+    func addFolders(folders: [Folder], filterType: VaultFilterType) -> VaultListArrangedDataBuilder {
+        if filterType == .allVaults {
+            arrangedData.folders = folders
```
matt-livefront: 🤔 Isn't it possible to also have folders be shown if the filter type isn't all vaults?

BitwardenShared/Core/Vault/Helpers/VaultListArrangedDataBuilder.swift
```diff
@@ -0,0 +1,295 @@
...
+        if await shouldIncludeTOTP(cipher: cipher) {
+            arrangedData.totpItemsCount += 1
+        }
+
+        return self
+    }
+
+    // MARK: Private methods
+
+    private func getHasPremiumFeaturesAccesss() async -> Bool {
```
matt-livefront: ⛏️ 
```suggestion
    private func getHasPremiumFeaturesAccess() async -> Bool {
```

BitwardenShared/Core/Vault/Helpers/VaultListBuilder.swift
fedemkr: Oh you're absolutely right about COW, I tend to forget the read-only behavior. Great catch, thanks! I'll make the change to remove the `inout`.

BitwardenShared/Core/Vault/Helpers/VaultListBuilder.swift
fedemkr: Oh makes sense, I'll make the change.

BitwardenShared/Core/Platform/Services/StateService.swift
```diff
@@ -1100,6 +1100,14 @@ extension StateService {
         try await pinProtectedUserKey(userId: nil)
     }
 
+    func safeDoesActiveAccountHavePremium() async -> Bool {
+        do {
+            return try await doesActiveAccountHavePremium()
+        } catch {
+            return false
```
matt-livefront: 🤔 Do we want to follow this pattern in the default case of `doesActiveAccountHavePremium()` (as opposed to a separate method)? It'd be nice to have one less thing that throws and has a default value. We could also add logging in the catch here if there's an error.

BitwardenShared/Core/Vault/Helpers/VaultListDirectorStrategyFactory.swift
```diff
@@ -0,0 +1,42 @@
...
+                clientService: clientService,
+                collectionService: collectionService,
+                errorReporter: errorReporter,
+                folderService: folderService,
+                stateService: stateService,
+                vaultListDataArranger: vaultListDataArranger
+            )
+        }
+
+        return MainVaultListDirectorStrategy(
```
matt-livefront: 🤔 There's a lot of overlap between `MainVaultListDirectorStrategy` and `MainVaultListGroupDirectorStrategy`. Maybe these will diverge at some point (?), but I wonder if it's beneficial at this point to break them into two separate objects.

BitwardenShared/Core/Vault/Services/CiphersClientWrapperService.swift
```diff
@@ -0,0 +1,77 @@
...
+                let decryptedCiphers = try await clientService.vault().ciphers().decryptList(
+                    ciphers: Array(ciphers[start ..< end])
+                )
+
+                guard !decryptedCiphers.isEmpty else {
+                    continue
+                }
+
+                for decryptedCipher in decryptedCiphers {
+                    try await onCipher(decryptedCipher)
```
matt-livefront: 🤔 Just curious, is there a benefit to the `!isEmpty` check here? `for...in` with an empty array will ultimately have the same result, right?

BitwardenShared/Core/Vault/Helpers/VaultListDirectors/MainVaultListDirectorStrategy.swift
```diff
@@ -0,0 +1,94 @@
...
+
+        if filter.addTOTPGroup {
+            builder = builder.addTOTPSection(from: &vaultListMetadata)
+        }
+
+        builder = try await builder
+            .addFavoritesSection(from: &vaultListMetadata)
+            .addTypesSection(from: &vaultListMetadata)
+            .addFoldersSection(from: &vaultListMetadata)
+            .addCollectionsSection(from: &vaultListMetadata)
```
matt-livefront: 🤔 Would it be easier to initialize the builder with `vaultListMetdata`? It seems like you wouldn't pass different metadata to any of these functions, so I wonder if we could give the builder it's data up-front and then have methods without parameters to add the sections.

BitwardenShared/Core/Vault/Helpers/VaultListDataArranger.swift
```diff
@@ -0,0 +1,130 @@
...
+
+            arrangedDataBuilder = arrangedDataBuilder
+                .addFolderItem(cipher: decryptedCipher, filter: filter, folders: folders)
+                .addFavoriteItem(cipher: decryptedCipher)
+                .addNoFolderItem(cipher: decryptedCipher)
+                .incrementCipherTypeCount(cipher: decryptedCipher)
+                .incrementCollectionCount(cipher: decryptedCipher)
+        }
+
+        return arrangedDataBuilder.build()
```
matt-livefront: ⚠️ Maybe this is still coming, but it looks like ciphers aren't currently sorted alphabetically.

BitwardenShared/Core/Vault/Helpers/VaultListArrangedDataBuilder.swift
fedemkr: Yes, I wanted that split so I don't need to add `cipher.deletedDate != nil` on each of the other cases. Trash items are the only ones that would go with the condition `cipher.deletedDate == nil`.

matt-livefront: Ah, yes, 🤦 moment.

BitwardenShared/Core/Vault/Helpers/VaultListArrangedDataBuilder.swift
fedemkr: I see the point that it's confusing and I should improve the methods names and also docs to clarify it.
Yes, my intention is to have optimizations on what should be added or not to the lists controlled by which methods are called.

I like the Builder pattern on this regard where you have an object and start building it up as you need step by step to construct it splitting the parts to a minimum so to have great flexibility and IMO better maintenance. Adding tests like this instead of with the `OptionSet` is really easy as you have really low cyclomatic complexity and is up to the caller to decide which methods to call to add the items indirectly "removing" the branches one should have with `OptionSet` or just with boolean parameters. Furthermore, to test the caller you only have to check whether a specific method was called and that's it.

Additionally, wanted to be able to arrange the data from different places like autofill extension items which have several different sections that require different data, so this gives the flexibility to build only what we need to in the next step end up building the vault list sections out of it.

Does this make more sense? Maybe given that I come from more OOP oriented languages this is not as Swift oriented as it could be.

KatherineInCode: It's definitely a very OOP architecture, and the Builder pattern hasn't traditionally been used much in Swift or Objective-C—the best I can think of is something like `DateComponents` or `URLComponents` where there's an object that defines all the pieces, which is then used to construct the object itself.

Which is not to say we shouldn't use it, to be clear. Just, we should make sure it's well-documented because it's outside of the usual platform paradigms.

BitwardenShared/Core/Vault/Helpers/VaultListArrangedDataBuilder.swift
fedemkr: Basically I wanted to avoid having the actual ciphers when not needed.
E.g. for the main vault screen we only need the count of TOTP ciphers not the actual ciphers objects so I thought saving such decrypted objects in memory for the count was a waste of resources.

BitwardenShared/Core/Vault/Helpers/VaultListArrangedDataBuilder.swift
fedemkr: Yes, maybe I'd need to improve how it's done as right now it's being covered by this part
https://github.com/bitwarden/ios/pull/1677/files/403691d4d9752b5f20f154bc3da52c9c2d95b9de#diff-9e3859c78be860e7265eb2165fae75cbbce0521ab7b93a66aeaa375f04e83853R127
I tried to keep the same logic as before but handling it in a per-cipher basis of the loop so I had to split it.

BitwardenShared/Core/Platform/Services/StateService.swift
fedemkr: Yes, I think from what the callers are doing it's a good idea to remove the `throws` from the original function and just log the error and return `false`. Nice one!

BitwardenShared/Core/Vault/Helpers/VaultListArrangedDataBuilder.swift
KatherineInCode: I do like trying to slimline what we actually hold in memory vs. what we count. Though I wonder if to Matt's point there might be some way to have a like, "slimline CipherListView" that *doesn't* have the full decrypted object in memory, but can still hang around to make for easier counting at the end rather than manual incrementing 🤔 and maybe we have some sort of lazy property on it that actually get the full decrypted object when we actually need it?

Spaghetti at the wall here.

BitwardenShared/Core/Vault/Helpers/VaultListDirectorStrategyFactory.swift
fedemkr: Yes, I added that now to show how the director strategies would behave for each vault list view of the app. I see some of the overlap that could be in a base object maybe but my intention was to have them split so we have a clear distinction on what each strategy is for as how the data is arranged and the sections are built which are different between the two. And even though I could clearly merge this in one object I would need to add more if branches which is the thing I'm trying to avoid to reduce cyclomatic complexity and decrease the number of test cases we have in our suite.
Thoughts?

BitwardenShared/Core/Vault/Services/CiphersClientWrapperService.swift
fedemkr: Yep, nice catch! I was doing something before in between those lines and removed it but forgot to remove the guard as well. Thanks!

BitwardenShared/Core/Vault/Helpers/VaultListDirectors/MainVaultListDirectorStrategy.swift
fedemkr: 🤔 That seems like a good improvement, I'll take a look with the other scenarios of future strategies but I think that could work very well, thanks!

BitwardenShared/Core/Vault/Helpers/VaultListDataArranger.swift
fedemkr: I may have missed some place but they should be being sorted directly in the `VaultListBuilder` on each section addition so the sort operation is faster as it only sorts what's already been filtered and per section where if you don't add the section then the sort operation doesn't need to happen.

BitwardenShared/Core/Vault/Helpers/VaultListArrangedDataBuilder.swift
fedemkr: 🤔 Wouldn't that be similar as now but with objects? I mean one would need to add the slim cipher view to a collection so it's later counted. I don't see much improvement over just incrementing a counter. Same number of calls to a method (increment vs add an object) and I believe similar chances of error of not calling such method. Perhaps a little bit safer given that you have the actual objects and more difficult to get duplicated counts but I feel the count safety should be covered by unit tests.

KatherineInCode: Point taken, especially if we have tests as the safety net. Sometimes none of the spaghetti you throw at the wall sticks 😅

matt-livefront: > Basically I wanted to avoid having the actual ciphers when not needed. E.g. for the main vault screen we only need the count of TOTP ciphers not the actual ciphers objects so I thought saving such decrypted objects in memory for the count was a waste of resources.

👍 This makes more sense now that I'm understanding this is counting stuff that doesn't exist in the builder metadata.

BitwardenShared/Core/Vault/Helpers/VaultListDataArranger.swift
matt-livefront: I see, maybe it's just `addGroupSection()` which is missing it.

BitwardenShared/Core/Vault/Helpers/VaultListDirectorStrategyFactory.swift
matt-livefront: Ok, yeah that makes sense!

BitwardenShared/Core/Vault/Helpers/VaultListDataArranger.swift
```diff
@@ -0,0 +1,130 @@
...
+            if filter.addTOTPGroup {
+                arrangedDataBuilder = await arrangedDataBuilder.incrementTOTPCount(cipher: decryptedCipher)
+            }
+
+            arrangedDataBuilder = arrangedDataBuilder
+                .addFolderItem(cipher: decryptedCipher, filter: filter, folders: folders)
+                .addFavoriteItem(cipher: decryptedCipher)
+                .addNoFolderItem(cipher: decryptedCipher)
+                .incrementCipherTypeCount(cipher: decryptedCipher)
+                .incrementCollectionCount(cipher: decryptedCipher)
```
matt-livefront: 🤔 One other thought I had, I think almost all of the grouping/counting/filtering of the data here happens on fields that aren't encrypted (e.g. `deletedDate`, `folderId`, `collectionId`, `CipherType`, etc). Instead of decrypting ciphers up-front, could we decrypt only the data that we end up showing in the list? Would that work? I think the TOTP group relies on having a decrypted TOTP key, but it seems like most or all of the other fields aren't encrypted. This might not work for the autofill lists but could be helpful for the main or group lists.

BitwardenShared/Core/Platform/Services/StateService.swift
fedemkr: Handled in #1689. I'll update this code when the other PR is merged.

BitwardenShared/Core/Vault/Helpers/VaultListDataArranger.swift
fedemkr: I'm not 100% convinced on `Factory` or `Initializer` for the logic being involved inside. So I've changed it to `VaultListDataPreparator` and the struct to `VaultListPreparedData`. How does that sound?

KatherineInCode: I like it!

BitwardenShared/Core/Vault/Extensions/CipherListViewExtensionsTests.swift
```diff
@@ -0,0 +1,52 @@
...
+
+class CipherListViewExtensionsTests: BitwardenTestCase {
+    // MARK: Tests
+
+    /// `passesRestrictItemTypesPolicy(_:)` passes the policy when there are no organization IDs.
+    func test_passesRestrictItemTypesPolicy_noOrgIds() {
+        XCTAssertTrue(CipherListView.fixture().passesRestrictItemTypesPolicy([]))
+    }
+
+    /// `passesRestrictItemTypesPolicy(_:)` passes the policy when the cipher type is no `.card`.
```
matt-livefront: ⛏️ 
```suggestion
    /// `passesRestrictItemTypesPolicy(_:)` passes the policy when the cipher type is not `.card`.
```

BitwardenShared/Core/Vault/Extensions/VaultListSection+SnapshotExtensions.swift
matt-livefront, line 1: ❓ Do we want/need this in the app target? Or should this only be in the testing target (in a TestHelpers folder)?

BitwardenShared/Core/Vault/Helpers/VaultListDataPreparator.swift
```diff
@@ -0,0 +1,142 @@
+import BitwardenKit
+import BitwardenSdk
+
+/// A object that prepares vault list data for the sections builder `VaultListSectionsBuilder`.
+///
+/// This decrypts and process data iteratively in batches to improve time and memory on the overall
+/// grouping/filtering/preparation.
+protocol VaultListDataPreparator { // sourcery: AutoMockable
+    /// Prepares data for the vault list builder based.
```
matt-livefront: ⛏️ Is this needed?

```suggestion
    /// Prepares data for the vault list builder.
```

BitwardenShared/Core/Vault/Helpers/VaultListDataPreparator.swift
```diff
@@ -0,0 +1,142 @@
...
+    /// - Parameters:
+    ///   - ciphers: An array of `Cipher` objects to be processed.
+    ///   - collections: An array of `Collection` objects to be processed.
+    ///   - folders: An array of `Folder` objects to be processed.
+    ///   - filter: A `VaultListFilter` object that defines the filtering criteria for the vault list.
+    /// - Returns: An optional `VaultListPreparedData` object containing the prepared data for the vault list.
```
matt-livefront, line 15: 🎨 Is it worth defining what returning `nil` here means? That's it's empty or contains no ciphers?
```suggestion
    /// - Returns: An optional `VaultListPreparedData` object containing the prepared data for the vault list. Returns `nil` if the vault is empty.
```

BitwardenShared/Core/Vault/Helpers/VaultListDataPreparatorTests.swift
```diff
@@ -0,0 +1,339 @@
...
+        XCTAssertTrue(vaultListPreparedDataBuilder.prepareFoldersFoldersFilterTypeCalled)
+        XCTAssertTrue(vaultListPreparedDataBuilder.prepareCollectionsCollectionsFilterTypeCalled)
+        XCTAssertFalse(vaultListPreparedDataBuilder.incrementCipherDeletedCountCalled)
+        XCTAssertTrue(vaultListPreparedDataBuilder.incrementTOTPCountCipherCalled)
+        XCTAssertTrue(vaultListPreparedDataBuilder.addFolderItemCipherFilterFoldersCalled)
+        XCTAssertTrue(vaultListPreparedDataBuilder.addFavoriteItemCipherCalled)
+        XCTAssertTrue(vaultListPreparedDataBuilder.addNoFolderItemCipherCalled)
+        XCTAssertTrue(vaultListPreparedDataBuilder.incrementCipherTypeCountCipherCalled)
+        XCTAssertTrue(vaultListPreparedDataBuilder.incrementCollectionCountCipherCalled)
+        XCTAssertNotNil(result)
```
matt-livefront: 🤔 While these tests are thorough, they are pretty tedious to try and follow. I'm wondering if it wouldn't be easer to comprehend if we tested the actual implementation of the builder rather than using a mock? Or have some combination of each?

If we had a way of snapshotting the representation of `VaultListPreparedData`, you could pass in some data and get an output that shows the actual data that will be sent to the next stage of building the list, without just asserting which methods were called on the builder. You would be testing more than one thing and there could be other downsides to this. And maybe there's some challenges that I'm not foreseeing, but a visual output of the resulting data seems like it could be a useful and comprehensive test.

BitwardenShared/Core/Vault/Helpers/VaultListDirectors/MainVaultListDirectorStrategy.swift
```diff
@@ -0,0 +1,88 @@
...
+    /// The service for managing the collections for the user.
+    let collectionService: CollectionService
+    /// The service used by the application to report non-fatal errors.
+    let errorReporter: ErrorReporter
+    /// The service used to manage syncing and updates to the user's folders.
+    let folderService: FolderService
+    /// The service used by the application to manage account state.
+    let stateService: StateService
+    /// The helper used to prepare data for the vault list builder.
+    let vaultListDataPreparator: VaultListDataPreparator
```
matt-livefront: 🎨 Looks like `clientService`, `errorReporter`, and `stateService` aren't used within this type.

BitwardenShared/Core/Vault/Helpers/VaultListDirectors/MainVaultListGroupDirectorStrategy.swift
```diff
@@ -0,0 +1,81 @@
...
+    /// The service used by the application to handle encryption and decryption tasks.
+    let clientService: ClientService
+    /// The service for managing the collections for the user.
+    let collectionService: CollectionService
+    /// The service used by the application to report non-fatal errors.
+    let errorReporter: ErrorReporter
+    /// The service used to manage syncing and updates to the user's folders.
+    let folderService: FolderService
+    /// The service used by the application to manage account state.
+    let stateService: StateService
```
matt-livefront: 🎨 Looks like clientService, errorReporter, and stateService aren't used within this type.

BitwardenShared/Core/Vault/Helpers/VaultListPreparedDataBuilder+AddItemTests.swift
```diff
@@ -0,0 +1,289 @@
+// swiftlint:disable:this file_name
```
matt-livefront, line 1: 🤔 Any reason not to have the file name match the test class name? Typically we only use `+` in the file name for extensions. Might be easier to make them the same in this case.

BitwardenShared/Core/Vault/Helpers/VaultListPreparedDataBuilder.swift
```diff
@@ -0,0 +1,312 @@
...
+
+/// A factory protocol to make vault list prepared data builders.
+protocol VaultListPreparedDataBuilderFactory { // sourcery: AutoMockable
+    /// Makes a builder for `VaultListPreparedData`.
+    func make() -> VaultListPreparedDataBuilder
+}
+
+// MARK: - DefaultVaultListPreparedDataBuilderFactory
+
+/// The default implemetnation of `VaultListPreparedDataBuilderFactory`.
```
matt-livefront: ⛏️ 
```suggestion
/// The default implementation of `VaultListPreparedDataBuilderFactory`.
```

BitwardenShared/Core/Vault/Helpers/VaultListPreparedDataBuilder.swift
```diff
@@ -0,0 +1,312 @@
...
+        case .identity:
+            preparedData.countPerCipherType[.identity, default: 0] += 1
+        case .login:
+            preparedData.countPerCipherType[.login, default: 0] += 1
+        case .secureNote:
+            preparedData.countPerCipherType[.secureNote, default: 0] += 1
+        case .sshKey:
+            preparedData.countPerCipherType[.sshKey, default: 0] += 1
+        }
+
```
matt-livefront: 🎨 If you add an extension to `CipherType` to initialize from a `CipherListType`, I think you could simplify this:

```suggestion
        preparedData.countPerCipherType[CipherType(cipher.type), default: 0] += 1
```

BitwardenShared/Core/Vault/Helpers/VaultListSectionsBuilder+FolderTests.swift
```diff
@@ -0,0 +1,250 @@
...
+    // MARK: Tests
+
+    /// `addFoldersSection(nestedFolderId:)` adds the folders section to the list of sections
+    /// with the count of ciphers per folder.
+    func test_addFoldersSection_noNestedFolderId() async throws {
+        setUpSubject(
+            withData: VaultListPreparedData(
+                folders: [
+                    .fixture(id: "1", name: "folder1"),
+                    .fixture(id: "2", name: "afolder2"),
```
matt-livefront, line 44: ❓ Just curious, what's the significance of using `afolder2` here vs `folder2`?

BitwardenShared/Core/Vault/Helpers/VaultListSectionsBuilder.swift
```diff
@@ -0,0 +1,295 @@
...
+    func addTOTPSection() -> VaultListSectionsBuilder {
+        sections.append(VaultListSection(
+            id: "TOTP",
+            items: preparedData.totpItemsCount > 0
+                ? [
+                    VaultListItem(
+                        id: "Types.VerificationCodes",
+                        itemType: .group(.totp, preparedData.totpItemsCount)
+                    ),
+                ] : [],
```
matt-livefront: 🤔 Do we still want to add the Verification Codes section if there's no TOTP items? I think we filtered that out previously.

BitwardenShared/UI/Vault/Vault/VaultList/VaultListItem.swift
```diff
@@ -39,11 +44,16 @@ public struct VaultListItem: Equatable, Identifiable, Sendable, VaultItemWithDec
...
+    /// What's used to sort `VaultListItem`s depending on its item type.
+    var sortValue: String {
+        return switch itemType {
+        case let .cipher(cipherListView, _):
+            cipherListView.name
+        case .group:
+            ""
+        case let .totp(name, model):
+            name + (model.cipherListView.type.loginListView?.username ?? "") + "\(model.id)"
+        }
```
matt-livefront, line 56: 🤔 Instead of implementing this `sortValue`, I wonder if we could implement `Comparable` and build the sorting into that based on the item type? Then I think we could just call `.sorted()` on an array of `VaultListItem`s.

BitwardenShared/Core/Platform/Utilities/TestHelpers/MockCallOrderHelper.swift
```diff
@@ -0,0 +1,11 @@
+/// Helper class to save calls makes to mocks in order.
```
KatherineInCode: ⛏️ 
```suggestion
/// Helper class to save calls made to mocks, in order.
```

BitwardenShared/Core/Vault/Extensions/VaultListSection+SnapshotExtensions.swift
```diff
@@ -0,0 +1,64 @@
+// MARK: - VaultListSection
+
+extension VaultListSection {
```
KatherineInCode: 🤔 Do we want to put these behind an `#if DEBUG` or something to make sure they're only used in testing? (It's also possible we're using it in the app somewhere and I've just missed it)

BitwardenShared/Core/Vault/Services/PolicyServiceTests.swift
```diff
@@ -696,4 +701,62 @@ class PolicyServiceTests: BitwardenTestCase { // swiftlint:disable:this type_bod
         let policies = await subject.getOrganizationIdsForRestricItemTypesPolicy()
         XCTAssertTrue(policies.isEmpty)
     }
+
+    /// `passesRestrictItemTypesPolicy(cipher:)` returns `true` when the `.removeCardPolicy` is enabled
+    /// and cipher passes restricted organization ids.
+    @MainActor
+    func test_passesRestrictItemTypesPolicy_cipherPassesRestrictedOrganizationIds() async {
```
KatherineInCode, line 708: 🤔 These feel like strange things to be in this PR. Is there something weird that happened with a merge from main?

BitwardenShared/Core/Vault/Helpers/VaultListSectionsBuilder+FolderTests.swift
fedemkr: I wanted to mix the order of the folders' names so I could corroborate they're properly ordered at the end, that's why I added a `a` at the beginning so then it needs to be moved to be ordered. I could use different names if you like.

BitwardenShared/Core/Vault/Services/PolicyServiceTests.swift
fedemkr: When I merge the changes on restrict item types policy I moved things a bit so they could accommodate better to the current approach in this PR. This also allows to decrease the number of combination tests done in the caller as how it was previously implemented.
Also I had to leave the previous implementation of that in place until the vault list refactor is applied to all places like search and the autofill.

BitwardenShared/Core/Vault/Helpers/VaultListSectionsBuilder+FolderTests.swift
matt-livefront: Ah, no that makes sense. The first time I saw it, I thought it was just an extra character that got inserted, but then I saw it repeated which made me curious.

BitwardenShared/Core/Vault/Helpers/VaultListSectionsBuilder.swift
fedemkr: Nice catch, I missed the filter which removed the section on 0 items. Thanks!

BitwardenShared/Core/Vault/Extensions/VaultListSection+SnapshotExtensions.swift
fedemkr: Moving it to `TestHelpers` folders.

BitwardenShared/UI/Vault/Vault/VaultList/VaultListItem.swift
fedemkr: 🤔 Isn't it implementing what the Sort descriptor does based on the `sortValue` property?

BitwardenShared/Core/Vault/Helpers/VaultListDataPreparatorTests.swift
fedemkr: Thanks for pointing this out, changed the approach as the same used in the `*VaultListDirectorStrategyTests` by using the `MockCallOrderHelper` to check which ones have been called. Do you feel that would be simple enough to follow?

Also my goal here is actually not to test the builder's implementation to simplify maintenance as if someone changes the implementation of one of the methods in the builder then we should only update the builder's tests instead of repercussions in other places. Moreover in this case there would multiple places to update as methods are used in combination.

Some other thing I could do is in the builder mock to add mock data to `VaultListPreparedData` and snapshot it as you mentioned but I feel like it's a bit outside of scope on the responsibilities of the mock and it could easily get out-of-sync with logic that is not 1 to 1 on method called vs `VaultListPreparedData` property.

BitwardenShared/UI/Vault/Vault/VaultList/VaultListItem.swift
matt-livefront: Yeah, that's fair, they are pretty similar. Using a sort descriptor gives you more flexibility on the property to sort, but we only sort alphabetically. So Comparable might end up being simpler, but I'm good with the sort descriptor approach.

BitwardenShared/Core/Vault/Helpers/VaultListDataPreparatorTests.swift
matt-livefront: That's fair, I think using `MockCallOrderHelper` helps to make these tests clearer!

Gerardv514:
> I'm seeing an empty TOTP and Favorites section if those items don't have any ciphers. Should those be filtered out?
> 

Should also hide types if there are none associated with it. If my mom doesn't use ssh keys (she has no clue what they even are) why show it.

fedemkr:
Hi @Gerardv514, thanks for pointing that out but that's out of scope of this PR which should leave the UI displaying the same as before the refactor. Additionally, that doesn't seem like a bug but a change in the existing feature.

The Bitwarden Community Forums has a Feature Requests section for submitting, voting for, and discussing requests like this one: https://community.bitwarden.com/c/feature-requests/

Please sign up on our forums and search to see if this request already exists (https://community.bitwarden.com/signup). If so, you can vote for it and contribute to any discussions about it. If not, you can re-create the request there so that it can be properly tracked.

PR #1691, opened 2025-06-25
[PM-19308] Handle restrict item types policy

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-19308 

## 📔 Objective

When the policy is turned on:

- The policy will be applied to the organization’s and individual vault
- Items of the card item type will be hidden (not deleted)
- The card item group will be hidden
- The card option in creating a new item will be hidden
- If the policy is turned on and off again, the card items will re-appear.

If the user is part of multiple organizations, and only 1 (or more) organizations have turned on the policy and 1 (or more) organizations have turned off the policy:
All of the above requirements apply, except:

- Users can continue to access and edit card item types that are stored in organizations with the policy turned off
- The card group will appear if users have card items stored in a collection for an organization with the policy turned off
 
## 📸 Screenshots

https://github.com/user-attachments/assets/d9aed18c-b5cd-4057-9ab2-2a799e991519

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


BitwardenShared/Core/Vault/Models/Enum/CipherType.swift
```diff
@@ -64,6 +64,9 @@ extension CipherType {
     /// These are the cases of `CipherType` that the user can use to create a cipher.
     static let canCreateCases: [CipherType] = [.login, .card, .identity, .secureNote]
 
+    /// These are the cases of `CipherType` that the user can use to create a cipher based on restrict item type policy.
+    static let canCreateCasesRestricted: [CipherType] = [.login, .identity, .secureNote]
```
fedemkr: 🤔 What do you think about renaming it to be explicit about the policy? Something like `canCreateRestrictItemTypesPolicyCases`

BitwardenShared/Core/Vault/Repositories/VaultRepository.swift
```diff
@@ -560,6 +561,9 @@ class DefaultVaultRepository { // swiftlint:disable:this type_body_length
                         isMatchingCipher(cipher) &&
                         (cipherFilter?(cipher) ?? true)
                 }
+                .filter { cipher in
+                    self.filterBasedOnRestrictItemTypesPolicy(cipher: cipher, restrictItemTypesOrgIds)
+                }
```
fedemkr: 🎨 You can just put this filter inside the previous one with an `&&`

BitwardenShared/Core/Vault/Repositories/VaultRepository.swift
```diff
@@ -809,8 +813,10 @@ class DefaultVaultRepository { // swiftlint:disable:this type_body_length
         collections: [Collection],
         folders: [Folder] = []
     ) async throws -> [VaultListSection] {
+        let restrictItemTypesOrgIds = await getRestrictItemTypesOrgIds()
         let ciphers = try await clientService.vault().ciphers().decryptList(ciphers: ciphers)
             .filter(filter.filterType.cipherFilter)
+            .filter { filterBasedOnRestrictItemTypesPolicy(cipher: $0, restrictItemTypesOrgIds) }
```
fedemkr: 🎨 Combine the filters.

BitwardenShared/Core/Vault/Repositories/VaultRepository.swift
```diff
@@ -850,6 +856,33 @@ class DefaultVaultRepository { // swiftlint:disable:this type_body_length
...
+    ///  - Returns: `true` if the cipher is allowed by the policy, `false` otherwise.
+    ///
+    func filterBasedOnRestrictItemTypesPolicy(cipher: CipherListView, _ restrictItemTypesOrgIds: [String]) -> Bool {
+        // If the restriction list is empty or the cipher is not a card, return true.
+        guard !restrictItemTypesOrgIds.isEmpty, cipher.type.isCard else { return true }
+        // If the cipher is not associated with an organization, return true.
+        guard let orgId = cipher.organizationId, !orgId.isEmpty else { return false }
+        // If the cipher's organization ID is not in the restriction list, return true.
+        return !restrictItemTypesOrgIds.contains(orgId)
+    }
```
fedemkr: ⚠️ Shouldn't `SSHKey` item type be into consideration along side `Card` as well? I mean from what I can see in the [can create restricted cases ](https://github.com/bitwarden/ios/pull/1691/files#diff-f373b65deba9465354d83ade7ca172684785ee2c0b8de1ebbfb03052a20043b3R68)`sshKey` is not one of them.
🤔 I think the code is pretty self-explanatory on what is doing on each line, I'd just remove the interline comments; unless you feel any of them is necessary as it's not obvious what the code is doing.

BitwardenShared/Core/Vault/Repositories/VaultRepository.swift
```diff
@@ -865,8 +898,10 @@ class DefaultVaultRepository { // swiftlint:disable:this type_body_length
         folders: [Folder],
         filter: VaultListFilter
     ) async throws -> [VaultListSection] {
+        let restrictItemTypesOrgIds = await getRestrictItemTypesOrgIds()
         let ciphers = try await clientService.vault().ciphers().decryptList(ciphers: ciphers)
             .filter(filter.filterType.cipherFilter)
+            .filter { filterBasedOnRestrictItemTypesPolicy(cipher: $0, restrictItemTypesOrgIds) }
```
fedemkr: ⛏️ Combine filters.

BitwardenShared/Core/Vault/Repositories/VaultRepository.swift
```diff
@@ -926,14 +961,20 @@ class DefaultVaultRepository { // swiftlint:disable:this type_body_length
...
             VaultListItem(id: "Types.SecureNotes", itemType: .group(.secureNote, typesSecureNoteCount)),
             VaultListItem(id: "Types.SSHKeys", itemType: .group(.sshKey, typesSSHKeyCount)),
         ]
 
+        // Only show the card section if there are cards and restrictItemTypes policy is not enabled.
+        if typesCardCount == 0,
+           !restrictItemTypesOrgIds.isEmpty {
+            types.remove(at: 1)
+        }
+
```
fedemkr: 🎨 Instead of removing by index which is fragile given that if someone changes the order of the item types it will end up removing a different item type, I would split the array and add the card items only when necessary instead of removing it afterwards.
```swift
var types = [
    VaultListItem(id: "Types.Logins", itemType: .group(.login, typesLoginCount)),
]

// Only show the card section if there are cards and restrictItemTypes policy is not enabled.
if typesCardCount != 0 || restrictItemTypesOrgIds.isEmpty {
    types.append(VaultListItem(id: "Types.Cards", itemType: .group(.card, typesCardCount)))
}

types.append(contentsOf: [
    VaultListItem(id: "Types.Cards", itemType: .group(.card, typesCardCount)),
    VaultListItem(id: "Types.Identities", itemType: .group(.identity, typesIdentityCount)),
    VaultListItem(id: "Types.SecureNotes", itemType: .group(.secureNote, typesSecureNoteCount)),
    VaultListItem(id: "Types.SSHKeys", itemType: .group(.sshKey, typesSSHKeyCount)),
]
```

BitwardenShared/Core/Vault/Services/PolicyService.swift
```diff
@@ -20,6 +20,13 @@ protocol PolicyService: AnyObject {
     ///
     func fetchTimeoutPolicyValues() async throws -> (action: SessionTimeoutAction?, value: Int)?
 
+    /// Get all active policies of a given type that apply to the active user.
+    ///
+    /// - Parameter policyType: The policy to check.
+    /// - Returns: Active policies that apply to the user.
+    ///
+    func getActiveUserPolicies(_ policyType: PolicyType) async -> [Policy]
+
```
fedemkr: 🤔 🎨 From what I can see this is being used to get the restricted item policy type policies to then just map the org IDs of such policies on each occasions. So in order not to repeat the same mapping code I would just change this method to actually return an array of Organization IDs for the specific policy you need that and then add the logic here for that. So the tests to get the policies and map the org id from such policies, if any all remain inside the `PolicyService[Tests]`, which makes it more cohesive and if, for some reason the `Policy` struct changes you will only need to modify the `PolicyService` instead of all the places this gets called.
Something like:
`func getOrganizationIdsForRestricItemTypesPolicy() async -> [String]`
or if you think it's going to be used with other policy types then you can just pass that as a parameter, but I wouldn't return the actual `Policy` to the caller in this case.

BitwardenShared/UI/Platform/Application/Extensions/View.swift
```diff
@@ -155,7 +156,8 @@ extension View {
                 Divider()
             }
 
-            ForEach(CipherType.canCreateCases.reversed(), id: \.hashValue) { type in
+            let itemTypes = restrictItemTypes ? CipherType.canCreateCasesRestricted : CipherType.canCreateCases
+            ForEach(itemTypes, id: \.hashValue) { type in
```
fedemkr: ⚠️ Beware the previous for each was using the array `reversed()`.

BitwardenShared/Core/Platform/Models/Enum/FeatureFlag.swift
```diff
@@ -58,6 +58,11 @@ extension FeatureFlag: @retroactive CaseIterable {
         rawValue: "pm-15493-restrict-item-deletion-to-can-manage-permission"
     )
 
+    /// A feature flag to enable the removal of card item types.
+    static let removeCardPolicy = FeatureFlag(
+        rawValue: "pm-16442-remove-card-item-type-policy"
+    )
+
```
fedemkr, line 65: ❓ Is this different than the `PolicyType -> restrictItemTypes` or the same? If so, I feel like they should be named the same. Additionally, from what I can see in the `canCreateCasesRestricted` it's not only cards, but `sshKey` types as well that are restricted, or is that only for creation?

AuthenticatorBridgeKit/Mocks/MockAuthenticatorBridgeItemService.swift
```diff
@@ -11,6 +11,8 @@ public class MockAuthenticatorBridgeItemService: AuthenticatorBridgeItemService
 
     public init() {}
 
+    public func checkForLogout() async throws {}
```
KatherineInCode: 🤔 Is this necessary? Or is it just an artifact of a previous protocol that's since been changed?

BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessor.swift
```diff
@@ -241,6 +242,18 @@ extension VaultListProcessor {
...
 
+    /// Checks if the restrict item types policy is enabled.
+    ///
+    private func checkRestrictItemTypesPolicy() async {
+        state.isRemoveCardPolicyFeatureFlagEnabled = await services.configService.getFeatureFlag(.removeCardPolicy)
+        if state.isRemoveCardPolicyFeatureFlagEnabled {
+            state.restrictItemTypesOrgIds = await services.policyService
+                .getActiveUserPolicies(.restrictItemTypes).map { policy in
+                    policy.organizationId
+                }
```
matt-livefront: 🤔 Instead of loading the feature flag and policy here, would it be cleaner and potentially more flexible to expose a method in the repository which returns the available cipher types that can be added? And then that list could be passed to the `addVaultItemFloatingActionMenu`?

BitwardenShared/Core/Vault/Models/Enum/PolicyType.swift
```diff
@@ -40,6 +40,9 @@ enum PolicyType: Int, Codable {
     /// If enabled, the setting to "Unlock with Pin" is hidden.
     case removeUnlockWithPin = 14
 
+    /// If enabled, restricts the types of items that can be shown in the vault.
+    case restrictItemTypes = 15
```
KatherineInCode, line 44: 🤔 Maybe this is YAGNI because the only type we restrict is cards, but would it make sense to be more explicit about that, in case we add the ability to restrict other item types in the future?

BitwardenShared/UI/Vault/Vault/VaultGroup/VaultGroupView.swift
```diff
@@ -85,17 +85,22 @@ struct VaultGroupView: View {
                 )
 
                 Group {
+                    let restrictItemTypes = store.state.isRestrictItemPolicyEnabled
                     switch newItemButtonType {
```
matt-livefront: 🤔 I think we could also do something similar here, by reusing the same repository method which returns the available list of types that could be added. We already support hiding the button if `newItemButtonType` returns `nil`. So within `newItemButtonType`, we could map from `VaultListGroup` to `CipherType` using `CipherType(group:)` and check if that exists in the available types? Something like this:

```
    var newItemButtonType: NewItemButtonType? {
+       if let cipherType = CipherType(group: group), !availableCipherTypes.contains(cipherType) {
+           return nil
+       }

        switch group {
        case .card, .identity, .login, .secureNote:
            return .button
        case .collection, .folder, .noFolder:
            return .menu
        case .sshKey, .totp, .trash:
            return nil
        }
    }
```

BitwardenShared/Core/Vault/Models/Enum/PolicyType.swift
andrebispo5: Server and clients made the change to be generic, so I just followed the same logic for the mobile side of things to keep consistent.

BitwardenShared/Core/Vault/Repositories/VaultRepository.swift
```diff
@@ -850,6 +856,33 @@ class DefaultVaultRepository { // swiftlint:disable:this type_body_length
...
+    /// - Parameters:
+    ///  - cipher: The cipher to check against the policy.
+    ///  - restrictItemTypesOrgIds: The list of organization IDs that are restricted by the policy.
+    ///  - Returns: `true` if the cipher is allowed by the policy, `false` otherwise.
+    ///
+    func filterBasedOnRestrictItemTypesPolicy(cipher: CipherListView, _ restrictItemTypesOrgIds: [String]) -> Bool {
+        // If the restriction list is empty or the cipher is not a card, return true.
+        guard !restrictItemTypesOrgIds.isEmpty, cipher.type.isCard else { return true }
+        // If the cipher is not associated with an organization, return true.
+        guard let orgId = cipher.organizationId, !orgId.isEmpty else { return false }
```
matt-livefront: ⛏️ 
```suggestion
        // If the cipher is not associated with an organization, return false.
        guard let orgId = cipher.organizationId, !orgId.isEmpty else { return false }
```

AuthenticatorBridgeKit/Mocks/MockAuthenticatorBridgeItemService.swift
andrebispo5: No clue, it was the lint tool that changed the formatting on it. 🤔

BitwardenShared/Core/Vault/Repositories/VaultRepository.swift
andrebispo5: SSHKey [was not there before](https://github.com/bitwarden/ios/pull/1691/files#diff-f373b65deba9465354d83ade7ca172684785ee2c0b8de1ebbfb03052a20043b3L65), it must be being added somewhere else.

BitwardenShared/Core/Vault/Repositories/VaultRepository.swift
andrebispo5: Ye I was on the fence on what approach to use, but you make a good point 👍

BitwardenShared/Core/Platform/Models/Enum/FeatureFlag.swift
andrebispo5: The names are different because when the ticket and the feature flag were created, the intention was only restrict cards, therefore the name. But down the road `server` and `client` changed to be more dynamic and allow other types to be restricted. That is why the name don't match, but the feature flag will be remove in time. 
The sshKey type was explained [here](https://github.com/bitwarden/ios/pull/1691#discussion_r2169266927).

AuthenticatorBridgeKit/Mocks/MockAuthenticatorBridgeItemService.swift
KatherineInCode: The formatting? Or adding it altogether?

BitwardenShared/Core/Vault/Models/Enum/PolicyType.swift
KatherineInCode: Makes sense

AuthenticatorBridgeKit/Mocks/MockAuthenticatorBridgeItemService.swift
andrebispo5: that method was already there, it was only formatted to remove the blank line in the body.

KatherineInCode: It was removed entirely elsewhere; the PR diff shows it being readded?

BitwardenShared/Core/Vault/Services/PolicyService.swift
andrebispo5: You are right, I didn't put much thought into it because on android we already had this method that returned the policies and my first reaction was to add it on iOS.

BitwardenShared/Core/Vault/Repositories/VaultRepository.swift
fedemkr: Oh so sorry, my bad SSH Keys are not allowed to be created in iOS.

AuthenticatorBridgeKit/Mocks/MockAuthenticatorBridgeItemService.swift
andrebispo5: oh sorry, you are right it was a merge with main that removed it and now it was added again. ty

BitwardenShared/Core/Vault/Repositories/TestHelpers/MockVaultRepository.swift
```diff
@@ -66,6 +66,8 @@ class MockVaultRepository: VaultRepository {
 
     var getTOTPKeyIfAllowedToCopyResult: Result<String?, Error> = .success(nil)
 
+    var getItemTypesUserCanCreate: [BitwardenShared.CipherType] = CipherType.canCreateCases
```
fedemkr: ⛏️ Rename to have result suffix so it's consistent with other properties, it doesn't matter if the result is not actually a `Result<...>`.
```suggestion
    var getItemTypesUserCanCreateResult: [BitwardenShared.CipherType] = CipherType.canCreateCases
```

BitwardenShared/Core/Vault/Repositories/VaultRepository.swift
```diff
@@ -926,13 +962,18 @@ class DefaultVaultRepository { // swiftlint:disable:this type_body_length
         let typesSecureNoteCount = activeCiphers.lazy.filter { $0.type == .secureNote }.count
         let typesSSHKeyCount = activeCiphers.lazy.filter { $0.type == .sshKey }.count
 
-        let types = [
-            VaultListItem(id: "Types.Logins", itemType: .group(.login, typesLoginCount)),
-            VaultListItem(id: "Types.Cards", itemType: .group(.card, typesCardCount)),
+        var types = [VaultListItem(id: "Types.Logins", itemType: .group(.login, typesLoginCount))]
+
+        // Only show the card section if there are cards and restrictItemTypes policy is not enabled.
```
fedemkr: ⛏️ I think you meant "or"
```suggestion
        // Only show the card section if there are cards or restrictItemTypes policy is not enabled.
```

BitwardenShared/Core/Vault/Repositories/VaultRepository.swift
```diff
@@ -1120,6 +1161,17 @@ extension DefaultVaultRepository: VaultRepository {
         return organizations.first(where: { $0.id == id })
     }
 
+    func getItemTypesUserCanCreate() async -> [CipherType] {
+        let itemTypes: [CipherType] = CipherType.canCreateCases.reversed()
```
fedemkr, line 1165: 🤔 I'm not 100% sure whether we should be using `reversed` here or not... On one hand the order of the array is based on which UI that's presented therefore it should be done by the processor or state of such view. But on the other hand we're already ordering other types like ciphers, folders and others in the `VaultRepository` for the vault list section items. I guess for the time being this is fine given it centralizes the operation on the current callers.
What do you think? @KatherineInCode @matt-livefront

BitwardenShared/Core/Vault/Repositories/VaultRepositoryTests.swift
```diff
@@ -1087,6 +1087,59 @@ class VaultRepositoryTests: BitwardenTestCase { // swiftlint:disable:this type_b
...
+                organizationId: "org1",
+                type: .restrictItemTypes,
+            ),
+        ]
+
+        let result = await subject.getItemTypesUserCanCreate()
+        XCTAssertEqual(
+            result,
+            CipherType.canCreateCases.filter { $0 != .card }.reversed(),
+        )
```
fedemkr: 🎨 IMO tests should not use the same logic as what's testing as it could be a bit hard to catch bugs on the implementation given that it would use the same as in the test method. Basically this would test the logic against itself instead of a known-good result.
I believe the expected result in this case would need to be the explicit expected enum cases ordered in an array: `[.sshKey, .secureNote, .identity, .login]`.
Same applies to the other tests for `getItemTypesUserCanCreate`.

BitwardenShared/Core/Vault/Repositories/VaultRepositoryTests.swift
```diff
@@ -2824,6 +2877,72 @@ class VaultRepositoryTests: BitwardenTestCase { // swiftlint:disable:this type_b
...
+                    id: "Items",
+                    items: [.fixture(cipherListView: .init(cipher: cipherOrg2))],
+                    name: Localizations.items
+                ),
+            ]
+        )
+    }
+
+    /// `vaultListPublisher(group:filter:)` returns a publisher for the vault list items not filtering cards from
+    /// individual and organizations vaults if restrict item type policy is enabled not enabled.
```
fedemkr: ⛏️ I believe you just meant `not enabled`:
```suggestion
    /// `vaultListPublisher(group:filter:)` returns a publisher for the vault list items not filtering cards from
    /// individual and organizations vaults if restrict item type policy is not enabled.
```

BitwardenShared/UI/Vault/Vault/VaultGroup/VaultGroupProcessor.swift
```diff
@@ -173,6 +174,12 @@ final class VaultGroupProcessor: StateProcessor<
         state.canShowVaultFilter = await services.vaultRepository.canShowVaultFilter()
     }
 
+    /// Checks available item types user can create.
+    ///
+    private func checkItemTypesUserCanCreate() async {
```
fedemkr: ⛏️ Rather than "checking" I believe this is "loading" the item types.

BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessor.swift
```diff
@@ -241,6 +242,12 @@ extension VaultListProcessor {
         state.canShowVaultFilter = await services.vaultRepository.canShowVaultFilter()
     }
 
+    /// Checks available item types user can create.
+    ///
+    private func checkItemTypesUserCanCreate() async {
```
fedemkr: ⛏️ Same as the other one IMO this is "loading" instead of "checking" item types.

BitwardenShared/Core/Vault/Repositories/VaultRepository.swift
andrebispo5: It is being used reversed everywhere atm, that is why I just moved it inside the repo. Since this is a Q2 blocker I would like to keep it simple for now.

BitwardenShared/Core/Vault/Repositories/VaultRepository.swift
andrebispo5: TY changed the logic and forgot the comment.

PR #1694, opened 2025-06-26
[PM-22811] fix: Implement invalid_grant on 400 response behavior while refreshing token

fedemkr:
## 🎟️ Tracking

[PM-22811](https://bitwarden.atlassian.net/browse/PM-22811)

## 📔 Objective

Log the user out automatically when getting `invalid_grant` 400 response when refreshing the token.
This was causing for a weird error to be displayed when trying to sync the vault and the user had perform some account change that modified the security stamp and the `access_token` was invalid/expired having to refresh the token.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-22811]: https://bitwarden.atlassian.net/browse/PM-22811?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenShared/Core/Platform/Services/API/AccountTokenProvider.swift
```diff
@@ -54,19 +68,34 @@ actor AccountTokenProvider: TokenProvider {
...
         _ = try await refreshTask.value
     }
+
+    func setupDelegate(delegate: AccountTokenProviderDelegate) async {
+        accountTokenProviderDelegate = delegate
+    }
+}
+
+protocol AccountTokenProviderDelegate: AnyObject {
+    func onRefreshTokenError(error: Error) async throws
```
matt-livefront: ⛏️ Should we add docs to this?

BitwardenShared/Core/Platform/Services/API/AccountTokenProvider.swift
```diff
@@ -1,11 +1,25 @@
 import Networking
 
+// MARK: - AccountTokenProvider
+
+/// A more specific `TokenProvider` protocol to use and ease testing.
+protocol AccountTokenProvider: TokenProvider {
+    /// Sets up the delegate to use in this token provider.
+    /// - Parameter delegate: The delegate to use.
+    func setupDelegate(delegate: AccountTokenProviderDelegate) async
```
matt-livefront: ⛏️ Pretty minor, but I think we've used `setDelegate(_:)` in other cases if we want to be consistent.

BitwardenShared/UI/Platform/Application/AppProcessor.swift
```diff
@@ -652,6 +666,17 @@ public extension AppProcessor {
     }
 }
 
+// MARK: - AccountTokenProviderDelegate
+
+extension AppProcessor: AccountTokenProviderDelegate {
+    func onRefreshTokenError(error: any Error) async throws {
+        if let identityTokenError = error as? IdentityTokenRefreshRequestError,
+           identityTokenError == .invalidGrant {
```
matt-livefront: 🎨 This could be simplified slightly:
```suggestion
        if case IdentityTokenRefreshRequestError.invalidGrant = error {
```

BitwardenKit/Core/Platform/Services/API/Errors/ServerError.swift
```diff
@@ -4,7 +4,7 @@ import Foundation
 
 /// An enumeration of server errors.
 ///
-public enum ServerError: Error, Equatable, CustomNSError {
+public enum ServerError: NetworkingError, Equatable, CustomNSError {
```
KatherineInCode: 🤔 It's a little strange to me to consider a Server Error as a Networking Error. My mental model is that they're two different things.

BitwardenShared/Core/Platform/Services/API/APIService.swift
```diff
@@ -39,6 +39,8 @@ class APIService {
     /// Initialize an `APIService` used to make API requests.
     ///
     /// - Parameters:
+    ///   - accountTokenProvider: The `AccountTokenProvider` to use. This is helpful for testing.
+    ///   The default will be built by default.
```
KatherineInCode, line 43: 🎨 TIL parameter comments can be multiline

BitwardenShared/Core/Platform/Services/API/APIService.swift
```diff
@@ -107,4 +110,10 @@ class APIService {
             tokenProvider: accountTokenProvider
         )
     }
+
+    /// Sets up the account token provider delegate.
+    /// - Parameter delegate: The delegate to use.
+    func setupAccountTokenProviderDelegate(delegate: AccountTokenProviderDelegate) async {
```
KatherineInCode: 🎨 This doesn't seem to do any setup, just setting; following Matt's point elsewhere, I think this should be just `setAccountTokenProviderDelegate`. (Ideally it would actually just be a property that can be set, rather than a function, but it doesn't look like you can do that with an await. Sigh.)

If we want to keep it as "set up", though, "set up" is the verb form ("setup" is the noun form), so it should more properly be `setUpAccountTokenProviderDelegate`.

BitwardenKit/Core/Platform/Services/API/Errors/ServerError.swift
fedemkr: I could rename the protocol to something else, but that's how it was called in `Error+Networking` boolean check. Would something like `NonLoggableError` make more sense?

BitwardenShared/Core/Platform/Services/API/APIService.swift
fedemkr: Yeah, specially helpful with our line length restrictions hehe

BitwardenShared/Core/Platform/Services/API/APIService.swift
fedemkr: Already changed to `set ` thanks!

BitwardenKit/Core/Platform/Services/API/Errors/ServerError.swift
KatherineInCode: I do think `NonLoggableError` captures what it does a bit more explicitly—"these are errors that aren't logged". But likewise, I think it might make sense, then, to instead of having the loggers ask `isNetworkingError` they explicitly call `isLoggableError` or something like that. 🤔

fedemkr: Totally agree, making the change 😄

KatherineInCode (approved):

PR #1724, opened 2025-07-07
[PM-23496] Add Sourcery library to automatically generate mocks

fedemkr:
## 🎟️ Tracking

[PM-23496](https://bitwarden.atlassian.net/browse/PM-23496)

## 📔 Objective

Add [Sourcery](https://github.com/krzysztofzablocki/Sourcery) and configure it to be able to automatically generate mocks.
This library is powered by [Stencil](https://github.com/stencilproject/Stencil) and we have one that acts as a template for the generated mocks: [AutoMockable.stencil](https://github.com/bitwarden/ios/compare/PM-23496/add-sourcery?expand=1#diff-eb95026c2c22d4c9a725e0d36109f38587d026a92c902fa6807f7be568e06cc2).

In terms of organization I've split what we use globally as the stencils/templates in `root -> Sourcery` and then specific targets configurations in each target. For the time being I've only configured `BitwardenShared` and `AuthenticatorShared` which can be seen in `BitwardenShared -> Sourcery -> sourcery.yml` or alike for BWA.
Additionally, the generated mocks for `BitwardenShared` will be located in `BitwardenShared -> Sourcery -> Generated` folder in a single file called `AutoMockable.generated.swift`.

### Usage

Add `// sourcery: AutoMockable` comment to any protocol so the mock is automatically generated.
Contributing docs PR: https://github.com/bitwarden/contributing-docs/pull/631

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-23496]: https://bitwarden.atlassian.net/browse/PM-23496?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/Sourcery/sourcery.yml
+++ b/BitwardenShared/Sourcery/sourcery.yml
@@ -12,4 +12,7 @@ exclude:
   - Tests
   - TestHelpers
   - Fixtures
-  
+
+args:
+  autoMockableImports: ["BitwardenKit", "BitwardenSdk", "Combine"]
+  autoMockableTestableImports: ["BitwardenShared"]
--- a/Sourcery/Templates/AutoMockable.stencil
+++ b/Sourcery/Templates/AutoMockable.stencil
@@ -1,12 +1,15 @@
 // swiftlint:disable line_length
 // swiftlint:disable variable_name
 
-import BitwardenKit
-import BitwardenSdk
-import Combine
 import Foundation
 
-@testable import BitwardenShared
+{% for import in argument.autoMockableImports %}
+import {{ import }}
+{% endfor %}
+
+{% for import in argument.autoMockableTestableImports %}
+@testable import {{ import }}
+{% endfor %}
 
 {% macro swiftifyMethodName name %}{{ name | replace:"(","_" | replace:")","" | replace:":","_" | replace:"`","" | snakeToCamelCase | lowerFirstWord }}{% endmacro %}
 
--- a/AuthenticatorShared/Sourcery/sourcery.yml
+++ b/AuthenticatorShared/Sourcery/sourcery.yml
@@ -13,3 +13,6 @@ exclude:
   - TestHelpers
   - Fixtures
   
+args:
+  autoMockableImports: ["BitwardenKit", "BitwardenSdk", "Combine"]
+  autoMockableTestableImports: ["AuthenticatorShared"]


BitwardenShared/Sourcery/sourcery.yml
matt-livefront, line 1: This file is showing up in the app's bundle. I think you can prevent this by removing it from a build phase (by specifying it's path and setting `buildPhase: none ` similar to `swiftgen-pm.yml `) or by moving it to the root of the repo.
Screenshot 2025-07-07 at 2 01 47 PM

https://github.com/bitwarden/ios/blob/5ab1cf6de528cc7e806702c73dc00d416c18f923/project-pm.yml#L165-L166

Sourcery/Templates/AutoMockable.stencil
```diff
@@ -0,0 +1,133 @@
...
+        }
+    }
+    var {% call underlyingMockedVariableName variable %}: {{ variable.typeName }}!
+{% endmacro %}
+
+{% macro underlyingMockedVariableName variable %}underlying{{ variable.name|upperFirstLetter }}{% endmacro %}
+{% macro mockedVariableName variable %}{{ variable.name }}{% endmacro %}
+
+{% for type in types.protocols where type.based.AutoMockable or type|annotated:"AutoMockable" %}{% if type.name != "AutoMockable" %}
+class {{ type.name }}Mock: {{ type.name }} {
```
matt-livefront: 🤔 It looks like this names the mock class `<thing-to-mock>Mock`, but we've been using the convention `Mock<thing-to-mock>`. Do we want to update this to match our existing conventions to avoid any confusion?

matt-livefront (commented):
I ran into an issue where I needed to bootstrap after the mock file was initially created. I think this is a case where we'll want to specify the generated file in the sources for Xcodegen and mark it as optional. It looks like you did this for the `Generated` directory, but I'm not sure if that will apply for anything that's created in that directory after the fact. We may also need to specify it as an output file in the Sourcery build phase to get Xcode to correctly pick it up when it is initially generated. And we may want to add `basedOnDependencyAnalysis: false`.

```
diff --git a/project-pm.yml b/project-pm.yml
index e92140496..af86a9abc 100644
--- a/project-pm.yml
+++ b/project-pm.yml
@@ -398,6 +398,9 @@ targets:
             PATH="/opt/homebrew/bin:$PATH"
           fi
           mint run sourcery --config BitwardenShared/Sourcery/sourcery.yml
+        basedOnDependencyAnalysis: false
+        outputFiles:
+          - $(SRCROOT)/BitwardenShared/Sourcery/Generated/AutoMockable.generated.swift
   BitwardenSharedTests:
     type: bundle.unit-test
     platform: iOS
@@ -418,6 +421,8 @@ targets:
       - path: GlobalTestHelpers
       - path: BitwardenShared/Sourcery/Generated
         optional: true
+      - path: BitwardenShared/Sourcery/Generated/AutoMockable.generated.swift
+        optional: true
     dependencies:
       - target: Bitwarden
       - target: BitwardenShared
```


> Finally, in terms of CI I had to explicitly call Sourcery generation before bootstrapping in the `test.yml` workflow so the generated mocks are available to be used in the tests. Moreover, had to move installing Homebrew dependencies earlier so `Mint` is available for Sourcery.

If you make the changes above, I'd be curious if that also fixes CI without needing to make the additional workflow changes.

Sourcery/Templates/AutoMockable.stencil
fedemkr: I wasn't sure if we wanted to keep the distinction to know it's a mock being automatically generated but I believe you're correct and we should fix this so it's a more seamless integration on what we have now. I'll make the change thanks!

PR #1738, opened 2025-07-10
[PM-13607] Implement SDK client-managed state repository registration

fedemkr:
## 🎟️ Tracking

[PM-13607](https://bitwarden.atlassian.net/browse/PM-13607)

## 📔 Objective

Implement SDK client-managed state repository registration, particularly `BitwardenSdk.CipherRepository`.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-13607]: https://bitwarden.atlassian.net/browse/PM-13607?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenShared/Core/Vault/Repositories/SdkCipherRepository.swift
```diff
@@ -0,0 +1,46 @@
+import BitwardenKit
+import BitwardenSdk
+
+/// `CipherRepository` implementation to be used on SDK client-managed state.
+final class SdkCipherRepository: @unchecked Sendable, BitwardenSdk.CipherRepository {
```
matt-livefront: 🤔 Instead of `@unchecked Sendable` should we make this `@MainActor`? I think defaulting things to `@MainActor` will ultimately make the strict concurrency migration easier.

BitwardenShared/Core/Vault/Repositories/SdkCipherRepository.swift
```diff
@@ -0,0 +1,46 @@
...
+    /// The service used to manage syncing and updates to the user's ciphers.
+    let cipherService: CipherService
```
matt-livefront: 🤔 `CipherService` currently mediates between the API and data layers. Since the `CipherRepository` implementation seems like it's just interacting with the data layer, would it be better to use `CipherDataStore` here instead?

BitwardenShared/Core/Vault/Repositories/SdkCipherRepository.swift
```diff
@@ -0,0 +1,46 @@
...
+    /// - Parameters:
+    ///   - cipherService: The service used to manage syncing and updates to the user's ciphers.
+    ///   - errorReporter: The service used by the application to report non-fatal errors.
+    init(cipherService: CipherService, errorReporter: ErrorReporter) {
+        self.cipherService = cipherService
+        self.errorReporter = errorReporter
+    }
+
+    func get(id: String) async throws -> BitwardenSdk.Cipher? {
+        try await cipherService.fetchCipher(withId: id)
```
matt-livefront: 🤔 `cipherService.fetchCipher(withId:)` internally fetches the active user ID from the state. Since we're creating a `SdkCipherRepository` for each client (and user), should we assign a user ID to this repository and use that instead of relying on the active user ID? That might help prevent any issues where some data is inadvertently saved to the wrong user.

BitwardenShared/Core/Vault/Repositories/SdkCipherRepository.swift
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    func has(id: String) async throws -> Bool {
+        let cipher = try? await cipherService.fetchCipher(withId: id)
```
matt-livefront: 🤔 Since this method can throw, would it be better to throw any errors here rather than ignoring them? I don't think it will throw if the cipher doesn't exist.

BitwardenShared/Core/Vault/Repositories/SdkCipherRepository.swift
```diff
@@ -0,0 +1,46 @@
...
+    func remove(id: String) async throws {
+        try await cipherService.deleteCipherWithLocalStorage(id: id)
+    }
+
+    func set(id: String, value: BitwardenSdk.Cipher) async throws {
+        guard id == value.id else {
+            errorReporter.log(
+                error: BitwardenError.dataError("CipherRepository: Trying to update a cipher with mismatch IDs")
+            )
+            return
```
matt-livefront: 🤔 Maybe this isn't known yet since I think there's still ongoing work in the SDK around this, but would there be any ramifications of logging the error here and not adding or updating a value in the database? For example, the SDK thinks we saved the Cipher but really we didn't nor did we throw?

This probably shouldn't happen in practice, but I wonder if we should either 1) throw or 2) update our data store APIs to accept the `id` separately from the `Cipher`, matching the method signature here. Or maybe the SDK signature should change to ensure this error isn't possible? I'm not sure we need to make a change here right now, but something to consider as the SDK team (and us) builds out this functionality.

BitwardenShared/Core/Vault/Repositories/TestHelpers/MockSdkCipherRepository.swift
```diff
@@ -0,0 +1,42 @@
...
+        try hasResult.get()
+    }
+
+    func list() async throws -> [BitwardenSdk.Cipher] {
+        try listResult.get()
+    }
+
+    func remove(id: String) async throws {
+        if let removeError {
+            throw removeError
```
matt-livefront: 🎨 We could still use the `Result` pattern here for errors in methods that don't return a value. For example, `try removeResult.get()` and not return anything.

```
    var removeResult: Result<Void, Error> = .success(())

    func remove(id: String) async throws {
        removeReceivedId = id
        try removeResult.get()
    }
```

BitwardenShared/Core/Platform/Services/ClientService.swift
```diff
@@ -339,18 +345,39 @@ class DefaultClientBuilder: ClientBuilder {
...
+            Client(settings: settings)
+        }
+        guard userId != nil else {
+            return client
+        }
+
+        client
+            .platform()
+            .state()
+            .registerCipherRepository(store: sdkCipherRepository)
```
matt-livefront: 🤔 Do we want this builder to also do configuration? We have existing logic in `ClientService` which loads the feature flags for the newly created client:

https://github.com/bitwarden/ios/blob/e07b310605ba49a379d10145a0ae4bc18754eb94/BitwardenShared/Core/Platform/Services/ClientService.swift#L257-L259

Should we have all of this client configuration together? I'm not sure if it make sense to do that here vs in `ClientService`. On one hand, I like breaking it out of `ClientService`. On the other, this was originally created to be able to mock building the client and adding the `clientMaker` closure feels a little redundant to give a builder type a maker type 😄.

BitwardenShared/Core/Vault/Repositories/SdkCipherRepository.swift
fedemkr: Yes I think that should be fine, nice catch!

BitwardenShared/Core/Vault/Repositories/SdkCipherRepository.swift
fedemkr: I wouldn't put these methods under `@MainActor` as they could potentially block or delay other work that needs to happen in the main actor.
I think we could make it a plain `actor` though. What do you think?

BitwardenShared/Core/Vault/Repositories/SdkCipherRepository.swift
fedemkr: Wow, that's a really nice catch and will help prevent a lot of issues. Thanks for catching it. Also `SdkCipherRepository` is currently a singleton so only one is being created, it's just registering for each client. So we might need to do a factory for them as the Android approach with the user per each SDK instance.

♻️ Another approach would be for the SDK to send us the `userId` on each method call so we don't need to have multiple cipher repositories and can use the same for all users. Or some injected object we can query for the current user ID the SDK is using.

I think I can implement the factory for now and discuss the other options with the SDK team for ongoing/future repositories.
Thoughts?

BitwardenShared/Core/Vault/Repositories/SdkCipherRepository.swift
fedemkr: Android team brought this up as well, and the SDK team is looking into doing the validation in their side so it's not necessary in the platform. However, I agree we can just throw here so it's explicit that the `set(...)` method failed instead of wrapping the error and short-circuiting.

BitwardenShared/Core/Vault/Repositories/SdkCipherRepository.swift
matt-livefront: I don't think it needs to be an `actor` since it doesn't manage any shared state. I wouldn't be too concerned about putting it on the main actor since it's just passing data through and not doing any processing. Apple's stance seems to be to put things on the main actor by default and then move things off of it as and where needed. I do think long-term we could make this type be `Sendable`, so maybe `@unchecked Sendable` is ok for now. We just need to make sure we remove `@unchecked` when we can so that we don't lose out on any of the compiler checks once we adopt strict concurrency checking.

BitwardenShared/Core/Vault/Repositories/SdkCipherRepository.swift
matt-livefront: That seems like a good plan! I agree if the SDK could pass us `userId`, that might be a good solution so we don't need to have multiple instances.

BitwardenShared/Core/Platform/Services/ClientService.swift
fedemkr: 100% agree, I'll move the registration to the `ClientService`.

BitwardenShared/Core/Vault/Repositories/SdkCipherRepository.swift
fedemkr: Yeah agree, it's kind of confusing why UniFFI requires all types to be `Sendable` maybe it's something on the interface with Rust. Not sure which would be the right action here, if you feel `@MainActor` is safer I can make the change but feels a bit odd as well. @KatherineInCode what are your thoughts on this?
I could also just leave the warning and address this when we adopt strict concurrency...

KatherineInCode: I suspect the reason everything from Rust is Sendable has to do with [the concurrency assumptions of Rust](https://mozilla.github.io/uniffi-rs/0.28/udl/interfaces.html#concurrent-access)—and the way to communicate those assumptions as best as possible to Swift is as `Sendable` objects. It therefore lets us know that these objects need to be safe across concurrency boundaries (and marking `@MainActor` I believe more or less does that by preventing the object from being used anywhere but within one set of boundaries).

In this particular case, I agree that long-term this needs to be `Sendable`—but when we adopt strict concurrency we're going to have some relatively deep architectural changes to likely make because of needing to make a number of things `Sendable` that are currently relatively inimical to it. To that end, I think leaving it as a warning isn't necessarily a bad thing, since it's the sort of thing we'll have to go through and deal with eventually before moving to strict concurrency anyways.

BitwardenShared/Core/Platform/Services/ClientServiceTests.swift
```diff
@@ -215,6 +220,18 @@ final class ClientServiceTests: BitwardenTestCase { // swiftlint:disable:this ty
...
 
+    /// `client(for:)` registers the SDK cipher repository.
+    func test_client_registersCipherRepository() async throws {
+        stateService.activeAccount = .fixture(profile: .fixture(userId: "1"))
+
+        let auth = try await subject.auth()
+        let client = try XCTUnwrap(clientBuilder.clients.first)
+        XCTAssertIdentical(auth, client.authClient)
+        XCTAssertTrue(sdkRepositoryFactory.makeCipherRepositoryUserIdCalled)
+        XCTAssertTrue(sdkRepositoryFactory.makeCipherRepositoryUserIdCalled)
```
matt-livefront: ⛏️ Duplicate line, or maybe this should assert the user ID?
```suggestion
        XCTAssertTrue(sdkRepositoryFactory.makeCipherRepositoryUserIdCalled)
```

BitwardenShared/Core/Vault/Repositories/SdkCipherRepository.swift
fedemkr: Sounds good, I'll leave the warnings then so we can discuss how to fix this when we tackle pre-work to enable strict concurrency. Thanks for the feedback!

matt-livefront (approved):

PR #1744, opened 2025-07-10
[PM-23629] Restrict exported items based on policy

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-23629

## 📔 Objective

When restricted item types policy is enabled, the credit card item(s), which are hidden in the UI, should not be exported/shouldn’t be included in the json data.

## 📸 Screenshots

https://github.com/user-attachments/assets/8b923875-0d93-4bc2-bcf6-56fee8b45d68

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/BitwardenShared/Core/Vault/Services/PolicyServiceTests.swift
+++ b/BitwardenShared/Core/Vault/Services/PolicyServiceTests.swift
@@ -786,64 +786,6 @@ class PolicyServiceTests: BitwardenTestCase { // swiftlint:disable:this type_bod
         policyDataStore.fetchPoliciesResult = .success([.fixture(type: .restrictItemTypes)])
 
         let restrictedTypes = await subject.getRestrictedItemCipherTypes()
-        XCTAssertTrue(restrictedTypes.isEmpty)    
-    }
-    
-    /// `passesRestrictItemTypesPolicy(cipher:)` returns `true` when the `.removeCardPolicy` is enabled
-    /// and cipher passes restricted organization ids.
-    @MainActor
-    func test_passesRestrictItemTypesPolicy_cipherPassesRestrictedOrganizationIds() async {
-        configService.featureFlagsBool[.removeCardPolicy] = true
-        let result: Policy = .fixture(organizationId: "1", type: .restrictItemTypes)
-        stateService.activeAccount = .fixture()
-        organizationService.fetchAllOrganizationsResult = .success([.fixture(id: "1")])
-        policyDataStore.fetchPoliciesResult = .success([result])
-
-        let passes = await subject.passesRestrictItemTypesPolicy(cipher: .fixture(
-            organizationId: "2",
-            type: .card(.fixture())
-        ))
-        XCTAssertTrue(passes)
-    }
-
-    /// `passesRestrictItemTypesPolicy(cipher:)` returns `false` when the `.removeCardPolicy` is enabled
-    /// and cipher doesn't pass restricted organization ids.
-    @MainActor
-    func test_passesRestrictItemTypesPolicy_cipherOrganizationPartOfRestrictedOnes() async {
-        configService.featureFlagsBool[.removeCardPolicy] = true
-        let result: Policy = .fixture(organizationId: "2", type: .restrictItemTypes)
-        stateService.activeAccount = .fixture()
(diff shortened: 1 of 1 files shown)


BitwardenShared/Core/Vault/Services/ExportVaultServiceTests.swift
```diff
@@ -274,6 +274,85 @@ final class ExportVaultServiceTests: BitwardenTestCase {
         )
     }
 
+    /// Test that no ciphers are filtered when restrictedTypes is empty
+    ///
+    func test_fileContent_restrictedTypes_empty() async throws {
+        let fileType = ExportFileType.json
```
fedemkr: ⛏️ Is this variable necessary? Couldn't we just pass `ExportFileType.json` to the `format:` parameter? Same applies to other tests.

BitwardenShared/Core/Vault/Services/ExportVaultServiceTests.swift
fedemkr, line 1: 🎨 In order to be aligned to other tests in the suite, could you add the symbolic name of what's being tested in the test function docs? E.g. [in here](https://github.com/bitwarden/ios/blob/main/BitwardenShared/Core/Auth/Services/AuthServiceTests.swift#L98)

BitwardenShared/Core/Vault/Services/ExportVaultService.swift
```diff
@@ -38,10 +38,13 @@ protocol ExportVaultService: AnyObject {
...
     ///
-    /// - Parameter format: The format to use for vault export.
+    /// - Parameters:
+    ///   - format: The format of the exported file.
+    ///   - restrictedTypes: An array of `CipherType` that should be excluded from the export.
+    ///
     /// - Returns: A string representing the file content.
     ///
-    func exportVaultFileContents(format: ExportFileType) async throws -> String
+    func exportVaultFileContents(format: ExportFileType, restrictedTypes: [CipherType]) async throws -> String
```
fedemkr: 🤔 Instead of passing the `restrictedTypes` here, what do you think about having the `PolicyService` in the `DefultExportVaultService` and call it to get the restricted types. You could even have a function in `PolicyService` that returns the restricted types so you encapsulate the logic there, e.g. `func restrictedItemCipherTypes() -> [CipherType]`.
By doing this we can leave the processor simple and concentrate the logic in the services.

BitwardenShared/Core/Vault/Services/ExportVaultServiceTests.swift
andrebispo5: Yes for sure, this must be an older test class and I followed the other test cases already present and didn't noticed the difference, I'll update the rest of the class as well.

BitwardenShared/Core/Vault/Services/ExportVaultService.swift
andrebispo5: It took a bit of work to redo everything, but I really like your idea — it's much cleaner!

BitwardenShared/Core/Vault/Services/TestHelpers/MockPolicyService.swift
```diff
@@ -42,6 +42,13 @@ class MockPolicyService: PolicyService {
         policyAppliesToUserPolicies.map(\.organizationId)
     }
 
+    func getRestrictedItemCipherTypes() async -> [BitwardenShared.CipherType] {
+        guard policyAppliesToUserPolicies.isEmpty else {
+            return [.card]
+        }
+        return []
+    }
+
```
fedemkr: 🎨 IMO we shouldn't add any further logic to the method like checking if the policy applies or not. This should have a result variable the caller can set and is directly returned by this.
```swift
var getRestrictedItemCipherTypesResult: Result<[BitwardenShared.CipherType], Void>() = .success([])
...
func getRestrictedItemCipherTypes() async -> [BitwardenShared.CipherType] {
    getRestrictedItemCipherTypesResult
}
```
Doing it like this keeps the mocks really simple and you don't need to check the mock's implementation to get the logic in which something is returned.

fedemkr (commented):
LGTM, just a minor improvement. Also thanks for doing the refactor and improving the docs.

PR #1750, opened 2025-07-11
[PM-23406] Update AuthenticatorSyncService to use a separate SDK client

matt-livefront:
## 🎟️ Tracking

[PM-23406](https://bitwarden.atlassian.net/browse/PM-23406)

## 📔 Objective

Updates `AuthenticatorSyncService` to use a separate version of the SDK client for syncing TOTP codes to the authenticator app.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-23406]: https://bitwarden.atlassian.net/browse/PM-23406?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
+++ b/BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
@@ -359,6 +359,7 @@ actor DefaultAuthenticatorSyncService: NSObject, AuthenticatorSyncService {
                 email: account.profile.email,
                 privateKey: encryptionKeys.encryptedPrivateKey,
                 signingKey: nil,
+                securityState: nil,
                 method: .decryptedKey(decryptedUserKey: authenticatorKey)
             )
         )
--- a/BitwardenShared/Core/Platform/Services/ServiceContainer.swift
+++ b/BitwardenShared/Core/Platform/Services/ServiceContainer.swift
@@ -853,6 +853,10 @@ public class ServiceContainer: Services { // swiftlint:disable:this type_body_le
             clientBuilder: clientBuilder,
             configService: configService,
             errorReporter: errorReporter,
+            sdkRepositoryFactory: DefaultSdkRepositoryFactory(
+                cipherDataStore: dataStore,
+                errorReporter: errorReporter
+            ),
             stateService: stateService
         )
         let authenticatorSyncService = DefaultAuthenticatorSyncService(


BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
```diff
@@ -319,20 +318,57 @@ actor DefaultAuthenticatorSyncService: NSObject, AuthenticatorSyncService {
...
+    private func initializeOrganizationCrypto(userId: String) async throws {
+        let organizationKeysById = try await organizationService.fetchAllOrganizations(userId: userId)
+            .reduce(into: [String: String]()) { result, organization in
+                guard let key = organization.key else { return }
+                result[organization.id] = key
+            }
+        try await clientService.crypto().initializeOrgCrypto(
+            req: InitOrgCryptoRequest(organizationKeys: organizationKeysById)
+        )
+    }
```
fedemkr: 🤔 Unless I'm missing something this is the same as the one in `OrganizationService` but with a specific `userId`:
https://github.com/bitwarden/ios/blob/f4f2fe624c86c7948a8d341f0bb76136f168d6a7/BitwardenShared/Core/Vault/Services/OrganizationService.swift#L85-L107

What do you think about reusing the same as in `OrganizationService` but passing a `userId: String? = nil`? So if it's `nil` then it grabs the active user and if not it uses the parameter one which would be for this case.

BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift
```diff
@@ -319,20 +318,57 @@ actor DefaultAuthenticatorSyncService: NSObject, AuthenticatorSyncService {
...
+    ///
+    /// - Parameter userId: The userId the account being used to initialize the SDK.
+    ///
+    private func unlockVaultWithAuthenticatorVaultKey(userId: String) async throws {
+        let authenticatorKey = try await keychainRepository.getAuthenticatorVaultKey(userId: userId)
+
+        let account = try await stateService.getAccount(userId: userId)
+        let encryptionKeys = try await stateService.getAccountEncryptionKeys(userId: userId)
+
+        try await clientService.crypto().initializeUserCrypto(
```
fedemkr: 🤔 Unless I'm missing something this won't use a separate SDK client but [reuse](https://github.com/bitwarden/ios/blob/main/BitwardenShared/Core/Platform/Services/ClientService.swift#L253) the one for the current `userId` if it exists. Did you mean to create a new SDK client for a particular user? If so, a bit of a refactor is needed to have multiple SDK clients for the same user.

matt-livefront: The difference here is that I'm creating a separate `DefaultClientService` that is injected into `AuthenticatorSyncService` to manage the SDK clients for the sync service. Since this is separate from the rest of app, that's what ends up creating a separate SDK client. The advantage of this is that it's a minimal change. The disadvantage is that it's 1) not obvious and 2) we have to be careful about calling out to other services which will have a different client service. For reason 2) we have to duplicate the SDK unlock methods here as opposed to calling out to `AuthRepository` / `OrganizationService` to reuse that existing functionality. Given that I think the unlock mechanism for authenticator sync will end up changing, this seemed like a decent tradeoff vs designing some other way for `DefaultClientService` to vend out separate SDK instances. I can add some comments to make this a little clearer.

https://github.com/bitwarden/ios/blob/7e2127991bdd4eda11c8528c64da28ff993129c2/BitwardenShared/Core/Platform/Services/ServiceContainer.swift#L813-L822

fedemkr: Oh sorry missed that and thanks for the added comments!
🤔 I'm not quite sure if I'm 100% on board with this as is a bit hidden and not much scalable if we need to start duplicating logic for SDK client related logic. We'd need to be extra careful when reviewing changes to this file not to introduce any service/repository that directly/indirectly uses an SDK client.
Having said that I think it's a very minimal change as you said and the other way I can think of solving this would involve a lot of refactoring so I believe it's ok for now and we can revisit in the future if we start adding more functionality to this.
@KatherineInCode what do you think?

KatherineInCode: I think it's okay for now, too—though any and all comments of the "be careful here in the future" would be good. My understanding from [the Slack conversation](https://bitwarden.slack.com/archives/C095CNYSMMW/p1752155521267379?thread_ts=1752092703.065989&cid=C095CNYSMMW) is that if we make changes to how we handle keys in the future, we can update a lot of this to avoid the issue entirely.

fedemkr (dismissed):

PR #1751, opened 2025-07-11
Switch to Homebrew-installed Sourcery due to Xcode 26/macOS SDK build failures

bo2themax:
## Background

When running bitwarden on Xcode 26 (and previously on Xcode 16.4), the Mint-installed Sourcery fails to compile against the macOS SDK:

```bash
/private/var/folders/…/Sourcery/.build/checkouts/swift-llbuild/lib/Basic/Subprocess.cpp:89:12:
error: static declaration of ‘posix_spawn_file_actions_addchdir’ follows non-static declaration
89 | static int posix_spawn_file_actions_addchdir(…);
|            ^
/Applications/Xcode-beta.app/…/MacOSX26.0.sdk/usr/include/spawn.h:72:9: note: previous declaration is here
72 | int     posix_spawn_file_actions_addchdir(…);
|         ^
1 error generated.
```

This stems from a name collision in `swift-llbuild`’s `Subprocess.cpp` (declaring `static posix_spawn_file_actions_addchdir`) versus the same symbol in Apple’s SDK. Upstream hasn’t merged a fix yet, and it may be some time before it lands. https://github.com/krzysztofzablocki/Sourcery/issues/1425

## Benefits
- Stability: Homebrew provides a pre-built, tested binary that doesn’t trigger macOS SDK conflicts.
- Speed: Instant availability in CI without waiting for Mint to compile from source.
- Low maintenance: Avoided patching swift-llbuild ourselves.

## Alternatives
- Wait for upstream swift-llbuild PR to land and release a new Sourcery tag.


vvolkgang:
Hi! Thanks for bringing it to our attention. Given homebrew's [rolling release](https://docs.brew.sh/Brew-Bundle-and-Brewfile#versions) approach preventing us from pinning versions, for the time being we'll wait for the upstream fix.

PR #1764, opened 2025-07-17
[PM-18211] Add alert when tapping on a decryption failure item

matt-livefront:
## 🎟️ Tracking

[PM-18211](https://bitwarden.atlassian.net/browse/PM-18211)

## 📔 Objective

- Updates the SDK usages of `decryptList(ciphers:)` to `decryptListWithFailures(ciphers:)`
- Any ciphers which fail to decrypt will display in the vault list with a `[error: cannot decrypt]` name in the list that the cipher would have been in (i.e. this will show the decryption failure cipher within the folder, collection, or the cipher type group that it was in).
- Tapping on a cipher which failed to decrypt will show an alert informing the user to contact support and provide the cipher's ID.

## 📸 Screenshots

Screenshot 2025-07-15 at 12 02 07 PM

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-18211]: https://bitwarden.atlassian.net/browse/PM-18211?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenShared/Core/Vault/Extensions/CipherListView+Extensions.swift
```diff
@@ -18,3 +18,56 @@ extension CipherListView {
         return !restrictItemTypesOrgIds.contains(orgId)
     }
 }
+
+extension CipherListView {
+    var isDecryptionFailure: Bool {
+        name == Localizations.errorCannotDecrypt
+    }
+
+    init(cipherDecryptFailure cipher: Cipher) {
```
matt-livefront, line 27: ℹ️ In order to get this working, I'm taking the `Cipher`s which fail to decrypt and converting them to `CipherListView`s with a name of `[error: cannot decrypt]`. By converting them to `CipherListView`'s, I didn't have  to modify much of our existing logic that handles filtering and grouping ciphers to build the sections within the vault list. I think as an alternative to this, we could:

1. Move this logic into the SDK. Instead of having the SDK return successful `CipherListView`s and failed `Cipher`s, it could return all `CipherListView`s and we could differentiate with a `isDecryptionFailure` property on `CipherListView`.
2. Create a protocol (e.g. `CipherListItem`) which `CipherListView` and `Cipher` could conform to. And then update the filtering and grouping logic to use this protocol instead of `CipherListView` directly.

BitwardenShared/UI/Vault/Extensions/Alert+Vault.swift
```diff
@@ -4,6 +4,23 @@ import UIKit
...
+    ///
+    static func cipherDecryptionFailure(cipherId: String?) -> Alert {
+        Alert(
+            title: Localizations.decryptionError,
+            message: Localizations.bitwardenCouldNotDecryptTheVaultItemDescriptionLong + "\n\n" + (cipherId ?? ""),
+            alertActions: [
+                AlertAction(title: Localizations.close, style: .default),
+            ]
+        )
+    }
```
fedemkr: 🤔 Not sure if you've already talked it with Product but couldn't we add a "Copy" button or "Share" button in here so the user can share the `cipherId` with Customer Success easier if they need to?

BitwardenShared/Core/Vault/Extensions/CipherListView+Extensions.swift
```diff
@@ -18,3 +18,56 @@ extension CipherListView {
...
+        let type: CipherListViewType = switch cipher.type {
+        case .card:
+            .card(CardListView(brand: cipher.card?.brand))
```
fedemkr: ⛏️  `cipher.card?.brand` is encrypted so I think you can directly put `nil` in here as none of the brand logic would apply if it's not decrypted.

fedemkr (commented):
⚠️ `VaultRepository -> ciphersAutofillPublisher` in `.all` mode (when the user autofills from a field's context menu) will use the new strategies way of loading the ciphers in the iOS Autofill extension. So I think it's needed to add the tap item handling in the `VaultAutofillListProcessor` as well for non-decrypted items.
Unless there shouldn't be such non-decrypted items displayed in the Autofill extension ever; in that case we'd need to filter them out from there.

BitwardenShared/UI/Vault/Extensions/Alert+Vault.swift
matt-livefront: Good idea! I'll ask product about this.

matt-livefront:
@fedemkr I talked with product and added a copy button to the alerts and added the alert to `VaultAutofillListProcessor`.

PR #1848, opened 2025-08-11
[PM-24668] Update credit card numbers to display as groups of 4-digits

vgrassia:
## 📔 Objective

I'm not an iOS developer, but I figured I'd take a stab at fixing the formatting around credit card numbers when they are unmasked. This change shows the credit card number in 4-digit groups when unmasked. It is consistent with the functionality in our Browser extensions and Desktop apps.

- [Test Build](https://github.com/bitwarden/ios/actions/runs/16889032613)

## 📸 Screenshots

image

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


pamperer562580892423:
@vgrassia:

> I'm not an iOS developer, but I figured I'd take a stab at fixing the formatting around credit card numbers when they are unmasked. This change shows the credit card number in 4-digit groups when unmasked. It is consistent with the functionality in our Browser extensions and Desktop apps.

I do hope, you're also _not an Android developer_... if you know what I mean. 😉

fedemkr (changes requested):
👍 Nice job and thanks for updating this!
🎨 However, there's no need to create a new `CreditCardText` you can keep using the `PasswordText` and just add a get only property in the State that returns the `cardNumber` applying `formattedCreditCardNumber()` which then you can use in the `password` argument of the view.

BitwardenShared/UI/Platform/Application/Extensions/String.swift
```diff
@@ -137,4 +137,28 @@ extension String {
     func withoutAutomaticEmailLinks() -> String {
         replacingOccurrences(of: "@", with: "\u{2060}@")
     }
+
+    /// Formats a credit card number by inserting spaces every 4 digits.
+    /// Only formats if the string contains only digits, otherwise returns the original string.
+    ///
+    /// - Returns: A formatted credit card number with spaces every 4 digits, or the original string if not a valid card number.
+    ///
+    func formattedCreditCardNumber() -> String {
```
KatherineInCode, line 146: 🤔 This assumes 16-digit card numbers that are broken into four 4-digit chunks. However, American Express cards are 15-digits broken up as 4-6-5. Is this something we're concerned with here?

fedemkr: Nice catch! I think we'd need to discuss this with Product. However, even though it's not how it's expected to be presented, the way it's done in here aligns with how other clients are handling this as far as I know.
Additionally, I'd move the formatting logic for each card to the SDK so it's centralized for all clients.

BitwardenShared/UI/Platform/Application/Views/PasswordText.swift
```diff
@@ -45,9 +45,9 @@ struct PasswordText: View {
                 SharedAsset.Colors.textPrimary.swiftUIColor
             }
 
-            // Add a zero-width space (U+200B) after each character to ensure text will wrap on any
-            // character boundary.
-            var characterString = AttributedString("\(character)\(String.zeroWidthSpace)")
+            // For spaces, don't add zero-width space to preserve formatting
```
KatherineInCode: 🤔 I don't suppose there's a way we could have a test for this change?

vgrassia: I removed this change as it wasn't necessary after testing.

BitwardenShared/UI/Platform/Application/Extensions/String.swift
vgrassia: Where exactly would this go in the SDK? `crates/bitwarden-vault/src/cipher/card.rs`?

fedemkr: I don't know, we'd need to discuss and plan it with the SDK team; as it could be just an additional property in the decrypted cipher view returned by the SDK or some function we can call on the clients helpers or something else I'm not thinking right now.
However, for the time being this change is fine IMO.

Gerardv514:
If the card number is copied and then pasted, does it maintain the spacing? What if it's pasted into a form field that will accept spacing or not accept spacing.

Will the pasting be dynamic?

fedemkr:
Hi @Gerardv514 the copy button gets [the value without spacing](https://github.com/bitwarden/ios/pull/1848/files#diff-79819f10fceb3f4e98ed2edba0464481e275c180232e2c8fb0aaf47a5d87ccaaR63) to be copied, so this is only visual on the current view field.

KatherineInCode:
@vgrassia 🤔 It looks like you might need to update some of the screenshots for tests, as well

vgrassia:
> @vgrassia 🤔 It looks like you might need to update some of the screenshots for tests, as well

Updated, thanks for pointing that out.

KatherineInCode (approved):

PR #1857, opened 2025-08-14
[PM-24715] Fix iOS autofill extension app initialization URLs

fedemkr:
## 🎟️ Tracking

[PM-24715](https://bitwarden.atlassian.net/browse/PM-24715)

## 📔 Objective

🍒 Same change as in RC fix #1858 

Fix iOS autofill extension app initialization to load environment URLs and config when going through the flows without user interaction. This was causing the app to logout when session timeout was set to Never, Region not set to US and doing any API calls in the flows; like in certain Fido2 autofill scenarios. So instead of performing the API call to the correct server it was defaulting to US, returning a 401 and then when refreshing a 400 with `invalid_grant` logging out the user.

Additionally, needed to change the flows without user interaction be async so we can load the URLs and config correctly.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-24715]: https://bitwarden.atlassian.net/browse/PM-24715?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenAutoFillExtension/CredentialProviderViewController.swift
+++ b/BitwardenAutoFillExtension/CredentialProviderViewController.swift
@@ -68,10 +68,12 @@ class CredentialProviderViewController: ASCredentialProviderViewController {
             return
         }
 
-        initializeApp(
-            with: DefaultCredentialProviderContext(.autofillCredential(credentialIdentity, userInteraction: false))
-        )
-        provideCredential(for: recordIdentifier)
+        Task {
+            await initializeAppWithoutUserInteraction(
+                with: DefaultCredentialProviderContext(.autofillCredential(credentialIdentity, userInteraction: false))
+            )
+            provideCredential(for: recordIdentifier)
+        }
     }
 
     @available(iOSApplicationExtension 17.0, *)
@@ -82,12 +84,14 @@ class CredentialProviderViewController: ASCredentialProviderViewController {
                 provideCredentialWithoutUserInteraction(for: passwordIdentity)
             }
         case let passkeyRequest as ASPasskeyCredentialRequest:
-            initializeApp(
-                with: DefaultCredentialProviderContext(
-                    .autofillFido2Credential(passkeyRequest, userInteraction: false)
+            Task {
+                await initializeAppWithoutUserInteraction(
+                    with: DefaultCredentialProviderContext(
+                        .autofillFido2Credential(passkeyRequest, userInteraction: false)
+                    )
                 )
-            )
-            provideFido2Credential(for: passkeyRequest)
+                provideFido2Credential(for: passkeyRequest)
+            }
         default:
             if #available(iOSApplicationExtension 18.0, *),
                let otpRequest = credentialRequest as? ASOneTimeCodeCredentialRequest,
@@ -161,6 +165,15 @@ class CredentialProviderViewController: ASCredentialProviderViewController {
                 await appProcessor.start(appContext: .appExtension, navigator: self, window: nil)
             }
         }
+    }
+    
+    /// Sets up and initializes the app without user interaction.
+    /// - Parameter context: The context that describes how the extension is being used.
+    private func initializeAppWithoutUserInteraction(
+        with context: CredentialProviderContext
+    ) async {
+        initializeApp(with: context)
+        await appProcessor?.prepareEnvironmentConfig()
     }
 
     /// Attempts to provide the credential with the specified ID to the extension context to handle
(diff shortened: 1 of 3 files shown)


BitwardenAutoFillExtension/CredentialProviderViewController.swift
```diff
@@ -162,6 +166,15 @@ class CredentialProviderViewController: ASCredentialProviderViewController {
...
         }
     }
+    
+    /// Sets up and initializes the app without user interaction.
+    /// - Parameter context: The context that describes how the extension is being used.
+    private func initializeAppWithoutUserInteraction(
+        with context: CredentialProviderContext
+    ) async {
+        initializeApp(with: context)
+        await appProcessor?.prepareEnvironmentConfig()
```
matt-livefront, line 176: 👍 Nice find on this!

🤔 Instead of creating a separate `initializeAppWithoutUserInteraction` function, since we already have logic in `initializeApp(with:)` which handles the user interaction case, would it make sense to combine these and handle the non-interactive case in an else block?

https://github.com/bitwarden/ios/blob/4c74bca517825d05f10bedca5bff679cb7ef4a46/BitwardenAutoFillExtension/CredentialProviderViewController.swift#L163-L167

I think we originally added the task within `initializeApp(with:)` to avoid having to create the task for each `initializeApp` call, but seeing how that has become problematic, maybe we're better off making it async and moving the Task creation outside of `initializeApp `?

BitwardenShared/UI/Platform/Application/AppProcessor.swift
```diff
@@ -183,8 +183,7 @@ public class AppProcessor {
         )
 
         await services.migrationService.performMigrations()
-        await services.environmentService.loadURLsForActiveAccount()
-        _ = await services.configService.getConfig()
+        await prepareEnvironmentConfig()
```
matt-livefront, line 186: 🤔 If there's an app upgrade, and you open the extension using one of the non-interactive flows prior to running the app, we'd skip performing any migrations here, which could be problematic. Would it be better to pass a flag to `AppProcessor.start()` indicating if user interaction is available as opposed to breaking out the environment config pieces? It's maybe more branching, but might be better for maintainability long-term to ensure `start()` is always called.

BitwardenAutoFillExtension/CredentialProviderViewController.swift
fedemkr: Thanks! 
I thought about that as well but given that this is an RC fix, I wanted something which modified as little as possible the code with minimal side effects on other flows to decrease QA work to cover the changes.
I will certainly add a ♻️ task to review this and totally agree we should just make the whole initialization asynchronous.
Would you be ok with this as is for now?

BitwardenShared/UI/Platform/Application/AppProcessor.swift
fedemkr: Nice catch!
I think that if migrations are needed I would throw an error in the flow and ask the user to open the main app to perform such migrations. I feel it's safer to run them in the main app process given the restrictions we have in the autofill extension process and that it may run in the background where several things are omitted/break.
Would you agree? If so, I'll create a ♻️ task for it and we can handle that in a future PR.

Regarding the `start()` being called for everything, I would very much like to refactor that part of the iOS extensions and the main app so we don't have that many branching for maintainability but I wouldn't do that as part of this PR; as it goes beyond scope IMO.

BitwardenAutoFillExtension/CredentialProviderViewController.swift
fedemkr: Task created here: https://bitwarden.atlassian.net/browse/PM-24773

matt-livefront: Yep, sounds good!

BitwardenShared/UI/Platform/Application/AppProcessor.swift
matt-livefront: I'm kind of torn on migrations. I agree that it would be safer to do them in the main app, but it kind of depends on the migration. Given that users are more likely to interact with the autofill extension than the main app, it would be annoying to be interrupted from an autofill flow and forced to open the main app, especially if the migration was for something minor which could easily be done in the extension. 🤔

fedemkr: I totally see your point and will create a Jira task for us to keep discussing it.

PR #1873, opened 2025-08-21
[PM-21862] Provide SDK with account keys

fedemkr:
## 🎟️ Tracking

PM-21862

## 📔 Objective

Implement passing AEAD (Authenticated Encryption with Associated Data) keys to the SDK crypto initialization and storing them.

> [!WARNING]  
> Ongoing work, missing tests and also missing crypto re-initialization from syncing responses.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
+++ b/BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
@@ -604,6 +604,7 @@ extension DefaultAuthRepository: AuthRepository {
 
         try await stateService.setAccountEncryptionKeys(
             AccountEncryptionKeys(
+                accountKeys: nil,
                 encryptedPrivateKey: registrationKeys.privateKey,
                 encryptedUserKey: nil
             )
@@ -802,6 +803,7 @@ extension DefaultAuthRepository: AuthRepository {
         resetPasswordAutoEnroll: Bool
     ) async throws {
         let account = try await stateService.getActiveAccount()
+        let accountPrivateKeys: PrivateKeysResponseModel?
         let email = account.profile.email
         let kdf = account.kdf
         let requestUserKey: String
@@ -816,8 +818,12 @@ extension DefaultAuthRepository: AuthRepository {
             requestPasswordHash = passwordResult.passwordHash
             requestUserKey = passwordResult.newKey
             requestKeys = nil
+            accountPrivateKeys = accountKeys.accountKeys
             encryptedPrivateKey = accountKeys.encryptedPrivateKey
         } else {
+            let accountKeys = try? await stateService.getAccountEncryptionKeys()
+            accountPrivateKeys = accountKeys?.accountKeys
+
             let keys = try await clientService.auth().makeRegisterKeys(
                 email: email,
                 password: password,
@@ -848,6 +854,7 @@ extension DefaultAuthRepository: AuthRepository {
 
         try await accountAPIService.setPassword(requestModel)
         try await stateService.setAccountEncryptionKeys(AccountEncryptionKeys(
+            accountKeys: accountPrivateKeys,
             encryptedPrivateKey: encryptedPrivateKey,
             encryptedUserKey: requestUserKey
         ))
@@ -1084,15 +1091,9 @@ extension DefaultAuthRepository: AuthRepository {
         let encryptionKeys = try await stateService.getAccountEncryptionKeys()
 
         try await clientService.crypto().initializeUserCrypto(
-            req: InitUserCryptoRequest(
-                userId: account.profile.userId,
-                kdfParams: account.kdf.sdkKdf,
-                email: account.profile.email,
-                privateKey: encryptionKeys.encryptedPrivateKey,
-                signingKey: nil,
-                securityState: nil,
-                method: method
-            )
+            account: account,
+            encryptionKeys: encryptionKeys,
+            method: method
         )
 
         switch method {
@@ -1146,6 +1147,7 @@ extension DefaultAuthRepository: AuthRepository {
 
         let encryptionKeys = try await stateService.getAccountEncryptionKeys()
         let newEncryptionKeys = AccountEncryptionKeys(
+            accountKeys: encryptionKeys.accountKeys,
             encryptedPrivateKey: encryptionKeys.encryptedPrivateKey,
             encryptedUserKey: updatePasswordResponse.newKey
         )
--- a/BitwardenShared/Core/Auth/Services/KeychainRepository.swift
+++ b/BitwardenShared/Core/Auth/Services/KeychainRepository.swift
@@ -1,4 +1,5 @@
 import BitwardenKit
+import BitwardenSdk
 import Foundation
 
 // MARK: - KeychainItem
@@ -7,6 +8,9 @@ import Foundation
 enum KeychainItem: Equatable {
     /// The keychain item for a user's access token.
     case accessToken(userId: String)
+
+    /// The keychain item for a user's account security state.
+    case accountSecurityState(userId: String)
 
     /// The keychain item for a user's vault key for Authenticator syncing.
     case authenticatorVaultKey(userId: String)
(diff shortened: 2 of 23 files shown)


BitwardenShared/Core/Platform/Utilities/KeyManagementTypes.swift
```diff
@@ -0,0 +1,16 @@
+import BitwardenSdk
+
+/// A private key, encrypted with a symmetric key.
+typealias WrappedPrivateKey = EncString
+
+/// A public key, signed with the accounts signature key.
+typealias SignedPublicKey = String
+
+/// A public key in base64 encoded SPKI-DER.
+typealias UnsignedPublicKey = [UInt8]
```
quexten: Question: Should this be string? I'm not exactly sure how the conversion from response model to bytes works in swift.

BitwardenShared/Core/Platform/Services/CryptoClient.swift
```diff
@@ -0,0 +1,29 @@
...
+        account: Account,
+        encryptionKeys: AccountEncryptionKeys,
+        method: InitUserCryptoMethod
+    ) async throws {
+        try await initializeUserCrypto(
+            req: InitUserCryptoRequest(
+                userId: account.profile.userId,
+                kdfParams: account.kdf.sdkKdf,
+                email: account.profile.email,
+                privateKey: encryptionKeys.encryptedPrivateKey,
```
quexten: Nit (non-blocking): This should probably already use the accountKeys.publicKeyEncryptionKeyPair.wrappedPrivateKey if available. These keys are the same, but I imagine the local state will in the future drop encryptionKeys.encryptedPrivateKey?

BitwardenShared/Core/Platform/Services/KeyConnectorService.swift
```diff
@@ -131,6 +131,7 @@ extension DefaultKeyConnectorService: KeyConnectorService {
 
         try await stateService.setAccountEncryptionKeys(
             AccountEncryptionKeys(
+                accountKeys: nil,
```
quexten, line 134: Nit (not-blocking): I think we can already store privateKeyResponseModel here by constructing it from the existing private key. That would allow us to drop the encryptedPrivateKey property quicker.

BitwardenShared/Core/Auth/Services/KeychainRepository.swift
```diff
@@ -8,6 +9,9 @@ enum KeychainItem: Equatable {
     /// The keychain item for a user's access token.
     case accessToken(userId: String)
 
+    /// The keychain item for a user's account security state.
+    case accountSecurityState(userId: String)
```
quexten: AccountSecurityState and UserSigningKey don't need to be protected by the keychain. As far as I understand, we don't store the userPrivateKey here either. They should be stored the same way as userPrivateKey.

quexten (commented):
Nice work!

Some nits, and a question around the keychain. I don't think we need to store the signing key and the security state in the keychain. The security state is signed, by the signing key, and the signing key is encrypted by the userkey, and they are safe to store in the same way as the user private key, which - as far as I can tell - is not in the keychain.

BitwardenShared/Core/Platform/Services/CryptoClient.swift
fedemkr: It's already using that as when [initializing](https://github.com/bitwarden/ios/pull/1873/files#diff-a821c10a939658c3cf6af3c3e884bfa5b12b2beba49b2ee2f6e1a7487916a8e6R22) the `AccountEncryptionKeys` I set the `encryptedPrivateKey` to the one in `accountKeys` if present, and to `privateKey` otherwise.
Would we want to maintain them separate for some reason or can I still reuse the same property for both private keys depending on the context?

BitwardenShared/Core/Platform/Utilities/KeyManagementTypes.swift
fedemkr: It should work but we could just try it when we have the server part deployed somewhere I can target and check the response. I also could try locally if it's urgent to know.
Additionally, if you have a serialized json response I could use, I can quickly check this in a test.

BitwardenShared/Core/Platform/Services/CryptoClient.swift
quexten: I think in the medium term we can drop `encryptionKeys.encryptedPrivateKey` entirely, since it's the same as `accountKeys.publicKeyEncryptionKeyPair.wrappedPrivateKey `

fedemkr: Yes we can definitely drop in a future PR given that now we'd need to do a migration in order to move the stored `encryptionKeys.encryptedPrivateKey` to `accountKeys.publicKeyEncryptionKeyPair.wrappedPrivateKey` and we wouldn't gain much now.

BitwardenShared/Core/Vault/Services/API/Fixtures/syncWithAccountKeysV2Profile.json
```diff
@@ -0,0 +1,44 @@
...
+    "premiumFromOrganization": false,
+    "masterPasswordHint": null,
+    "culture": "en-US",
+    "twoFactorEnabled": false,
+    "key": "key",
+    "privateKey": "private key",
+    "accountKeys": {
+      "publicKeyEncryptionKeyPair": {
+        "wrappedPrivateKey": "WRAPPED_PRIVATE_KEY",
+        "publicKey": [1, 2, 3, 4, 5],
```
quexten: Question / note: This is a base64 encoded string on the server https://github.com/bitwarden/server/blob/8a39481fec87848e19cdd92d620d3e23b97901cb/src/Api/KeyManagement/Models/Requests/PublicKeyEncryptionKeyPairRequestModel.cs#L9. Is it intended to be an array of bytes here?

quexten (commented):
Changes look reasonable. My only question is around the type of public key in the state here. I'm not sure if this is intended.

BitwardenShared/Core/Vault/Services/API/Fixtures/syncWithAccountKeysV2Profile.json
fedemkr: Nice catch! You're right, I saw an object in the clients repo with that type and I mistakenly applied it here directly missing it was being converted from a string in the response. Fixed and removed the `UnsignedPublicKey` typealias as it's no longer use.

BitwardenShared/Core/Vault/Models/Response/KeyManagement/PublicKeyEncryptionKeyPairResponseModel.swift
```diff
@@ -0,0 +1,12 @@
...
+///
+struct PublicKeyEncryptionKeyPairResponseModel: Codable, Equatable {
+    /// The wrapped private key.
+    let wrappedPrivateKey: WrappedPrivateKey
+
+    /// The public key.
+    let publicKey: String
+
+    /// The signed public key.
+    let signedPublicKey: SignedPublicKey?
```
matt-livefront: ⛏️ Do we want to alphabetize these similar to other models?

BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
```diff
@@ -816,8 +818,12 @@ extension DefaultAuthRepository: AuthRepository {
             requestPasswordHash = passwordResult.passwordHash
             requestUserKey = passwordResult.newKey
             requestKeys = nil
+            accountPrivateKeys = accountKeys.accountKeys
             encryptedPrivateKey = accountKeys.encryptedPrivateKey
         } else {
+            let accountKeys = try? await stateService.getAccountEncryptionKeys()
+            accountPrivateKeys = accountKeys?.accountKeys
```
matt-livefront: 🤔 Do you think this would end up throwing in the expected use cases? Would it better to use `try` instead of `try?` here? If so, we could also move getting the encryption keys outside if the `if` since we do that in both cases.

BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
```diff
@@ -604,6 +604,7 @@ extension DefaultAuthRepository: AuthRepository {
 
         try await stateService.setAccountEncryptionKeys(
             AccountEncryptionKeys(
+                accountKeys: nil,
```
matt-livefront, line 607: 🤔 Is it possible to construct a `PrivateKeysResponseModel` here with the keys we have or does the server need to do that? If I'm understanding correctly, I think this is similar to https://github.com/bitwarden/ios/pull/1873#discussion_r2293886040.

fedemkr: I think it could be partially possible as I'm not sure if we'd always have all the necessary things to do that. Like the `securityState`. Also there are parts that check whether there are `accountKeys` so I think it could be safer to have `accountKeys` only when the server sends it.
Additionally, we may need migrations to use this instead of the `encryptedPrivateKey` if we'd like to drop it.
Thoughts?

matt-livefront: That sounds reasonable to me! I wasn't sure, so I just wanted to check, especially after seeing the other comment.

PR #1753, opened 2025-07-11
[PM-13621] ci: Add cron job to regularly update SPM dependencies in project files

KatherineInCode:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-13621

## 📔 Objective

This creates a new cron job in the pattern of our "update public suffix list" job that checks the SPM packages defined in our project files for updates, and if they exist creates a PR for it. The `update-dependencies.sh` script can also be run offline if desired.

This should hopefully let us keep better tabs on those updates, and keep our dependencies up to date.

This has also been an opportunity for me to experiment with Claude Code, in terms of writing the bash script.

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

PR #1877, opened 2025-08-22
[PM-20288] New workflow to update the sdk

LRNcardozoWDF:
## 🎟️ Tracking

PM-20288

## 📔 Objective

Adding a new workflow able to update the sdk to a given version and open a PR. This will be used by the sdk-internal repo when PRs are merged. The ability to test SDK PRs will be introduced in a future PR.

Based on https://github.com/bitwarden/android/pull/5742, with the addition of the downgrade detection.

Test runs:

* Happy path - Update to new version with success https://github.com/bitwarden/ios/actions/runs/17439403637
* Error state - Updating to the same version in `main` - https://github.com/bitwarden/ios/actions/runs/17435544778
*. Downgrade detection - https://github.com/bitwarden/ios/pull/1904

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/.github/workflows/sdlc-sdk-update.yml
+++ b/.github/workflows/sdlc-sdk-update.yml
@@ -146,6 +146,28 @@ jobs:
           echo "sdk-swift-ref=$SDK_SWIFT_REF" >> $GITHUB_OUTPUT
           echo "sdk-internal-ref=$SDK_INTERNAL_REF" >> $GITHUB_OUTPUT
 
+      - name: Detect downgrade and prevent updating to the current version
+        id: detect-downgrade
+        env:
+          GH_TOKEN: ${{ steps.app-token.outputs.token }}
+          _CURRENT_SDK_SWIFT_REF: ${{ steps.get-current-sdk.outputs.sdk-swift-ref }}
+          _NEW_SDK_SWIFT_REF: ${{ inputs.sdk-swift-ref }}
+        run: |
+          if [ "$_CURRENT_SDK_SWIFT_REF" = "$_NEW_SDK_SWIFT_REF" ]; then
+            echo "::error::Provided sdk-swift ref is the same as the current version in main."
+            exit 1
+          fi
+
+          COMPARE_RESULT=$(gh api "repos/bitwarden/sdk-swift/compare/$_CURRENT_SDK_SWIFT_REF...$_NEW_SDK_SWIFT_REF" --jq '.status')
+
+          if [ "$COMPARE_RESULT" = "behind" ]; then
+            echo "::warning::The new SDK version ($_NEW_SDK_SWIFT_REF) is older than the current version ($_CURRENT_SDK_SWIFT_REF)"
+            echo "downgrading=true" >> $GITHUB_OUTPUT
+          else
+            echo "✅ New SDK version is newer - proceeding with update"
+            echo "downgrading=false" >> $GITHUB_OUTPUT
+          fi
+
       - name: Update SDK Version
         env:
           _SDK_VERSION: ${{ inputs.sdk-version }}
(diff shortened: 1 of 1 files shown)


Scripts/update-sdk-version.sh
```diff
@@ -0,0 +1,36 @@
...
+FILES=(
+  "project-bwa.yml"
+  "project-bwk.yml"
+  "project-pm.yml"
+)
+
+for file in "${FILES[@]}"; do
+    if [[ -f "$file" ]]; then
+        echo "🔧 Updating revision in $file..."
+        sed -i.bak -E "/^packages:/,/^[^[:space:]]/ {
```
KatherineInCode: 🤔 As with [my PR for updating all SPM dependencies](https://github.com/bitwarden/ios/pull/1753), instead of using `sed` you should be able to use `yq` to update the YAML.

.github/workflows/sdlc-sdk-update.yml
```diff
@@ -0,0 +1,165 @@
...
+        default: Test
+      #
+      sdk-package:
+        description: "SDK Package ID"
+        required: true
+        default: "BitwardenSdk"
+      sdk-version:
+        description: "SDK Version"
+        required: true
+        default: "2a6609428275c758fcda5383bfb6b3166ec29eda"
```
fedemkr: 🤔 By having a default, couldn't it misfire and go back to this revision instead of updating to the one in fact needed? IMO this should be `required` with no default so every time this flow is executed a revision must be passed.

Scripts/update-sdk-version.sh
```diff
@@ -0,0 +1,32 @@
...
+    echo "Example: $0 BitwardenSdk 2a6609428275c758fcda5383bfb6b3166ec29eda"
+    exit 1
+fi
+
+SDK_PACKAGE="$1"
+SDK_VERSION="$2"
+FILES=(
+  "project-bwa.yml"
+  "project-bwk.yml"
+  "project-pm.yml"
```
fedemkr: 🎨 I think we should take this opportunity to centralize the SDK revision in another file which these three can reference; so then we just need to update one place.

KatherineInCode: That is, ostensibly, covered by [PM-19516](https://bitwarden.atlassian.net/browse/PM-19516)

fedemkr: Indeed but it's a small change so perhaps we could do it right here; which would also simplify this script.

KatherineInCode: I don't disagree, it just might be worth updating that ticket to that effect as well (and/or marking this PR as handling that, if it pulls out more than just the SDK into that common yml)

vvolkgang: That will slightly change how some of the bits in this PR are handled so I opened the following separately to sort it: https://github.com/bitwarden/ios/pull/1899

.github/workflows/sdlc-sdk-update.yml
vvolkgang: (1) Defaults helps us test these workflows and (2) they communicate the type of value / format we expect - GitHub unfortunately doesn't handle long descriptions gracefully nor provide a way to set tooltip like instructions.

Given that we control when this is executed (either automatically or manually) and that it will just create a PR which has to be reviewed, providing the wrong inputs isn't much of a concern. Nonetheless I added some bits to detect downgrades and communicate them in the PR title / description: https://github.com/bitwarden/ios/pull/1905

.github/workflows/sdlc-sdk-update.yml
```diff
@@ -1,16 +1,236 @@
...
+        run: |
+          SDK_SWIFT_REF=$(git show origin/main:project-common.yml | yq '.packages.BitwardenSdk.revision')
+          if [ -z "$SDK_SWIFT_REF" ]; then
+            echo "::error::Failed to get current SDK version from main branch."
+            exit 1
+          fi
+
+          echo "👀 sdk-swift ref: $SDK_SWIFT_REF"
+
+          COMMIT_MESSAGE=$(gh api "repos/bitwarden/sdk-swift/commits/$SDK_SWIFT_REF" --jq '.commit.message')
```
fedemkr, line 135: 🤔 Currently we're using the `unstable` branch of the `sdk-swift` repo to update the revision. Are we changing to `main` now?

PR #1906, opened 2025-09-03
[PM-24115] New account switcher on iOS 26

KatherineInCode:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-24115

## 📔 Objective

This creates a new account switcher bottom sheet that is used on iOS 26 devices, and moves away from having an overlay view. Devices still on iOS <26 will still see the old overlay view, however.

Things left to do on this draft:
- Backfill tests
- Some accessibility concerns with the sheet at large font sizes
- Some other potential visual tweaks
- Other cleanup of dead ends and such detritus

However, it made sense to get a draft out there now to let what I have be reviewed in case there are any major things that need changed, as that could greatly impact the remaining items.

## 📸 Screenshots

Simulator Screenshot - iPhone 16 Pro - 2025-09-02 at 13 18 02

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/BitwardenShared/UI/Auth/ProfileSwitcher/ProfileSwitcherSheet.swift
+++ b/BitwardenShared/UI/Auth/ProfileSwitcher/ProfileSwitcherSheet.swift
@@ -12,6 +12,7 @@ public struct ProfileSwitcherSheet: View {
 
             SectionView(Localizations.selectAccount) {
                 ProfileSwitcherAccountsView(store: store)
+                    .clipShape(RoundedRectangle(cornerRadius: 8))
                 if store.state.showsAddAccount {
                     addAccountRow
                 }


BitwardenShared/UI/Auth/ProfileSwitcher/ProfileSwitcherCoordinator.swift
```diff
@@ -0,0 +1,69 @@
...
+
+    func navigate(to route: ProfileSwitcherRoute, context: AnyObject?) {
+        switch route {
+        case .dismiss:
+            stackNavigator?.dismiss()
+        case .open:
+            let processor = ProfileSwitcherProcessor(coordinator: asAnyCoordinator(),
+                                                     handler: handler,
+                                                     services: services,
+                                                     state: handler.profileSwitcherState)
```
matt-livefront: ⛏️ For consistency, I think it's more common that we would indent these properties a single tab width:
```suggestion
            let processor = ProfileSwitcherProcessor(
                coordinator: asAnyCoordinator(),
                handler: handler,
                services: services,
                state: handler.profileSwitcherState
            )
```

BitwardenShared/UI/Auth/ProfileSwitcher/ProfileSwitcherHandler.swift
```diff
@@ -160,7 +183,11 @@ private extension ProfileSwitcherHandler {
     /// - Parameter account: The `ProfileSwitcherItem` long pressed by the user.
     ///
     func didLongPressProfileSwitcherItem(_ account: ProfileSwitcherItem) async {
-        profileSwitcherState.isVisible = false
+        if #available(iOS 26, *) {
+            dismissProfileSwitcher()
+        } else {
+            profileSwitcherState.isVisible = false
+        }
```
matt-livefront: 🎨 Seeing this repeated a few times, would it be worth extracting the show/dismiss of the profile switcher to a separate method and then handling the iOS 26 vs pre-iOS 26 behavior within it?

BitwardenShared/UI/Auth/ProfileSwitcher/ProfileSwitcherProcessor.swift
```diff
@@ -0,0 +1,47 @@
...
+
+    // MARK: Private Properties
+
+    /// The `Coordinator` that handles navigation.
+    private let coordinator: AnyCoordinator<ProfileSwitcherRoute, Void>
+
+    /// The services used by this processor.
+    private let services: Services
+
+    private let handler: ProfileSwitcherHandler
```
matt-livefront, line 25: ⛏️ Can you add docs for this?

BitwardenShared/UI/Auth/ProfileSwitcher/ProfileSwitcherProcessorTests.swift
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    // MARK: Tests
+
+    /// `receive(action:)` passes the action to the handler.
+    @MainActor
+    func test_receive() {
+        handler.profileSwitcherState.isVisible = true
+        subject.receive(.dismissTapped)
+        XCTAssertFalse(handler.profileSwitcherState.isVisible)
```
matt-livefront: 🤔 Will this test fail on iOS 26+, since iOS 26+ calls `dismissProfileSwitcher()` instead of toggling `isVisible`?

BitwardenShared/UI/Auth/ProfileSwitcher/ProfileSwitcherSheet.swift
```diff
@@ -0,0 +1,208 @@
+import BitwardenKit
+import BitwardenResources
+import SwiftUI
+
+public struct ProfileSwitcherSheet: View {
```
matt-livefront, line 5: 🎨 There's some overlap between this and `ProfileSwitcherView`. Is it possible to share any of that code?

KatherineInCode:
> if you long-press on an account in the switcher, it should present an action sheet

Hm. Dang it, I checked that at some point, but I guess I broke it since then. I'll take a look.

BitwardenShared/UI/Auth/ProfileSwitcher/ProfileSwitcherProcessorTests.swift
KatherineInCode: Hm, yes. Would you recommend a branch in the test itself? Or possibly an `@available` test for it? 🤔 I don't know that we have a good sense of a pattern of how to handle these sorts of things right now.

BitwardenShared/UI/Auth/ProfileSwitcher/ProfileSwitcherSheet.swift
KatherineInCode: It's a good thought. My hesitation would be if we want to change how the two objects looked over time—and I was kind of figuring when we get rid of the `View`, then we delete it and just the `Sheet` remains. But realistically, it'll be a while before that happens, so...I'll see what I can do.

BitwardenShared/UI/Auth/ProfileSwitcher/ProfileSwitcherProcessorTests.swift
matt-livefront: I think either would be fine. Given that we don't plan to use `ProfileSwitcherProcessor` for pre-iOS 26, maybe the `@available` check is easiest?

KatherineInCode: The `@available` check does not work as expected 🤔 It looks like [Apple prefers](https://developer.apple.com/documentation/xctest/methods-for-skipping-tests) a `guard #available` so I'll take that tack instead.

BitwardenResources/Localizations/en.lproj/Localizable.strings
```diff
@@ -1278,6 +1278,8 @@
 "WhenUsingTwoStepVerification" = "When using 2-step verification, you’ll enter your username and password and a code generated in this app.";
 "YesSetDefault" = "Yes, set default";
 "YouCanUpdateYourDefaultAnytimeInSettings" = "You can update your default anytime in settings.";
+"SelectAccount" = "Select account";
+"AccountsPluralNoun" = "Accounts";
```
fedemkr: 🤔 Do we have an "Accounts" key already being used for something else? I wasn't able to find it. I think we could just add a comment to indicate it's a noun, and with that test whether Crowdin detects it to be shown to translators. AFAIK it does but doesn't hurt to double check. Additionally, I think it'd be pretty unlikely to have "Accounts" as a verb alone without any other context around.

BitwardenShared/UI/Auth/AuthCoordinator.swift
```diff
@@ -270,8 +278,16 @@ final class AuthCoordinator: NSObject, // swiftlint:disable:this type_body_lengt
     /// - Parameter rehydratableTarget: The rehydratable target, if any to restore after unlocking if needed.
     private func completeAuth(rehydratableTarget: RehydratableTarget? = nil) {
         if stackNavigator?.isPresenting == true {
-            stackNavigator?.dismiss {
-                self.delegate?.didCompleteAuth(rehydratableTarget: rehydratableTarget)
+            if let presented = (stackNavigator as? ViewLoggingNavigationController)?.visibleViewController,
```
fedemkr: 🤔 What do you think about using `UINavigationController` here instead of `ViewLoggingNavigationController`? This way if we ever need to wrap the `ViewLoggingNavigationController` or have a different navigation controller for some other stuff that presents the switcher we'll be covered as well. Additionally, we don't need anything specific from `ViewLoggingNavigationController` so we can use the base class instead.

BitwardenShared/UI/Auth/ProfileSwitcher/ProfileSwitcherAccountsView.swift
```diff
@@ -0,0 +1,91 @@
...
+    /// - Parameters:
+    ///   - accountProfile: A `ProfileSwitcherItem` to display in row format.
+    ///   - showDivider: Should the cell show a divider at the bottom.
+    ///
+    @ViewBuilder
+    private func profileSwitcherRow(
+        accountProfile: ProfileSwitcherItem,
+        showDivider: Bool = true
+    ) -> some View {
+        let isActive = (accountProfile.userId == store.state.activeAccountId)
```
fedemkr: ⛏️ I think you can omit the parenthesis
```suggestion
        let isActive = accountProfile.userId == store.state.activeAccountId
```

BitwardenShared/UI/Auth/ProfileSwitcher/ProfileSwitcherHandler.swift
```diff
@@ -32,6 +32,10 @@ protocol ProfileSwitcherHandler: AnyObject {
     /// The `State` for a toast view.
     var toast: Toast? { get set }
 
+    /// Dismisses the profile switcher; this is used on iOS >=26 for making sure the sheet
+    /// is dismissed appropriately; on iOS <26, `profileSwitcherState.isVisible` is used instead.
+    func dismissProfileSwitcher()
```
fedemkr: 🤔 Would it be worth to mark them with available on 26 onwards? So we force the caller to check it's on 26. Same for `showProfileSwitcher`.

BitwardenShared/UI/Auth/ProfileSwitcher/ProfileSwitcherHandler.swift
```diff
@@ -266,8 +307,24 @@ private extension ProfileSwitcherHandler {
...
+            }
+            return
+        }
+        if #available(iOS 26, *) {
+            // This has to happen before the account switch event is handled, otherwise in the share extension,
+            // the stack navigator believes it's not presenting anything, and the dismiss becomes a no-op, leaving
+            // the profile switcher on the screen.
+            // Making sure we do the dismiss *first* solves the problem.
+            dismissProfileSwitcher()
+        }
```
fedemkr: 🎨 The guard else clause it's doing the same as when the guard passes for iOS 26+. I believe we could avoid the repetition by dismissing before the guard:
```suggestion
        if #available(iOS 26, *) {
            // This has to happen before the account switch event is handled, otherwise in the share extension,
            // the stack navigator believes it's not presenting anything, and the dismiss becomes a no-op, leaving
            // the profile switcher on the screen.
            // Making sure we do the dismiss *first* solves the problem.
            dismissProfileSwitcher()
        }
        guard account.userId != profileSwitcherState.activeAccountId || showPlaceholderToolbarIcon else {
            return
        }
```

BitwardenShared/UI/Auth/ProfileSwitcher/ProfileSwitcherModule.swift
```diff
@@ -0,0 +1,31 @@
+import Foundation
+
+// MARK: - ProfileSwitcherModule
+
+/// An object that builds coordinators for the profile switcher views on iOS >=26.
+@MainActor
+protocol ProfileSwitcherModule {
```
fedemkr: 🤔 Should this be marked as available on 26+ so it forces the caller to create it only on these situations?

BitwardenShared/UI/Auth/ProfileSwitcher/ProfileSwitcherSheetTests.swift
```diff
@@ -0,0 +1,357 @@
...
+            secondAlternate,
+        ]
+        let current = processor.state.activeAccountProfile!
+        processor.state = ProfileSwitcherState(
+            accounts: alternateAccounts + [current],
+            activeAccountId: current.userId,
+            allowLockAndLogout: true,
+            isVisible: true
+        )
+        let addAccountRow = try subject.inspect().find(button: "")
```
fedemkr: ⛏️ Really tiny but I think this test is a bit fragile to find a button by an empty string. Could the accessibility identifier be used here instead to find it? Which is also more straightforward when someone else reads it to know it refers to the add account row?

BitwardenShared/UI/Auth/ProfileSwitcher/TestHelpers/MockProfileSwitcherHandler.swift
```diff
@@ -0,0 +1,46 @@
+import BitwardenKit
+import BitwardenKitMocks
+@testable import BitwardenShared
+
+// MARK: - MockProfileSwitcherHandler
+
+class MockProfileSwitcherHandler: ProfileSwitcherHandler {
```
fedemkr: 🤔 Is `sourcery: AutoMockable` having problems with the handler to make a manual mock of it?

BitwardenShared/UI/Auth/AuthCoordinator.swift
KatherineInCode: I think there's value in tracking that the user opened/closed the account switcher window, which is why I used it.

BitwardenShared/UI/Auth/ProfileSwitcher/ProfileSwitcherSheetTests.swift
KatherineInCode: It's literally a copy from `ProfileSwitcherViewTests`. But I can see about updating tests in both files to be more explicit.

BitwardenShared/UI/Auth/ProfileSwitcher/ProfileSwitcherModule.swift
KatherineInCode: It's an interesting thought (same with the other ones noted). I'll take a look.

BitwardenShared/UI/Auth/ProfileSwitcher/TestHelpers/MockProfileSwitcherHandler.swift
KatherineInCode: I don't know; I forgot we had that in there, and defaulted to making a manual mock. I'll see if sourcery can handle it.

BitwardenResources/Localizations/en.lproj/Localizable.strings
KatherineInCode: I wasn't able to find it, either, and decided to be explicit with the key. However, moving that to a comment seems reasonable, so I'll see what I can do there, especially as then I can add the context of "this is always maximally plural, even if there are only one or two accounts".

BitwardenShared/UI/Auth/ProfileSwitcher/ProfileSwitcherModule.swift
KatherineInCode: So I was able to do this on the protocol functions, but doing it on the Module ended up meaning that paths weren't getting tested anymore (because they'd only run on iOS 26, and tests are currently run only on iOS 18.6). There's also no technical reason to only do the sheet on iOS 26; it ends up working fine if called on iOS 18—right now the difference is just a design decision, but I could see us moving to the sheet for the account switcher on pre-26 iOS as well.

Once we move to iOS 26 for testing, though, it might be worth revisiting this if we really are concerned with it getting called in the pre-26 flows.

BitwardenShared/UI/Auth/AuthCoordinator.swift
fedemkr: The `ViewLoggingNavigationController` will be used but I was talking here specifically for the casting; so if at some point that is changed for whatever reason this flow will still work successfully.

BitwardenShared/UI/Auth/ProfileSwitcher/ProfileSwitcherModule.swift
fedemkr: Sounds reasonable, thanks for taking a look!

BitwardenShared/UI/Auth/AuthCoordinator.swift
KatherineInCode: Oh, yeah, that makes sense.

BitwardenShared/UI/Auth/ProfileSwitcher/ProfileSwitcherSheetTests.swift
KatherineInCode: Ah, the variable is poorly named, which I'll fix. It's intentionally trying to find the button for an account that has an empty name. And because every account button has the accessibility identifier of "AccountCell" (that our accessibility identifiers aren't unique is a separate issue that I've brought up before, but QA prefers them that way for automated tests), we wouldn't be able to uniquely identify it that way (and I can't find a ViewInspector API that looks at accessibility IDs—just labels and IDs, which are both different).

fedemkr: Ohhh, nice catch! Now it makes sense, thanks Katherine!

KatherineInCode:
Hm. Yes, they should, and they were at some point. I'll take a look.

PR #1911, opened 2025-09-04
[PM-25284] Fix invalid URL on CXP flow when URL is IP address with port

fedemkr:
## 🎟️ Tracking

[PM-25284](https://bitwarden.atlassian.net/browse/PM-25284)

## 📔 Objective

Fix CXP JSON decoder so it takes into consideration URIs where it has no scheme, are IP addresses and have a port. This was causing the URL(string:) initializer to return `nil` and breaking the decoding process. This fixes it by prefixing `http://` or `https://` accordingly depending on if it's an IP address or a different type of URI that can't be converted to `URL` directly.

Example of URL that was causing problems: `192.168.0.1:8080`.

> [!NOTE]
> A new decoder was added that can configure property names that are array of URLs to be taken into account for the fix. This will check whether the URL is an IP address and add the `http://` prefix to it so it's able to be automatically decoded to `URL` without problems.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-25284]: https://bitwarden.atlassian.net/browse/PM-25284?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenKit/Core/Platform/Extensions/String+Extensions.swift
+++ b/BitwardenKit/Core/Platform/Extensions/String+Extensions.swift
@@ -15,7 +15,7 @@ public extension String {
             return url.absoluteString
         }
 
-        if !self.hasPrefix("http") {
+        if !hasPrefix("http") {
             return "https://\(self)"
         }
 
--- a/BitwardenKit/Core/Platform/Extensions/String+Extensions.swift
+++ b/BitwardenKit/Core/Platform/Extensions/String+Extensions.swift
@@ -1,0 +1,24 @@
+import Foundation
+
+/// Extension helpers for `String`.
+public extension String {
+    /// Tries to convert the string to an `URL`, if it can't then it tries to fix it by adding
+    /// a http scheme prefix if it's an IP address or https scheme prefix in any other case.
+    /// - Returns: The same `String` if it can be converted to an `URL` or the fixed
+    /// `String` otherwise.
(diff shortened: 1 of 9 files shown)


KatherineInCode (approved):

fedemkr:
I really like your idea @matt-livefront  but I think it's already covered by the IP address with port test:
https://github.com/bitwarden/ios/blob/5938abf3ec2ab4ae0c0440bee696b7273b032244/BitwardenKit/Core/Platform/Extensions/StringExtensionsTests.swift#L16-L19
If Apple fixed that issue then the test would fail as it would simply return the same URI string: `192.168.0.1:8080` instead of `http://192.168.0.1:8080`. So I believe we're already covered.

matt-livefront:
> I really like your idea @matt-livefront but I think it's already covered by the IP address with port test:
> 
> https://github.com/bitwarden/ios/blob/5938abf3ec2ab4ae0c0440bee696b7273b032244/BitwardenKit/Core/Platform/Extensions/StringExtensionsTests.swift#L16-L19
> 
> 
> If Apple fixed that issue then the test would fail as it would simply return the same URI string: `192.168.0.1:8080` instead of `http://192.168.0.1:8080`. So I believe we're already covered.

Ah, I missed that, thanks!

PR #1960, opened 2025-09-19
[PM-26006] Create a stringsdict to use for plurals

KatherineInCode:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-26006

## 📔 Objective

⚠️ This PR is being held until after discussions with the team on if this is the path we want to go forward with localization ⚠️ 

This creates a `Localizable.stringsdict` in `BitwardenResources` that can be used for better pluralization handling going forward. As well, one entry for minutes has been added as an example.

The main caveat with using a `stringsdict` is because of [an unmerged PR in SwiftGen](https://github.com/SwiftGen/SwiftGen/pull/1128), languages that don't have the key in their own `stringsdict` will default to a user-hostile string of `Plural format key` without the parameterized information; this means CrowdIn pushes and pulls will be essential to avoid user issues.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


fedemkr (commented):
🤔 Looks good although I think that given we have the plural replacement now we should remove the old localizations for `oneMinute`, `twoMinutes`, so on. And if they are somewhere else used in the code, they should be updated as well in this PR so we don't mix both approaches for the singular <-> plural keys.

KatherineInCode:
That makes sense to me; I was going for the smallest unit of change as demonstration, but since we've decided to go with this, I'll update it accordingly. I still want to stick with just minutes right now—a later PR can handle some of the other obvious ones (seconds, days, hours, etc.)

PR #1961, opened 2025-09-19
[PM-25567] Update profile switcher toolbar button to full circle

matt-livefront:
## 🎟️ Tracking

[PM-25567](https://bitwarden.atlassian.net/browse/PM-25567)

## 📔 Objective

Updates the profile switcher toolbar button to be the full circle on iOS 26+.

## 📸 Screenshots

Screenshot 2025-09-19 at 3 43 03 PM

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-25567]: https://bitwarden.atlassian.net/browse/PM-25567?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/UI/Platform/Application/Extensions/View+Backport.swift
+++ b/BitwardenShared/UI/Platform/Application/Extensions/View+Backport.swift
@@ -19,6 +19,17 @@ extension View {
 /// Adapted from https://davedelong.com/blog/2021/10/09/simplifying-backwards-compatibility-in-swift/
 ///
 extension Backport where Content: View {
+    /// On iOS 26+, configures the button style to use `glassProminent`.
+    ///
+    @available(iOS, deprecated: 26.0, message: "Use `buttonStyle(.glassProminent)` instead.")
+    func buttonStyleGlassProminent() -> some View {
+        if #available(iOS 26, *) {
+            return content.buttonStyle(.glassProminent)
+        } else {
+            return content
+        }
+    }
+
     /// On iOS 16+, configures the scroll view to dismiss the keyboard immediately.
     ///
     func dismissKeyboardImmediately() -> some View {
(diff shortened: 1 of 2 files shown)


KatherineInCode:
One of the things I noticed while fiddling with this is that the animation when going into another screen is a little weird, with the border disappearing, but not the square around the letters:

https://github.com/user-attachments/assets/59519ba5-b08d-433d-b73e-386d04a33911

This might be in the category of "things we can't do anything about", much like the search bar pop-in also demonstrated there. But it still feels kind of jarring to me, even if it's harder to see in realtime.

matt-livefront:
> One of the things I noticed while fiddling with this is that the animation when going into another screen is a little weird, with the border disappearing, but not the square around the letters:
> 
> This might be in the category of "things we can't do anything about", much like the search bar pop-in also demonstrated there. But it still feels kind of jarring to me, even if it's harder to see in realtime.

@KatherineInCode Ah, good catch. I made a change to remove the square around the letters. That seems to fix the most glaring issue. You can still see the letters fade out, but it isn't as obvious. It seems like the tint on the button is removed immediately without an animation when the transition happens, but the button's contents are faded out. I'm not sure if there's anything else we can do about this though, other than hope that a future iOS update starts animating the tint change.

BitwardenShared/UI/Platform/Application/Extensions/View+Backport.swift
```diff
@@ -19,6 +19,17 @@ extension View {
 /// Adapted from https://davedelong.com/blog/2021/10/09/simplifying-backwards-compatibility-in-swift/
 ///
 extension Backport where Content: View {
+    /// On iOS 26+, configures the button style to use `glassProminent`.
+    ///
+    @available(iOS, deprecated: 26.0, message: "Use `buttonStyle(.glassProminent)` instead.")
```
KatherineInCode, line 24: 👍🏻 I like the deprecation message. We should remember to do that sort of thing more often when applicable

KatherineInCode:
@matt-livefront Yes, I do think that looks a lot better, and is a fair bit harder to see. Though I suspect the tint being removed immediately is probably related to why the search bar also has its pop-in/pop-out. I think that's reasonable, though at some point we may want to consider circling back on the design, especially if Apple doesn't do anything about animating tint because they want people to stop using tint.

KatherineInCode (approved):

PR #1973, opened 2025-09-23
[PM-25993] Automatically select default collection on adding a new cipher

fedemkr:
## 🎟️ Tracking

[PM-25993](https://bitwarden.atlassian.net/browse/PM-25993)

## 📔 Objective

Automatically select default user collection when adding a new cipher if present.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-25993]: https://bitwarden.atlassian.net/browse/PM-25993?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenShared/UI/Vault/VaultItem/AddEditItem/AddEditItemProcessorTests.swift
```diff
@@ -1128,6 +1128,129 @@ class AddEditItemProcessorTests: BitwardenTestCase {
...
+        XCTAssertEqual(subject.state.collectionIds, [])
+    }
+
+    /// `perform(_:)` with `.fetchCipherOptions` fetches the ownership options for a cipher from the repository
+    /// and doesn't select the default user collection when personal ownership policy is disabled.
+    @MainActor
+    func test_perform_fetchCipherOptions_defaultUserCollectionPersonalOwnershipDisabled() async {
+        let collections: [CollectionView] = [
+            .fixture(id: "1", name: "Design", organizationId: "1", type: .sharedCollection),
+            .fixture(id: "2", name: "Engineering", organizationId: "1", type: .sharedCollection),
```
matt-livefront, line 1200: ❓ Is this the same as the test above? Should one of the collection's type be `defaultUserCollection`?

BitwardenShared/UI/Vault/VaultItem/AddEditItem/AddEditItemProcessorTests.swift
```diff
@@ -1128,6 +1128,129 @@ class AddEditItemProcessorTests: BitwardenTestCase {
...
+                rehydrationHelper: rehydrationHelper,
+                stateService: stateService,
+                totpService: totpService,
+                vaultRepository: vaultRepository
+            ),
+            state: CipherItemState(
+                existing: CipherView.fixture(),
+                hasPremium: false
+            )!
+        )
```
matt-livefront: 🤔 Would it work to override the state here vs re-initializing the subject?
```suggestion
        subject.state = CipherItemState(
            existing: CipherView.fixture(),
            hasPremium: false
        )!
```

BitwardenShared/UI/Vault/VaultItem/AddEditItem/AddEditItemProcessorTests.swift
fedemkr: Indeed, one of them should be `.defaultUserCollection`; nice catch!

BitwardenShared/UI/Vault/VaultItem/AddEditItem/AddEditItemProcessorTests.swift
```diff
@@ -1128,6 +1128,111 @@ class AddEditItemProcessorTests: BitwardenTestCase {
         XCTAssertTrue(subject.state.canAssignToCollection)
     }
 
+    /// `perform(_:)` with `.fetchCipherOptions` fetches the ownership options for a cipher from the repository
+    /// and selects the default user collection when there's any on adding mode.
+    @MainActor
+    func test_perform_fetchCipherOptions_defaultUserCollection() async {
+        let collections: [CollectionView] = [
+            .fixture(id: "1", name: "Design", organizationId: "1", type: .defaultUserCollection),
```
KatherineInCode: 🤔 Would it make sense to have a test where the default collection is not the first item in the collections? As well, is it possible for there to be two collections both marked as default (as the data certainly seems to indicate)? Should we test/handle that scenario as well?

fedemkr: As far as I understand there can be only one default user collection per organization. However, there could be multiple default user collections generally if the user belongs to multiple organizations. For this scenario in particular where a single organization is selected then only one default collection should be present or it would be wrongly configured server side and I think there isn't much we can do about it.
I will add another test thought to check that only the first one gets selected if by any chance multiple are present.

KatherineInCode: Yeah, I agree that the business rules shouldn't allow it—but the API does (something something "make illegal states unrepresentable"), so I feel like making sure we have a test for that circumstance is a good idea.

KatherineInCode (approved):

Issue #1912, opened 2025-09-04
[PM-25483] Missing plural forms


mKoonrad:
### Steps To Reproduce

In many languages, there is more than one plural form. App currently does not support plural forms. There is no way to translate some strings correctly. You need to add full support for plurals. Then Crowdin translator will have opportunity to translate plurals in specific country format.

Example English:
1 character
2 character**s**,
10 character**s**
etc.

Polish:
1 znak
2 znak**i**
5 znak**ów**

### Expected Result

Adding plural forms for following strings using Localizable.stringsdict file:
https://github.com/bitwarden/ios/blob/713f8b526153f7e2db513b7058d043e857d72abc/BitwardenResources/Localizations/en.lproj/Localizable.strings#L394
https://github.com/bitwarden/ios/blob/713f8b526153f7e2db513b7058d043e857d72abc/BitwardenResources/Localizations/en.lproj/Localizable.strings#L982
https://github.com/bitwarden/ios/blob/713f8b526153f7e2db513b7058d043e857d72abc/BitwardenResources/Localizations/en.lproj/Localizable.strings#L1097
https://github.com/bitwarden/ios/blob/713f8b526153f7e2db513b7058d043e857d72abc/BitwardenResources/Localizations/en.lproj/Localizable.strings#L1119

### Actual Result

Missing plural forms. Can't translate some strings correctly.

### Additional Context

[https://developer.apple.com/documentation/xcode/localizing-strings-that-contain-plurals](https://developer.apple.com/documentation/xcode/localizing-strings-that-contain-plurals)

### Build Version

any

### What server are you connecting to?

N/A

### Issue Tracking Info

- [x] I understand that work is tracked outside of Github. A PR will be linked to this issue should one be opened to address it, but Bitwarden doesn't use fields like "assigned", "milestone", or "project" to track progress.

cylayabw:
Thank you for your report! We use Crowdin to manage our translations, and these are commonly done by our amazing translation community!

Please take a look at the localization section of our contributing ​documentation for further information.

https://contributing.bitwarden.com/contributing/#localization-l10n 

New translations can be added, or existing ones can be up-/down-voted.

mKoonrad:
[...]

PR #1999, opened 2025-09-28
Add plurals

mKoonrad:
## 🎟️ Tracking

#1912 

## 📔 Objective

I added plurals by stringsdict file.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines


AuthenticatorShared/Core/Platform/Models/Enum/ClearClipboardValue.swift
```diff
@@ -44,11 +44,11 @@ enum ClearClipboardValue: Int, Menuable {
         case .never:
             Localizations.never
         case .tenSeconds:
-            Localizations.tenSeconds
+            Localizations.XSeconds(10)
```
KatherineInCode: 🤔 I'm curious why this came out as `XSeconds` when minutes is `xMinutes`. The latter is our preferred casing.

BitwardenResources/Localizations/en.lproj/Localizable.stringsdict
```diff
@@ -1,22 +1,236 @@
...
-		<dict>
-			<key>NSStringFormatSpecTypeKey</key>
-			<string>NSStringPluralRuleType</string>
-			<key>NSStringFormatValueTypeKey</key>
-			<string>d</string>
-			<key>one</key>
-			<string>%d minute</string>
-			<key>other</key>
-			<string>%d minutes</string>
+		<key>XMinutes</key>
```
KatherineInCode: 🤔 Would it be possible to alphabetize these by key? I know we aren't the most consistent with it in the `strings` file, but it would be helpful to start out with good habits like that.

AuthenticatorShared/Core/Platform/Models/Enum/ClearClipboardValue.swift
mKoonrad: May you explain that better? This file contains minutes and seconds time option. Shorter and longer values. Both options.

KatherineInCode: The keys in both cases have the same casing of "XMinutes" and "XSeconds". I wonder why SwiftGen is casing the generated functions differently as `xMinutes` vs. `XSeconds`.

BitwardenShared/UI/Vault/Extensions/AlertVaultTests.swift
```diff
@@ -48,7 +48,7 @@ class AlertVaultTests: BitwardenTestCase { // swiftlint:disable:this type_body_l
         XCTAssertEqual(subject.title, Localizations.decryptionError)
         XCTAssertEqual(
             subject.message,
-            Localizations.bitwardenCouldNotDecryptOneVaultItemDescriptionLong
+            Localizations.BitwardenCouldNotDecryptXVaultItemsDescriptionLong(1)
```
KatherineInCode: 🤔 I don't see any changes to SwiftGen configuration; how did you change how it cases the functions it generates just for some of the items in the stringsdict file? Because as stands, this looks like it would fail to build.

KatherineInCode:
🤔 It looks like some other issues have come up with conflicts and the like. Do you mind if I take this PR and run with it (though probably in a separate PR because GitHub, I'll have to see) to clean some remaining stuff up and get all the ducks in a row?

mKoonrad:
Feel free to edit!

KatherineInCode:
After spending some time futzing with git and the code, I determined that I thought it would be easier to break out adding these plurals into a couple PRs, potentially, so as to get a more-focused look at them, and handle some ancillary improvements that can be made at the same time. I'll still keep this open as a way of keeping tabs on what still might need to be migrated, as we can keep merging `main` into it, which *should* hopefully Just Work.

KatherineInCode:
@mKoonrad I think with https://github.com/bitwarden/ios/pull/2199 I've finished pulling in all the plurals you identified here, aside from the ones that got shifted around in the interim. Can you check to make sure nothing got missed? And then if you're okay with it, we can close this PR.

mKoonrad:
@KatherineInCode First of all, thank you for your work! 🎆

It's done, it's works, but in some cases transaltors could transalte it in the wrong way.

Some sentences were split into two parts:
1. The static part (the main sentence) was in the top-level key.
2. The dynamic part (the number and the noun) was inside the plural variations.

**Why this was bad**: Translators in Crowdin only saw small fragments like `%d items `or `%d characters` without knowing the rest of the sentence. In many languages (like Polish), the words before the number need to change their form (grammar/declension) based on that number. When the string is fragmented, it is impossible to translate it with 100% grammatical accuracy.

Crowdin links the same `%d items` translations for part of sentences and standalone string.

I suggest to move entire sentence into plural rules. What improved:
- **Better Context**: Translators now see the complete sentence in Crowdin (e.g., "Your PIN must be at least %d characters" instead of just "%d characters").
- **Grammatical Correctness**: It allows for different word orders and correct declension across all languages.
- **Easier Localization**: Translators don't have to guess where the fragment is being used.

Current:
image
image

After improvement:

image

Does this make sense to you? Should I prepare PR or do you want to take care of it?

KatherineInCode:
> Crowdin links the same `%d items` translations for part of sentences and standalone string.

Ah, dang, I didn't quite realize that. Yeah, that makes sense. I can swing back through on this. Though it does look like Crowdin provides the context of where things are used in the places where it doesn't do the matching, as well 🤔 but yes, that makes sense. I'll circle back to it when I can.

KatherineInCode:
@mKoonrad I've been thinking about this over the holidays, and I'm curious. What would be your recommendation for how best to handle sentences with two different (number + noun)s in them?

For instance, take the following English sentence:

> The app will play X songs and show Y pictures.

In the way I've done it before at other jobs (that admittedly had translations into much fewer languages), and did with the stringsdict here, it would be to break out subdictionaries. This is [what the Apple docs indicate](https://developer.apple.com/documentation/xcode/localizing-strings-that-contain-plurals), though now it's just in a single line and doesn't have a clear example; [this page](https://localit.io/docs/ios-stringsdict-format-guide/) shows some examples of it. That looks like:

> The app will play `[songs]` and show `[pictures]`

where `[songs]` and `[pictures]` would break into "0 song / N songs" and "0 picture / N pictures", respectively.

However, this can cause problems because of CrowdIn's translation memory, where "N songs" gets assumed to be the same unit everywhere, even though in some languages it would inflect differently if it were the subject vs. the object.

As well, there are languages like Hungarian that have [polypersonal agreement](https://en.wikipedia.org/wiki/Polypersonal_agreement), where the verb is inflected also in accordance with the count of the objects. As well, as you indicated, Polish also has grammatical features similar to this. How best to handle this?

My initial instinct is to also pull in the verbs:

> The app will `[play songs]` and `[show pictures]`

But that 1) relies way too much on English word order, and 2) in this case, I completely ignored that it's actually "will play" and "will show"—that is, future tense, which some languages inflect in the verb itself instead of having an auxiliary verb like English. It very much wouldn't work.

So I'm a little stymied, short of breaking it up into two sentences (which may not always be an option, depending exactly on how it's constructed, thinking particularly of the "X hours and Y minutes" sorts of constructions).

Do you have any thoughts/opinions/advice on the matter? Could something fancy be done with the "nested dependencies" as they're called in that example page I linked earlier?

mKoonrad:
@KatherineInCode Sorry for the late reply! For now, I recommend the standard solution.
The app will play `X songs` and show `X pictures`

Any special string will get format updates.

Regarding to Crowdin memory:
Unfortunately, we can't fully prevent translator errors. In all other cases, we should aim to make their job easier (e.g. by providing full sentences in plural forms rather than using variables, comments etc.).

KatherineInCode:
@mKoonrad I think at this point, we have everything migrated to the stringsdict file, and this can be closed?

mKoonrad:
@KatherineInCode Done :)

PR #2007, opened 2025-09-30
[PM-26387] Fix lint warnings

matt-livefront:
## 🎟️ Tracking

[PM-26387](https://bitwarden.atlassian.net/browse/PM-26387)

## 📔 Objective

Fixes the remaining lint warnings.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/AuthenticatorShared/UI/Vault/ItemList/ItemList/ItemListView.swift
+++ b/AuthenticatorShared/UI/Vault/ItemList/ItemList/ItemListView.swift
@@ -713,7 +713,7 @@ struct ItemListView_Previews: PreviewProvider { // swiftlint:disable:this type_b
                                             )
                                         ),
                                     ],
-                                    name: "example.com",
+                                    name: "example.com"
                                 ),
                             ])
                         )
--- a/BitwardenKit/Core/Platform/Extensions/StringTests.swift
+++ b/BitwardenKit/Core/Platform/Extensions/StringTests.swift
@@ -146,7 +146,6 @@ class StringTests: BitwardenTestCase {
         XCTAssertEqual(expected, subject.whitespaceRemoved())
     }
 
-
     /// `withoutAutomaticEmailLinks()` returns the string with email addresses appropriately modified.
     func test_withoutAutomaticEmailLinks() {
         let subject = "person@example.com"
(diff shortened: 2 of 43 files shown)


AuthenticatorShared/UI/Vault/ItemList/ItemList/ItemListView.swift
```diff
@@ -713,7 +713,7 @@ struct ItemListView_Previews: PreviewProvider { // swiftlint:disable:this type_b
                                             )
                                         ),
                                     ],
-                                    name: "example.com",
+                                    name: "example.com"
```
KatherineInCode: 🤔 For some reason, I thought we had final commas turned on? Did that change? And if this is just for parameter lists, I kind of think we should be consistent about it, especially now that Xcode 26 allows it. Thoughts?

fedemkr: I personally don't like trailing commas when there are no further items after that (nor see the benefit of it). I can live with the ones in the array lists as there are a lot of them and it'd be tedious to change them all, but IMO we shouldn't use that for function parameters (as in this case).

KatherineInCode: Ah, I prefer trailing commas because it makes diffs easier to read and makes it easier to rearrange the orders of things (or insert/delete items), because you never have to futz with the commas as well.

matt-livefront: > For some reason, I thought we had final commas turned on?

I think SwiftFormat didn't previously complain about whether you included the trailing comma or not for parameter lists. It is configurable though, so we could enable either option.

I do like trailing commas for the reasons @KatherineInCode noted. Parameter lists would take some getting used to, but it might be nice for consistency. That would be a large change though, so maybe we can consider it and enable it separately if we want to do that.

KatherineInCode: It might be reasonable to just turn the rule off entirely right now, and address the inconsistency later after some discussion?

fedemkr: Oh that's a nice thought, I don't mind dealing with the commas but if it's easier for review and code rearrangement then let's configure and use it for sure 👍

KatherineInCode (approved):

PR #2041, opened 2025-10-14
[PM-26735] Prevent push notification logout when changing KDF settings

matt-livefront:
## 🎟️ Tracking

[PM-26735](https://bitwarden.atlassian.net/browse/PM-26735)

## 📔 Objective

Prevents a logout push notification from logging the user out when changing KDF settings. When changing KDF settings, the backend will send a logout push notification with a reason of `kdfChange: 0`. If this is received by the app, and the app has the `forceUpdateKdfSettings` feature flag enabled, the user shouldn't be logged out since the new KDF settings can be used immediately following a sync.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-26735]: https://bitwarden.atlassian.net/browse/PM-26735?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/Core/Platform/Services/NotificationService.swift
+++ b/BitwardenShared/Core/Platform/Services/NotificationService.swift
@@ -93,6 +93,9 @@ class DefaultNotificationService: NotificationService {
     /// The service used by the application to handle authentication tasks.
     private let authService: AuthService
 
+    /// The service to get server-specified configuration.
+    private let configService: ConfigService
+
     /// The service used by the application to report non-fatal errors.
     private let errorReporter: ErrorReporter
 
@@ -116,6 +119,7 @@ class DefaultNotificationService: NotificationService {
     ///   - appIdService: The service used by the application to manage the app's ID.
     ///   - authRepository: The repository used by the application to manage auth data for the UI layer.
     ///   - authService: The service used by the application to handle authentication tasks.
+    ///   - configService: The service to get server-specified configuration.
     ///   - errorReporter: The service used by the application to report non-fatal errors.
     ///   - notificationAPIService: The API service used to make notification requests.
     ///   - refreshableApiService: The API service used to refresh tokens.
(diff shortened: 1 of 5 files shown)


BitwardenShared/Core/Platform/Services/NotificationService.swift
```diff
@@ -217,7 +223,16 @@ class DefaultNotificationService: NotificationService {
                 try await refreshableApiService.refreshAccessToken()
                 try await syncService.fetchSync(forceSync: true)
             case .logOut:
-                guard let data: UserNotification = notificationData.data() else { return }
+                guard let data: LogoutNotification = notificationData.data() else { return }
+
+                if data.reason == .kdfChange,
+                   // TODO: PM-26960 Remove user ID check with forceUpdateKdfSettings feature flag.
+                   data.userId == userId,
+                   await configService.getFeatureFlag(.forceUpdateKdfSettings) {
```
matt-livefront, line 231: ℹ️ One thing to note here is that this only works for the active user. We don't currently have a way to check a feature flag for inactive users. Adding that would introduce a ripple effect of other issues, for example we'd need the ability to make authenticated network requests (to update the config) for inactive users.

fedemkr, line 231: 🤔 Shouldn't we stick with the user ID check when removing the feature flag? Or will that property of the notification go away from the API when the FF is removed?

matt-livefront: The user ID check is checking that the user in the push notification is the active user. We have the ability to logout an inactive user. But we can't check the feature flags for the inactive user. So once we remove the feature flag check, we can also remove the user ID check.

fedemkr: Oh makes sense, saw your comment after posting this sorry hehe

PR #2054, opened 2025-10-20
[PM-26606] Fix most strict warnings in AuthenticatorBridgeKit

KatherineInCode:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-26606
https://bitwarden.atlassian.net/browse/PM-26770
https://bitwarden.atlassian.net/browse/PM-26754

## 📔 Objective

This turns on strict concurrency warnings in `AuthenticatorBridgeKit` and `AuthenticatorBridgeKitTests`, and then fixes most of the warnings, partly through `@preconcurrency` and partially through using a `DispatchQueue` to enforce thread safety.

There is one warning remaining in `AuthenticatorBridgeKitTests`, relating to `TestData`; this will be covered in [PM-27143](https://bitwarden.atlassian.net/browse/PM-27143).

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-27143]: https://bitwarden.atlassian.net/browse/PM-27143?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/actions/boot-simulator/action.yml
+++ b/.github/actions/boot-simulator/action.yml
@@ -1,0 +1,51 @@
+name: 'Boot iOS Simulator'
+description: 'Boots an iOS simulator with retry logic and timeout handling'
+author: 'Bitwarden'
+
+inputs:
+  simulator-name:
+    description: 'Name of the simulator device (e.g., iPhone 17 Pro)'
+    required: true
+  simulator-version:
+    description: 'iOS version for the simulator (e.g., 26.0)'
+    required: true
+
+runs:
+  using: 'composite'
+  steps:
+    - name: Boot Simulator
+      shell: bash
+      env:
+        _SIMULATOR_NAME: ${{ inputs.simulator-name }}
+        _SIMULATOR_VERSION: ${{ inputs.simulator-version }}
+      run: |
+        echo "Listing simulator devices:"
+        SIMULATORS=$(xcrun simctl list devices "iOS $_SIMULATOR_VERSION" available)
+        echo "$SIMULATORS"
+        DEVICE_ID=$(echo "$SIMULATORS" | grep "$_SIMULATOR_NAME" | head -1 | sed -E 's/.*\(([A-F0-9-]+)\).*/\1/')
+        echo "--------------------------------"
+        echo "👀 Updating dyld_shared_cache:"
+        xcrun simctl runtime dyld_shared_cache update --all
+        echo "--------------------------------"
+        echo "👀 Booting simulator: $_SIMULATOR_NAME ($_SIMULATOR_VERSION) with ID: $DEVICE_ID"
+
+        i=1
+        max=5
+        while [ $i -le $max ]; do
+            echo "👀 Launching Simulator (Attempt $i/$max)"
+            # Brewfile installs coreutils which includes gtimeout
+            if gtimeout -s SIGKILL 2m xcrun simctl bootstatus "$DEVICE_ID" -b; then
+                echo "✅ Simulator booted successfully"
+                exit 0
+            else
+                echo "❌ Simulator boot timed out. Shutting down..."
+                xcrun simctl shutdown "$DEVICE_ID" || true # prevents "Unable to shutdown device in current state: Shutdown" error
+
+                echo "🔁 Retrying in 5s..."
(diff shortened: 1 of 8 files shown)


AuthenticatorBridgeKit/AuthenticatorBridgeDataStore.swift
```diff
@@ -20,8 +20,11 @@ public enum AuthenticatorBridgeStoreType {
 private let authenticatorBridgeModelName = "Bitwarden-Authenticator"
 
 /// A data store that manages persisting data across app launches in Core Data.
+/// This is currently marked `@unchecked Sendable` because of how we ensure thread safety of the `backgroundContext`
+/// property. Once we have a minimum version of iOS 16 or higher, we can migrate to the `Synchronization` framework
+/// and make this more properly `Sendable`.
 ///
-public class AuthenticatorBridgeDataStore {
+public final class AuthenticatorBridgeDataStore: @unchecked Sendable {
```
fedemkr: 🤔 I was wondering whether we should move this to be an `actor` and then create `persistentContainer.newBackgroundContext` per function instead of reusing the same background context all the time which would fix these warnings as well.
Although not sure if that could cause any race condition on non-atomic usage of it by any of the functions.
Perhaps it's just better this way for now not to change a lot of stuff at once.

KatherineInCode: My understanding is that generally [with CoreData](https://developer.apple.com/documentation/coredata/setting-up-a-core-data-stack) you only want the one `NSManagedObjectContext` per `NSPersistentContainer`, which in this case is specific to the BWA-PM shared CoreData.

Another option was moving the initialization of the context into the initializer, but I assumed there was a good reason we made it `lazy` to begin with, probably along the performance line.

Moving it to an actor might not be a bad idea, though 🤔 Because looking at it again, I don't know that it makes sense for it to be nonisolated. The other option would be to make it run on the main actor, since that's the default that Swift is moving towards.

I can poke at it a bit.

KatherineInCode: I asked Claude for the implications of migrating the class to be an actor, and it generated:

> Core Data Specific Concerns
>
> The biggest concern would be how actor isolation interacts with Core Data's queue-based concurrency model. Core Data contexts have their own internal queues, and actor isolation might conflict with this, potentially leading to:
>
> • Deadlocks if Core Data operations try to synchronously access actor-isolated properties
> • Performance issues from double-queuing (actor queue + Core Data queue)
>
> Recommendation
>
> For this specific use case with Core Data, I'd recommend keeping it as a class because:
>
> 1. The current solution works well and maintains API compatibility
> 2. Core Data's queue-based concurrency model is already well-established
> 3. The performance overhead of making all access async might not be worth it
> 4. The manual synchronization is well-contained and documented
>
> The actor model shines more for business logic objects rather than data store wrappers around frameworks like Core Data that have their own concurrency models.

KatherineInCode: Asking Claude more questions, its generation very much wants this class to be nonisolated, actually. And doing a bit of digging, I think that makes sense, especially as `NSManagedObjectContext` is itself marked nonisolated. Strict concurrency seems to interact weirdly with Core Data 🫠

fedemkr: Alright we can keep the `class` for now and then check whether we find a more up-to-date solution for this that is not a workaround to properly use it. I wouldn't use `@MainActor` as for long-running operations it will affect UI performance.
I'll research more about this during the week.

KatherineInCode: After some back-and-forth with Claude and with digging around in Apple forums and such, I think the best option is for the `AuthenticatorBridgeDataStore` to be nonisolated, so I've marked it explicitly such, to avoid issues if/when we switch the default isolation to `MainActor` in line with current Swift defaults.

PR #2066, opened 2025-10-23
[PM-27219] Fix update and set master password errors

matt-livefront:
## 🎟️ Tracking

[PM-27219](https://bitwarden.atlassian.net/browse/PM-27219)

## 📔 Objective

This fixes two errors that were occurring around updating and setting a master password during login:

**TDE user with reset password on trusted device**
 
When a TDE user has their password reset on a trusted device, the app was displaying the update master password screen before the SDK was initialized (`initializeUserCrypto`). When updating the password, the SDK was throwing throw following error:

> BitwardenSdk.BitwardenError.MobileCrypto(BitwardenSdk.CryptoClientError.Crypto(message: "Missing Key for Id: User"))

This was fixed by moving the update master password screen to be after the SDK was initialized. We already had a check for this in place, so the `AuthError.requireUpdatePassword` error that occurs before initialization could be removed.

**JIT provisioned user in an organization with master password decryption (non-TDE)**

When a new user is created in a master password organization with SSO, the user authenticates with SSO and then has to set a master password. In this scenario, the user doesn't have account encryption keys until after they set a master password. `StateServiceError.noEncryptedPrivateKey` was being thrown before the master password was set.

## 📸 Screenshots

| Before | After |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/6255fae1-0713-473d-88b4-068db228527f" /> | <video src="https://github.com/user-attachments/assets/ffd466e4-2933-4e6e-af20-4557409eb980" /> | 

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-27219]: https://bitwarden.atlassian.net/browse/PM-27219?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
```diff
@@ -815,21 +815,22 @@ extension DefaultAuthRepository: AuthRepository {
         resetPasswordAutoEnroll: Bool,
     ) async throws {
         let account = try await stateService.getActiveAccount()
-        let accountKeys = try await stateService.getAccountEncryptionKeys()
```
matt-livefront, line 818: ℹ️ `getAccountEncryptionKeys()` can't be called here because in the case where the user hasn't set a password yet, their encryption keys don't exist and this throws. This ultimately needed to be moved into the TDE `if` block below so it doesn't apply to non-TDE users.

BitwardenShared/Core/Auth/Services/AuthService.swift
```diff
@@ -661,11 +658,6 @@ class DefaultAuthService: AuthService { // swiftlint:disable:this type_body_leng
                     throw AuthError.requireDecryptionOptions
                 }
 
-                // User need to update password
-                if response.forcePasswordReset {
-                    throw AuthError.requireUpdatePassword
```
matt-livefront, line 666: ℹ️ This error was being thrown too early before the SDK was initialized. We have a similar check after auth completes, which is a better place to handle this:

https://github.com/bitwarden/ios/blob/29cd1685e681a875f754fff49c8e20b67903f53d/BitwardenShared/UI/Auth/Extensions/AuthRouter%2BRedirects.swift#L33-L34

BitwardenShared/UI/Platform/Application/Extensions/UIViewController.swift
```diff
@@ -18,7 +18,7 @@ extension UIViewController {
     ) {
         var presentedChild = presentedViewController
         var availablePresenter: UIViewController? = self
-        while presentedChild != nil {
+        while presentedChild != nil, presentedChild?.isBeingDismissed == false {
```
matt-livefront, line 21: ℹ️ I was occasionally seeing a "VC tried to present on a VC not in the view hierarchy" error if the view controller was presented too quickly after the SSO `ASWebAuthenticationSession` is dismissed despite the 0.5 second delay that we have in place. This seemed to fix that from occurring.

BitwardenShared/UI/Auth/Login/SingleSignOn/SingleSignOnProcessor.swift
```diff
@@ -220,7 +221,6 @@ extension SingleSignOnProcessor: SingleSignOnFlowDelegate {
                     // Attempt to unlock the vault with tde.
                     try await services.authRepository.unlockVaultWithDeviceKey()
                     await coordinator.handleEvent(.didCompleteAuth)
-                    coordinator.navigate(to: .dismiss)
```
fedemkr, line 223: 🤔 Why is it not necessary to `dismiss` anymore? Is it handled by `didCompleteAuth`? Or this line https://github.com/bitwarden/ios/pull/2066/files#diff-5aabf9f289fa2c9bc5c8c922348ac48551b824e4aa91bcea6aaca7c7a354fce2R471?

matt-livefront: It depends 😄. The most common case is that `didCompleteAuth` will dismiss any presented views and then navigate you to the vault. That happens as part of swapping the root view controller [here](https://github.com/bitwarden/ios/blob/main/BitwardenShared/UI/Platform/Application/Utilities/RootViewController.swift#L21). If you complete auth, but then need to update your master password, the auth router will present the update master password view on top of this one.

I think it's easier to let the router/coordinator handle dismissals in these auth flows. One issue I ran into was if you complete auth, this processor tells the coordinator to dismiss, but then the router is presenting the update master password view. In this case, the dismissal here was actually dismissing the update master password screen. So we have a bit more control if we dismiss from the coordinator/router as opposed to individual processors when there could be different places that are navigated from here.

The dismissal in `showLanding()` that you linked is primarily used if you log out from `UpdateMasterPasswordProcessor`: https://github.com/bitwarden/ios/pull/2066/files#diff-5443cbfbef761fa47c7c5e50eeab9d3cc5a8ca4826a1a508090e571f32faf154R94 or other views. Any time you would be navigating back to the landing page from a presented view.

fedemkr: Awesome thanks for the explanation Matt, makes total sense and I agree on letting the router/coordinator handle dismissals on these flows 👍

PR #1939, opened 2025-09-16
[PM-24290] Improve autofill cipher lists performance

fedemkr:
## 🎟️ Tracking

[PM-24290](https://bitwarden.atlassian.net/browse/PM-24290)

## 📔 Objective

Improve autofill cipher lists performance on `.passwords` and `.combinedSingleSection` modes by applying the refactor to use the data/section builders.

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

[PM-24290]: https://bitwarden.atlassian.net/browse/PM-24290?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2091, opened 2025-10-31
[PM-27190] Improve performance on Autofill Passwords + Fido2 list

fedemkr:
## 🎟️ Tracking

[PM-27190](https://bitwarden.atlassian.net/browse/PM-27190)

## 📔 Objective

Refactors the ciphers autofill publisher to use the new approach for the `.combinedMultipleSections` flow (autofilling Fido2 + Passwords credentials shown in different sections).
This is being done in order to improve the overall performance of the list loading, which will load faster and consume less memory given the credentials are being decrypted and filtered in bulk. This is similar to #1939.

⛏️ Also fixed some warnings in the project.

> [!IMPORTANT]  
> Added a `WORKAROUND` when getting the URI on this flow as Apple doesn't seem to be sending the service identifiers thus we can't get the URI from them to filter the vault. Therefore the **relying party identifier** is taken as the URI to filter in order to fix this scenario until Apple fixes the issue. Check `CredentialProviderContext -> uri` (it was moved there in order to add tests)

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-27190]: https://bitwarden.atlassian.net/browse/PM-27190?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenShared/Core/Autofill/Utilities/CredentialProviderContext/CredentialProviderContext.swift
```diff
@@ -89,6 +91,33 @@ public struct DefaultCredentialProviderContext: CredentialProviderContext {
...
+            // If at some point the OS fixes that problem, then this would be automatically sorted out
+            // as it wouldn't enter this block of the flow and use the previous working approach
+            // of getting it from the `serviceIdentifiers`.
+            if case let .autofillFido2VaultList(_, passkeyParameters) = extensionMode,
+               !passkeyParameters.relyingPartyIdentifier.isEmpty {
+                return passkeyParameters.relyingPartyIdentifier
+            }
+
+            return nil
+        }
```
fedemkr, line 110: 🤔  Even though Claude has a point in here about adding a ticket TODO to clear this; I may be inclined to leave this code there in case Apple fixes the issue and then breaks it again in a future release. It seems like a good fallback when the `uri` is not provided from the `service identifiers`. What do you think? @KatherineInCode @matt-livefront

KatherineInCode: I think it's fine not having an explicit TODO for this sort of thing. I feel like we've had "check if this is still broken if we're on a future version of iOS" comments before 🤔

BitwardenShared/Core/Autofill/Utilities/CredentialProviderContext/CredentialProviderContextTests.swift
```diff
@@ -294,6 +294,96 @@ class CredentialProviderContextTests: BitwardenTestCase {
...
+        let subject = DefaultCredentialProviderContext(.autofillVaultList(identifiers))
+        XCTAssertEqual(subject.uri, "https://first.com")
+    }
+
+    /// `getter:uri` returns relying party identifier as fallback for autofillFido2VaultList
+    /// when service identifiers are empty.
+    func test_uri_autofillFido2VaultList_relyingPartyFallback() {
+        let parameters = MockPasskeyCredentialRequestParameters(relyingPartyIdentifier: "passkey.example.com")
+        let subject = DefaultCredentialProviderContext(.autofillFido2VaultList([], parameters))
+        XCTAssertEqual(subject.uri, "passkey.example.com")
```
matt-livefront: ❓ In the case of `ASCredentialServiceIdentifier` with type `domain`, we prepend "https://" to the domain. Should we do the same here?

BitwardenShared/Core/Vault/Helpers/VaultListSectionsBuilder.swift
```diff
@@ -110,6 +115,32 @@ class DefaultVaultListSectionsBuilder: VaultListSectionsBuilder {
...
+            ))
+        }
+
+        if !preparedData.fido2Items.isEmpty, let rpID {
+            vaultListData.sections.append(VaultListSection(
+                id: Localizations.passkeysForX(rpID),
+                items: preparedData.fido2Items.sorted(using: VaultListItem.defaultSortDescriptor),
+                name: Localizations.passkeysForX(rpID),
+            ))
+        }
```
matt-livefront: 🤔 It looks like this adds "Password for X" as the first section with "Passkeys for X" as the second section. Previously, those were reversed, with "Passkeys for X" at the top. Is reversing these expected?

BitwardenShared/Core/Autofill/Utilities/CredentialProviderContext/CredentialProviderContextTests.swift
fedemkr: I think it's not required but perhaps I'm missing thinking about some filter scenario where that could make some trouble. I'll add it; nice catch!

PR #2101, opened 2025-11-04
[PM-26063] Consolidate ViewInspector test helpers

matt-livefront:
## 🎟️ Tracking

[PM-26063](https://bitwarden.atlassian.net/browse/PM-26063)

## 📔 Objective

Consolidates the `InspectableView` `ViewInspector` test helper into a shared `ViewInspectorTestHelper` framework.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-26063]: https://bitwarden.atlassian.net/browse/PM-26063?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/project-bwk.yml
+++ b/project-bwk.yml
@@ -294,8 +294,6 @@ targets:
         GENERATE_INFOPLIST_FILE: YES
     sources:
       - path: ViewInspectorTestHelpers
-        excludes:
-          - "**/Info.plist"
     dependencies:
       - package: ViewInspector
 


project-bwk.yml
```diff
@@ -284,4 +285,17 @@ targets:
...
+    type: framework
+    platform: iOS
+    settings:
+      base:
+        ENABLE_TESTING_SEARCH_PATHS: YES
+        GENERATE_INFOPLIST_FILE: YES
+    sources:
+      - path: ViewInspectorTestHelpers
+        excludes:
+          - "**/Info.plist"
```
fedemkr: 🤔 Should we generate the Info.plist and then exclude it? Would it be the same not to generate it?

KatherineInCode: ~~Frameworks require an info.plist file, at least; I think this excludes it from sources like `TestHelpers` does, though I don't think that's strictly necessary 🤔 it just keeps it from appearing in Xcode~~

Oh, I see. Hm. That I'm less familiar with—that's what I get for looking before my morning tea 😅

matt-livefront: In other places (like TestHelpers) we include a generic `Info.plist` in the folder but then exclude it from sources. Excluding it from sources probably isn't necessary, since Xcodegen should recognize it as an `Info.plist` file.

As an alternative to creating an `Info.plist` file, I enabled `GENERATE_INFOPLIST_FILE` which lets Xcode generate an `Info.plist` file from the project file. Since we don't need to customize the `Info.plist` file, this seems simpler. In this case, there is no `Info.plist` file on disk (in the repo at least), so excluding it really does nothing. The excludes line is something I copied over but never removed.

@fedemkr Are you suggesting removing `excludes: "**/Info.plist"` or something else?

fedemkr: Makes sense. Yes, perhaps we could remove `excludes: "**/Info.plist"` if it's not necessary then; so the `yml` is cleaner.

KatherineInCode: I think the balance is whether we want the yml cleaner (by not having the exclusion line) or Xcode cleaner (by excluding the Info plist there)

matt-livefront: @KatherineInCode I think `GENERATE_INFOPLIST_FILE: YES` is the best of both worlds. Clean yml and there's no Info.plist in the repo or shown in Xcode's navigator.

KatherineInCode (approved):

PR #2130, opened 2025-11-12
Add information about mocks to Claude configuration

KatherineInCode:
## 📔 Objective

I have noticed that Claude often generates comments that states that mock files aren't in the `BitwardenKitMocks` (etc.) target, but are instead in a `Mocks` folder. This is because Claude has insufficient training data to understand our Xcodegen configurations.

This attempts to update the Claude configurations to specify that we put mock files in `Mocks` directories, and also to call out our preferred usage of Sourcery for mock generation.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -38,7 +38,7 @@ Core directives for maintaining code quality and consistency in the Bitwarden iO
 1. Follow existing code style in surrounding files
 2. Write tests alongside implementation
    1. Prefer using Sourcery with the `AutoMockable` attribute for creating mocks.
-   2. If a file needs to be created for a mock, it is put in a `Mocks` folder, as our Xcodegen configuration will put that in the correct Xcode target.
+   2. If a manual mock implementation is needed (not using Sourcery), place it in a `Mocks` folder within the appropriate framework directory, as our Xcodegen configuration will automatically include it in the correct test target.
 3. Add DocC to everything except protocol implementations and mocks
 4. Validate against architecture guidelines
 


.claude/CLAUDE.md
```diff
@@ -37,6 +37,8 @@ Core directives for maintaining code quality and consistency in the Bitwarden iO
 
 1. Follow existing code style in surrounding files
 2. Write tests alongside implementation
+   1. Prefer using Sourcery with the `AutoMockable` attribute for creating mocks.
+   2. If a manual mock implementation is needed (not using Sourcery), place it in a `Mocks` folder within the appropriate framework directory, as our Xcodegen configuration will automatically include it in the correct test target.
```
fedemkr, line 41: 🤔 I think this is `true` only for mocks that are for `BitwardenKit` or `AuthenticatorBridgeKit` targets. However, on other targets like `BitwardenShared` or `AuthenticatorShared` we have no `Mocks` folder; mocks are under `TestHelpers` folders and belong to their tests targets. Perhaps we should make that explicit here as well as it might create `Mocks` folders in those targets to follow this rule which would not be the current approach.

KatherineInCode: Yeah, I don't love how this is phrased or implemented, especially as we *do* have other places we talk about mocks, and it might be better to put it all there.

Long-term, I'd like to migrate all the things into `Mocks` folders for consistency, though.

fedemkr: Yes, I agree we should have the same approach throughout the projects in the future. However for the time being I believe this should be updated with today's state of where we put mocks so Claude is not confused on where the mocks are per each target/project.

PR #2132, opened 2025-11-12
[PM-23729] Refactor searches to use new approach with vault list builders

fedemkr:
## 🎟️ Tracking

[PM-23729](https://bitwarden.atlassian.net/browse/PM-23729)

## 📔 Objective

Refactor the search lists code (both in main app as in Autofill extension flows) to use the new approach with the vault list builders and director strategies.
This will improve the performance on the usage of the searches and how we decrypt ciphers for them.
This also removes a lot of code from the `VaultRepository` and its tests given that everything has been refactored to the new approach.
Additionally, no need for vault list publishers specific for searches, we can use `vaultListPublisher` directly now given that I've moved the `searchText` filter to the `VaultListFilter` to easily pass it around and reuse new builders structure.

Furthermore, all folding operations with `diacriticInsensitive` options have locale `.current` now to consolidate how we do that throughout the app as we were sometimes using `nil` (system locale) and sometimes `.current`. So we decided to go with `.current` as it's the current locale the user is having in their app as it may differ from the system one.

> [!NOTE]  
> There will be a future PR to add debouncing to the search text changes to further improve the behavior.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-23729]: https://bitwarden.atlassian.net/browse/PM-23729?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/Core/Vault/Helpers/VaultListDataPreparator.swift
+++ b/BitwardenShared/Core/Vault/Helpers/VaultListDataPreparator.swift
@@ -255,7 +255,7 @@ struct DefaultVaultListDataPreparator: VaultListDataPreparator {
                 return
             }
 
-            if filter.addTOTPGroup {
+            if filter.options.contains(.addTOTPGroup) {
                 preparedDataBuilder = await preparedDataBuilder.incrementTOTPCount(cipher: decryptedCipher)
             }
 
--- a/BitwardenShared/Core/Vault/Repositories/VaultRepository.swift
+++ b/BitwardenShared/Core/Vault/Repositories/VaultRepository.swift
@@ -791,8 +791,6 @@ extension DefaultVaultRepository: VaultRepository {
         case .all:
             try await vaultListPublisher(
                 filter: VaultListFilter(
-                    addTOTPGroup: false,
-                    addTrashGroup: false,
                     filterType: .allVaults,
                     group: group,
                 ),
@@ -800,8 +798,6 @@ extension DefaultVaultRepository: VaultRepository {
         case .combinedMultipleSections, .combinedSingleSection, .passwords:
             try await vaultListPublisher(
                 filter: VaultListFilter(
-                    addTOTPGroup: false,
-                    addTrashGroup: false,
                     filterType: .allVaults,
                     mode: mode,
                     rpID: rpID,
@@ -811,8 +807,6 @@ extension DefaultVaultRepository: VaultRepository {
         case .totp:
             try await vaultListPublisher(
                 filter: VaultListFilter(
-                    addTOTPGroup: false,
-                    addTrashGroup: false,
                     group: .totp,
                 ),
             )
--- a/BitwardenShared/UI/Vault/Vault/VaultItemSelection/VaultItemSelectionProcessor.swift
+++ b/BitwardenShared/UI/Vault/Vault/VaultItemSelection/VaultItemSelectionProcessor.swift
@@ -238,22 +238,11 @@ class VaultItemSelectionProcessor: StateProcessor<
                 filter: VaultListFilter(
                     filterType: .allVaults,
                     group: .login,
+                    options: [.isInPickerMode],
                     searchText: searchName,
                 ),
             ) {
-                guard let items = vaultListData.sections.first?.items,
-                      !items.isEmpty else {
-                    state.vaultListSections = []
-                    continue
-                }
-
-                state.vaultListSections = [
-                    VaultListSection(
-                        id: Localizations.matchingItems,
-                        items: items,
-                        name: Localizations.matchingItems,
-                    ),
-                ]
+                state.vaultListSections = vaultListData.sections
             }
         } catch {
             coordinator.showAlert(.defaultAlert(title: Localizations.anErrorHasOccurred))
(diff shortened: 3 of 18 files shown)


BitwardenShared/Core/Vault/Helpers/CipherMatchResult.swift
```diff
@@ -0,0 +1,13 @@
...
+enum CipherMatchResult {
+    /// The cipher is an exact match for the "query".
+    case exact
+
+    /// The cipher is a close match for the "query".
+    case fuzzy
+
+    /// The cipher doesn't match the "query".
+    case none
+}
```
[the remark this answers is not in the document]
fedemkr: I don't agree with this @claude. This is just the cipher match result enum and should not know about the internal implementations of how the search operation determines whether something is `exact` `fuzzy` or `none` as it goes beyond the scope of information this enum should have. Additionally, different search strategies could exist which could mark something as `fuzzy` where in others could be just `none` so I don't see the need to add documentation on implementation behavior details here.

BitwardenShared/Core/Vault/Helpers/VaultListDataPreparator.swift
```diff
@@ -106,15 +130,15 @@ struct DefaultVaultListDataPreparator: VaultListDataPreparator {
         let cipherMatchingHelper = await cipherMatchingHelperFactory.make(uri: uri)
 
         var preparedDataBuilder = vaultListPreparedDataBuilderFactory.make()
-        let restrictedOrganizationIds = await prepareRestrictedOrganizationIds(builder: preparedDataBuilder)
 
-        await ciphersClientWrapperService.decryptAndProcessCiphersInBatch(
+        await decryptAndProcessCiphersInBatch(
```
matt-livefront, line 134: 👍 I like the simplification of moving some of this shared logic into `decryptAndProcessCiphersInBatch`.

BitwardenShared/Core/Vault/Helpers/VaultListDataPreparator.swift
```diff
@@ -319,3 +422,26 @@ struct DefaultVaultListDataPreparator: VaultListDataPreparator {
...
     }
 }
+
+// MARK: VaultListPreparedDataBuilder
+
+private extension VaultListPreparedDataBuilder {
+    func addItemsForCombinedMultipleSections(
+        decryptedCipher: CipherListView,
+        withFido2Credentials fido2Credentials: [CipherView]?,
+    ) async -> VaultListPreparedDataBuilder {
```
matt-livefront, line 432: 🤔 Should this be here in an extension or was it meant to be in `VaultListPreparedDataBuilder`?

BitwardenShared/Core/Vault/Repositories/VaultRepository.swift
```diff
@@ -304,29 +271,11 @@ extension VaultRepository {
...
-            searchText: searchText,
-            group: nil,
-            filter: filter,
-        )
-    }
 }
 
 /// A default implementation of a `VaultRepository`.
 ///
-class DefaultVaultRepository { // swiftlint:disable:this type_body_length
```
matt-livefront, line 329: 🎉

BitwardenShared/UI/Vault/Vault/VaultItemSelection/VaultItemSelectionProcessor.swift
```diff
@@ -231,12 +234,15 @@ class VaultItemSelectionProcessor: StateProcessor<
...
+            for try await vaultListData in try await services.vaultRepository.vaultListPublisher(
+                filter: VaultListFilter(
+                    filterType: .allVaults,
+                    group: .login,
+                    searchText: searchName,
+                ),
             ) {
-                guard !items.isEmpty else {
+                guard let items = vaultListData.sections.first?.items,
+                      !items.isEmpty else {
```
matt-livefront: 🤔 Could we refactor this (not necessarily as part of this) to assign `vaultListData.sections` directly to `state.vaultListSections` without removing the first items from the section and rebuilding a `VaultListSection`? It looks like we use `Localizations.matchingItems` as the section title, so maybe that adds a wrinkle to this?

BitwardenShared/Core/Vault/Helpers/VaultListDataPreparator.swift
fedemkr: I intended to do this here as it was just a "helper" function to combine the same logic applied in the list as in the search for `.combinedMultipleSections` flow. However, if you feel this should be in the builder I can definitely move it there.

matt-livefront: I can see the argument for keeping it here, so I think it's fine. 👍

BitwardenShared/UI/Vault/Vault/VaultItemSelection/VaultItemSelectionProcessor.swift
fedemkr: At the moment didn't do that for that reason but revisit it and ended up refactoring the  `VaultListFilter` to have an `OptionSet` property where we can pass this kind of stuff alongside moving some flags there.

PR #2153, opened 2025-11-19
[PM-28540] Simplify project files with Xcodegen templates

matt-livefront:
## 🎟️ Tracking

[PM-28540](https://bitwarden.atlassian.net/browse/PM-28540)

## 📔 Objective

When I added Sourcery to BitwardenKit (https://github.com/bitwarden/ios/pull/2136), I wondered if we could simplify some of the duplication in our project files.

This is an attempt at using Xcodegen's [templates](https://github.com/yonaskolb/XcodeGen/blob/master/Docs/ProjectSpec.md#target-template) to share some common logic between targets.

There's two categories of templates that I've added:

- "Sources" templates: a template that contains our common regex patterns for the source files used in different types of targets:
  - `CommonTarget`: this is meant for app or framework targets and mainly includes all files other than tests and test helpers (mocks, fixtures, etc).
  - `MocksTarget`: a target containing mocks and fixtures.
  - `TestTarget`: a unit testing target which excludes ViewInspector and snapshot tests.
  - `SnapshotTestTarget`: a test target for snapshot tests.
  - `ViewInspectorTestTarget`: a test target for ViewInspector tests.
- "Build phase" templates:
  - `SourceryTarget`: a template for a target that needs a Sourcery pre-build script.

The templates are defined in [project-common.yml](https://github.com/bitwarden/ios/blob/51342b0eb8f1e4ad710ac4ddcc9cc59f83d6046b/project-common.yml#L28).

Going forward, adding Sourcery to a target is simplified to adding `SourceryTarget` to the list of a target's templates (and adding the generated file to any test targets):

```diff
diff --git a/project-bwk.yml b/project-bwk.yml
--- a/project-bwk.yml
+++ b/project-bwk.yml
@@ -121,4 +121,5 @@
     templates:
       - CommonTarget
+      - SourceryTarget
     templateAttributes:
       sourcesPath: BitwardenKit
```

<details><summary>Diff of the project files before and after:</summary>
<p>

The only noticeable change here of the project files is the removal of the `gitignore` files, which I intentionally excluded. So even though this changes a bunch of stuff, I hope I've minimized any potential issues 🤞. 

```diff
diff -U0 before/Bitwarden.pbxproj Bitwarden.xcodeproj/project.pbxproj
--- before/Bitwarden.pbxproj	2025-11-18 16:36:29
+++ Bitwarden.xcodeproj/project.pbxproj	2025-11-19 09:25:25
@@ -3158 +3157,0 @@
-		20E4EE0C07D3EE8586C46233 /* .gitignore */ = {isa = PBXFileReference; path = .gitignore; sourceTree = "<group>"; };
@@ -4607 +4605,0 @@
-		A6C964FC3085FA0C093F0C63 /* .gitignore */ = {isa = PBXFileReference; path = .gitignore; sourceTree = "<group>"; };
@@ -7463 +7460,0 @@
-				20E4EE0C07D3EE8586C46233 /* .gitignore */,
@@ -8724 +8720,0 @@
-				A6C964FC3085FA0C093F0C63 /* .gitignore */,

diff -U0 before/Authenticator.pbxproj Authenticator.xcodeproj/project.pbxproj
--- before/Authenticator.pbxproj	2025-11-18 16:36:53
+++ Authenticator.xcodeproj/project.pbxproj	2025-11-19 09:25:25
@@ -980 +979,0 @@
-		3CEC58C44DB13461B95880A5 /* .gitignore */ = {isa = PBXFileReference; path = .gitignore; sourceTree = "<group>"; };
@@ -1008 +1006,0 @@
-		4D0217410E4A8906F80DEDD9 /* .gitignore */ = {isa = PBXFileReference; path = .gitignore; sourceTree = "<group>"; };
@@ -1209 +1206,0 @@
-		C203B0C946484B15F519EFE6 /* .gitignore */ = {isa = PBXFileReference; path = .gitignore; sourceTree = "<group>"; };
@@ -1472 +1468,0 @@
-				3CEC58C44DB13461B95880A5 /* .gitignore */,
@@ -1528 +1523,0 @@
-				C203B0C946484B15F519EFE6 /* .gitignore */,
@@ -2278 +2272,0 @@
-				4D0217410E4A8906F80DEDD9 /* .gitignore */,

diff -U0 before/BitwardenKit.pbxproj BitwardenKit.xcodeproj/project.pbxproj
--- before/BitwardenKit.pbxproj	2025-11-18 16:36:16
+++ BitwardenKit.xcodeproj/project.pbxproj	2025-11-19 09:25:23
@@ -1309 +1308,0 @@
-		DBBC174DA9DFDD9E2F708A65 /* .gitignore */ = {isa = PBXFileReference; path = .gitignore; sourceTree = "<group>"; };
@@ -1669 +1667,0 @@
-				DBBC174DA9DFDD9E2F708A65 /* .gitignore */,
```

</p>
</details> 

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-28540]: https://bitwarden.atlassian.net/browse/PM-28540?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/project-common.yml
+++ b/project-common.yml
@@ -46,13 +46,19 @@ targetTemplates:
           - "**/__Snapshots__/*"
           - "**/GoogleService-Info.*.plist"
         buildPhase: none
+  # Adds common include/excludes for mocks targets.
+  MocksTarget:
+    sources:
+      - path: ${sourcesPath}
+        includes:
+          - "**/Fixtures/*"
+          - "**/Mocks/*"
   # Adds common include/excludes for snapshot test targets.
   SnapshotTestTarget:
     sources:
       - path: ${sourcesPath}
         includes:
           - "**/*SnapshotTests.*"
-          - "**/Fixtures/*"
           - "**/TestHelpers/*"
   # Adds the Sourcery pre-build script.
   SourceryTarget:
@@ -78,7 +84,6 @@ targetTemplates:
           - "**/*ViewInspectorTests.*"
         includes:
           - "**/*Tests.*"
-          - "**/Fixtures/*"
           - "**/TestHelpers/*"
   # Adds common include/excludes for ViewInspector test targets.
   ViewInspectorTestTarget:
@@ -86,5 +91,4 @@ targetTemplates:
       - path: ${sourcesPath}
         includes:
           - "**/*ViewInspectorTests.*"
-          - "**/Fixtures/*"
           - "**/TestHelpers/*"
--- a/project-bwa.yml
+++ b/project-bwa.yml
@@ -203,11 +203,11 @@ targets:
       base:
         ENABLE_TESTING_SEARCH_PATHS: YES
         GENERATE_INFOPLIST_FILE: YES
+    templates:
+      - MocksTarget
+    templateAttributes:
+      sourcesPath: AuthenticatorShared
     sources:
-      - path: AuthenticatorShared
-        includes:
-          - "**/Fixtures/*"
-          - "**/Mocks/*"
       - path: AuthenticatorShared/Sourcery/Generated/AutoMockable.generated.swift
         optional: true
     dependencies:
--- a/project-bwk.yml
+++ b/project-bwk.yml
@@ -99,11 +99,10 @@ targets:
         ENABLE_TESTING_SEARCH_PATHS: YES
         INFOPLIST_FILE: AuthenticatorBridgeKit/MocksInfo.plist
         SWIFT_STRICT_CONCURRENCY: complete
-    sources:
-      - path: AuthenticatorBridgeKit
-        includes:
-          - "**/Fixtures/*"
-          - "**/Mocks/*"
+    templates:
+      - MocksTarget
+    templateAttributes:
+      sourcesPath: AuthenticatorBridgeKit
     dependencies:
       - target: AuthenticatorBridgeKit
       - target: BitwardenKit
@@ -135,11 +134,11 @@ targets:
       base:
         ENABLE_TESTING_SEARCH_PATHS: YES
         INFOPLIST_FILE: BitwardenKit/Application/Mocks/Support/Info.plist
+    templates:
+      - MocksTarget
+    templateAttributes:
+      sourcesPath: BitwardenKit
     sources:
-      - path: BitwardenKit
-        includes:
-          - "**/Fixtures/*"
-          - "**/Mocks/*"
       - path: BitwardenKit/Sourcery/Generated/AutoMockable.generated.swift
         optional: true
     dependencies:
--- a/project-pm.yml
+++ b/project-pm.yml
@@ -376,11 +376,11 @@ targets:
       base:
         ENABLE_TESTING_SEARCH_PATHS: YES
         GENERATE_INFOPLIST_FILE: YES
+    templates:
+      - MocksTarget
+    templateAttributes:
+      sourcesPath: BitwardenShared
     sources:
-      - path: BitwardenShared
-        includes:
-          - "**/Fixtures/*"
-          - "**/Mocks/*"
       - path: BitwardenShared/Sourcery/Generated/AutoMockable.generated.swift
         optional: true
     dependencies:


project-common.yml
```diff
@@ -25,3 +25,70 @@ packages:
...
+          - "**/*Tests.*"
+          - "**/.gitignore"
+          - "**/Fixtures/*"
+          - "**/GoogleService-Info.*.plist"
+          - "**/Mocks/*"
+          - "**/Sourcery/Generated/*"
+          - "**/TestHelpers/*"
+          - "**/Tests/*"
+          - "**/__Snapshots__/*"
+          - "**/sourcery.yml"
```
fedemkr, line 43: 🤔 Would it make sense to order this alphabetically? I think it's easier then to check if a rule exists.

project-common.yml
```diff
@@ -25,3 +25,70 @@ packages:
...
+          - "**/__Snapshots__/*"
+          - "**/GoogleService-Info.*.plist"
+        buildPhase: none
+  # Adds common include/excludes for mocks targets.
+  MocksTarget:
+    sources:
+      - path: ${sourcesPath}
+        includes:
+          - "**/Fixtures/*"
+          - "**/Mocks/*"
```
fedemkr, line 55: 🤔 Should the auto generated Sourcery optional file be added here as well?
```
- path: ${sourcesPath}/Sourcery/Generated/AutoMockable.generated.swift
        optional: true
```

project-common.yml
```diff
@@ -25,3 +25,70 @@ packages:
   ViewInspector:
     url: https://github.com/nalexn/ViewInspector
     exactVersion: 0.10.3
+targetTemplates:
+  # Adds common include/excludes for app and framework targets.
+  CommonTarget:
+    sources:
+      - path: ${sourcesPath}
+        excludes:
```
fedemkr, line 33: 🤔 I think the `- "**/swiftgen-{app-abbreviated-name}.yml" may be missing here as it was deleted from the other `yml` excludes but was not added here.

fedemkr (commented):
Amazing work Matt, love it! 🎉 👏 
Just have a few questions.

project-common.yml
matt-livefront: What would you propose for ordering this? I think it's already ordered lexicographically based on ASCII codes. I wasn't sure the best way to handle symbols and lower vs upper case, so that made the most sense to me, but I'm open to other ordering suggestions.

project-common.yml
matt-livefront: It didn't seem to make a difference in the generated project file and it's not currently being added as a resource file. I think Xcodegen will ignore YAML files, but I can add it back to be safe.

project-common.yml
matt-livefront: The only issue with this is that not all targets using the `MocksTarget` template integrate Sourcery. `AuthenticatorBridgeKitMocks` doesn't use Sourcery (yet), which causes the build to fail if that file doesn't exist. I guess we could add Sourcery for that target or keep it as-is for now. Do you have a preference?

project-common.yml
matt-livefront: Ah, I'm remember why I did this. The `swiftgen-*.yml` files were duplicated in the sources list (both as an exclude and with a build phase of `none`, so I removed the duplicate. But they still exist:

https://github.com/bitwarden/ios/blob/d553a80abb954614f81c6d27961aa6bbcd2ebb78/project-bwa.yml#L96-L97
https://github.com/bitwarden/ios/blob/d553a80abb954614f81c6d27961aa6bbcd2ebb78/project-pm.yml#L166-L167

 The `swiftgen-*.yml` files are in the project's root directly, so I don't think they can be added as a wildcard to the templates. I think they need to be referenced directly, and that's probably better done from the respective target.

fedemkr: Oh missed that, thanks perfect then!

project-common.yml
fedemkr: Oh you're right, I got confused between the `- "**/*Tests.*"` and `- "**/Tests/*"` but yeah it has the asterisk. We can leave it as it's now. Thanks for checking it.

project-common.yml
fedemkr: I would like the bridge to follow the same pattern at some point but currently there are only 2 mocks in there, so applying it now may be a bit overkill.
Maybe we can revisit this if we have more work being done in the bridge. Does that sound good to you?

matt-livefront: Yep, I like that.

PR #2179, opened 2025-12-03
[PM-14031] Improve search memory performance

fedemkr:
## 🎟️ Tracking

[PM-14031](https://bitwarden.atlassian.net/browse/PM-14031)

## 📔 Objective

Improve the performance of searches throughout the app, specially regarding memory usage. This is particularly useful for iOS extensions where several searches in a row could potentially crash the extension given the memory limits (this also depends on the vault size).

- Moved filtering of whether the cipher is deleted and also the group check to be done as pre filter on encrypted ciphers so less ciphers get decrypted.
- Added a publisher for the `VaultListFilter` from the processor, so any time something from the filter like the search text query changes a new filter value gets published. Therefore each time the search starts, the publisher is subscribed only once for the three internal ones (filter, ciphers, available fido2 credentials) which saves a lot of memory piling up vs the previous approach.
  - A new `SearchProcessorMediator` has been added to handle the subscription, cancellation and handling search results with a delegate to share the logic amongst the different vault list processors.
- The `preparedData` now is cleaned after `build` returns in `DefaultVaultListSectionsBuilder` to make Swift clean the internal arrays faster. This has a large caveat given that I had to make it optional, so if the caller calls any of the methods after `build` nothing will be done; which is kind of OK as nothing should be called after `build` in the builder but something to raise up. 🤔 I could add a DEBUG warning/assertion here just in case it happens.

> [!NOTE]  
> These changes could also be done to other cipher list flows but this PR only focuses on searches to decrease the testing surface. Additional tasks track the necessary changes on other parts of the app.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-14031]: https://bitwarden.atlassian.net/browse/PM-14031?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenShared/Core/Vault/Helpers/VaultListSectionsBuilder.swift
```diff
@@ -101,7 +101,7 @@ class DefaultVaultListSectionsBuilder: VaultListSectionsBuilder { // swiftlint:d
     /// The service used by the application to report non-fatal errors.
     let errorReporter: ErrorReporter
     /// Vault list data prepared to  be used by the builder.
-    let preparedData: VaultListPreparedData
+    var preparedData: VaultListPreparedData?
```
matt-livefront: 🤔 Instead of making this optional/nil after calling `build`, could you set it to a new instance of the struct? That would avoid the need for adding the `guard`s. The downside being you wouldn't be able to differentiate an initially empty prepared data struct and one that has already been "built", but maybe that isn't a needed requirement.

BitwardenShared/UI/Vault/Vault/AutofillList/VaultAutofillListProcessor.swift
```diff
@@ -94,8 +99,13 @@ class VaultAutofillListProcessor: StateProcessor<// swiftlint:disable:this type_
         )
         self.coordinator = coordinator
         self.services = services
+        searchProcessorMediator = services.searchProcessorMediatorFactory.make()
+
         super.init(state: state)
 
+        searchProcessorMediator.setDelegate(self)
+        searchProcessorMediator.setAutofillListMode(autofillListMode)
```
matt-livefront: 🎨 What do you think about providing the delegate and list mode parameters to `make()` rather than calling separate functions? It seems like that would help enforce that those are required (if they are) and prevent them from being missed.

BitwardenShared/UI/Vault/Utilities/SearchProcessorMediator.swift
```diff
@@ -0,0 +1,168 @@
...
+    /// Deinitializes this processor mediator.
+    deinit {
+        vaultSearchListSubscriptionTask?.cancel()
+        vaultSearchListSubscriptionTask = nil
+        searchProcessorMediatorDelegate = nil
+    }
+
+    // MARK: Methods
+
+    func onFilterChanged(_ filter: VaultListFilter) {
```
matt-livefront: 🎨 The method name `onFilterChanged(_:)` feels like it would be used to provide a closure for when the filter changes. In this case, it's setting the filter. Would it be better to name this `setFilter(_:)` similar to `setAutofillListMode(_:)`?

BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessor.swift
```diff
@@ -72,13 +76,19 @@ final class VaultListProcessor: StateProcessor<
...
+
         super.init(state: state)
+
+        searchProcessorMediator.setDelegate(self)
     }
 
     deinit {
         reviewPromptTask?.cancel()
+
+        searchProcessorMediator.stopSearching()
```
matt-livefront: 🤔 How critical is it that `stopSearching()` is called in `deinit`? I noticed that this is missing from `VaultGroupProcessor` and `VaultItemSelectionProcessor`. On one hand, I think it's good to be explicit, but on the other hand, this is easy to miss. Given that `SearchProcessorMediator` also cancels the task in `deinit`, I wonder if we need the extra logic to explicitly stop searching, or if we could rely on deinit to cancel the task?

BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessor.swift
```diff
@@ -654,3 +659,12 @@ extension VaultListProcessor: ProfileSwitcherHandler {
...
     }
 }
+
+// MARK: - SearchProcessorMediatorDelegate
+
+extension VaultListProcessor: SearchProcessorMediatorDelegate {
+    func onNewSearchResults(data: VaultListData) {
+        let items = data.sections.first?.items ?? []
+        state.searchResults = items
+    }
```
matt-livefront: 🤔 Given that the delegate is only a single method, and there's minimal logic in the delegate, would it be simpler to pass a closure to `SearchProcessorMediator` instead of using delegation? Using delegation has a few downsides in that you have to 1) remember to set the delegate and 2) it separates the logic associated with `SearchProcessorMediator`. If you foresee other methods being added though, then I could see delegation making more sense.

BitwardenShared/UI/Vault/Utilities/SearchProcessorMediator.swift
```diff
@@ -0,0 +1,168 @@
...
+            guard let self else { return }
+
+            do {
+                let publisher = try await vaultRepository.vaultSearchListPublisher(
+                    mode: autofillListMode,
+                    filterPublisher: vaultListFilterPublisher,
+                )
+
+                for try await vaultListData in publisher {
+                    guard !Task.isCancelled else { break }
```
matt-livefront: 🤔 It looks like there's a memory leak after searching from the group view and then returning to the list view. I'm wondering if it's caused by this strong self reference here. Maybe it's because of the missing `stopSearching()` in the processor's deinit? But I also wonder if it'd be better to keep the self reference weak rather than converting it to a strong self immediately for the duration of the task, so it doesn't create a retain cycle.

matt-livefront (commented):
This is great, just have a few suggestions and a potential memory leak that I noticed.

BitwardenShared/UI/Vault/Vault/AutofillList/VaultAutofillListProcessor.swift
fedemkr: I started doing it that way but it was a bit tricky to get the `autofillListMode` as it needs the `super.init` to be called (same happens with delegate `self`). Whereas if I put the `factory.make` after the `super.init` it errors given that `searchProcessorMediator` has not been initialized before.
I could get the mode from `(appExtensionDelegate as? AutofillAppExtensionDelegate)?.autofillListMode ?? .passwords` but I didn't think it was clean to repeat that logic in the init when we have the computed property. Moreover, doing it in the coordinator would end up in something similar repeating such logic.
Additionally, the list mode is not required in all cases; it's only used in this processor, so it felt right to have it as a set function can call it when needed.

BitwardenShared/UI/Vault/Utilities/SearchProcessorMediator.swift
fedemkr: Makes sense, changing it to `updateFilter(_:)` to indicate that the same instance can/may update the filter several times as I see `set...` as it may indicate a one-time operation.

BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessor.swift
fedemkr: Removed as it's unnecessary per the reasons you mentioned.

BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessor.swift
fedemkr: I went for the delegate because I was thinking we could add some other functions like `onError(_:)` if we'd need to handle that particularly on the processors or some more particular features we could integrate in there if necessary.
By YAGNI I could go with your approach but because of a similar explanation like [the above comment](https://github.com/bitwarden/ios/pull/2179#discussion_r2624629868) we could forget to set the closure as it couldn't be set before `super.init`.
What do you think?

BitwardenShared/UI/Vault/Utilities/SearchProcessorMediator.swift
fedemkr: 🤔 I may be missing something but which strong `self` reference are you seeing? I marked it as weak [here](https://github.com/bitwarden/ios/pull/2179/changes#diff-71d71656775c8f1833a60e490c39a23b5e54d38ddc8c625378e78b7f49aeb8cfR105).

BitwardenShared/UI/Vault/Vault/AutofillList/VaultAutofillListProcessor.swift
matt-livefront: Ah, needing access to `self` makes sense why you did it the way you did. Hm, you could use a lazy var, but it's not as nice as just initializing everything in `init` 🤔 

```
private lazy var searchProcessorMediator: SearchProcessorMediator = {
    let searchProcessorMediator = services.searchProcessorMediatorFactory.make()
    searchProcessorMediator.setDelegate(self)
    searchProcessorMediator.setAutofillListMode(autofillListMode)
    return searchProcessorMediator
}()
```

Alternatively, maybe it makes sense in this case to provide the autofill mode and delegate/closure as parameters to `startSearching()`? It looks like both of those things are only used within the `startSearching()` function so that may simplify some things (no set method, no instance variable).

BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessor.swift
matt-livefront: I think providing a closure to handle search results as a parameter to `startSearching()` could be a good fit. That would require it from the API perspective and alleviate the need for accessing self before the processor is initialized. What do you think about that?

BitwardenShared/UI/Vault/Vault/AutofillList/VaultAutofillListProcessor.swift
fedemkr: At the very beginning I started like that and then changed to this approach. However, I see it simplifies a lot the structure/logic so I agree I'll make the change. Thanks for the feedback Matt! 🎉

BitwardenShared/UI/Vault/Utilities/SearchProcessorMediator.swift
matt-livefront: In this case, `[weak self]` only makes the capture weak at the moment the closure is created. Once `guard let self` runs, which will happen immediately once the task starts, the weak self reference is promoted to a strong one for the rest of the closure. 

Using `[weak self]`/`guard let self` works better for closures that won't run right away and don't contain long-running loops. If `self` is deallocated before the closure starts, then the guard will return, and the weak reference won't be promoted to a strong one. 

However, the task here will end up making a strong reference to self and then because it's subscribing to a publisher, the task is going to stay alive until the publisher completes, maintaining a strong reference the entire time.

I think you could restructure it like this to avoid a strong reference:

```
    func startSearching() {
        vaultSearchListSubscriptionTask?.cancel()
        vaultSearchListSubscriptionTask = Task { [weak self, autofillListMode, vaultListFilterPublisher] in
            do {
                let publisher = try await self?.vaultRepository.vaultSearchListPublisher(
                    mode: autofillListMode,
                    filterPublisher: vaultListFilterPublisher,
                )
                guard let publisher else { return }

                for try await vaultListData in publisher {
                    guard !Task.isCancelled else { break }

                    await self?.searchProcessorMediatorDelegate?.onNewSearchResults(data: vaultListData)
                }
            } catch {
                self?.errorReporter.log(error: error)
            }
        }
    }
```

fedemkr: Really nice catch, changed!

BitwardenShared/UI/Vault/Vault/VaultGroup/VaultGroupProcessor.swift
```diff
@@ -158,8 +164,10 @@ final class VaultGroupProcessor: StateProcessor<
             if !isSearching {
                 state.searchText = ""
                 state.searchResults = []
+                searchProcessorMediator.stopSearching()
                 searchTotpExpirationManager?.configureTOTPRefreshScheduling(for: [])
             }
+            searchProcessorMediator.startSearching(mode: nil, onNewSearchResults: searchResultsReceived)
```
matt-livefront: 🤔 I'm still seeing a memory leak with `searchProcessorMediator` because the `onNewSearchResults` closure is holding a implicit strong self reference.. I guess that was one benefit of the delegate, sorry. We could either use weak self in the closure or switch back to the delegate.


```suggestion
            searchProcessorMediator.startSearching(mode: nil) { [weak self] data in
                self?.searchResultsReceived(data: data)
            }
```

Image

fedemkr: Not sure if it's because I'm using Xcode 26.2 with iOS 26.1 simulator but can't reproduce the memory leak; the group processor doesn't appear after I close such view.
Nonetheless added the weak self there as it seems like the right approach to do there. Thanks for the catch!

matt-livefront: Did you also perform a search from the group view? That was key for reproducing it for me, since the closure isn't captured by the task until after a search is performed. Changes look great!

fedemkr: Oh I think I did the search on the main vault and missed searching again in the group. That surely was it. Nonetheless, tested again and now it doesn't leak 🎉 Thanks Matt!

PR #2181, opened 2025-12-04
[PM-28339] Convert data store publisher tests to use async iterators

matt-livefront:
## 🎟️ Tracking

[PM-28339](https://bitwarden.atlassian.net/browse/PM-28339)

## 📔 Objective

After making the changes in https://github.com/bitwarden/ios/pull/2165, some of the data store publisher tests would randomly fail. Since the data store publishers were using the background managed object context, this introduced a race condition in the tests. If the database update was made before the initial database results were published, only the most recent change would end up being published.

```
    waitFor(results.count == 2)
    XCTAssertEqual(results[0], initialItems) // this would fail because results[0] could already have `replacedItems`
    XCTAssertEqual(results[1], replacedItems)
```

I initially started adding more `waitFor` calls to wait for the initial results to be published. Converting these tests to use async iterators is similar to waiting for each result and results in a simpler and more efficient test.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-28339]: https://bitwarden.atlassian.net/browse/PM-28339?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/AuthenticatorBridgeKit/Tests/AuthenticatorBridgeItemServiceTests.swift
+++ b/AuthenticatorBridgeKit/Tests/AuthenticatorBridgeItemServiceTests.swift
@@ -355,7 +355,7 @@ final class AuthenticatorBridgeItemServiceTests: AuthenticatorBridgeKitTestCase 
         let initialItems = AuthenticatorBridgeItemDataView.fixtures().sorted { $0.id < $1.id }
         try await subject.insertItems(initialItems, forUserId: "userId")
 
-        var iterator = try await subject.sharedItemsPublisher().values.makeAsyncIterator()
+        var iterator = try await subject.sharedItemsPublisher().valuesWithTimeout().makeAsyncIterator()
 
         let firstValue = try await iterator.next()
         XCTAssertEqual(firstValue, initialItems)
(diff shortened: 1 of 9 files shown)


AuthenticatorBridgeKit/Tests/AuthenticatorBridgeItemServiceTests.swift
```diff
@@ -355,21 +355,15 @@ final class AuthenticatorBridgeItemServiceTests: AuthenticatorBridgeKitTestCase
...
+
+        let firstValue = try await iterator.next()
+        XCTAssertEqual(firstValue, initialItems)
 
         try await subject.replaceAllItems(with: [], forUserId: "userId")
 
-        waitFor(results.count == 2)
-        XCTAssertEqual(results[0], initialItems)
-        XCTAssertEqual(results[1], [])
+        let secondValue = try await iterator.next()
```
KatherineInCode, line 365: ❓ My concern with this is that there's no timeout. If a next value never comes (because someone broke it), then it just hangs. Is there a way we can wrap this in a timeout of some sort, to avoid that?

matt-livefront: Great idea! I added a publisher extension `valuesWithTimeout()` which can be used in place of `values` to stream the publisher's values. If no value is published within the timeout window, an error is thrown.

KatherineInCode (approved):

PR #2190, opened 2025-12-10
[PM-17241] Incorporate Bitwarden FIDO SDK/passkey-rs changes

iinuwa:
## 🎟️ Tracking

[PM-17241](https://bitwarden.atlassian.net/browse/PM-17241)

Depends on https://github.com/bitwarden/sdk-internal/pull/606.
## 📔 Objective

This addresses the breaking changes introduced by syncing our fork of passkey-rs with upstream.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-17241]: https://bitwarden.atlassian.net/browse/PM-17241?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/UI/Autofill/Utilities/Fido2UserInterfaceHelper.swift
+++ b/BitwardenShared/UI/Autofill/Utilities/Fido2UserInterfaceHelper.swift
@@ -67,10 +67,6 @@ protocol Fido2UserInterfaceHelper: Fido2UserInterface {
     /// Sets up the delegate to use on Fido2 user verification flows and to get information of the FIdo2 flow.
     /// - Parameter fido2UserInterfaceHelperDelegate: The delegate to use
     func setupDelegate(fido2UserInterfaceHelperDelegate: Fido2UserInterfaceHelperDelegate)
-
-    /// Sets up the user verification preference of the current Fido2 flow.
-    /// - Parameter userVerificationPreference: User verification preference to set up.
-    func setupCurrentUserVerificationPreference(userVerificationPreference: Uv)
 }
 
 // MARK: - DefaultFido2UserInterfaceHelper
@@ -93,11 +89,6 @@ class DefaultFido2UserInterfaceHelper: Fido2UserInterfaceHelper {
     )
     private(set) var fido2CreationOptions: BitwardenSdk.CheckUserOptions?
     private(set) var fido2CredentialNewView: BitwardenSdk.Fido2CredentialNewView?
-
-    /// The user verification preference of the current Fido2 flow.
-    /// This is needed as a workaround to be used in `isVerificationEnabled`
-    /// as currently we don't have the UV value provided by the SDK.
-    private var userVerificationPreference: Uv?
 
     /// Initializes a `DefaultFido2UserInterfaceHelper`.
     /// - Parameter fido2UserVerificationMediator: Mediator which manages user verification on Fido2 flows
@@ -187,18 +178,8 @@ class DefaultFido2UserInterfaceHelper: Fido2UserInterfaceHelper {
     }
 
     func isVerificationEnabled() -> Bool {
-        guard let userVerificationPreference else {
-            return false
-        }
-
-        switch userVerificationPreference {
-        case .discouraged:
-            return false
-        case .preferred:
-            return true
-        case .required:
-            return true
-        }
+        // TODO: PM-30454 We should consider subsuming this into the SDK, since this value is constant across all our current clients
+        true
     }
 
     func pickedCredentialForAuthentication(result: Result<CipherView, Error>) {
@@ -214,10 +195,6 @@ class DefaultFido2UserInterfaceHelper: Fido2UserInterfaceHelper {
         fido2UserVerificationMediator.setupDelegate(
             fido2UserVerificationMediatorDelegate: fido2UserInterfaceHelperDelegate,
         )
-    }
-
-    func setupCurrentUserVerificationPreference(userVerificationPreference: Uv) {
-        self.userVerificationPreference = userVerificationPreference
     }
 
     // MARK: Private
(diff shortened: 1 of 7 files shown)


BitwardenShared/UI/Autofill/Utilities/Fido2UserInterfaceHelper.swift
```diff
@@ -195,7 +195,7 @@ class DefaultFido2UserInterfaceHelper: Fido2UserInterfaceHelper {
         case .discouraged:
             return false
         case .preferred:
-            return await fido2UserVerificationMediator.isPreferredVerificationEnabled()
+            return true
```
fedemkr: ❓ Why does this return `true` at all times now?

iinuwa: Good question. This part of the class is in fulfillment of the PasskeyAuthenticator trait from passkey-rs. That is intended to be a static configuration for whether user verification would ever be available for an authenticator, not whether it's available for this specific call. (It essentially enables an early return from the library for processing UV.) Cf https://github.com/1Password/passkey-rs/pull/74#issuecomment-3422453500

This dynamic call should instead deferred to the `checkUser()` function, which we're still doing in `checkEnforceRequiredVerification()` at L234 (called from checkUser() at L144)

fedemkr: Oh alright, have you tested that the flows keep working as expected for `.preferred` UV? Particularly with session timeout -> **Never**.
If so, given that's the case then in the same file this should be removed alongside with their calls as part of this change as they're not needed anymore and just were there to workaround having UV in `isVerificationEnabled` which was a wrong assumption:

`private var userVerificationPreference: Uv?` ([code](https://github.com/bitwarden/ios/blob/main/BitwardenShared/UI/Autofill/Utilities/Fido2UserInterfaceHelper.swift#L100))
`func setupCurrentUserVerificationPreference(userVerificationPreference: Uv)` ([code](https://github.com/bitwarden/ios/blob/main/BitwardenShared/UI/Autofill/Utilities/Fido2UserInterfaceHelper.swift#L73))

iinuwa: On the simulator, I can confirm that the iPhone prompted with FaceID when user ID after removing the code referenced above and setting the session timeout to "Never".

(Curiously, it seems that when I set `userVerification: discouraged`, iOS is upgrading the request to `required`. `preferred` is still set as preferred.)

BitwardenShared/UI/Autofill/Utilities/Fido2UserInterfaceHelper.swift
```diff
@@ -186,19 +177,8 @@ class DefaultFido2UserInterfaceHelper: Fido2UserInterfaceHelper {
...
-
-        switch userVerificationPreference {
-        case .discouraged:
-            return false
-        case .preferred:
-            return await fido2UserVerificationMediator.isPreferredVerificationEnabled()
-        case .required:
-            return true
-        }
+    func isVerificationEnabled() -> Bool {
```
KatherineInCode, line 180: 🤔 Is this function likely to change in the future, to have more logic in it? If not, it would make sense to me to just delete it, and cascade up the `true`.

iinuwa: No, not on iOS, where using passkeys requires setting up at least one local OS auth method. After thinking about this, I agree, I think it would be good to push it into the SDK.

This method exists because the SDK is passing through a required method to fulfill a `passkey-rs` trait, so we can't remove it without changing it for all consumers of the SDK. I think there are two reasons we should push this into the SDK:

- I believe that all of our clients should have a way to do user verification, even it's just redoing whatever auth/unlock ceremony they have enabled for their vault. So this would just become a static value inside the SDK.
- We are only using this method to:
  - set the `options.uv` value in `authenticatorGetInfo`, which is merely a hint for platforms, and to
  - set the `options.uv` value in `checkUser` on `authenticatorMakeCredential` and `authenticatorGetAssertion`—which is always ignored in our implementation, since we pass the WebAuthn `UserVerificationRequirement` enum instead.[^1]

For those reasons, and because of the confusion having this method introduces, I think removing this method from the SDK trait and letting the SDK abstract over it would be beneficial.

[^1]: To expand on this point, passkey-rs only sends the CTAP2 `options.uv` boolean, which loses context on whether the UV request can be ignored or not (e.g. `preferred` is not represented.) So instead, we send the original WebAuthn UV preference, `discouraged`, `required`, or `preferred` (found in `options.authenticatorSelection.userVerification` in `navigator.credential.create()` and `userVerification` in `navigator.credential.get()`) which gives more context. 

But the current implementation, where `isVerificationEnabled()` always returns `true` produces the same behavior as pushing this SDK trait method, so I'd recommend following that up in a separate PR in the SDK.

---

I think I should clarify two other things that I am _not_ talking about changing:

- changing `checkUser(options: BitwardenSdk.CheckUserOptions, ...)` to `checkUser(shouldVerify: bool, ...)`. Even though we could decide whether we should prompt the user for verification from the SDK, if we change this to a simple bool, it still loses information on how hard we should try to verify the user. For example, the current iOS implementation will prompt for UV on both `preferred` and `required`, but if set to `required` and no verification methods are available, it will start the PIN setup flow, but it will just return `userVerified: false` if only `preferred`.

- removing `shouldThrowEnforcingRequiredVerification` parameter from `checkUser()`. I'm a little fuzzy on this one. It seems that we don't set this flag when calling this from the SDK, since the SDK will check this and return an appropriate CTAP2 error code. But we call this method from other places that don't go through the SDK FIDO methods (`VaultAutofillListProcessor`, `AddEditItemProcessor`), so this method is used there. I don't know enough about these two call sites to know whether this is useful or not.

abergs: A ticket has been created to clean up / refine ´isVerificationEnabled´ so that we can unblock this PR https://bitwarden.atlassian.net/browse/PM-30454

iinuwa:
This was implemented in #2222. Closing this PR.

PR #2227, opened 2025-12-30
Review Code Triggered by labeled event

theMickster:
## 📔 Objective

We have implemented an opt-in for Claude Code Review using the `ai-review` label and must trigger the reusable workflow when the label is added. The reusable workflow already contains the PR label validation.

And found that `CODEOWNERS` file was also not in alignment with our other repos.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -40,3 +40,8 @@
 **/*.dockerignore @bitwarden/team-appsec @bitwarden/dept-bre
 **/entrypoint.sh @bitwarden/team-appsec @bitwarden/dept-bre
 **/docker-compose.yml @bitwarden/team-appsec @bitwarden/dept-bre
+
+# Claude related files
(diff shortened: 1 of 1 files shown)


.github/workflows/review-code.yml
```diff
@@ -2,7 +2,7 @@ name: Code Review
 
 on:
   pull_request:
-    types: [opened, synchronize, reopened, ready_for_review]
+    types: [opened, labeled]
```
vvolkgang, line 5: Kudos for reducing this list! Do we even need `opened`? That would help reduce the Actions Run spam.

theMickster: Yes; still needed because you can open a PR with a label and the labeled event is not fired, the opened event is instead. For now, these are the changes that we've coordinated together for the caller repos.

SaintPatrck (approved):

PR #2169, opened 2025-11-28
[PM-28855] Update credential identities store on cipher changes on iOS extensions

fedemkr:
## 🎟️ Tracking

[PM-28855](https://bitwarden.atlassian.net/browse/PM-28855)

## 📔 Objective

Add cipher changes subscription to update credential identities store on iOS extensions.
This differs from the current approach where all ciphers/credentials get updated to be done on individual cipher CUD helping memory efficiency.
As you can see the new `CipherChangePublisher` publishes only when "one" cipher operation was done, but it doesn't when there's a batch operation like replacing all ciphers to avoid memory problems on the extensions.

> [!IMPORTANT]  
> This is a solution for iOS 17+ versions as they have incremental updates on the credential identities store. I didn't add the fallback as it'd imply replacing all identities from all ciphers which means loading and decrypting all (or most) ciphers which may crash on large vaults or have the extension use more memory leaving less for other operations like searching before crashing. The user can always force a sync on the main app to update the OS store and have autofill working for that specific credential again.

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

[PM-28855]: https://bitwarden.atlassian.net/browse/PM-28855?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2046, opened 2025-10-15
[PM-24290] Decrease intensive memory usage on iOS extensions by not syncing ASStore

fedemkr:
## 🎟️ Tracking

[PM-24290](https://bitwarden.atlassian.net/browse/PM-24290)

## 📔 Objective

Decrease intensive memory usage on iOS extensions to reduce likelihood of crashes.

On the `AutofillCredentialService`, we were subscribing to vault unlock status changes and when unlocked syncing credentials with the OS `ASStore`. This loads all credentials into memory which was causing the memory to spike a lot. Therefore the iOS Autofill extension was crashing as soon as it loaded on large vaults.

By doing this fix, which simply ignores syncing identities on iOS extensions, memory usage goes down a lot.
As an example: on a 5k ciphers vault it went from ~150MB to ~90MB. 
So it drastically improves the experience on the extensions and reduces the likelihood of crashing.

Also moved the call to replace all credentials in the identity store to be explicitly called when the extension is activated which is the specific case where we want that to run in the `.appExtension` context.

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

[PM-24290]: https://bitwarden.atlassian.net/browse/PM-24290?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2235, opened 2026-01-06
[PM-30337] Revert AutofillCredentialService cipher change subscription to fix autofill crash

matt-livefront:
## 🎟️ Tracking

[PM-30337](https://bitwarden.atlassian.net/browse/PM-30337)

## 📔 Objective

Fixes a crash in the autofill extension and `CipherChangePublisher` by reverting changes temporarily from #2169 and #2046.

#2046 was added to decrease memory usage by not syncing the ASCredentialIdentityStore in the extension. This had the downside of not adding new credentials created by the extension to the store. #2169 was added to attempt to only sync new items to the store. That ended up introducing this crash in `CipherChangePublisher`.

`CipherChangePublisher` was publishing cipher changes without respecting the subscriber's demand. This led to a Combine backpressure violation which resulted in a "Received an output without requesting demand" crash. I was able to fix the crash in https://github.com/bitwarden/ios/commit/c99e553f212b4b6fb5b5fb2245d1fe94ad85b6e1. While testing these changes, I was seeing another issue with the publisher. Following a sync in the extension, when we bulk delete/insert ciphers, the publisher was only publishing the deletes and not the inserts. This led to `AutofillCredentialService` removing all credential identities from the store which would cause no autofill suggestions the next time autofill was invoked. I thought the publisher was supposed to ignore bulk deletes/inserts, so I'm not sure if this is something we didn't run into previously or if some other Core Data change is affecting this. For now, I think it's safest to revert these changes while reevaluating incremental cipher change updates to the store.  

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-30337]: https://bitwarden.atlassian.net/browse/PM-30337?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


KatherineInCode:
🤔 This looks like it only reverts calling `subscribeToCipherChanges()` in `AutoFillCredentialService`. Does that alone just basically obviate calling everything else added in those two PRs? Or is there something else I'm missing?

matt-livefront:
> 🤔 This looks like it only reverts calling `subscribeToCipherChanges()` in `AutoFillCredentialService`. Does that alone just basically obviate calling everything else added in those two PRs? Or is there something else I'm missing?

@KatherineInCode Correct, I only removed where we call `subscribeToCipherChanges()`. That's enough to remove the incremental updates based on updates from `CipherChangePublisher`.

With the guard removed, that will enable the full syncing of the identity store which we had previously:

https://github.com/bitwarden/ios/blob/3b33335dbed677b32165e6e7e3996f919a197ec6/BitwardenShared/Core/Autofill/Services/AutofillCredentialService.swift#L203-L207

I don't think it's worth reverting everything else at this time since I suspect we'll want to reenable the subscription/incremental updates once we can fix the issue with only receiving deletes.

PR #2239, opened 2026-01-08
[PM-23300] feat: Store session values to keychain so they can be shared between app and autofill

KatherineInCode:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-23300

## 📔 Objective

This moves the storage of several values related to session length from UserDefaults to the shared keychain. This allows the app and the autofill extension (and other extensions) to share these values, which allows things like session timeout to be shared between both of them. As well, this makes these values more secure.

With the `UserSession*` protocols, I did break the alphabetization convention by pairing getters and setters together, then alphabetizing by underlying property; this also descended into the tests. I tend to prefer this sort of grouping, because then it's easier to see the parallelism rather than related methods being much farther apart from each other. However, if we want to follow our coding guidelines more strictly here, I can rearrange to put all the getters together then all the setters.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/Bitwarden.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/Bitwarden.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -1,5 +1,5 @@
 {
-  "originHash" : "bde128d30b1cfe9a2a1705335aa08196602e4a15eef690d93c4edbbb654cf005",
+  "originHash" : "93ec2eb77f36826178e4c8677e95e70ec909a1b40aa8e2cb1ccaa066a778be58",
   "pins" : [
     {
       "identity" : "abseil-cpp-binary",


BitwardenShared/Core/Auth/Repositories/AuthRepositoryTests.swift
```diff
@@ -926,13 +932,16 @@ class AuthRepositoryTests: BitwardenTestCase { // swiftlint:disable:this type_bo
...
             shortName.profile.userId: true,
         ]
-        stateService.vaultTimeout = [
-            anneAccount.profile.userId: .never,
-            beeAccount.profile.userId: .never,
-            empty.profile.userId: .never,
-            shortEmail.profile.userId: .never,
-            shortName.profile.userId: .fifteenMinutes,
-        ]
+        userSessionStateService.getVaultTimeoutClosure = { userId in
```
matt-livefront: 🤔 This probably needs a weak self to avoid a retain cycle, since self has a reference to `userSessionStateService` which has a closure with a strong reference back to self.

BitwardenShared/Core/Auth/Services/KeychainRepository.swift
```diff
@@ -382,10 +403,13 @@ extension DefaultKeychainRepository {
             .authenticatorVaultKey(userId: userId),
             .biometrics(userId: userId),
             // Exclude `deviceKey` since it is used to log back into an account.
+            .lastActiveTime(userId: userId),
             .neverLock(userId: userId),
             // Exclude `pendingAdminLoginRequest` since if a TDE user is logged out before the request
             // is approved, the next login for the user will succeed with the pending request.
             .refreshToken(userId: userId),
+            .unsuccessfulUnlockAttempts(userId: userId),
+            .vaultTimeout(userId: userId),
```
matt-livefront: 🤔 I believe we don't want to always clear `vaultTimeout` when a user logs out, specifically for the soft-logout case so that it isn't reset each time your vault times out (I don't think we clear it for a hard logout today either though).

BitwardenShared/Core/Auth/Services/KeychainRepository.swift
```diff
@@ -477,3 +501,50 @@ extension DefaultKeychainRepository: BiometricsKeychainRepository {
...
+    }
+
+    // MARK: Unsuccessful Unlock Attempts
+
+    func getUnsuccessfulUnlockAttempts(userId: String) async throws -> Int {
+        do {
+            let stored = try await getValue(for: .unsuccessfulUnlockAttempts(userId: userId))
+            return Int(stored) ?? 0
+        } catch {
+            return 0
```
matt-livefront: ❓ It looks like for other methods, if `getValue` throws, we throw the error. But this one catches it and returns 0 instead. Is that expected or should this also throw?

🎨 One pattern we've used elsewhere in `StateService` is to catch any errors, log the error to the error reporter, and then return a default value. Maybe that would work here (or up a level in `StateService`).

BitwardenShared/Core/Platform/Services/StateService+UserSessionTests.swift
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    override func tearDown() {
+        super.tearDown()
+
+        appSettingsStore = nil
+        dataStore = nil
+        errorReporter = nil
+        keychainRepository = nil
+        subject = nil
```
matt-livefront, line 49: ⛏️  Missing `userSessionKeychainRepository`:

```suggestion
        subject = nil
        userSessionKeychainRepository = nil
```

BitwardenShared/UI/Auth/VaultUnlock/VaultUnlockProcessorTests.swift
```diff
@@ -32,6 +33,13 @@ class VaultUnlockProcessorTests: BitwardenTestCase { // swiftlint:disable:this t
         coordinator = MockCoordinator()
         errorReporter = MockErrorReporter()
         stateService = MockStateService()
+        userSessionStateService = MockUserSessionStateService()
+
+        userSessionStateService.getVaultTimeoutReturnValue = .fifteenMinutes
+        userSessionStateService.getUnsuccessfulUnlockAttemptsReturnValue = 0
+        userSessionStateService.setUnsuccessfulUnlockAttemptsClosure = { attempts, _ in
+            self.userSessionStateService.getUnsuccessfulUnlockAttemptsReturnValue = attempts
```
matt-livefront: 🤔 Should this use a weak self?

BitwardenShared/UI/Auth/VaultUnlock/VaultUnlockProcessorTests.swift
```diff
@@ -556,7 +561,7 @@ class VaultUnlockProcessorTests: BitwardenTestCase { // swiftlint:disable:this t
         subject.state.masterPassword = "password"
         stateService.activeAccount = .fixtureAccountLogin()
         subject.state.unsuccessfulUnlockAttemptsCount = 4
-        await stateService.setUnsuccessfulUnlockAttempts(5)
+        userSessionStateService.getUnsuccessfulUnlockAttemptsReturnValue = 3
```
matt-livefront: ⛏️ Is `getUnsuccessfulUnlockAttempts()` called from within this test? It looks like it might just be checking the value set in the processor's state (which would have previously been set from `userSessionStateService`. I wonder if we can delete this line?

Bitwarden.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -123,7 +123,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/bitwarden/sdk-swift",
       "state" : {
-        "revision" : "4a418afe73d335f1e4ddba75d7b681737a6ebdf0"
+        "revision" : "e04fe2ad586c152d319dab6a21d26c3698295333"
```
matt-livefront: ⛏️ Looks like this might have been changed by mistake. `4a418afe73d335f1e4ddba75d7b681737a6ebdf0` is the latest SDK version commit.

BitwardenShared/Core/Auth/Services/KeychainRepository.swift
```diff
@@ -477,3 +501,50 @@ extension DefaultKeychainRepository: BiometricsKeychainRepository {
...
+    func getLastActiveTime(userId: String) async throws -> Date? {
+        let stored = try await getValue(for: .lastActiveTime(userId: userId))
+        guard let timeInterval = TimeInterval(stored) else {
+            return nil
+        }
+        return Date(timeIntervalSince1970: timeInterval)
+    }
+
+    func setLastActiveTime(_ date: Date?, userId: String) async throws {
+        let value = date.map(\.timeIntervalSince1970).map(String.init(describing:)) ?? ""
```
matt-livefront: 🤔 Using `String(describing:)` should be fine but makes me a little uneasy since `String(describing:)` is meant for debug output and isn't guaranteed that it won't change. Would using something like `date.map { String($0.timeIntervalSince1970) }` be equivalent?

BitwardenShared/Core/Auth/Services/KeychainRepository.swift
KatherineInCode: Good call; I was erring on the side of wiping things, but that makes sense

Bitwarden.xcworkspace/xcshareddata/swiftpm/Package.resolved
KatherineInCode: Sigh. That file. Good catch.

BitwardenShared/Core/Auth/Services/KeychainRepository.swift
KatherineInCode: That is a significantly better way to do it, yes 🤦🏼‍♀️

BitwardenShared/Core/Auth/Services/KeychainRepository.swift
KatherineInCode: I did that because when it was in `AppSettingsStore`, it didn't return an optional; if it wasn't in UserDefaults, that just [returned 0](https://developer.apple.com/documentation/foundation/userdefaults/integer(forkey:)). I could see the argument for consistency, though.

Probably makes sense to follow the same pattern with the other two, as well. However, for things like the vault timeout, an absence of value isn't really an error, it's just "the user hasn't set this yet". 🤔

matt-livefront: > However, for things like the vault timeout, an absence of value isn't really an error, it's just "the user hasn't set this yet".

That's a good point. For vault timeout, we have logic in StateService that returns a default value. So maybe it makes sense to follow that pattern here? Let the storage layer throw for any error/if it doesn't exist and let `StateService` return a default value if and when it makes sense?

KatherineInCode: Agreed; that keeps the keychain layer simpler, and consolidates the "how to handle lack of thing" logic into the state service.

BitwardenShared/UI/Auth/VaultUnlock/VaultUnlockProcessorTests.swift
KatherineInCode: Good catch

BitwardenShared/UI/Auth/VaultUnlock/VaultUnlockProcessorTests.swift
KatherineInCode: In practice I don't think it will matter (the `tearDown` will nil out the mock and break the retain cycle), but I'll update it still, on the principle of the matter.

BitwardenShared/Core/Auth/Repositories/AuthRepositoryTests.swift
KatherineInCode: In practice the `tearDown` should handle it, I think, but it makes sense to break it on the principle of the matter either way

matt-livefront: Ah, right, `tearDown` should break the cycle.

matt-livefront (approved):

PR #2240, opened 2026-01-09
[PM-27098] feat: Update some plurals to full-sentence forms to handle polypersonal agreement languages better

KatherineInCode:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-27098

## 📔 Objective

[As noted in comments on a previous PR](https://github.com/bitwarden/ios/pull/1999#issuecomment-3672028969), the previous form of handling plurals in strings by breaking out just the plural had some issues. The first is that CrowdIn Translation Memory would cause issues with pluralized nouns being taken out of subject/object context, which doesn't work for languages that decline nouns differently for subjects and objects. The second is that languages with [polypersonal agreement](https://en.wikipedia.org/wiki/Polypersonal_agreement) also require inflecting the verb congruently with the number of items, and we weren't allowing that.

Per that recommendation, I've pulled things out into full sentences, so that translators can provide better translations for the array of possibilities.

Nota bene: I elected even in circumstances where it is just a full sentence to leave the sentence-ending punctuation out of the pluralized forms. This allows us to remain consistent in multi-sentence situations, and for the base string to clearly indicate the sentence patterns. I'm not, however, wedded to that, I would just like to be consistent.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/BitwardenResources/Localizations/en.lproj/Localizable.stringsdict
+++ b/BitwardenResources/Localizations/en.lproj/Localizable.stringsdict
@@ -6,59 +6,59 @@
 	<key>BitwardenCouldNotDecryptXVaultItemsDescriptionLong</key>
 	<dict>
 		<key>NSStringLocalizedFormatKey</key>
-		<string>%#@vaultitems@. Copy and share this error report with customer success to avoid additional data loss.</string>
-		<key>vaultitems</key>
+		<string>%#@BitwardenCouldNotDecryptVaultItems@ Copy and share this error report with customer success to avoid additional data loss.</string>
+		<key>BitwardenCouldNotDecryptVaultItems</key>
 		<dict>
 			<key>NSStringFormatSpecTypeKey</key>
 			<string>NSStringPluralRuleType</string>
 			<key>NSStringFormatValueTypeKey</key>
 			<string>d</string>
 			<key>one</key>
-			<string>Bitwarden could not decrypt %d vault item</string>
+			<string>Bitwarden could not decrypt %d vault item.</string>
 			<key>other</key>
-			<string>Bitwarden could not decrypt %d vault items</string>
+			<string>Bitwarden could not decrypt %d vault items.</string>
 		</dict>
 	</dict>
 	<!-- A message indicating the minimum length of a master password in characters. -->
 	<key>MasterPasswordMustBeAtLeastXCharactersLong</key>
 	<dict>
 		<key>NSStringLocalizedFormatKey</key>
-		<string>%#@masterpasswordcharacters@.</string>
-		<key>masterpasswordcharacters</key>
+		<string>%#@MasterPasswordMustBeAtLeastCharactersLong@</string>
+		<key>MasterPasswordMustBeAtLeastCharactersLong</key>
 		<dict>
 			<key>NSStringFormatSpecTypeKey</key>
 			<string>NSStringPluralRuleType</string>
 			<key>NSStringFormatValueTypeKey</key>
 			<string>d</string>
 			<key>one</key>
-			<string>Master password must be at least %d character long</string>
+			<string>Master password must be at least %d character long.</string>
 			<key>other</key>
-			<string>Master password must be at least %d characters long</string>
+			<string>Master password must be at least %d characters long.</string>
 		</dict>
 	</dict>
 	<!-- A warning if we have determined that a password has been exposed in a data breach. -->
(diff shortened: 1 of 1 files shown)


KatherineInCode:
@mKoonrad I'd love to get your eyes on this, also, to make sure I'm not missing anything, and these updates to the plurals look reasonable to you.

BitwardenResources/Localizations/en.lproj/Localizable.stringsdict
```diff
@@ -23,34 +23,34 @@
 	<key>MasterPasswordMustBeAtLeastXCharactersLong</key>
 	<dict>
 		<key>NSStringLocalizedFormatKey</key>
-		<string>Master password must be at least %#@characters@ long.</string>
-		<key>characters</key>
+		<string>%#@masterpasswordcharacters@.</string>
+		<key>masterpasswordcharacters</key>
```
fedemkr: 🤔 Is it a requirement that this is fully in lowercase? If not, I feel like PascalCase is better/faster to read.
How do you feel about using the same as the main localization key here? in this case:
```
<string>%#@MasterPasswordMustBeAtLeastXCharactersLong@.</string>
<key>MasterPasswordMustBeAtLeastXCharactersLong</key>
```

Also, do you know if this `<string>` - `<key>` pair is presented in Crowdin somehow? Because if not, we could just simplify this to a generic way we can repeat in any localization:
```
<string>%#@pluralizedValue@.</string>
<key>pluralizedValue</key>
```

KatherineInCode: The all-lowercase isn't required; that's just what [I had done before](https://github.com/bitwarden/ios/blob/main/BitwardenResources/Localizations/en.lproj/Localizable.stringsdict#L10). I'm fine if we want to move to camel case, though (which is, at least, what [this guide to the format](https://localit.io/docs/ios-stringsdict-format-guide/) does). I'm hesitant to reuse a key in a substring, though, just to avoid any potential conflict.

I'm not inclined to do the generic way, as well, because of the potential issue with translation memory, which is one of the things we're trying to avoid by not putting the `%#@items@` in the middle of a sentence, in addition to handling polypersonal agreement better.

It does show up in CrowdIn. Here's the vault items in Portuguese:
Screenshot 2026-01-12 at 3 42 09 PM

Also of interest here is that substrings end up as separate items to be translated—which makes sense, since my understanding is that you can nest substrings fairly deep if you really wanted to, and that provides a more-unified format for it, from a UX perspective. But that might want to inform how we structure things 🤔

fedemkr: Makes sense.
However, IMO we should have a convention to make substring keys to avoid unwanted repetition/conflicts. That's why I was saying to use the same convention we use to build the localization key to build the substring key, i.e. join the substring value in Pascal Case as it's the best way to guarantee we won't repeat the same substring key in other localizations that are different. This ends up being the original localization key when the whole value is the substring as well. 
Moreover, it could be something else that has some sort of convention we could use based off the substring value as well if the parent localization key has conflicts but I strongly believe we should have some sort of convention for it to avoid potential problems and be faster to add new keys without checking the others substring keys.
Thoughts?

KatherineInCode: I agree that a convention would be helpful, though in a fair number of these, it's just the first sentence—which is what's already captured in the key for the whole phrase—that needs translated for number, so we might get substring keys that match the whole-phrase key 🤔 Though I suppose if we drop the X(/Y/Z/etc.) out of the key, that might accomplish what we want.

I'll swing through and update things CamelCase and see what I can do about establishing a more-obvious convention.

KatherineInCode:
@fedemkr @mKoonrad 
I've updated the way we handle the substring keys. They are now PascalCase versions of the string being replaced, *without* the X/Y/Z variable(s). This should avoid conflicts in the cases where the substring is the same as the full string.

I also pulled the punctuation into the pluralized string itself. This is for two reasons:
1) Languages like Spanish that use bracketed punctuation, such as with `¿Qué pasa?`, to be consistent in those cases, such as if we needed to pluralize a question.
2) It allows us to adjust the punctuation according to number, such as "There are no items to delete" and "Would you like to delete one item?" I don't think this is *likely*, mind, but it seemed prudent to allow ourselves the capability.

mKoonrad:
@KatherineInCode I just translated these strings on Crowdin. Great job 👍

PR #2241, opened 2026-01-09
[PM-28263] Clear PIN on key rotation

matt-livefront:
## 🎟️ Tracking

[PM-28263](https://bitwarden.atlassian.net/browse/PM-28263)

## 📔 Objective

If the user's key is rotated they will be logged out of the device. Upon logging back in, if the user has a previous PIN configured, that PIN should be cleared.

We can detect that a PIN is invalid if `enrollPinWithEncryptedPin(encryptedPin:)` fails at login. And then the existing PIN can be cleared.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-28263]: https://bitwarden.atlassian.net/browse/PM-28263?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
+++ b/BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
@@ -1276,6 +1276,9 @@ extension DefaultAuthRepository: AuthRepository {
             await flightRecorder.log("[Auth] enrollPinWithEncryptedPin failed: \(error), clearing existing PIN keys")
             // If `enrollPinWithEncryptedPin` fails, the user's key was likely rotated and the
             // existing PIN keys need to be removed since they are no longer valid.
+            // Note: We handle all errors broadly here because the SDK doesn't provide specific
+            // error types to distinguish key rotation failures from other errors. Clearing the
+            // PIN keys on any error is the safest approach to maintain data consistency.
             try await stateService.clearPins()
             // Return `nil` instead of throwing to avoid erroring out of the unlock process.
             return nil
--- a/BitwardenShared/Core/Auth/Repositories/AuthRepositoryTests.swift
+++ b/BitwardenShared/Core/Auth/Repositories/AuthRepositoryTests.swift
@@ -1998,6 +1998,54 @@ class AuthRepositoryTests: BitwardenTestCase { // swiftlint:disable:this type_bo
         XCTAssertEqual(stateService.manuallyLockedAccounts["1"], false)
     }
 
+    /// `unlockVaultWithBiometrics()` clears the PIN if enrolling the PIN fails.
+    func test_unlockVaultWithBiometrics_enrollPinWithEncryptedPinError() async throws {
+        let account = Account.fixture()
+        clientService.mockCrypto.enrollPinWithEncryptedPinResult = .failure(BitwardenTestError.example)
+        stateService.activeAccount = account
+        stateService.accountEncryptionKeys = [
+            "1": AccountEncryptionKeys(
+                accountKeys: .fixtureFilled(),
+                encryptedPrivateKey: "PRIVATE_KEY",
+                encryptedUserKey: "USER_KEY",
+            ),
+        ]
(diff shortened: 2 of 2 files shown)


BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
```diff
@@ -1230,50 +1230,55 @@ extension DefaultAuthRepository: AuthRepository {
...
-            else {
-                break
-            }
+        } else if pinProtectedUserKey == nil, pinUnlockRequiresPasswordAfterRestart {
+            // If the user has a PIN (encryptedPin), but requires master password after restart, set
+            // the PIN-protected user key in memory for future unlocks prior to app restart.
+            try await stateService.setPinProtectedUserKeyToMemory(enrollPinResponse.pinProtectedUserKeyEnvelope)
+            await flightRecorder.log("[Auth] Set PIN-protected user key in memory")
+        }
+    }
```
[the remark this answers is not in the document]
matt-livefront: > If encryptedPin exists → always migrate to PIN-protected user key envelope

This was originally set up just for the `pin` case. Once the PIN was migrated, the unlock method would change to `pinEnvelope`. So at that time, the migration was just done once and then you'd be in a different branch of the switch statement. After adding the `authRequest`/`deviceKey`/`keyConnector` cases, we should have added a check to see if the migration had already been done to prevent unnecessary work.

To support the key rotation fix, I needed to add the `enrollPinWithErrorHandling` call to all switch cases. There was already similarities between the two branches in the switch statement, so unifying them made the most sense to me and I think is clearer overall without separate branches based on the unlock method.

BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
```diff
@@ -1230,50 +1230,58 @@ extension DefaultAuthRepository: AuthRepository {
...
+    /// - Parameter encryptedPin: The encrypted PIN to enroll.
+    /// - Returns: The `EnrollPinResponse` if enrollment succeeds, or `nil` if it fails (allowing the
+    ///   unlock process to continue without erroring).
+    /// - Throws: An error if clearing the PIN keys fails.
+    ///
+    private func enrollPinWithErrorHandling(encryptedPin: String) async throws -> EnrollPinResponse? {
+        do {
+            return try await clientService.crypto().enrollPinWithEncryptedPin(encryptedPin: encryptedPin)
+        } catch {
+            await flightRecorder.log("[Auth] enrollPinWithEncryptedPin failed: \(error), clearing existing PIN keys")
```
fedemkr, line 1276: 🤔 Should we also log the error to the `errorReporter` for the case it's not a key rotation scenario and something else is making this throw an error? Is the SDK not going to throw a specific error to avoid some sort of attack like enumeration attacks or could they provide the specific error? That would be useful to know when to log it in the error reporter and when just to use this specific logic.

matt-livefront: If we log _any_ error here to the `errorReporter`, will we be able to notice or catch if the error starts changing? My concern would be that unexpected errors would get lost among the expected errors. Unless maybe the volume of errors changes, but we'd still be reporting errors that aren't true errors (at least in the sense of errors that need to be resolved).

The SDK throws the following error if enrollPinWithEncryptedPin fails:

```
(lldb) po error
▿ BitwardenError
  ▿ MobileCrypto : CryptoClientError
    ▿ Crypto : 1 element
      - message : "The cipher\'s MAC doesn\'t match the expected value"
```

I didn't feel like this was specific enough to key off of. Do you think we can be confident enough to know that this error isn't going to change in the SDK? We could look for either this error type or this error + the error message if you think that's better.

fedemkr: Talked with KM team, and for now it's ok to handle it like this as they will introduce changes for the crypto crate regarding error handling.

PR #2246, opened 2026-01-13
[PM-30568] VaultRepository bulk cipher share

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-30568

## 📔 Objective

This pull request adds support for bulk sharing ciphers and migrating personal vault items to an organization's default collection. It introduces new repository and service methods for these features, implements the logic for preparing and sharing ciphers, and provides comprehensive tests to ensure correct behavior, including handling attachments and error cases.

### Bulk sharing ciphers

* Added the `bulkShareCiphers` method to the `VaultRepository` protocol and implemented it in `DefaultVaultRepository`, allowing multiple ciphers to be prepared and shared with an organization and specific collections. This includes ensuring cipher keys and migrating attachments without encryption keys.
* Added the `bulkShareCiphersWithServer` method to the `CipherService` protocol to support server-side bulk sharing and updating local data.
* Updated test helpers (`MockVaultRepository`) and added tests to verify bulk sharing logic, including attachment migration and handling empty encryption contexts. 

### Personal vault migration

* Added the `migratePersonalVault` method to the `VaultRepository` protocol and implemented logic to move all personal vault items (including deleted items) to an organization's default collection, with error handling for missing default collections. 
* Updated test helpers and added tests to verify migration behavior, including cases with no personal items and missing default collections. 

### Supporting changes

* Added a convenience initializer to `BitwardenSdk.Cipher` for constructing a cipher from a mini response model, preserving original fields not present in the response.
* Minor update to `MockClientCiphers` to support new method signature for bulk share preparation.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/Bitwarden.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/Bitwarden.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -123,7 +123,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/bitwarden/sdk-swift",
       "state" : {
-        "revision" : "facfd7458bd6b3c8e3f568df15ab5811e3f33155"
+        "revision" : "a8fc6fb0eaa89cdde84a6b2343bfa818ecfe9833"
       }
     },
     {
--- a/BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
+++ b/BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
@@ -174,6 +174,13 @@ protocol AuthRepository: AnyObject {
     ///
     func requestOtp() async throws
 
+    /// Revokes the current user's access to an organization.
+    ///
+    /// - Parameters:
+    ///   - organizationId: The ID of the organization the user is revoking access from.
+    ///
+    func revokeSelfFromOrganization(organizationId: String) async throws
+
     /// Gets the `SessionTimeoutAction` for a user.
     ///
     ///  - Parameter userId: The userId of the account. Defaults to the active user if nil.
@@ -796,6 +803,10 @@ extension DefaultAuthRepository: AuthRepository {
         try await accountAPIService.requestOtp()
     }
 
+    func revokeSelfFromOrganization(organizationId: String) async throws {
+        try await organizationAPIService.revokeSelfFromOrganization(organizationId: organizationId)
+    }
+
     func sessionTimeoutValue(userId: String?) async throws -> SessionTimeoutValue {
         try await vaultTimeoutService.sessionTimeoutValue(userId: userId)
     }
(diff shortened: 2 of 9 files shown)


BitwardenShared/Core/Vault/Repositories/VaultRepository.swift
```diff
@@ -474,6 +493,51 @@ extension DefaultVaultRepository: VaultRepository {
...
+            )
+
+        guard let firstContext = encryptionContexts.first else {
+            return
+        }
+
+        try await cipherService.bulkShareCiphersWithServer(
+            encryptionContexts.map(\.cipher),
+            collectionIds: newCollectionIds,
+            encryptedFor: firstContext.encryptedFor,
```
matt-livefront: 🤔 This is pretty minor, but when I read this guard initially, I thought it was only keeping the first item in the `encryptionContexts` array and discarding the rest. After seeing that this is only for the `encryptedFor` argument, I wonder if this would be clearer to extract the `encryptedFor` property in the guard?

```suggestion
        guard let encryptedFor = encryptionContexts.first?.encryptedFor else {
            return
        }

        try await cipherService.bulkShareCiphersWithServer(
            encryptionContexts.map(\.cipher),
            collectionIds: newCollectionIds,
            encryptedFor: encryptedFor,
```

BitwardenShared/Core/Vault/Extensions/BitwardenSdk+Vault.swift
```diff
@@ -359,6 +359,51 @@ extension BitwardenSdk.Cipher {
             data: nil,
         )
     }
+
+    /// Creates a Cipher from a CipherMiniResponseModel, preserving fields from the original cipher
```
fedemkr: ⛏️ Add ticks to reference objects.
```suggestion
    /// Creates a `Cipher` from a `CipherMiniResponseModel`, preserving fields from the original cipher
```

BitwardenShared/Core/Vault/Repositories/VaultRepositoryTests.swift
```diff
@@ -1016,6 +1120,75 @@ class VaultRepositoryTests: BitwardenTestCase { // swiftlint:disable:this type_b
...
+            EncryptionContext(encryptedFor: "user-1", cipher: .fixture(id: "1")),
+            EncryptionContext(encryptedFor: "user-1", cipher: .fixture(id: "2")),
+            EncryptionContext(encryptedFor: "user-1", cipher: .fixture(id: "4")),
+        ])
+
+        try await subject.migratePersonalVault(to: "target-org")
+
+        // Verify that bulkShareCiphers was called with all 3 personal ciphers (including deleted).
+        XCTAssertEqual(cipherService.bulkShareCiphersWithServerCiphers.count, 1)
+        XCTAssertEqual(cipherService.bulkShareCiphersWithServerCiphers.first?.count, 3)
```
matt-livefront: ⛏️ This verifies that 3 ciphers were shared, but not necessary which ciphers. I think it'd be nice to verify the cipher IDs that were shared, and also the ciphers passed to `prepareCiphersForBulkShare` to ensure the filtering logic works correctly.

BitwardenShared/Core/Vault/Repositories/VaultRepository.swift
```diff
@@ -643,6 +707,34 @@ extension DefaultVaultRepository: VaultRepository {
...
+            try await clientService.vault().ciphers().decrypt(cipher: cipher)
+        }
+
+        // Fetch collections and find the default collection for the organization.
+        let collections = try await fetchCollections(includeReadOnly: false)
+        guard let defaultCollection = collections.first(where: { collection in
+            collection.organizationId == organizationId && collection.type == .defaultUserCollection
+        }), let collectionId = defaultCollection.id else {
+            throw BitwardenError.dataError("No default collection found for organization \(organizationId)")
+        }
```
fedemkr: ⛏️ A bit minor but I would invert the order of this to fetch the collection first before decrypting all the ciphers as if fetching the default collection throws then we don't waste resources decrypting all the ciphers which is the heaviest operation of the two.

BitwardenShared/Core/Vault/Services/CipherService.swift
```diff
@@ -206,6 +218,40 @@ extension DefaultCipherService {
...
+            // Find the original cipher to preserve fields not returned by the API.
+            let originalCipher = ciphersById[cipherResponse.id]
+
+            // The API doesn't return all fields, so preserve them from the original cipher.
+            let updatedCipher = Cipher(
+                cipherMiniResponseModel: cipherResponse,
+                collectionIds: collectionIds,
+                originalCipher: originalCipher,
+            )
+            try await cipherDataStore.upsertCipher(updatedCipher, userId: userId)
```
fedemkr, line 251: 🤔 Would we have any problems if let's say we bulk share 10 ciphers and when we're in this loop the 4th cipher throws when `cipherDataStore.upsertCipher(...)` gets called? So only 3 out of 10 would have been updated locally.
Would it need any special handling for that case or is it alright given that the user would only need to force sync again?

andrebispo5: I believe it is ok in this case because the server got the right info and on sync it will be updated anyway.

fedemkr: Ok, although I'm wondering if there are any potential problems server side if the user tries to migrate the vault again because when this throws I think it will display an error to the user which may want to trigger the flow again. Therefore given the local data has not been fully updated the same ciphers already migrated on the server will be bulk shared again with the server. 
So I'm not sure what the server does but would it be fine if the same ciphers get bulk shared again with the server after an error and retry from the user in this situation?

PR #2265, opened 2026-01-20
[PM-29709] Fix vault migration endpoint

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-29709

## 📔 Objective

This pull request updates the bulk cipher sharing functionality to ensure cipher IDs are included in the request payload, which is necessary for proper identification during bulk operations. The changes primarily affect the request model and its initialization, as well as related tests and the service layer.

**Bulk Share Model and Request Model Updates:**

* The `BulkShareCiphersRequestModel` now uses `CipherRequestModel` instead of `CipherCreateRequestModel` for its `ciphers` property, allowing for inclusion of cipher IDs in bulk share requests.
* The initializer for `BulkShareCiphersRequestModel` passes `includeId: true` to `CipherRequestModel`, ensuring each cipher's ID is included in the request.
* The `CipherRequestModel` struct adds an optional `id` field, with documentation noting its use for bulk share operations.
* The `CipherRequestModel` initializer gains a new `includeId` parameter, defaulting to `false`, and sets the `id` field accordingly. [[1]](diffhunk://#diff-e5913cdba04fac8a65783e78455f7c1dad1fe941810b3420961301ab604f4525L76-R83) [[2]](diffhunk://#diff-e5913cdba04fac8a65783e78455f7c1dad1fe941810b3420961301ab604f4525R94)

**Testing and Service Layer Adjustments:**

* The bulk share ciphers request tests are updated to expect the cipher IDs in the JSON payload, verifying the new request format.
* The cipher service code is refactored to move the dictionary creation for cipher ID lookup after the API call, ensuring consistency with the updated request flow.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/BitwardenShared/Core/Vault/Models/Request/CipherRequestModel.swift
+++ b/BitwardenShared/Core/Vault/Models/Request/CipherRequestModel.swift
@@ -29,6 +29,11 @@ struct CipherRequestModel: JSONRequestBody {
     /// The folder identifier.
     let folderId: String?
 
+    /// The cipher's identifier.
+    ///
+    /// - Note: This is only included for bulk share operations where the ID needs to be in the request body.
+    let id: String?
+
     /// Identity data if the cipher is a identity.
     let identity: CipherIdentityModel?
 
@@ -73,7 +78,9 @@ extension CipherRequestModel {
     /// - Parameters:
     ///   - cipher: The `Cipher` used to initialize a `CipherRequestModel`.
     ///   - encryptedFor: The user ID who encrypted the `cipher`.
-    init(cipher: Cipher, encryptedFor: String? = nil) {
+    ///   - includeId: Whether to include the cipher's ID in the request model. Defaults to `false`.
+    ///
+    init(cipher: Cipher, encryptedFor: String? = nil, includeId: Bool = false) {
         self.init(
             attachments2: cipher.attachments?.reduce(into: [String: AttachmentRequestModel]()) { result, attachment in
                 guard let id = attachment.id else { return }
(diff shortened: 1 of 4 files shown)


BitwardenShared/Core/Vault/Services/CipherService.swift
```diff
@@ -231,35 +215,25 @@ extension DefaultCipherService {
...
         let response = try await cipherAPIService.bulkShareCiphers(
             ciphers,
             collectionIds: collectionIds,
             encryptedFor: encryptedFor,
         )
 
+        // Create a dictionary for quick lookup of original ciphers by ID.
+        let ciphersById = Dictionary(uniqueKeysWithValues: ciphers.compactMap { cipher in
+            cipher.id.map { ($0, cipher) }
+        })
```
fedemkr, line 235: 🤔 Why is it necessary to move this after the server call?

BitwardenShared/Core/Vault/Models/Request/CipherRequestModel.swift
```diff
@@ -84,6 +91,7 @@ extension CipherRequestModel {
             favorite: cipher.favorite,
             fields: cipher.fields?.map(CipherFieldModel.init),
             folderId: cipher.folderId,
+            id: includeId ? cipher.id : nil,
```
fedemkr, line 94: 🤔 Which requests need the `id` to be `nil` when the `cipher.id` is not `nil`? I'm asking to see if we can just always set it as the `cipher.id` instead of checking the boolean.

BitwardenShared/Core/Vault/Services/CipherService.swift
andrebispo5: I just moved it because if the server call fails we won't be making unnecessary computations.

PR #2268, opened 2026-01-20
[PM-28150] fix: Add initial vault sync in autofill extension if needed

matt-livefront:
## 🎟️ Tracking

[PM-28150](https://bitwarden.atlassian.net/browse/PM-28150)

## 📔 Objective

This fixes an issue where an account is logged into the app from the autofill extension which didn't previously trigger a vault sync. This would result in it looking like no matching ciphers were found in the extension.

 If the extension is loaded for an account which hasn't performed a sync (based on the last sync time), the vault will sync before showing any matching ciphers for autofill. This also implements a loading state while the vault is syncing.

## 📸 Screenshots

https://github.com/user-attachments/assets/d4366f28-48db-4014-951e-1d39621bd4a3

| Loading | Error |
|--------|--------|
| Screenshot 2026-01-20 at 3 51 09 PM | Screenshot 2026-01-20 at 3 54 27 PM | 

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-28150]: https://bitwarden.atlassian.net/browse/PM-28150?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/UI/Vault/Vault/AutofillList/VaultAutofillListProcessor.swift
+++ b/BitwardenShared/UI/Vault/Vault/AutofillList/VaultAutofillListProcessor.swift
@@ -243,7 +243,7 @@ class VaultAutofillListProcessor: StateProcessor<// swiftlint:disable:this type_
     /// Fetches initial sync if necessary, checking if the user has synced before.
     private func fetchInitialSyncIfNecessary() async {
         do {
-            guard await isInitialSyncRequired() else { return }
+            guard await services.stateService.isInitialSyncRequired() else { return }
 
             if !state.loadingState.isLoading {
                 // Set the loading state if not loading. This transitions to the loading state on a
@@ -340,20 +340,6 @@ class VaultAutofillListProcessor: StateProcessor<// swiftlint:disable:this type_
                 }
             },
         )
-    }
-
-    /// Checks if an initial sync is required (user hasn't synced before).
-    ///
-    /// - Returns: `true` if initial sync is needed, `false` otherwise.
-    ///
-    private func isInitialSyncRequired() async -> Bool {
-        do {
-            let lastSyncTime = try await services.stateService.getLastSyncTime()
-            return lastSyncTime == nil
-        } catch {
-            services.errorReporter.log(error: error)
-            return false
-        }
     }
 
     /// Refreshes the vault group's TOTP Codes.
@@ -455,7 +441,7 @@ class VaultAutofillListProcessor: StateProcessor<// swiftlint:disable:this type_
                     vaultItemsTotpExpirationManager?.configureTOTPRefreshScheduling(for: sections.flatMap(\.items))
                 }
 
-                if await isInitialSyncRequired(), case .loading = state.loadingState {
+                if await services.stateService.isInitialSyncRequired(), case .loading = state.loadingState {
                     // Sync hasn't completed yet, store the sections in the loading state.
                     state.loadingState = .loading(sections)
                 } else {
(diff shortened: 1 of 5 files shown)


BitwardenShared/UI/Vault/Vault/AutofillList/VaultAutofillListProcessor.swift
```diff
@@ -311,19 +342,34 @@ class VaultAutofillListProcessor: StateProcessor<// swiftlint:disable:this type_
...
+    ///
+    private func isInitialSyncRequired() async -> Bool {
+        do {
+            let lastSyncTime = try await services.stateService.getLastSyncTime()
+            return lastSyncTime == nil
+        } catch {
+            services.errorReporter.log(error: error)
+            return false
+        }
+    }
```
fedemkr: 🤔 I feel like this could be moved to the `StateService`. What do you think?

BitwardenShared/UI/Vault/Vault/AutofillList/VaultAutofillListProcessor.swift
```diff
@@ -239,6 +240,36 @@ class VaultAutofillListProcessor: StateProcessor<// swiftlint:disable:this type_
...
+                state.loadingState = .loading(nil)
+            }
+
+            try await services.vaultRepository.fetchSync(forceSync: false, isPeriodic: false)
+
+            if case let .loading(sections) = state.loadingState, let sections {
+                // If the sections have already been loaded and stored in the loading state, display
+                // them immediately, otherwise wait for the vault publisher to populate the sections.
+                state.loadingState = .data(sections)
+            }
```
fedemkr, line 261: 🤔 Just to understand this better in the case where this happens:

1. Loading
2. Publisher loads sections into loading state
3. FetchSync completes
4. Goes here and the `loadingState` is replaced by data with the aforementioned publisher's sections
  a. isn't this data old? i.e. pre-sync? Or will it be displayed by a brief moment until the publisher gets the new data which will replace this state with the up-to-date sections? If so, does the user ever see this previous data sections?

matt-livefront: This is primarily to handle the case of either an empty vault or no autofill matches after the sync occurs.

1. Loading
2. Publisher loads initial sections into the loading state. Since sync hasn't occurred yet, this is an empty list of sections.
3. FetchSync fetches an empty vault from the API or a populated vault with no autofill matches.
    1. If the vault is empty, no data will be inserted into the database and the publisher won't publish again. So `fetchInitialSyncIfNecessary()` can use the previous data (`[]`) received from the publisher and transition to the data state.
    2. If there's ciphers in the vault, but no autofill matches, the publisher will publish an empty list of sections. If an empty list of sections is received in `streamAutofillItems()`, there isn't a good way to determine if the list is empty because a) the sync hasn't completed and there's nothing in the database or b) the sync has completed, but the vault is either empty or contains no matches. So this is updated in the loading state, and once sync completes in `fetchInitialSyncIfNecessary()`, we can transition to the data state.

I don't think you would see old data, especially since we're only syncing if we haven't synced before. It's mainly here just to catch any data that is published before `fetchInitialSyncIfNecessary()` completes.

BitwardenShared/UI/Vault/Vault/AutofillList/VaultAutofillListProcessor.swift
matt-livefront: Yeah, I can move it.

BitwardenShared/UI/Vault/Vault/AutofillList/VaultAutofillListProcessor.swift
fedemkr: Sounds perfect, thanks for the explanation!

PR #2278, opened 2026-01-23
[PM-31141] feat: Add organizationId to event logs

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31141

## 📔 Objective

This pull request adds support for associating events with an organization by introducing an optional `organizationId` field to event tracking models and APIs. The changes span the core event data structures, service interfaces, and tests to ensure that events can now be linked to organizations when relevant.

**Core event model and API changes:**

- Added an optional `organizationId` property to `EventData` and `EventRequestModel`, allowing events to be associated with an organization.
- Updated the `EventAPIService` to include `organizationId` when posting events to the server.

**Service interface and implementation updates:**

- Extended the `EventService` protocol and its default implementation to accept an optional `organizationId` parameter in the `collect` method, enabling event collection with organization context. 
- Updated `MockEventService` for testing to support the new `organizationId` parameter.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

### DIFF
--- a/BitwardenShared/Core/Platform/Services/EventService.swift
+++ b/BitwardenShared/Core/Platform/Services/EventService.swift
@@ -97,10 +97,20 @@ class DefaultEventService: EventService {
                 return
             }
 
+            let organizationIds = organizations.map(\.id)
+
+            if let organizationId, !organizationIds.contains(organizationId) {
+                return
+            }
+
             if let cipherId {
                 guard let cipher = try await cipherService.fetchCipher(withId: cipherId),
-                      let orgId = cipher.organizationId,
-                      organizations.map(\.id).contains(orgId) else {
+                      let cipherOrgId = cipher.organizationId,
+                      organizationIds.contains(cipherOrgId) else {
+                    return
+                }
+
+                if let organizationId, cipherOrgId != organizationId {
                     return
                 }
             }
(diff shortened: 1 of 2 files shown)


BitwardenShared/Core/Platform/Services/EventService.swift
```diff
@@ -85,7 +87,7 @@ class DefaultEventService: EventService {
 
     // MARK: Methods
 
-    func collect(eventType: EventType, cipherId: String?) async {
+    func collect(eventType: EventType, cipherId: String?, organizationId: String?) async {
```
fedemkr, line 90: 🤔 When `organizationId` is not `nil` should this function check whether that specific organization has `useEvents` on? Because in here it's checking that there is at least one organization with that and the org id of the cipher matches one that has `useEvents` but I'm not sure if we should also check with the `organizationId` parameter as from what I can see there's no indication they must match and there could potentially be a case where there's one organization with `useEvents` on that is not the `organizationId` parameter but is the `cipher`'s organization id one.
Maybe it's just fine as there could be cases for that but wanted to double check.

andrebispo5: yes it makes sense, I'll add the check for use events.

PR #2293, opened 2026-01-28
[PM-26177] feat: Add device auth key to keychain repository

iinuwa:
## 🎟️ Tracking

[PM-26177](https://bitwarden.atlassian.net/browse/PM-26177)

## 📔 Objective

Adds the device auth key and its metadata to the keychain repository. These will be used in later PRs for Device Auth Key implementation.

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

[PM-26177]: https://bitwarden.atlassian.net/browse/PM-26177?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2297, opened 2026-01-28
[PM-26177] feat: Create stub DeviceAuthKeyService

iinuwa:
## 🎟️ Tracking

[PM-26177](https://bitwarden.atlassian.net/browse/PM-26177)

## 📔 Objective

This adds a stub of the DeviceAuthKey service, which will be used in future PRs to create and assert passkey credentials using a key bound to the local device keychain.

Depends on #2293.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-26177]: https://bitwarden.atlassian.net/browse/PM-26177?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenShared/Core/Auth/Services/DeviceAuthKeychainRepository.swift
fedemkr, line 41: 🤔 Do we always have the record and the metadata at the same time when calling this function? Or should it be split?

BitwardenShared/Core/Auth/Services/DeviceAuthKeyService.swift
```diff
@@ -0,0 +1,116 @@
...
+    func getDeviceAuthKeyMetadata(userId: String) async throws -> DeviceAuthKeyMetadata?
+}
+
+// MARK: - DefaultDeviceAuthKeyService
+
+/// Implementation of DeviceAuthKeyService
+struct DefaultDeviceAuthKeyService: DeviceAuthKeyService {
+    // MARK: Properties
+
+    private let deviceAuthKeychainRepository: DeviceAuthKeychainRepository
```
fedemkr, line 122: ⛏️ Add docs.

BitwardenShared/Core/Auth/Services/DeviceAuthKeyService.swift
```diff
@@ -0,0 +1,116 @@
...
+
+    // MARK: Initializers
+
+    init(
+        deviceAuthKeychainRepository: DeviceAuthKeychainRepository,
+    ) {
+        self.deviceAuthKeychainRepository = deviceAuthKeychainRepository
+    }
```
fedemkr, line 138: ⛏️ Add docs.

BitwardenShared/Core/Auth/Services/DeviceAuthKeyService.swift
```diff
@@ -0,0 +1,116 @@
...
+        throw DeviceAuthKeyError.notImplemented
+    }
+
+    func createDeviceAuthKey(
+        masterPasswordHash: String,
+        overwrite: Bool,
+        userId: String,
+    ) async throws -> DeviceAuthKeyRecord {
+        throw DeviceAuthKeyError.notImplemented
+    }
```
fedemkr, line 158: ⛏️ Could a TODO be added here as well to track which Jira task(s) will implement these functions?
`// TODO: PM-{number} implement this`

BitwardenShared/Core/Auth/Services/DeviceAuthKeyService.swift
```diff
@@ -0,0 +1,116 @@
+import BitwardenSdk
+import Foundation
+
+// MARK: - DeviceAuthKeyService
+
+/// Service to manage the device auth key.
+protocol DeviceAuthKeyService { // sourcery: AutoMockable
```
fedemkr, line 8: 🤔 Typically on the services that have functions that involve passing the `userId` we make it optional and if `nil` then we just get the active user ID on the implementation. Furthermore we add an extension with the same function without the `userId` parameter so it just calls the original one with user `nil` as a shortcut for usage.
Not sure about the whole/remaining implementation but it could be good to have the same approach here if it makes sense.

BitwardenShared/Core/Auth/Services/DeviceAuthKeyService.swift
```diff
@@ -0,0 +1,116 @@
...
+        try await deviceAuthKeychainRepository.getDeviceAuthKey(userId: userId)
+    }
+}
+
+// MARK: - DeviceAuthKeyError
+
+enum DeviceAuthKeyError: Error {
+    case missingOrInvalidKey
+    case notImplemented
+}
```
fedemkr, line 197: ⛏️ Add docs for the enum and the cases.

BitwardenShared/Core/Auth/Services/KeychainRepository.swift
```diff
@@ -346,7 +346,9 @@ class DefaultKeychainRepository: KeychainRepository {
             throw BitwardenError.dataError("JSON string contains invalid UTF-8 encoding.")
         }
 
-        return try JSONDecoder.defaultDecoder.decode(T.self, from: jsonData)
+        let value = try JSONDecoder.defaultDecoder.decode(T.self, from: jsonData)
+
+        return value
```
fedemkr: ⛏️ I think we can return directly here instead of assigning to `value` and then returning.

BitwardenShared/Core/Auth/Services/DeviceAuthKeyService.swift
KatherineInCode: That makes sense; I've pushed up that change

BitwardenShared/Core/Auth/Services/DeviceAuthKeychainRepository.swift
KatherineInCode: My understanding is that yes, we'll always have both when setting, because the metadata is the unencrypted version of a small selection of fields from the full record, which is used to quick-check things instead of always doing the expensive decryption process each time.

iinuwa: Correct, another reason is so that we can have separate access controls for the metadata vs. the keys: we can look up the metadata and decide whether the DAK matches the request before prompting them for for FaceID/TouchID to unlock the key.

PR #2300, opened 2026-01-28
[PM-26177] feat: Wire up device auth key to autofill service

iinuwa:
## 🎟️ Tracking
[PM-26177](https://bitwarden.atlassian.net/browse/PM-26177)

## 📔 Objective

This updates the AutofillCredentialService to try to fulfill incoming requests using the device auth key before unlocking the vault and trying passkeys from there.

The DeviceAuthKey service is responsible for checking the record identifier to make sure that it only answers a request if the cipher ID matches.

Depends on #2297.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-26177]: https://bitwarden.atlassian.net/browse/PM-26177?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/Core/Auth/Services/DeviceAuthKeyServiceTests.swift
+++ b/BitwardenShared/Core/Auth/Services/DeviceAuthKeyServiceTests.swift
@@ -71,6 +71,7 @@ final class DeviceAuthKeyServiceTests: BitwardenTestCase {
             })
         defer { publisher.cancel() }
 
+        // TODO: PM-26177 Once the create flow is working, this can become a `try` instead of `try?`
         _ = try? await subject.createDeviceAuthKey(
             masterPasswordHash: "hashedPassword",
             overwrite: false,
--- a/BitwardenShared/Core/Autofill/Services/AutofillCredentialServiceTests.swift
+++ b/BitwardenShared/Core/Autofill/Services/AutofillCredentialServiceTests.swift
@@ -737,9 +737,9 @@ class AutofillCredentialServiceTests: BitwardenTestCase { // swiftlint:disable:t
     }
 
     /// `provideFido2Credential(for:autofillCredentialServiceDelegate:fido2UserVerificationMediatorDelegate:)`
-    /// does not succeed with device auth key if the feature flag is off.
+    /// skips device auth key logic if the feature flag is off.
     @available(iOS 18.0, *)
-    func test_provideFido2Credential_succeeds_deviceAuthKey_off() async throws {
+    func test_provideFido2Credential_skips_deviceAuthKey_featureFlagOff() async throws {
         stateService.activeAccount = .fixture()
         configService.featureFlagsBool[.deviceAuthKey] = false
         let passkeyIdentity = ASPasskeyCredentialIdentity.fixture(
@@ -760,18 +760,14 @@ class AutofillCredentialServiceTests: BitwardenTestCase { // swiftlint:disable:t
 
         deviceAuthKeyService.assertDeviceAuthKeyReturnValue = expectedAssertionResult
 
-        clientService.mockPlatform.fido2Mock
-            .clientFido2AuthenticatorMock
-            .getAssertionMocker
-            .throwing(BitwardenTestError.example)
+        _ = try await subject.provideFido2Credential(
+            for: passkeyRequest,
+            autofillCredentialServiceDelegate: autofillCredentialServiceDelegate,
+            fido2UserInterfaceHelperDelegate: fido2UserInterfaceHelperDelegate,
+        )
 
-        await assertAsyncThrows(error: BitwardenTestError.example) {
-            _ = try await subject.provideFido2Credential(
-                for: passkeyRequest,
-                autofillCredentialServiceDelegate: autofillCredentialServiceDelegate,
-                fido2UserInterfaceHelperDelegate: fido2UserInterfaceHelperDelegate,
-            )
-        }
+        XCTAssertFalse(deviceAuthKeyService.getDeviceAuthKeyMetadataCalled)
+        XCTAssertFalse(deviceAuthKeyService.assertDeviceAuthKeyCalled)
     }
 
     /// `provideOTPCredential(for:autofillCredentialServiceDelegate:repromptPasswordValidated:)`


BitwardenShared/Core/Auth/Services/DeviceAuthKeyService.swift
```diff
@@ -153,6 +160,12 @@ struct DefaultDeviceAuthKeyService: DeviceAuthKeyService {
         overwrite: Bool,
         userId: String?,
     ) async throws -> DeviceAuthKeyRecord {
+        let resolvedUserId = try await activeAccountStateProvider.userIdOrActive(userId)
+
+        var curVal = deviceAuthKeySubject.value
+        curVal[resolvedUserId] = true
+        deviceAuthKeySubject.send(curVal)
```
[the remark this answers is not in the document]
KatherineInCode: This will be taken care of after the real implementation is provided; for now, we can't do that to preserve the test.

BitwardenShared/Core/Auth/Services/DeviceAuthKeyService.swift
```diff
@@ -52,6 +53,9 @@ protocol DeviceAuthKeyService { // sourcery: AutoMockable
     ///   - userId: The user ID for the account to get device auth key metadata for.
     ///
     func getDeviceAuthKeyMetadata(userId: String?) async throws -> DeviceAuthKeyMetadata?
+
+    /// A publisher for the device auth key
+    func deviceAuthKeyPublisher() -> AnyPublisher<[String: Bool], Never>
```
fedemkr, line 60: ⛏️ Order functions alphabetically.

BitwardenShared/Core/Auth/Services/DeviceAuthKeyService.swift
```diff
@@ -153,6 +160,12 @@ struct DefaultDeviceAuthKeyService: DeviceAuthKeyService {
         overwrite: Bool,
         userId: String?,
     ) async throws -> DeviceAuthKeyRecord {
+        let resolvedUserId = try await activeAccountStateProvider.userIdOrActive(userId)
+
+        var curVal = deviceAuthKeySubject.value
+        curVal[resolvedUserId] = true
```
fedemkr, line 168: 🤔 Is this going to go back to `false` at some time? Maybe will be in `PM-26177` or is this the finish state of the key/value pair?

BitwardenShared/Core/Auth/Services/DeviceAuthKeyServiceTests.swift
```diff
@@ -61,6 +61,28 @@ final class DeviceAuthKeyServiceTests: BitwardenTestCase {
...
+    func test_createDeviceAuthKey_publisher() async throws {
+        var publishedValues = [[String: Bool]]()
+
+        let publisher = subject.deviceAuthKeyPublisher()
+            .sink(receiveValue: { value in
+                publishedValues.append(value)
+            })
+        defer { publisher.cancel() }
+
+        _ = try? await subject.createDeviceAuthKey(
```
fedemkr, line 75: 🤔 Should we leave it with `try` and just Skip the test, adding the TODO with the PM task?
If current way is preferred, I'd add a TODO with the PM task to update the test to not use `try?`

BitwardenShared/Core/Autofill/Services/AutofillCredentialService.swift
```diff
@@ -341,9 +353,27 @@ class DefaultAutofillCredentialService {
...
+                        credentialID: metadata.credentialId,
+                        userHandle: metadata.userHandle,
+                        recordIdentifier: metadata.cipherId,
+                    )
+                    identities.append(identity)
+                    deviceAuthKeyFlightRecorderInfo = ", including a device auth key"
+                } else {
+                    deviceAuthKeyFlightRecorderInfo = ""
+                }
+
```
fedemkr: 🎨 I'd just initialize with an empty string `var deviceAuthKeyFlightRecorderInfo: String = ""` and remove the `else` block so it's simpler.
⚠️ Also I'd not use `try?` to get the metadata as we'd not get any errors to Crashlytics/Flight recorder if that occurs. Is that intended?
🎨 Is it possible to extract this into a function so `replaceAllIdentities` doesn't grow that much? Or perhaps just do an extension/helper function that creates the `ASPasskeyCredentialIdentity` from the `metadata`.

BitwardenShared/Core/Autofill/Services/AutofillCredentialService.swift
```diff
@@ -442,15 +472,39 @@ extension DefaultAutofillCredentialService: AutofillCredentialService {
...
+                recordIdentifier: recordIdentifier,
+                userId: userId,
+            ) {
+                return ASPasskeyAssertionCredential(
+                    assertionResult: deviceResult,
+                    rpId: credentialIdentity.relyingPartyIdentifier,
+                    clientDataHash: passkeyRequest.clientDataHash,
+                )
+            }
+        }
```
fedemkr, line 521: 🤔 Can this be extracted into another function?
🎨 Can we not use `try?` so errors are caught in Crashlytics/Flight recorder?

BitwardenShared/Core/Autofill/Services/AutofillCredentialServiceTests.swift
```diff
@@ -680,6 +688,92 @@ class AutofillCredentialServiceTests: BitwardenTestCase { // swiftlint:disable:t
...
+        XCTAssertEqual(result.credentialID, expectedAssertionResult.credentialId)
+
+        // TODO: PM-26177 once SDK is updated for full PRF support we can include this
+        XCTAssertNil(result.extensionOutput)
+    }
+
+    /// `provideFido2Credential(for:autofillCredentialServiceDelegate:fido2UserVerificationMediatorDelegate:)`
+    /// does not succeed with device auth key if the feature flag is off.
+    @available(iOS 18.0, *)
+    func test_provideFido2Credential_succeeds_deviceAuthKey_off() async throws {
```
fedemkr: 🤔 The test function name has `succeeds` but the DocC says "does not succeed". Also, I think it's kind of mixing two concepts as `deviceAuthKey` off doesn't imply the test needs to throw an error. Perhaps it could just test that `deviceAuthKeyService.getDeviceAuthKeyMetadata` and `deviceAuthKeyService.assertDeviceAuthKey` haven't been called as the previous to this change normal paths are already tested on the other tests.

BitwardenShared/Core/Auth/Services/DeviceAuthKeyService.swift
KatherineInCode: We typically put publisher-creating functions in a separate section. cf. [SettingsService](https://github.com/bitwarden/ios/blob/main/BitwardenShared/Core/Platform/Repositories/SettingsRepository.swift#L101), [SendRepository](https://github.com/bitwarden/ios/blob/main/BitwardenShared/Core/Tools/Repositories/SendRepository.swift#L72). Though we can add the appropriate pragma mark for it.

BitwardenShared/Core/Auth/Services/DeviceAuthKeyService.swift
KatherineInCode: Once PM-26177 is done, the true/false flipping of it should be completed, yes. We are also allowing for the possibility that depending on how things with SDK development shake out, we might adjust exactly what this publisher emits.

BitwardenShared/Core/Autofill/Services/AutofillCredentialService.swift
KatherineInCode: The use of `try?` is intended so that we don't potentially interrupt the user flow, though thinking about it, I guess it would just catch, log it, and continue on, so it's probably okay. We can make that change.

Pulling out a custom initializer for the metadata is a good idea; we can handle that.

BitwardenShared/Core/Autofill/Services/AutofillCredentialServiceTests.swift
KatherineInCode: Yeah, figuring out how best to test "if the feature flag is off don't do the thing" was a tricky one, and I think the error was basically "this is the earliest thing we can easily check as having happened *after* the feature-flagged code". We can circle back on it, see if we can find something better and/or document it better.

PR #2313, opened 2026-01-30
[PM-29672] fix: Move error report metadata to the top to match flight recorder logs

matt-livefront:
## 🎟️ Tracking

[PM-29672](https://bitwarden.atlassian.net/browse/PM-29672)

## 📔 Objective

Moves the metadata (app/device info, server details, etc) to the top of the error report to match flight recorder logs. Previously, this was at the bottom of the report. This also adds the date at which the report was generated, which also matches the log start date in the flight recorder log.

Example report:
```
Bitwarden Error
🕒 Error Date: 2026-01-30T16:08:05-06:00
📝 Bitwarden 2026.1.1 (1)
📦 Bundle: com.8bit.bitwarden
📱 Device: iPhone17,1
🍏 System: iOS 26.2.1
🦀 SDK: 2.0.0-3989-0544dde
🌩️ Server: 2026.1.0 @ US
👤 User ID: <user ID>

BitwardenShared.StateServiceError.noAccounts
The operation couldn’t be completed. (BitwardenShared.StateServiceError error 0.)

Stack trace:
<stack trace>
```

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-29672]: https://bitwarden.atlassian.net/browse/PM-29672?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenKit/Core/Platform/Utilities/ErrorReportBuilder.swift
+++ b/BitwardenKit/Core/Platform/Utilities/ErrorReportBuilder.swift
@@ -98,9 +98,9 @@ extension DefaultErrorReportBuilder: ErrorReportBuilder {
         let formattedErrorDate = dateFormatter.string(from: timeProvider.presentTime)
         return await """
         Bitwarden Error
-        Error Date: \(formattedErrorDate)
+        🕒 Error Date: \(formattedErrorDate)
         \(appInfoService.appInfoWithoutCopyrightString)
-        User ID: \(userId)
+        👤 User ID: \(userId)
 
         \(error as NSError)
         \(error.localizedDescription)
--- a/BitwardenKit/Core/Platform/Services/FlightRecorder.swift
+++ b/BitwardenKit/Core/Platform/Services/FlightRecorder.swift
@@ -316,10 +316,10 @@ public actor DefaultFlightRecorder {
         let userId = await (try? stateService.getActiveAccountId()) ?? "n/a"
         let contents = await """
         Bitwarden iOS Flight Recorder
-        Log Start: \(dateFormatter.string(from: log.startDate))
-        Log Duration: \(log.duration.shortDescription)
+        🕒 Log Start: \(dateFormatter.string(from: log.startDate))
+        ⏳ Log Duration: \(log.duration.shortDescription)
         \(appInfoService.appInfoWithoutCopyrightString)
-        User ID: \(userId)\n\n
+        👤 User ID: \(userId)\n\n
         """
 
         let url = try fileURL(for: log)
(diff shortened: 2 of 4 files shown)


BitwardenKit/Core/Platform/Utilities/ErrorReportBuilder.swift
```diff
@@ -31,6 +31,16 @@ public struct DefaultErrorReportBuilder {
     /// The service used by the application to get info about the app and device it's running on.
     private let appInfoService: AppInfoService
 
+    /// The date formatter used to format the timestamp included in the error report.
+    private let dateFormatter: ISO8601DateFormatter = {
+        let dateFormatter = ISO8601DateFormatter()
+        dateFormatter.timeZone = .autoupdatingCurrent
```
KatherineInCode, line 37: 🤔 For the error report, do we want the local time zone, or do we actually want to use UTC? I think it's fine either way (and probably moot), but it still stood out to me

matt-livefront: Good question. This matches the flight recorder date formatter and because it's ISO 8601, you get both the local time and the timezone offset (`2026-01-30T16:08:05-06:00`), so best of both worlds? It seems nice to have the local time and timezone offset so you could get UTC if needed. I'm not tied to this though if you have a different idea.

fedemkr: I think I generally prefer `UTC` as well as we don't have any logic that depends on whether the local time is day/night or something like that which would imply not using UTC to analyze that.
Moreover, I think it could be useful to compare clients vs server logs using the same format which would be simpler, but very rarely used.
On the contrary, we could always just run a script that transforms it as we have the offset and it's not something we need to do usually.
So as long as the order of the events is correct, I'm fine with either of these for the time being.

matt-livefront: My reasoning behind using local time was for if a user says something happened in the app around 9pm local time, we can correlate which events happened in the log around that time. If you needed to convert to UTC, the timezone offset is there to do so. If you just log UTC time, you either need to also include the timezone offset or get that information from the user so it seemed nice to have local time + offset.

fedemkr: 👍  Makes sense to me, fine with keeping the offset.

KatherineInCode: Yep, that makes sense to me, too

PR #2322, opened 2026-02-03
[PM-31609] fix: Return nil instead of throwing if last active time hasn't been set yet

matt-livefront:
## 🎟️ Tracking

[PM-31609](https://bitwarden.atlassian.net/browse/PM-31609)

## 📔 Objective

Fixes a user displayed error that occurs when setting the vault timeout if the user's last active time hasn't been set yet. When this value was stored in UserDefaults, UserDefaults would return `nil` if there wasn't a value stored. But the keychain implementation throws an error if the items isn't found. This catches the item not found error and returns `nil`, matching the previous behavior.

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-31609]: https://bitwarden.atlassian.net/browse/PM-31609?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/Core/Platform/Services/StateService.swift
+++ b/BitwardenShared/Core/Platform/Services/StateService.swift
@@ -2297,11 +2297,7 @@ extension DefaultStateService: UserSessionStateService {
 
     func getLastActiveTime(userId: String?) async throws -> Date? {
         let userId = try userId ?? getActiveAccountUserId()
-        do {
-            return try await userSessionKeychainRepository.getLastActiveTime(userId: userId)
-        } catch KeychainServiceError.osStatusError(errSecItemNotFound) {
-            return nil
-        }
+        return try await userSessionKeychainRepository.getLastActiveTime(userId: userId)
     }
 
     func setLastActiveTime(_ date: Date?, userId: String?) async throws {
--- a/BitwardenShared/Core/Auth/Services/KeychainRepository+UserSessionTests.swift
+++ b/BitwardenShared/Core/Auth/Services/KeychainRepository+UserSessionTests.swift
@@ -57,6 +57,14 @@ final class KeychainRepositoryUserSessionTests: BitwardenTestCase {
         }
     }
 
+    /// `getLastActiveTime(userId:)` returns nil when the time has never been set.
+    func test_getLastActiveTime_itemNotFound() async throws {
+        let error = KeychainServiceError.osStatusError(errSecItemNotFound)
+        keychainService.searchResult = .failure(error)
+        let lastActiveTime = try await subject.getLastActiveTime(userId: "1")
+        XCTAssertNil(lastActiveTime)
+    }
+
     /// `setLastActiveTime(_:userId:)` stores the last active time with correct attributes.
     ///
     func test_setLastActiveTime() async throws {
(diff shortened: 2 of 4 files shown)


fedemkr (dismissed):

KatherineInCode (dismissed):

BitwardenShared/Core/Platform/Services/StateService.swift
```diff
@@ -2297,7 +2297,11 @@ extension DefaultStateService: UserSessionStateService {
 
     func getLastActiveTime(userId: String?) async throws -> Date? {
         let userId = try userId ?? getActiveAccountUserId()
-        return try await userSessionKeychainRepository.getLastActiveTime(userId: userId)
+        do {
+            return try await userSessionKeychainRepository.getLastActiveTime(userId: userId)
+        } catch KeychainServiceError.osStatusError(errSecItemNotFound) {
+            return nil
+        }
```
fedemkr: 🤔 This is fine, but I was wondering whether this should be directly placed in the `DefaultKeychainRepository` as it's not something that depends on whether the repository's function is called from a specific place but any time last active time is gotten from the repository it should consider this..

KatherineInCode: We [discussed that for unsuccessful unlock attempts](https://github.com/bitwarden/ios/pull/2239#discussion_r2737278578) and decided to pull it up into the state service for consistency with other values. I'd be fine with pushing it down, but then we'd need to push *all* of them down, in my opinion.

matt-livefront: > it's not something that depends on whether the repository's function is called from a specific place but any time last active time is gotten from the repository it should consider this..

True, but isn't the idea with exposing this through `StateService` that we'd prefer to always access it from `StateService` rather than from the keychain repository directly? So the logic here would cover all cases of accessing this item.

The more I think about this though, I see this as two separate things:

1) How the storage layer (UserDefaults / Keychain) handles the item not existing?
2) If the item doesn't exist, where do we want to assign a default value, if it makes sense to have a default value? Like we do for `getUnsuccessfulUnlockAttempts()`.

For 1) maybe it's better if we have the keychain follow the UserDefaults approach and return `nil` for any items that don't exist when fetching rather than throw? If we make this change, we probably want to follow this up with making this change for other keychain items as well.

For 2) I like the idea of letting the storage layer return `nil` and then letting `StateService` assign a default value if it makes sense. Similar to how we handle `getUnsuccessfulUnlockAttempts()` and `getVaultTimeout()` today.

fedemkr: You have a point there with the `StateService` access.
Thinking it again storage layer shouldn't handle default business values so they should be in `StateService` as you pointed out.
Returning `nil` on item not existing seems like a nice idea, we just need to take extra attention on items that can actually be `nil` as a correct value and is not the default; so only for those specific cases we'd need to keep throwing.
Maybe I'm overthinking it and the current solution is the most flexible for all situations.

KatherineInCode: I think with 1 I'm inclined to have it return `nil`, because I think most of the time if we don't have a value, it's just that we don't have a value, rather than there being a real issue we'd need to see in e.g. Crashlytics 🤔 

I agree on 2. The storage layer shouldn't know default values for things.

matt-livefront: I ended up moving the `errSecItemNotFound` logic back into `KeychainRepository` so that it returns `nil` if the date hasn't been set.

PR #2323, opened 2026-02-03
[PM-23299] fix: Improve time comparisons to prevent clock tampering

fedemkr:
## 🎟️ Tracking

[PM-23299](https://bitwarden.atlassian.net/browse/PM-23299)

## 📔 Objective

Improve time comparisons to prevent clock tampering on last sync time and last active time, so changing the current time on the OS settings doesn't affect Bitwarden.
This approach uses monotonic time (`CLOCK_MONOTONIC_RAW` <-> `mach_continuous_time()`) as a more robust way to address these flows.
This is particularly necessary now given the work on user key session sharing amongst the app and its extensions like Autofill (#2239)

### Notes

Privacy Info file adding `NSPrivacyAccessedAPICategorySystemBootTime` with `35F9.1` reason is no longer needed as it's only necessary for the absolute time => check  https://developer.apple.com/documentation/bundleresources/app-privacy-configuration/nsprivacyaccessedapitypes/nsprivacyaccessedapitype for the app privacy values.

> NSPrivacyAccessedAPICategorySystemBootTime
> The following APIs for accessing the system boot time require reasons for use.
> - [systemUptime](https://developer.apple.com/documentation/Foundation/ProcessInfo/systemUptime)
> - mach_absolute_time()

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-23299]: https://bitwarden.atlassian.net/browse/PM-23299?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenKit/Core/Platform/Services/Mocks/MockTimeProvider.swift
```diff
@@ -17,13 +17,15 @@ public class MockTimeProvider: TimeProvider {
     }
 
     public var timeConfig: TimeConfig
+    public var monotonicTime: TimeInterval
```
matt-livefront: ⛏️ Alphabetize these?

BitwardenKit/Core/Platform/Services/Mocks/MockTimeProvider.swift
```diff
@@ -17,13 +17,15 @@ public class MockTimeProvider: TimeProvider {
...
 
     public var presentTime: Date {
         timeConfig.date
     }
 
-    public init(_ timeConfig: TimeConfig) {
+    public init(_ timeConfig: TimeConfig, monotonicTime: TimeInterval = 0) {
```
matt-livefront: 🤔 Does it make sense to update the `mockTime(Date)` case to include the monotonic time as an associated value (e.g. `mockTime(Date, TimeInterval)`) rather than specify it separately here?

BitwardenKit/Core/Platform/Services/TimeProvider.swift
```diff
@@ -29,6 +37,10 @@ public class CurrentTime: TimeProvider {
         .now
     }
 
+    public var monotonicTime: TimeInterval {
+        ProcessInfo.processInfo.systemUptime
```
matt-livefront: ⚠️ I believe using `systemUptime` requires updating PrivacyInfo.xcprivacy.

https://developer.apple.com/documentation/foundation/processinfo/systemuptime#Discussion

BitwardenKit/Core/Platform/Services/TimeProvider.swift
```diff
@@ -12,6 +12,14 @@ public protocol TimeProvider: AnyObject {
     ///
     var presentTime: Date { get }
 
+    /// The monotonic time expressed as a `TimeInterval` since system boot.
+    ///
+    /// This time is based on `ProcessInfo.systemUptime` and cannot be affected by user
+    /// clock changes, making it suitable for measuring elapsed time intervals securely.
+    /// The value resets to 0 when the device reboots.
+    ///
+    var monotonicTime: TimeInterval { get }
```
matt-livefront, line 24: ⛏️ Should this be alphabetized relative to `presentTime`?

BitwardenShared/Core/Auth/Services/KeychainRepository.swift
```diff
@@ -21,6 +21,9 @@ enum KeychainItem: Equatable, KeychainStorageKeyPossessing {
     /// The keychain item for a user's last active time.
     case lastActiveTime(userId: String)
 
+    /// The keychain item for a user's last active monotonic time.
+    case lastActiveMonotonicTime(userId: String)
```
matt-livefront, line 40: ❓ Should this be cleared on logout? Should we add it to `deleteItems(for:)`?

https://github.com/bitwarden/ios/blob/44ecde167db431d57e4869b30295efaefe6e20a4/BitwardenShared/Core/Auth/Services/KeychainRepository.swift#L411-L424

BitwardenShared/Core/Platform/Services/StateServiceTests.swift
```diff
@@ -992,6 +992,66 @@ class StateServiceTests: BitwardenTestCase { // swiftlint:disable:this type_body
         XCTAssertNil(urls)
     }
 
+    /// `getLastActiveMonotonicTime(userId:)` returns the last active monotonic time for a user.
+    func test_getLastActiveMonotonicTime() async throws {
```
matt-livefront: ❓ I think the other `UserSessionStateService` tests are in [StateService+UserSessionTests.swift](https://github.com/bitwarden/ios/blob/44ecde167db431d57e4869b30295efaefe6e20a4/BitwardenShared/Core/Platform/Services/StateService%2BUserSessionTests.swift). Should we move the last active monotonic time tests there?

BitwardenShared/Core/Vault/Services/VaultTimeoutService.swift
```diff
@@ -189,12 +189,28 @@ class DefaultVaultTimeoutService: VaultTimeoutService {
...
+
+            // Use monotonic time for tamper-resistant timeout checking
+            if let lastActiveMonotonic {
+                let elapsedMonotonic = timeProvider.monotonicTime - lastActiveMonotonic
 
-            return timeProvider.presentTime.timeIntervalSince(lastActiveTime)
-                >= TimeInterval(vaultTimeout.seconds)
+                // Reboot detection: negative elapsed time means device rebooted
+                if elapsedMonotonic < 0 {
+                    return true // Force timeout after reboot (safe approach)
```
matt-livefront: 🤔 What happens if you reboot, wait, and then elapsedMonotonic is positive, and within the timeout window. You'd have to time it right, but is that a concern?

BitwardenShared/Core/Vault/Services/SyncService.swift
```diff
@@ -301,12 +302,49 @@ class DefaultSyncService: SyncService {
...
+        if elapsedMonotonic < 0 {
+            return true // Force sync after reboot (safer approach)
+        }
+
+        // Check if minimum interval has passed
+        if elapsedMonotonic < Constants.minimumSyncInterval {
+            return false
+        }
+
+        return nil
```
matt-livefront: ❓ This sync logic always trips me up for some reason, so just so I'm understanding this correctly. If `isPeriodic=true`, then the following applies:

- If `elapsedMonotonic < 0` due to reboot, we return `true` to force a sync (without even checking account revision date, is this expected?)
- If the minimum sync interval hasn't elapsed, we return `false` to _not_ sync
- If the minimum sync interval has elapsed, we return `nil` to check the revision date to see if a sync is needed

I think returning `nil` when the minimum sync interval has elapsed feels counterintuitive, but I see why if my understanding is correct. Perhaps the documentation could be updated to make this clearer. Or switching to an enum instead of optional bool.

Maybe something like this:
```
    /// - Returns:
    ///   - `true` if a periodic sync should occur.
    ///   - `false` it a periodic sync should not occur.
    ///   - `nil` if the minimum sync interval has elapsed, but the account revision date should be
    ///     checked before syncing.
```

BitwardenKit/Core/Platform/Services/TimeProvider.swift
fedemkr: Nice catch, I thought it was only for third-party SDK libraries, but you're right it applies to apps as well. I'll make the update.

andrebispo5:
@fedemkr do not forget to add the proper label to the PR 👍

BitwardenKit/UI/Platform/Application/Utilities/PreviewTimeProvider.swift
```diff
@@ -7,6 +7,9 @@ public class PreviewTimeProvider: TimeProvider {
     /// A fixed date to use for previews.
     public var fixedDate: Date
 
+    /// THe current monotonic time.
```
KatherineInCode: ⛏️ 
```suggestion
    /// The current monotonic time.
```

BitwardenShared/Core/Auth/Services/KeychainRepository.swift
```diff
@@ -45,6 +48,7 @@ enum KeychainItem: Equatable, KeychainStorageKeyPossessing {
              .authenticatorVaultKey,
              .deviceKey,
              .lastActiveTime,
+             .lastActiveMonotonicTime,
```
KatherineInCode, line 71: ⛏️ Should be alphabetized

BitwardenShared/Core/Auth/Services/KeychainRepositoryTests.swift
```diff
@@ -572,4 +572,84 @@ final class KeychainRepositoryTests: BitwardenTestCase { // swiftlint:disable:th
         // Verify add was NOT called
         XCTAssertNil(keychainService.addAttributes)
     }
+
+    // MARK: UserSessionKeychainRepository Tests
```
KatherineInCode: 🤔 Should these go in the `KeychainRepository+UserSessionTests` instead?

BitwardenShared/Core/Platform/Services/StateService.swift
```diff
@@ -231,6 +231,13 @@ protocol StateService: AnyObject {
     ///
     func getLastSyncTime(userId: String?) async throws -> Date?
 
+    /// Gets the monotonic time of the last sync for a user.
```
KatherineInCode, line 234: 🤔 Is there a reason to also put these in `StateService` instead of keeping it segregated out into the `UserSessionStateService`?

BitwardenShared/Core/Platform/Services/Stores/AppSettingsStore.swift
```diff
@@ -212,6 +212,13 @@ protocol AppSettingsStore: AnyObject {
     ///
     func lastSyncTime(userId: String) -> Date?
 
+    /// Gets the monotonic time of the last sync for the user ID.
```
KatherineInCode, line 218: 🤔 Does it make sense to keep it in both UserDefaults and the Keychain? Shouldn't we just keep it in the Keychain?

BitwardenShared/Core/Platform/Services/TestHelpers/MockStateService.swift
```diff
@@ -63,6 +63,7 @@ class MockStateService: StateService, ActiveAccountStateProvider { // swiftlint:
     var getBiometricAuthenticationEnabledResult: Result<Void, Error> = .success(())
     var lastSyncTimeByUserId = [String: Date]()
     var lastSyncTimeSubject = CurrentValueSubject<Date?, Never>(nil)
+    var lastSyncMonotonicTimeByUserId = [String: TimeInterval?]()
```
KatherineInCode: ⛏️ Alphabetization

BitwardenShared/Core/Platform/Services/TestHelpers/MockStateService.swift
```diff
@@ -608,6 +614,11 @@ class MockStateService: StateService, ActiveAccountStateProvider { // swiftlint:
         lastSyncTimeByUserId[userId] = date
     }
 
+    func setLastSyncMonotonicTime(_ monotonicTime: TimeInterval?, userId: String?) async throws {
```
KatherineInCode, line 646: ⛏️ Alphabetizing/grouping

BitwardenShared/Core/Vault/Services/SyncService.swift
fedemkr: Yes, exactly. I've updated it to use an enum instead as it's easier to follow that way. Thanks!

BitwardenShared/Core/Platform/Services/StateService.swift
fedemkr: I was following the same we do for `getLastSyncTime` as you can see there. Maybe you got confused with `getLastActiveTime` which is more specific to the session.

BitwardenShared/Core/Platform/Services/Stores/AppSettingsStore.swift
fedemkr: The one in User defaults is the last time a synchronization has been done and the one in the keychain is the last time the app was active. They are two different values. They were added in the same place where we're using the "old" OS normal clock values.

BitwardenShared/Core/Platform/Services/StateService.swift
KatherineInCode: Ah, yes, I missed that there was a difference between active time and sync time here.

BitwardenKit/Core/Platform/Services/TimeProvider.swift
```diff
@@ -8,10 +8,40 @@ import Foundation
...
+    /// If the divergence between the two exceeds the threshold, tampering is detected.
+    ///
+    /// - Parameters:
+    ///   - lastMonotonicTime: The previously stored monotonic time (seconds since boot).
+    ///   - lastWallClockTime: The previously stored wall-clock time.
+    ///   - divergenceThreshold: Maximum allowed divergence between the two clocks in seconds.
+    /// - Returns: A `TamperResistantTimeResult` containing tampering detection status and elapsed times.
+    ///
+    func calculateTamperResistantElapsedTime(
+        since lastMonotonicTime: TimeInterval,
```
KatherineInCode: 🤔 Does it make sense for these properties to not be alphabetized?

BitwardenKit/Core/Platform/Services/TimeProvider.swift
```diff
@@ -25,13 +55,112 @@ public protocol TimeProvider: AnyObject {
...
 
+    public var monotonicTime: TimeInterval {
+        ProcessInfo.processInfo.systemUptime
+    }
+
+    // MARK: Init
+
     public init() {}
 
+    // MARK: - Methods
```
KatherineInCode: ⛏️ We typically only use the horizontal-rule `MARK` for denoting top-level objects in a file, not for sections inside of a top-level object
```suggestion
    // MARK: Methods
```

BitwardenKit/Core/Platform/Services/TimeProvider.swift
```diff
@@ -25,13 +55,119 @@ public protocol TimeProvider: AnyObject {
...
+}
+
+// MARK: - TamperResistantTimeResult
+
+/// Result of tamper-resistant elapsed time calculation.
+public struct TamperResistantTimeResult {
+    // MARK: Properties
+
+    /// Whether clock manipulation or device reboot was detected.
+    public let tamperingDetected: Bool
```
KatherineInCode: 🎨 Should these properties be alphabetized? As well, would it make sense to break this struct out into its own file?

Bitwarden/Application/Support/PrivacyInfo.xcprivacy
```diff
@@ -15,6 +15,12 @@
 			<array>
 				<string>1C8F.1</string>
 			</array>
+			<key>NSPrivacyAccessedAPIType</key>
+			<string>NSPrivacyAccessedAPICategorySystemBootTime</string>
+			<key>NSPrivacyAccessedAPITypeReasons</key>
+			<array>
+				<string>35F9.1</string>
+			</array>
 		</dict>
```
matt-livefront, line 18: 🤔 This looks off - both of these are grouped in a single `<dict>`. Shouldn't there be two dictionaries in the array?

It looks like Xcode only shows one of them:

Image

BitwardenKit/Core/Platform/Services/TimeProviderTests.swift
```diff
@@ -0,0 +1,240 @@
...
+            divergenceThreshold: 15.0,
+        )
+
+        XCTAssertFalse(result.tamperingDetected)
+        XCTAssertEqual(result.elapsedMonotonic, elapsedTime, accuracy: 0.1)
+        XCTAssertEqual(result.elapsedWallClock, elapsedTime, accuracy: 0.1)
+        XCTAssertEqual(result.effectiveElapsed, elapsedTime, accuracy: 0.1)
+    }
+
+    // MARK: - timeSince Tests
```
KatherineInCode: ⛏️  I don't think we want a horizontal-rule `MARK` here:
```suggestion
    // MARK: timeSince Tests
```

BitwardenKit/Core/Platform/Services/TimeProvider.swift
fedemkr: I'll order it but I don't think the struct should be in its own file as it's specifically related to the `TimeProvider` and it's pretty basic.

matt-livefront (dismissed):

BitwardenKit/Core/Platform/Services/TimeProvider.swift
```diff
@@ -57,12 +59,27 @@ public protocol TimeProvider: AnyObject {
 public class CurrentTime: TimeProvider {
     // MARK: Properties
 
+    /// Hardware-specific conversion ratio used to translate raw `mach_continuous_time()` ticks
+    /// into nanoseconds: `nanoseconds = ticks × numer / denom`.
```
matt-livefront: ❓ Can `clock_gettime_nsec_np(CLOCK_MONOTONIC_RAW)` be used instead to get nanoseconds instead of ticks?

> Prefer to use the equivalent clock_gettime_nsec_np(CLOCK_MONOTONIC_RAW) in nanoseconds.

https://developer.apple.com/documentation/kernel/1646199-mach_continuous_time#discussion

Bitwarden/Application/Support/PrivacyInfo.xcprivacy
```diff
@@ -16,6 +16,14 @@
 				<string>1C8F.1</string>
 			</array>
 		</dict>
+		<dict>
+			<key>NSPrivacyAccessedAPIType</key>
+			<string>NSPrivacyAccessedAPICategorySystemBootTime</string>
```
matt-livefront: 🤔 I wonder if we still need this after switching to `mach_continuous_time`? There's no warning in the docs as there was for `ProcessInfo.systemUptime`.

fedemkr: I have double checked and yes we can remove this section. It's a bit weird that is not required as it's very similar to the other one and it seems that it can be used for the same reason flagging in PrivacyInfo is necessary. But well if it's not required I won't add it here. Good catch!

BitwardenShared/Core/Auth/Services/KeychainRepository.swift
```diff
@@ -529,6 +548,44 @@ extension DefaultKeychainRepository: UserSessionKeychainRepository {
         try await keychainServiceFacade.setValue(value, for: BitwardenKeychainItem.lastActiveTime(userId: userId))
     }
 
+    func getLastActiveMonotonicTime(userId: String) async throws -> TimeInterval? {
+        do {
+            let stored = try await keychainServiceFacade.getValue(
+                for: BitwardenKeychainItem.lastActiveMonotonicTime(userId: userId),
+            )
+            return TimeInterval(stored)
+        } catch KeychainServiceError.osStatusError(errSecItemNotFound) {
```
matt-livefront: ❓ Should this also catch `KeychainServiceError.keyNotFound`?

BitwardenShared/Core/Auth/Services/KeychainRepository.swift
```diff
@@ -529,6 +548,44 @@ extension DefaultKeychainRepository: UserSessionKeychainRepository {
...
+        )
+    }
+
+    func getLastActiveBootEpoch(userId: String) async throws -> TimeInterval? {
+        do {
+            let stored = try await keychainServiceFacade.getValue(
+                for: BitwardenKeychainItem.lastActiveBootEpoch(userId: userId),
+            )
+            return TimeInterval(stored)
+        } catch KeychainServiceError.osStatusError(errSecItemNotFound) {
```
matt-livefront: ❓ Should this also catch `KeychainServiceError.keyNotFound`?

BitwardenShared/Core/Auth/Services/KeychainRepository.swift
```diff
@@ -529,6 +548,44 @@ extension DefaultKeychainRepository: UserSessionKeychainRepository {
...
+                for: BitwardenKeychainItem.lastActiveMonotonicTime(userId: userId),
+            )
+            return TimeInterval(stored)
+        } catch KeychainServiceError.osStatusError(errSecItemNotFound) {
+            return nil
+        }
+    }
+
+    func setLastActiveMonotonicTime(_ monotonicTime: TimeInterval?, userId: String) async throws {
+        let value = monotonicTime.map { String($0) } ?? ""
```
matt-livefront: ❓ If `monotonicTime` is `nil`, do we want to store the empty string as opposed to `nil`? Could this just be:

```suggestion
        let value = monotonicTime.map { String($0) }
```

BitwardenShared/Core/Auth/Services/KeychainRepository.swift
```diff
@@ -529,6 +548,44 @@ extension DefaultKeychainRepository: UserSessionKeychainRepository {
...
+                for: BitwardenKeychainItem.lastActiveBootEpoch(userId: userId),
+            )
+            return TimeInterval(stored)
+        } catch KeychainServiceError.osStatusError(errSecItemNotFound) {
+            return nil
+        }
+    }
+
+    func setLastActiveBootEpoch(_ bootEpoch: TimeInterval?, userId: String) async throws {
+        let value = bootEpoch.map { String($0) } ?? ""
```
matt-livefront: ❓ If `bootEpoch` is nil, do we want to store the empty string as opposed to nil? Could this just be:


```suggestion
        let value = bootEpoch.map { String($0) }
```

BitwardenShared/Core/Auth/Services/KeychainRepositoryTests.swift
```diff
@@ -92,6 +92,7 @@ final class KeychainRepositoryTests: BitwardenTestCase { // swiftlint:disable:th
                 BitwardenKeychainItem.accessToken(userId: "1").unformattedKey,
                 BitwardenKeychainItem.authenticatorVaultKey(userId: "1").unformattedKey,
                 BitwardenKeychainItem.biometrics(userId: "1").unformattedKey,
+                BitwardenKeychainItem.lastActiveMonotonicTime(userId: "1").unformattedKey,
```
matt-livefront, line 96: ❓ Should this also test `lastActiveBootEpoch`?

BitwardenShared/Core/Platform/Services/UserSessionStateService.swift
```diff
@@ -24,6 +24,38 @@ protocol UserSessionStateService { // sourcery: AutoMockable
     ///
     func setLastActiveTime(_ date: Date?, userId: String?) async throws
 
+    /// Gets the user's last active monotonic time within the app.
```
matt-livefront, line 29: 🎨 It looks like the other groups of methods in this protocol have `// MARK: Last Active Time` to separate out the methods. Should we do that here too?

BitwardenShared/Core/Vault/Services/VaultTimeoutService.swift
```diff
@@ -195,12 +201,45 @@ class DefaultVaultTimeoutService: VaultTimeoutService {
...
+            let lastActiveTime = try await userSessionStateService.getLastActiveTime(userId: userId)
+
+            // We need both times to calculate session timeout. If any of the times is not present
+            // then treat it as the session has timed out.
+            guard let lastActiveMonotonic, let lastActiveTime else {
+                return true
+            }
+
+            // Use monotonic time for tamper-resistant timeout checking
+            await flightRecorder.log("Checking timeout with monotonic time for userId: \(userId)")
```
matt-livefront: 🤔 This method gets called fairly frequently - is there value in logging this every time the method is called? Maybe we just log when a timeout occurs?

.typos.toml
```diff
@@ -24,3 +24,10 @@ Bitwarden = "Bitwarden"
 Rotateable = "Rotateable"
 RotateableKeySet = "RotateableKeySet"
 Sourcery = "Sourcery"
+
+[type.timeprovider]
+extend-glob = ["**/TimeProvider.swift"]
+
+[type.timeprovider.extend-words]
+# mach_timebase_info_data_t fields: numer (numerator) and denom (denominator)
+numer = "numer"
```
matt-livefront: ❓ Are these changes still necessary since `numer` was removed?

BitwardenShared/Core/Vault/Services/SyncService.swift
```diff
@@ -336,12 +365,59 @@ class DefaultSyncService: SyncService {
...
+        // Migration fallback: use wall-clock if monotonic time not available
+        guard let lastSyncMonotonic else {
+            await flightRecorder.log("Checking needs sync with wall-clock time for userId: \(userId)")
+            if lastSyncTime.addingTimeInterval(Constants.minimumSyncInterval) >= timeProvider.presentTime {
+                return .shouldNotSync
+            }
+            return .undefined
+        }
+
+        await flightRecorder.log("Checking needs sync with monotonic time for userId: \(userId)")
```
matt-livefront: ❓ This also ends up being called a lot, do we want to log this each time?

matt-livefront (commented):
I'm seeing that my account is locked whenever the app is backgrounded and foregrounded again, despite the timeout see to 15 minutes. It looks like `lastActiveMonotonic` is never being set.

BitwardenShared/Core/Vault/Services/VaultTimeoutService.swift
fedemkr: No need, missed to remove this when I removed the use of wall-clock check fallback when it was needed as something to check whether it was using monotonic or wall-clock check here. Now it just uses monotonic so no need for the log.

BitwardenShared/Core/Vault/Services/SyncService.swift
fedemkr: No need thanks!

fedemkr:
> I'm seeing that my account is locked whenever the app is backgrounded and foregrounded again, despite the timeout see to 15 minutes. It looks like lastActiveMonotonic is never being set.

It appears that with the move to the Keychain façade the `set` method for the keychain was calling the one with `Codable` instead of the `String` one because of the value being a `String?`. I've fixed that directly in the specific functions. I was going to add a new extension `set` method for `String?` however I don't want to potentially change all places that are currently calling `Codable` for this new one as I may break the `get` if it expects a json instead of a direct `String` value.

matt-livefront:
> > I'm seeing that my account is locked whenever the app is backgrounded and foregrounded again, despite the timeout see to 15 minutes. It looks like lastActiveMonotonic is never being set.
> 
> It appears that with the move to the Keychain façade the `set` method for the keychain was calling the one with `Codable` instead of the `String` one because of the value being a `String?`. I've fixed that directly in the specific functions. I was going to add a new extension `set` method for `String?` however I don't want to potentially change all places that are currently calling `Codable` for this new one as I may break the `get` if it expects a json instead of a direct `String` value.

Good catch. It looks looks like we might have benefited by mapping the optional to an empty string (`map { String($0) } ?? ""`), which I suggested we should remove. In that case, an optional string would be converted to an empty string and it wouldn't have used the Codable function. It looks like the other values probably don't use the Codable setValue, so it seems like we should be safe to add a setValue for `String?`. Should we add a ticket to confirm and do that work so we aren't bit by this in the future?

matt-livefront (dismissed):

KatherineInCode:
> > I'm seeing that my account is locked whenever the app is backgrounded and foregrounded again, despite the timeout see to 15 minutes. It looks like lastActiveMonotonic is never being set.
> 
> It appears that with the move to the Keychain façade the `set` method for the keychain was calling the one with `Codable` instead of the `String` one because of the value being a `String?`. I've fixed that directly in the specific functions. I was going to add a new extension `set` method for `String?` however I don't want to potentially change all places that are currently calling `Codable` for this new one as I may break the `get` if it expects a json instead of a direct `String` value.

I'm pretty sure adding specific `String?` get/sets should be fine; I don't think anything else was using or is using that quite yet, since it should have started failing when the façade was introduced 🤔 but I could be wrong; that we have a `Codable` layer on top of `String` with public interfaces to both leads to potential weirdness

matt-livefront (approved):

fedemkr:
@claude perform code review.

PR #2327, opened 2026-02-05
[PM-31654] fix: Handle premium subscription required on Archive group in the vault list

fedemkr:
## 🎟️ Tracking

[PM-31654](https://bitwarden.atlassian.net/browse/PM-31654)

## 📔 Objective

Handle premium subscription required on the Archive group row in the vault list.
Added an accessory icon to the vault list row view to display the "lock" icon in there.

## 📸 Screenshots

Archive premium subscription required

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-31654]: https://bitwarden.atlassian.net/browse/PM-31654?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/UI/Vault/Vault/VaultList/VaultListItemTests.swift
+++ b/BitwardenShared/UI/Vault/Vault/VaultList/VaultListItemTests.swift
@@ -39,7 +39,7 @@ class VaultListItemTests: BitwardenTestCase { // swiftlint:disable:this type_bod
 
     /// `accessoryIcon` returns nil for group items that don't require premium.
     func test_accessoryIcon_group_noPremiumRequired() {
-        let item = VaultListItem(id: "test", hasPremium: true, itemType: .group(.archive, 0))
+        let item = VaultListItem(id: "test", hasPremium: false, itemType: .group(.login, 0))
         XCTAssertNil(item.accessoryIcon)
     }
 
--- a/BitwardenShared/UI/Vault/Views/VaultListItemRow/VaultListItemRowView+SnapshotTests.swift
+++ b/BitwardenShared/UI/Vault/Views/VaultListItemRow/VaultListItemRowView+SnapshotTests.swift
@@ -37,6 +37,18 @@ class VaultListItemRowViewTests: BitwardenTestCase {
 
     // MARK: Snapshots
 
+    /// Test that the archive with premium subscription expired renders correctly.
+    @MainActor
+    func disabletest_snapshot_archivePremiumRequired() {
+        processor.state.iconBaseURL = .example
+        processor.state.item = VaultListItem(
+            id: "ID",
+            hasPremium: false,
+            itemType: .group(.archive, 0)
+        )
+        assertSnapshot(of: subject, as: .fixedSize())
+    }
+
     /// Test that the default view renders correctly.
     @MainActor
     func disabletest_snapshot_default() {
@@ -170,18 +182,6 @@ class VaultListItemRowViewTests: BitwardenTestCase {
             ),
         )
         processor.state.isFromExtension = true
-        assertSnapshot(of: subject, as: .fixedSize())
-    }
-
-    /// Test that the archive with premium subscription expired renders correctly.
-    @MainActor
-    func disabletest_snapshot_archivePremiumRequired() {
-        processor.state.iconBaseURL = .example
-        processor.state.item = VaultListItem(
-            id: "ID",
-            hasPremium: false,
-            itemType: .group(.archive, 0)
-        )
         assertSnapshot(of: subject, as: .fixedSize())
     }
 }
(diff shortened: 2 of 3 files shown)


BitwardenShared/UI/Vault/Vault/VaultList/VaultListItem.swift
```diff
@@ -40,6 +40,10 @@ public struct VaultListItem: Equatable, Identifiable, Sendable, VaultItemWithDec
     /// The identifier for the item.
     public let id: String
 
+    /// Whether the user has premium subscription.
+    /// This is only used on Archive group for now, so it's not being set for any other occassions.
+    public let hasPremium: Bool
```
matt-livefront, line 45: 🤔 One other option to adding this here might be to add it to an associated value of `VaultListGroup.archive`, similar to how we pass along folder and collection IDs. Then it would be clearer that it only applies to the archive group.

This might end up requiring some changes to where we have `group == .archive` checks though. But we might be able to avoid setting the `hasPremium` property in `VaultListProcessor` since it's already attached to the archive group that you tap on.

BitwardenShared/UI/Vault/Vault/VaultList/VaultListItemTests.swift
```diff
@@ -25,6 +25,42 @@ class VaultListItemTests: BitwardenTestCase { // swiftlint:disable:this type_bod
...
+
+    /// `accessoryIcon` returns nil for TOTP items.
+    func test_accessoryIcon_totp() {
+        let item = VaultListItem.fixtureTOTP(totp: .fixture())
+        XCTAssertNil(item.accessoryIcon)
+    }
+
+    /// `accessoryIcon` returns nil for group items that don't require premium.
+    func test_accessoryIcon_group_noPremiumRequired() {
+        let item = VaultListItem(id: "test", hasPremium: true, itemType: .group(.archive, 0))
```
matt-livefront: ❓ Should this item type use a group other than archive? Right now I think it's returning nil because group is archive but hasPremium = true, which contradicts the test name and comment.

```suggestion
    /// `accessoryIcon` returns nil for group items that don't require premium.
    func test_accessoryIcon_group_noPremiumRequired() {
        let item = VaultListItem(id: "test", hasPremium: false, itemType: .group(.login, 0))
```

BitwardenShared/UI/Vault/Views/VaultListItemRow/VaultListItemRowView+SnapshotTests.swift
```diff
@@ -172,4 +172,16 @@ class VaultListItemRowViewTests: BitwardenTestCase {
         processor.state.isFromExtension = true
         assertSnapshot(of: subject, as: .fixedSize())
     }
+
+    /// Test that the archive with premium subscription expired renders correctly.
+    @MainActor
+    func disabletest_snapshot_archivePremiumRequired() {
```
matt-livefront: ⛏️ Alphabetize this?

BitwardenShared/UI/Vault/Views/VaultListItemRow/VaultListItemRowView.swift
```diff
@@ -91,15 +91,35 @@ struct VaultListItemRowView: View {
...
                         Spacer()
-                        Text("\(count)")
-                            .styleGuide(.body)
-                            .foregroundColor(SharedAsset.Colors.textSecondary.swiftUIColor)
-                            .accessibilityIdentifier("GroupCountLabel")
+
+                        if let accessoryIcon = store.state.item.accessoryIcon {
+                            Image(decorative: accessoryIcon)
+                                .imageStyle(.rowIcon)
+                                .padding(.vertical, 19)
```
matt-livefront: 🤔 It looks like there may be extra vertical padding on the Archive row compared to Trash. Is that caused by the padding on the icon? I wonder if that's necessary, or if we can just rely on the overall padding on the row.

Image

BitwardenShared/UI/Vault/Vault/VaultList/VaultListItem.swift
fedemkr: Yes I thought about that as well, but I thought it was simpler to have it here and it could serve the purpose to any other item/group logic that needs `hasPremium` as it's not a particular feature for Archive but something that so far is used there.
However, if you still feel like it'd be better there I can make the change.

BitwardenShared/UI/Vault/Views/VaultListItemRow/VaultListItemRowView.swift
fedemkr: Yep changed, looks good without the padding.
image

BitwardenShared/UI/Vault/Vault/VaultList/VaultListItem.swift
matt-livefront: Makes sense!

PR #2329, opened 2026-02-06
[PM-31719] fix: Race condition causing EXC_BAD_ACCESS in FetchedResultsPublisher

matt-livefront:
## 🎟️ Tracking

[PM-31719](https://bitwarden.atlassian.net/browse/PM-31719)

## 📔 Objective

Fixes a race condition in FetchedResultsPublisher that caused EXC_BAD_ACCESS crashes when subscriptions were cancelled before or during Core Data fetch operations.

The crash occurred when NSFetchedResultsController was accessed from multiple queues simultaneously:
- The controller property could be set to nil from self.queue in cancel()
- While simultaneously being accessed from the Core Data context's queue in context.perform blocks
- This created a race condition where the controller could be deallocated while being accessed

To fix this, a strong reference to controller is captured in both context.perform closures using [weak self, controller] capture lists. This ensures:
  - The controller remains valid for the duration of the closure execution
  - Even if cancel() nils out self.controller, the closure's captured reference prevents premature deallocation
  - Thread-safe access to the controller across different dispatch queues

Fixes the following crashes in Crashlytics:

https://console.firebase.google.com/u/0/project/bitwarden-c2b35/crashlytics/app/ios:com.8bit.bitwarden/issues/3589af8f72862f293153d9cf7f13fbb3?time=7d&types=crash&sessionEventKey=25804dda3d7a42959ffb84fdaef19287_2181954787163989163

https://console.firebase.google.com/u/0/project/bitwarden-c2b35/crashlytics/app/ios:com.8bit.bitwarden/issues/6845c9fef11574b83f832e7d2d1b4aea?time=7d&types=crash&sessionEventKey=3529e07509b24874b1b2c19c86c63d3d_2181842795786969972

## ⏰ Reminders before review

- CI builds passed

## 🦮 Reviewer guidelines

[PM-31719]: https://bitwarden.atlassian.net/browse/PM-31719?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenKit/Core/Platform/Utilities/FetchedResultsPublisher.swift
+++ b/BitwardenKit/Core/Platform/Utilities/FetchedResultsPublisher.swift
@@ -137,10 +137,13 @@ private final class FetchedResultsSubscription<SubscriberType, ResultType, Outpu
 
         controller?.delegate = self
 
-        context.perform { [weak self] in
+        // Capture a strong reference to controller to prevent a race condition where cancel()
+        // could set self.controller to nil on self.queue while this perform block executes
+        // on the context's queue, causing EXC_BAD_ACCESS when accessing the deallocated controller.
+        context.perform { [weak self, controller] in
             do {
-                try self?.controller?.performFetch()
-                if self?.controller?.fetchedObjects != nil {
+                try controller?.performFetch()
+                if controller?.fetchedObjects != nil {
                     self?.queue.async {
                         self?.hasChangesToSend = true
                         self?.fulfillDemand()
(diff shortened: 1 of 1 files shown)


KatherineInCode (approved):
I'm a little surprised that it would still have an issue, even though the `controller` is optional in the block 🤔 but I'm probably missing some subtlety. Good catch!

matt-livefront:
> I'm a little surprised that it would still have an issue, even though the `controller` is optional in the block 🤔 but I'm probably missing some subtlety. Good catch!

The timing has to be just right. The optional will return `nil` if it's deallocated before it's accessed.

But one thread could call `controller?.performFetch()` which starts to fetch results. At the same time a different thread sets `controller` to `nil`. Because the first thread doesn't have a strong reference to `controller`, its reference count is now 0 and it can be deallocated. If `performFetch()` tries to access any freed memory, it crashes.

With the change, a strong reference to `controller` is captured in the closure, which will prevent it from being deallocated until the closure finishes.

PR #2359, opened 2026-02-20
[PM-32528] chore: Improve push notification logging

matt-livefront:
## 🎟️ Tracking

[PM-32528](https://bitwarden.atlassian.net/browse/PM-32528)

## 📔 Objective

Adds logging of push notifications to the flight recorder and improves error handling around decoding the push notification payload.

Push notification received:
> [Notification] Received push notification, type: syncCipherUpdate

Push notification skipped for inactive user:
> [Notification] Skipping syncCipherUpdate: userId does not match active user

Push notification decoding error:
> Error: BitwardenShared.PushNotificationDataError.payloadDecodingFailed(type: Optional(BitwardenShared.NotificationType.syncCipherUpdate), underlyingError: Swift.DecodingError.keyNotFound(CodingKeys(stringValue: "id", intValue: nil), Swift.DecodingError.Context(codingPath: [], debugDescription: "No value associated with key CodingKeys(stringValue: \"id\", intValue: nil) (\"id\").", underlyingError: nil)))

Changes:
- PushNotificationDataError: New error type thrown when a notification payload fails to decode. Embeds the notification type and the underlying DecodingError as NSUnderlyingErrorKey, which Crashlytics will surface with full field-level detail.
- PushNotificationData.data(): Changed from -> T? (silent nil on failure) to throws -> T, wrapping all failure modes in PushNotificationDataError.payloadDecodingFailed.
- NotificationService: Added FlightRecorder dependency. Logs the notification type at receipt time, and logs a userId mismatch message in each switch case where the notification is skipped. If a decode error occurs, it propagates to the existing errorReporter.log(error:) call in messageReceived.

[PM-32528]: https://bitwarden.atlassian.net/browse/PM-32528?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/Core/Platform/Services/NotificationService.swift
+++ b/BitwardenShared/Core/Platform/Services/NotificationService.swift
@@ -206,33 +206,25 @@ class DefaultNotificationService: NotificationService {
             case .syncCipherCreate,
                  .syncCipherUpdate:
                 let data: SyncCipherNotification = try notificationData.data()
-                if data.userId == userId {
+                if await checkActiveUser(notification: data, type: type, userId: userId) {
                     try await syncService.fetchUpsertSyncCipher(data: data)
-                } else {
-                    await flightRecorder.log("[Notification] Skipping \(type): userId does not match active user")
                 }
             case .syncFolderCreate,
                  .syncFolderUpdate:
                 let data: SyncFolderNotification = try notificationData.data()
-                if data.userId == userId {
+                if await checkActiveUser(notification: data, type: type, userId: userId) {
                     try await syncService.fetchUpsertSyncFolder(data: data)
-                } else {
-                    await flightRecorder.log("[Notification] Skipping \(type): userId does not match active user")
                 }
             case .syncCipherDelete,
                  .syncLoginDelete:
                 let data: SyncCipherNotification = try notificationData.data()
-                if data.userId == userId {
+                if await checkActiveUser(notification: data, type: type, userId: userId) {
                     try await syncService.deleteCipher(data: data)
-                } else {
-                    await flightRecorder.log("[Notification] Skipping \(type): userId does not match active user")
                 }
             case .syncFolderDelete:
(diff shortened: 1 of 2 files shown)


BitwardenShared/Core/Platform/Services/NotificationService.swift
```diff
@@ -192,28 +198,41 @@ class DefaultNotificationService: NotificationService {
...
             switch type {
             case .syncCipherCreate,
                  .syncCipherUpdate:
-                if let data: SyncCipherNotification = notificationData.data(), data.userId == userId {
+                let data: SyncCipherNotification = try notificationData.data()
+                if data.userId == userId {
                     try await syncService.fetchUpsertSyncCipher(data: data)
+                } else {
+                    await flightRecorder.log("[Notification] Skipping \(type): userId does not match active user")
                 }
```
fedemkr: 🎨 I see this pattern is repeated amongst most notifications. Would it make sense to centralize this in a helper function?
We could even add a protocol `NotificationWithUser`:
```swift
protocol NotificationWithUser {
    /// The user id that owns the cipher.
    let userId: String
}
```
Then conform to such protocol on the notifications where applicable.
Therefore we can just pass the argument to the validation function as the protocol:
```swift
private func checkActiveUser(
    notification: NotificationWithUser, 
    userId: String, 
    type: NotificationType,
) async -> Bool {
    guard notification.userId == userId else {
        await flightRecorder.log("[Notification] Skipping \(type): userId does not match active user")
        return false
    }
    return true
}
```
And then usage is centralized:
```swift
if checkActiveUser(notification: data, userId: userId, type: type) {
    try await syncService.fetchUpsertSyncCipher(data: data)
}
```

matt-livefront: Good idea!

PR #2361, opened 2026-02-20
[PM-31720] fix: Only update credential identity store when app is foregrounded

matt-livefront:
## 🎟️ Tracking

[PM-31720](https://bitwarden.atlassian.net/browse/PM-31720)

## 📔 Objective

In Crashlytics, there's a number of `ASCredentialIdentityStoreErrorDomain Code: 0` errors.

https://console.firebase.google.com/project/bitwarden-c2b35/crashlytics/app/ios:com.8bit.bitwarden/issues/994083f5f31b455b3450e50d39ab8650?time=7d&sessionEventKey=93d7d216d5d541238973f12dc31d5204_2181089556910192743

Most of these occur while the app is backgrounded. I was able to intermittently reproduce this by triggering updates for the identity store while the app is backgrounded.

This change ensures that the app is foregrounded before updating the identity store. If a vault change occurs while the app is backgrounded, the identity store will be updated again once the app is foregrounded, provided the vault is still unlocked.

[PM-31720]: https://bitwarden.atlassian.net/browse/PM-31720?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/Core/Platform/Services/NotificationCenterService.swift
+++ b/BitwardenShared/Core/Platform/Services/NotificationCenterService.swift
@@ -30,7 +30,7 @@ protocol NotificationCenterService: AnyObject {
 class DefaultNotificationCenterService: NotificationCenterService {
     // MARK: Properties
 
-    /// Cancellables for the foreground/background notification subscriptions.
+    /// Cancellables for any subscriptions owned by the service.
     private var cancellables = Set<AnyCancellable>()
 
     /// The subject tracking the app's current foreground state.


BitwardenShared/Core/Platform/Services/NotificationCenterService.swift
```diff
@@ -22,8 +26,14 @@ protocol NotificationCenterService: AnyObject {
 class DefaultNotificationCenterService: NotificationCenterService {
     // MARK: Properties
 
+    /// Cancellables for the foreground/background notification subscriptions.
+    private var cancellables = Set<AnyCancellable>()
+
+    /// The subject tracking the app's current foreground state.
+    private let isInForegroundSubject = CurrentValueSubject<Bool, Never>(true)
```
fedemkr: 🤔 Should it start with `false` to account on flows that don't have the app in foreground like Autofilling without user interaction? Not sure if some of such flows would actually trigger the credential identity store update but raising this as we may use this publisher's subject for other parts.

matt-livefront: I think defaulting this to `false` makes sense. My only concern is that you have to be careful with `willEnterForegroundNotification` on app startup. The initializer that that subscribes to that notification is called early enough on app startup that the notification does fire on a normal app startup in my testing. However, it doesn't fire during the autofill extension startup (I think because that app is already foregrounded when the CredentialProviderViewController methods are called). Given that we don't currently sync all identities in the extension, this is probably a moot point. `false` would make it so that if the app is launched in the background, it would still be `false` before the app is foregrounded, which is better. I can update it.

fedemkr (dismissed):

BitwardenShared/Core/Platform/Services/NotificationCenterService.swift
fedemkr: hmm interesting, alright let's keep that in mind in case we need the foreground in the extensions; hopefully we won't.

fedemkr: Should we add a clarification in the docs for `isInForegroundSubject` or the publisher to indicate this caveat?

matt-livefront: Yep, added.

BitwardenShared/Core/Platform/Services/NotificationCenterService.swift
```diff
@@ -22,8 +30,14 @@ protocol NotificationCenterService: AnyObject {
 class DefaultNotificationCenterService: NotificationCenterService {
     // MARK: Properties
 
+    /// Cancellables for the foreground/background notification subscriptions.
+    private var cancellables = Set<AnyCancellable>()
```
KatherineInCode, line 34: 🎨 If the cancellables are just for foreground/background notifications, does it make sense for the variable name to include that in some way, instead of just being `cancellables`?

matt-livefront: I could remove the "foreground/background notification" wording. It doesn't need to be scoped to those notifications specifically, that's just what's stored in there now. The distinction is that those notification subscriptions are owned by this service which is different than the other publishers which are owned by whatever subscribes to them.

KatherineInCode: That makes sense. I think it'd be better to update the comment, then.

fedemkr (approved):

PR #2369, opened 2026-02-23
[PM-32539] [PM-32540] bug: Fix Sends error dialogs

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32536
https://bitwarden.atlassian.net/browse/PM-32539
https://bitwarden.atlassian.net/browse/PM-32540

## 📔 Objective

Fix text on Sends error dialogs and improve email validation to be consistent with the Android client.

## 📸 Screenshots

Simulator Screenshot - iPhone 17 Pro - 2026-02-23 at 15 59 57
Simulator Screenshot - iPhone 17 Pro - 2026-02-23 at 15 59 47

### DIFF
--- a/BitwardenShared/UI/Auth/StartRegistration/StartRegistrationProcessor.swift
+++ b/BitwardenShared/UI/Auth/StartRegistration/StartRegistrationProcessor.swift
@@ -145,7 +145,7 @@ class StartRegistrationProcessor: StateProcessor<
                 throw StartRegistrationError.emailEmpty
             }
 
-            guard email.isValidEmail() else {
+            guard email.isValidEmail(useStrictValidation: false) else {
                 throw StartRegistrationError.invalidEmail
             }
 


BitwardenShared/UI/Auth/StartRegistration/StartRegistrationProcessor.swift
```diff
@@ -145,7 +145,7 @@ class StartRegistrationProcessor: StateProcessor<
                 throw StartRegistrationError.emailEmpty
             }
 
-            guard email.isValidEmail else {
+            guard email.isValidEmail() else {
```
matt-livefront: ❓ Should this use `useStrictValidation: false`?

BitwardenKit/Core/Platform/Extensions/String.swift
```diff
@@ -64,11 +64,20 @@ public extension String {
...
     ///
-    /// An email is considered valid if it has at least one `@` symbol in it.
-    var isValidEmail: Bool {
-        contains("@")
+    /// - Parameter useStrictValidation: If `true`, validates against a regex pattern requiring
+    ///   a properly formatted email (e.g., `user@domain.com`). If `false`, only checks for
+    ///   the presence of an `@` symbol. Defaults to `true`.
+    /// - Returns: `true` if the string is a valid email according to the validation mode.
+    ///
+    func isValidEmail(useStrictValidation: Bool = true) -> Bool {
```
matt-livefront, line 74: 🤔 In general, it's better to validate emails on the server rather than relying on a regex which could produce false negatives, preventing a user from using a valid email. Is this validation needed because the email are encrypted? I think it might be better to default `useStrictValidation` to `false` here and only require strict validation if really needed (I would avoid it for login / sign up for example).

andrebispo5: I agree with you but I was told to match the android behaviour: https://github.com/bitwarden/android/blob/3360999706f21bfc866d566459416e7e5d0f368d/ui/src/main/kotlin/com/bitwarden/ui/platform/base/util/StringExtensions.kt#L59

andrebispo5: I removed the strict validation on registration.

matt-livefront: Thanks. It looks like Android added the `useStrictValidation` flag because someone reported that their email wasn't passing the regex (https://github.com/bitwarden/android/issues/5687), which tells me that the regex is too restrictive. I still disagree with this in principle, but since it’s not part of the login flow and we need to match Android, I’m okay with it for consistency.

matt-livefront (approved):

PR #2374, opened 2026-02-24
[PM-24195] fix: Pass userId to refreshToken and setTokens

LRNcardozoWDF:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-24195

## 📔 Objective

There's a possible race condition when refreshing the token with that in mind I'm adding more logs and passing the userId of the active account.


BitwardenShared/UI/Vault/VaultItem/AddEditItem/AddEditItemProcessor.swift
```diff
@@ -698,7 +698,14 @@ final class AddEditItemProcessor: StateProcessor<// swiftlint:disable:this type_
...
             await coordinator.showErrorAlert(error: error)
-            services.errorReporter.log(error: error)
+            if case let ServerError.error(errorResponse: errorResponse) = error,
+               errorResponse.message.contains("Cipher was not encrypted for the current user") {
+                services.errorReporter.log(error: BitwardenError.generalError(
+                    type: "Save item failed",
+                    message: errorResponse.message,
+                    error: error,
+                ))
+            }
```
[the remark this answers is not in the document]
LRNcardozoWDF: @claude this is an old version, I already updated this. Review the last changes.

BitwardenShared/UI/Vault/VaultItem/AddEditItem/AddEditItemProcessor.swift
```diff
@@ -698,7 +698,16 @@ final class AddEditItemProcessor: StateProcessor<// swiftlint:disable:this type_
...
+            if case let ServerError.error(errorResponse: errorResponse) = error,
+               errorResponse.message.contains("Cipher was not encrypted for the current user") {
+                services.errorReporter.log(error: BitwardenError.generalError(
+                    type: "Save item failed",
+                    message: errorResponse.message,
+                    error: error,
+                ))
+            } else {
+                services.errorReporter.log(error: error)
+            }
```
fedemkr: 🤔 Is this the only endpoint that can throw this error? If not, can we extract to some other place more generic?
🎨 Instead of using `if case` you can use `catch` with the specific error and avoid the if else:
```swift
catch let ServerError.error(errorResponse):
```

BitwardenShared/Core/Platform/Services/API/AccountTokenProvider.swift
```diff
@@ -93,6 +108,19 @@ actor DefaultAccountTokenProvider: AccountTokenProvider {
...
+                        let error = AccountTokenProviderError(
+                            userIdBefore: userIdBefore,
+                            userIdAfter: userIdAfter,
+                        )
+                        errorReporter.log(error: error)
+                    }
+                } catch {
+                    errorReporter.log(error: error)
+                }
+
```
fedemkr: 🤔 If this would happen then I think we shouldn't be setting the tokens in the wrong account. So you could just add the check before `setTokens` and add the `userId: String?` as parameter to `setTokens` so you can pass the user id you have in there instead of getting it again inside the function:

```suggestion
                let userIdAfter = try await stateService.getActiveAccountId()
                guard userIdBefore == userIdAfter else {
                    let error = AccountTokenProviderError(
                        userIdBefore: userIdBefore,
                        userIdAfter: userIdAfter,
                    )
                    errorReporter.log(error: error)
                    return
                }

                try await tokenService.setTokens(
                    accessToken: response.accessToken,
                    refreshToken: response.refreshToken,
                    expirationDate: expirationDate,
                    userId: userIdBefore // <-- maybe rename to expectedUserId instead of userIdBefore
                )

```

🤔 Moreover, there could potentially be a race condition between these two as well:
```swift
                let userIdBefore = try await stateService.getActiveAccountId()

                let refreshToken = try await tokenService.getRefreshToken()
```
So perhaps you could add `userId: String?` to `getRefreshToken` as well thus you can pass it, so the `userId` used throughout this `refreshToken` is as stable as possible.

BitwardenShared/Core/Platform/Services/API/AccountTokenProvider.swift
```diff
@@ -81,6 +93,9 @@ actor DefaultAccountTokenProvider: AccountTokenProvider {
             defer { self.refreshTask = nil }
 
             do {
+                // TODO: PM-33074 Remove logs after confirmation that the error doesn't happen anymore.
+                let userIdBefore = try await stateService.getActiveAccountId()
```
fedemkr: 🤔 I think you can remove the comment if you start passing the user ID to the token service functions. Also I think `userIdBefore` should be renamed to `userIdToRefresh` or something like that so it's easier to know what it refers to.

BitwardenShared/UI/Vault/VaultItem/AddEditItem/AddEditItemProcessor.swift
LRNcardozoWDF: I didn't found other endpoints, I'll look into it again.

BitwardenShared/UI/Vault/VaultItem/AddEditItem/AddEditItemProcessor.swift
```diff
@@ -696,18 +696,19 @@ final class AddEditItemProcessor: StateProcessor<// swiftlint:disable:this type_
             return
         } catch UserVerificationError.cancelled {
             return
+        } catch ServerError.error(errorResponse: let errorResponse)
+            where errorResponse.message.contains("Cipher was not encrypted for the current user") {
+            let serverError = ServerError.error(errorResponse: errorResponse)
```
fedemkr: 🎨 No need to recreate the server error, you already have it in the `error` variable.

fedemkr (commented):
👍  Looks good to me, I'd wait for @KatherineInCode or @matt-livefront's approval as well in case I've missed anything as this touches the account tokens.

BitwardenShared/Core/Auth/Services/TestHelpers/MockKeychainRepository.swift
```diff
@@ -35,6 +35,10 @@ class MockKeychainRepository: KeychainRepository {
     var setServerCommunicationConfigCalledConfig: BitwardenSdk.ServerCommunicationConfig?
     var setServerCommunicationConfigCalledHostname: String? // swiftlint:disable:this identifier_name
 
+    // Track which userId was passed to get/set methods for testing
+    var getAccessTokenUserId: String?
+    var getRefreshTokenUserId: String?
```
matt-livefront: ⛏️ Does it make sense to keep these next to `getRefreshTokenResult` and `getRefreshTokenResult`. That's the pattern we've used elsewhere and keeps them alphabetized.

BitwardenShared/Core/Platform/Services/TokenService.swift
```diff
@@ -102,6 +132,24 @@ actor DefaultTokenService: TokenService {
         try await keychainRepository.setRefreshToken(refreshToken, userId: userId)
         await stateService.setAccessTokenExpirationDate(expirationDate, userId: userId)
     }
+
+    func getAccessToken(userId: String) async throws -> String {
+        try await keychainRepository.getAccessToken(userId: userId)
+    }
+
+    func getActiveAccountId() async throws -> String {
+        try await stateService.getActiveAccountId()
```
matt-livefront: 🤔 Instead of exposing this via `TokenService`, would it be better to pass `ActiveAccountStateProvider` (implemented by `StateService`) to `AccountTokenProvider`?

BitwardenShared/Core/Platform/Services/TestHelpers/MockTokenService.swift
```diff
@@ -10,15 +10,35 @@ class MockTokenService: TokenService {
     var getIsExternalResult: Result<Bool, Error> = .success(false)
     var refreshToken: String? = "REFRESH_TOKEN"
 
+    // Track which userId was used in explicit userId methods
+    var getAccessTokenCalledWithUserId: String?
+    var getRefreshTokenCalledWithUserId: String?
+    var setTokensCalledWithUserId: String?
+    var activeAccountId: String = "1"
+    var accessTokenByUserId: [String: String] = [:]
+    var refreshTokenByUserId: [String: String] = [:]
```
matt-livefront: ⛏️ Can you alphabetize these with the other properties?

BitwardenShared/Core/Platform/Services/TokenService.swift
```diff
@@ -102,6 +132,24 @@ actor DefaultTokenService: TokenService {
         try await keychainRepository.setRefreshToken(refreshToken, userId: userId)
         await stateService.setAccessTokenExpirationDate(expirationDate, userId: userId)
     }
+
+    func getAccessToken(userId: String) async throws -> String {
+        try await keychainRepository.getAccessToken(userId: userId)
+    }
+
+    func getActiveAccountId() async throws -> String {
```
matt-livefront: ⛏️ Can you alphabetize these methods?

BitwardenShared/Core/Platform/Services/API/AccountTokenProvider.swift
```diff
@@ -42,15 +45,18 @@ actor DefaultAccountTokenProvider: AccountTokenProvider {
     ///   - httpService: The service used to make the API call to refresh the access token.
     ///   - timeProvider: The service used to get the present time.
     ///   - tokenService: The service used to get the current tokens from.
+    ///   - errorReporter: The service used to report non-fatal errors.
     ///
     init(
         httpService: HTTPService,
         timeProvider: TimeProvider = CurrentTime(),
         tokenService: TokenService,
+        errorReporter: ErrorReporter,
```
KatherineInCode: 🤔 Should these be alphabetized?

BitwardenShared/Core/Platform/Services/StateService.swift
```diff
@@ -10,7 +10,7 @@ import Foundation
 
 /// A protocol for a `StateService` which manages the state of the accounts in the app.
 ///
-protocol StateService: AnyObject {
+protocol StateService: ActiveAccountStateProvider {
```
matt-livefront: ❓ Does `StateService` need to conform to `ActiveAccountStateProvider` here? It's already being done for `DefaultStateService`:

https://github.com/bitwarden/ios/blob/c342e8211e2bacf3ff01d286912a4c4987a9ebfc/BitwardenShared/Core/Platform/Services/StateService.swift#L1399-L1400

BitwardenShared/Core/Platform/Services/API/AccountTokenProvider.swift
```diff
@@ -42,12 +48,17 @@ actor DefaultAccountTokenProvider: AccountTokenProvider {
     ///   - httpService: The service used to make the API call to refresh the access token.
     ///   - timeProvider: The service used to get the present time.
     ///   - tokenService: The service used to get the current tokens from.
+    ///   - errorReporter: The service used to report non-fatal errors.
```
matt-livefront: ⛏️ Can you update these docs to match the parameter order and add `activeAccountStateProvider`?


```suggestion
    ///   - activeAccountStateProvider: The provider for the active account state.
    ///   - errorReporter: The service used to report non-fatal errors.
    ///   - httpService: The service used to make the API call to refresh the access token.
    ///   - timeProvider: The service used to get the present time.
    ///   - tokenService: The service used to get the current tokens from.
```

BitwardenShared/Core/Platform/Services/StateService.swift
LRNcardozoWDF: @matt-livefront I missed that, but that doesn't seem to help me. My problem is that I added `ActiveAccountStateProvider` conformance to `StateService` to fix a compile error in `APIService`, where `stateService` was being passed directly as `activeAccountStateProvider` to `DefaultAccountTokenProvider`. I can try to add `activeAccountStateProvider` as a separate parameter to `APIService.init`. Does that approach work for you? Or do you have a better approach?

matt-livefront: Ah, I see. Yeah I think it'd be better to pass it as a separate parameter to `APIService` then.

KatherineInCode: Agreed. It's a separate parameter.

matt-livefront (dismissed):

BitwardenShared/Core/Platform/Services/API/TestHelpers/APIService+Mocks.swift
```diff
@@ -7,17 +7,21 @@ import Networking
 extension APIService {
     convenience init(
         accountTokenProvider: AccountTokenProvider? = nil,
+        activeAccountStateProvider: ActiveAccountStateProvider = MockStateService(),
```
KatherineInCode: ⛏️ You should be able to use just a `MockActiveAccountStateProvider` here, I'd think

BitwardenShared/Core/Platform/Services/API/ConfigAPIServiceTests.swift
```diff
@@ -21,7 +21,7 @@ class ConfigAPIServiceTests: BitwardenTestCase {
 
         client = MockHTTPClient()
         stateService = MockStateService()
-        subject = APIService(client: client, stateService: stateService)
+        subject = APIService(activeAccountStateProvider: stateService, client: client, stateService: stateService)
```
KatherineInCode: ⛏️ Nominally, I'd prefer this to be a `MockActiveAccountStateProvider`

KatherineInCode (approved):

PR #2378, opened 2026-02-25
[PM-32742] feat: Server communication cookie bootstrap config models and environment

fedemkr:
## 🎟️ Tracking

[PM-32742](https://bitwarden.atlassian.net/browse/PM-32742)

## 📔 Objective

Update config models and environment URLs to prepare for server communication cookie bootstrap.

[PM-32742]: https://bitwarden.atlassian.net/browse/PM-32742?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenKit/Core/Platform/Models/Domain/EnvironmentURLDataTests.swift
```diff
@@ -248,4 +248,39 @@ class EnvironmentURLDataTests: XCTestCase {
...
+            webVault: URL(string: "https://web.vault.example.com"),
+        )
+        XCTAssertEqual(
+            subject.proxyCookieRedirectConnectorURL?.absoluteString,
+            "https://web.vault.example.com/proxy-cookie-redirect-connector.html",
+        )
+    }
+
+    /// `proxyCookieRedirectConnectorURL` does not contain a `#` sign, unlike other subpage URLs.
+    func test_proxyCookieRedirectConnectorURL_doesNotContainPoundSign() {
```
KatherineInCode: 🤔 If we're already verifying the URL by absolute string, above, do we also need to verify that it doesn't have a pound sign in it?

BitwardenKit/Core/Platform/Models/Domain/ServerConfig.swift
```diff
@@ -93,3 +98,78 @@ public struct EnvironmentServerConfig: Equatable, Codable, Sendable {
         vault = responseModel.vault
     }
 }
+
+// MARK: - CommunicationSettings
+
+/// Server communication configuration settings.
+///
+public struct CommunicationSettings: Equatable, Codable, Sendable {
```
KatherineInCode: 🎨 I somewhat feel like this should be named `ServerCommunicationSettings` to be more explicit about what it is (along with `ServerCommunicationBoostrapSettings`). I also think they would probably be best in a seaparate file, to segregate out data objects from business logic.

BitwardenKit/Core/Platform/Services/API/Response/ConfigResponseModelTests.swift
```diff
@@ -0,0 +1,135 @@
+import TestHelpers
+import XCTest
+
+@testable import BitwardenKit
+
+class ConfigResponseModelTests: BitwardenTestCase {
+    // MARK: Tests
+
+    /// `init` stores all provided fields.
+    func test_init_withAllFields() {
```
KatherineInCode: 🎨 Do we really need to have a suite of tests on a basic data struct with no business logic? It feels like these are testing "does Swift initialize objects" and "does Swift Codable work"; all of which would be effectively covered by other tests that create/use these objects in the normal course of logic.

fedemkr: My bad, missed that when I reviewed the Claude Code. Removing it now.

BitwardenKit/Core/Platform/Models/Domain/ServerConfig.swift
fedemkr: Made the renaming changes. What business logic are you referring to? I think these are all data objects. I followed the same in-file structure as with `ThirdPartyServerConfig` and `EnvironmentServerConfig`.

KatherineInCode: Hah, yeah, they're fine in `ServerConfig`. For some reason I thought they were in a different file.

BitwardenKit/Core/Platform/Services/API/Response/ConfigResponseModelTests.swift
```diff
@@ -0,0 +1,97 @@
+import TestHelpers
+import XCTest
+
+@testable import BitwardenKit
+
+class ConfigResponseModelTests: BitwardenTestCase {
+    // MARK: Tests
+
+    /// `JSONResponse` decoding includes the `communication` field with `ssoCookieVendor` bootstrap.
+    func test_jsonDecoding_withCommunication() throws {
```
KatherineInCode: 🎨 Do we need tests around decoding, especially as we're not overriding the `Codable` methods?

fedemkr: At the time I thought it could be useful if we were to change something on the model so the decoding would catch that but thinking it again it's pretty useless for almost all real scenarios of change as other tests in that file; so I've just removed the whole file.

PR #2382, opened 2026-02-26
[PM-29876] feat: Implement server communication config SDK services

fedemkr:
## 🎟️ Tracking

[PM-29876](https://bitwarden.atlassian.net/browse/PM-29876)

## 📔 Objective

Implement the SDK-required services for server communication configuration, building on the models introduced in PM-32742. It wires up the full service layer needed for cookie bootstrapping: secure Keychain storage, SDK repository bridge, an API service implementing the SDK's `ServerCommunicationConfigPlatformApi`, and a singleton coordinator that reacts to config changes and drives the SDK client.

**SDK repository bridge**

SdkServerCommunicationConfigRepository — a new `ServerCommunicationConfigRepository` conformance that bridges SDK get/save calls through `StateService`; save merges the cookie value from the incoming config into the locally persisted config to avoid overwriting an existing cookie.

**SDK platform API service**

`DefaultServerCommunicationConfigAPIService` — implements `ServerCommunicationConfigPlatformApi` to handle config requests forwarded from the SDK.

**Singleton coordinator**

`DefaultServerCommunicationConfigClientSingleton` — an actor that lazily creates and caches the `ServerCommunicationConfigClientProtocol`, observing `configService.configPublisher()` to call `setCommunicationType` whenever the server config changes. Merges persisted local cookie value into each update.

[PM-29876]: https://bitwarden.atlassian.net/browse/PM-29876?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/Core/Platform/Services/TestHelpers/MockStateService.swift
+++ b/BitwardenShared/Core/Platform/Services/TestHelpers/MockStateService.swift
@@ -8,7 +8,7 @@ import Foundation
 @testable import BitwardenShared
 @testable import BitwardenSharedMocks
 
-class MockStateService: StateService, ActiveAccountStateProvider, ServerCommunicationConfigStateService { // swiftlint:disable:this type_body_length lin_length
+class MockStateService: StateService, ActiveAccountStateProvider, ServerCommunicationConfigStateService { // swiftlint:disable:this type_body_length line_length
     var accessTokenExpirationDateByUserId = [String: Date]()
     var accountEncryptionKeys = [String: AccountEncryptionKeys]()
     var accountSetupAutofill = [String: AccountSetupProgress]()


BitwardenShared/Core/Platform/Services/ServerCommunicationConfigClientSingleton.swift
```diff
@@ -0,0 +1,114 @@
...
+
+        Task {
+            for try await result in try await configService.configPublisher() {
+                guard let serverConfig = result?.serverConfig else {
+                    continue
+                }
+
+                await updateSDKCommunicationType(serverConfig)
+            }
+        }
```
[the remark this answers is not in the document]
fedemkr: 🤔 Perhaps we'd need to not throw on the `configPublisher` function as the subject is `CurrentValueSubject<MetaServerConfig?, Never>` so I think it won't ever throw. What do you think @matt-livefront @KatherineInCode ? Perhaps to do in a different PR as it involves other places.

AuthenticatorShared/Core/Platform/Services/StateService.swift
```diff
@@ -180,6 +180,10 @@ actor DefaultStateService:
 
     // MARK: Methods
 
+    func clearServerCommunicationCookieValue(hostname: String) async throws {
+        // no-op
```
KatherineInCode: 🤔 It might be helpful to more explicitly note that BWA doesn't have a server communication cookie value to clear.

BitwardenKit/Core/Platform/Services/ConfigServiceTests.swift
```diff
@@ -509,6 +509,16 @@ final class ConfigServiceTests: BitwardenTestCase { // swiftlint:disable:this ty
...
     }
 
+    // MARK: Tests - Server communication cookie
+
+    /// `clearServerCommunicationCookieValue(hostname:)` delegates to the state service with the correct hostname.
+    func test_clearServerCommunicationCookieValue() async throws {
+        try await subject.clearServerCommunicationCookieValue(hostname: "example.com")
+
+        XCTAssertTrue(stateService.clearServerCommCookieValueCalled)
+        XCTAssertEqual(stateService.clearServerCommCookieValueHostname, "example.com")
```
KatherineInCode, line 518: 🤔 Is it necessary to both check if it's called and that it's called with the correct hostname? I'd think the latter to be sufficient to also confirm that it was called.

BitwardenShared/Core/Platform/Services/TestHelpers/MockServerCommunicationConfigAPIService.swift
```diff
@@ -0,0 +1,24 @@
+import BitwardenSdk
+import Combine
+
+@testable import BitwardenShared
+
+final class MockServerCommunicationConfigAPIService: ServerCommunicationConfigAPIService {
```
KatherineInCode, line 6: 🤔 Is there a reason we can't mark the protocol as AutoMockable?

BitwardenShared/Core/Platform/Services/TestHelpers/MockServerCommunicationConfigClient.swift
```diff
@@ -0,0 +1,56 @@
+import BitwardenSdk
+import TestHelpers
+
+@testable import BitwardenShared
+
+final class MockServerCommunicationConfigClient: ServerCommunicationConfigClientProtocol {
```
KatherineInCode, line 6: 🤔 Is there a reason we can't mark the protocol as Automockable?

BitwardenShared/Core/Platform/Services/ServerCommunicationConfigClientSingleton.swift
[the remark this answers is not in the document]
KatherineInCode: I think we're okay; I'm pretty sure we do `for try await result in try await [publisher]` in other places as well 🤔

BitwardenShared/Core/Platform/Services/ServerCommunicationConfigClientSingleton.swift
```diff
@@ -0,0 +1,116 @@
+import BitwardenKit
+import BitwardenSdk
+
+/// A singleton for `ServerCommunicationConfigClientProtocol`. Needed to break circular dependency.
```
KatherineInCode: 🤔 It might be helpful to explicitly denote the circular dependency in the comment, which could help expose some of the architecture to a future maintainer.

That said, if there's a circular dependency, I'm curious why we need a singleton object instead of just a `weak` or `unowned` reference.

BitwardenShared/Core/Platform/Services/StateService.swift
```diff
@@ -299,6 +299,13 @@ protocol StateService: AnyObject {
     ///
     func getReviewPromptData() async -> ReviewPromptData?
 
+    /// Gets the server communication configuration for a given hostname.
```
KatherineInCode: 🤔 Does it make sense to add these directly to the `StateService` and expose them to the entire app, or would it be better to segregate it out into a more-scoped  `ServerCommunicationStateService` protocol?

(Similar could be asked about the updates to `KeychainRepository`)

KatherineInCode (commented):
On the whole, looks reasonable. I just have a few comments about documentation, and then wondering if we could more firmly apply the principle of Interface Segregation.

BitwardenShared/Core/Platform/Services/TestHelpers/MockServerCommunicationConfigAPIService.swift
fedemkr: Maybe I missed something but last time I checked it wasn't doing the publisher mock well.

BitwardenShared/Core/Platform/Services/ServerCommunicationConfigClientSingleton.swift
fedemkr: I'll update the docs and the singleton is to be able to create the client lazily when needed and handle the config subscription in this place where it's specifically related to the server communication config client.

BitwardenShared/Core/Platform/Services/TestHelpers/MockServerCommunicationConfigClient.swift
fedemkr: As far as I recall `Automockable` can't be applied to SDK protocols as it won't generate it based on applying the mark to an extension of the protocol.

BitwardenShared/Core/Platform/Services/ServerCommunicationConfigClientSingleton.swift
```diff
@@ -0,0 +1,145 @@
...
+    /// - Parameters:
+    ///   - clientService: The service that handles common client functionality such as encryption and decryption.
+    ///   - configService: The service to get server-specified configuration.
+    ///   - environmentService: The service used by the application to manage the environment settings.
+    ///   - errorReporter: The service used by the application to report non-fatal errors.
+    ///   - sdkRepositoryFactory: The factory to create SDK repositories.
+    ///   - serverCommunicationConfigAPIService: The service that bridges server communication
+    ///   configuration requests from the SDK.
+    ///   - stateService: The service that provides state management functionality for the
+    ///   server communication configuration.
```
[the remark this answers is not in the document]
KatherineInCode: I do think it would be good to get this updated now, rather than later

BitwardenShared/Core/Auth/Services/KeychainRepository+ServerCommunicationConfigTests.swift
```diff
@@ -0,0 +1,133 @@
...
+        let configString = String(data: configData, encoding: .utf8)!
+        keychainService.setSearchResultData(string: configString)
+
+        let result = try await subject.getServerCommunicationConfig(hostname: "example.com")
+
+        XCTAssertNotNil(result)
+        guard case .direct = result?.bootstrap else {
+            XCTFail("Expected .direct bootstrap, got \(String(describing: result?.bootstrap))")
+            return
+        }
```
matt-livefront: 🤔 Could you instead use `XCTAssertEqual` here? It looks like `ServerCommunicationConfig` conforms to `Equatable`.

BitwardenShared/Core/Auth/Services/KeychainRepository.swift
```diff
@@ -74,6 +78,7 @@ enum KeychainItem: Equatable, KeychainStorageKeyPossessing {
              .lastActiveTime,
              .neverLock,
              .pendingAdminLoginRequest,
+             .serverCommunicationConfig,
```
matt-livefront: 🤔 Will this keychain value need to be accessed while the app is in the background and the device is locked? If so, `kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly` might be better, similar to the access token.

BitwardenShared/Core/Platform/Services/ServerCommunicationConfigClientSingleton.swift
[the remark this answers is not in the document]
matt-livefront: `configPublisher()` throws and returns an `AsyncThrowingPublisher` but there's nothing throwing in that function, so it seems like we could make it not throw, and use `AsyncPublisher` instead.

```
    public func configPublisher() async throws -> AsyncThrowingPublisher<AnyPublisher<MetaServerConfig?, Never>> {
        configSubject.eraseToAnyPublisher().values
    }
```

We could simplify to this:

```
func configPublisher() async -> AsyncPublisher<AnyPublisher<MetaServerConfig?, Never>>
```

BitwardenShared/Core/Auth/Services/KeychainRepository.swift
fedemkr: Nice catch Matt!

PR #2383, opened 2026-02-27
[PM-32845] bug: Fix trusted device not remembered for new SSO users 

andrebispo5:
## 🎟️ Tracking

<!-- Paste the link to the Jira or GitHub issue or otherwise describe / point to where this change is coming from. -->
https://bitwarden.atlassian.net/browse/PM-32845

## 📔 Objective

<!-- Describe what the purpose of this PR is, for example what bug you're fixing or new feature you're adding. -->
### Summary                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                    
  - Store accountKeys returned from /accounts/keys endpoint when creating new SSO users                                                                                                                                                             
  - Add SetAccountKeysResponseModel to capture the response instead of discarding it                                                                                                                                                                
  - Fixes bug where trusted device selection was not persisted across app restarts for TDE users                                                                                                                                                    
                                                                                                                                                                                                                                           
###  Root Cause                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                    
  The /accounts/keys endpoint returns account keys in the response, but iOS was using EmptyResponse and discarding this data. The accountKeys field was always set to nil when storing encryption keys, causing the device trust state to not persist.      
  Keychain getDeviceKey was also throwing exception trying to decode literal string value of the device key as json.                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                    
###  Solution                                                                                                      
                                                                                                                    
  1. Created SetAccountKeysResponseModel conforming to JSONResponse and AccountKeysResponseModelProtocol                                                                                                                                            
  2. Updated SetAccountKeysRequest to use the new response model                                                                                                                                                                                    
  3. Modified AccountAPIService.setAccountKeys() to return the response                                                                                                                                                                             
  4. Updated AuthRepository.createNewSsoUser() to store accountKeys from the response                                                                                                                                                               
                                                                                         
## 📸 Screenshots

<!-- Required for any UI changes; delete if not applicable. Use fixed width images for better display. -->


https://github.com/user-attachments/assets/b8041f1c-bb3d-41e5-ba12-5b53a1933c2d

### DIFF
--- a/BitwardenShared/Core/Auth/Services/KeychainRepositoryTests.swift
+++ b/BitwardenShared/Core/Auth/Services/KeychainRepositoryTests.swift
@@ -144,6 +144,20 @@ final class KeychainRepositoryTests: BitwardenTestCase { // swiftlint:disable:th
         )
     }
 
+    /// `deletePendingAdminLoginRequest` succeeds quietly.
+    ///
+    func test_deletePendingAdminLoginRequest_success() async throws {
+        let item = KeychainItem.pendingAdminLoginRequest(userId: "1")
+        keychainService.deleteResult = .success(())
+        let expectedQuery = await subject.keychainQueryValues(for: item)
+
+        try await subject.deletePendingAdminLoginRequest(userId: "1")
+        XCTAssertEqual(
+            keychainService.deleteQueries,
+            [expectedQuery],
+        )
+    }
+
     /// The service should generate a storage key for a` KeychainItem`.
     ///
     func test_formattedKey_biometrics() async {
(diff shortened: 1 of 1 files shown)


BitwardenShared/Core/Auth/Models/Response/SetAccountKeysResponseModelTests.swift
```diff
@@ -0,0 +1,76 @@
+import TestHelpers
+import XCTest
+
+@testable import BitwardenShared
+@testable import BitwardenSharedMocks
+
+// MARK: - SetAccountKeysResponseModelTests
+
+class SetAccountKeysResponseModelTests: BitwardenTestCase {
```
fedemkr: 🎨 I think you can remove the whole test class as it doesn't have any custom coding/decoding logic so it should work out of the box and there's no need to test that it conforms to the protocol.

BitwardenShared/Core/Auth/Services/KeychainRepository.swift
```diff
@@ -491,15 +491,25 @@ extension DefaultKeychainRepository {
     }
 
     func getDeviceKey(userId: String) async throws -> String? {
-        try await getValue(for: .deviceKey(userId: userId))
+        do {
+            let value: String = try await getValue(for: .deviceKey(userId: userId))
+            return value
+        } catch KeychainServiceError.keyNotFound {
+            return nil
+        }
```
fedemkr, line 499: 🎨  You can directly return here unless I'm missing something. Also, for the catch we're adding both of these to handle not found cases:
`catch KeychainServiceError.osStatusError(errSecItemNotFound), KeychainServiceError.keyNotFound`

BitwardenShared/Core/Auth/Services/KeychainRepository.swift
```diff
@@ -491,15 +491,25 @@ extension DefaultKeychainRepository {
...
     }
 
     func getPendingAdminLoginRequest(userId: String) async throws -> String? {
-        try await getValue(for: .pendingAdminLoginRequest(userId: userId))
+        do {
+            let value: String = try await getValue(for: .pendingAdminLoginRequest(userId: userId))
+            return value
+        } catch KeychainServiceError.keyNotFound {
+            return nil
+        }
```
fedemkr, line 512: 🎨  Same applies here as in `getDeviceKey`'s comment.

PR #2384, opened 2026-02-27
[PM-29877] feat: Handle acquire cookies and 302 redirection for SSO cookies vendor

fedemkr:
## 🎟️ Tracking

[PM-29877](https://bitwarden.atlassian.net/browse/PM-29877)
[PM-29878](https://bitwarden.atlassian.net/browse/PM-29878)

## 📔 Objective

Implements the acquire-cookies flow for the SSO Cookie Vendor bootstrap type. This covers the networking layer, HTTP handlers, hostname resolution, and deep link handling — everything except the UI.

- **HTTPService** — Added `shouldFollowRedirect` parameter to `end(_:)`. When enabled, a retryWith closure is passed to response handlers; when disabled (on retried calls), `nil` is passed to prevent infinite recursion.
- **NoRedirectSessionDelegate** — New URLSessionTaskDelegate that surfaces 302 responses to the handler chain instead of letting URLSession auto-follow them. Only 302s are blocked; 301, 303, etc. follow normally.
- **SSOCookieVendorRequestHandler** — New RequestHandler that injects stored SSO cookies as Cookie headers into outgoing requests when the server's bootstrap config is `.ssoCookieVendor.
- **SSOCookieVendorResponseHandler** — New ResponseHandler that intercepts 302 responses. If the host needs a cookie bootstrap, it calls acquireCookie on the SDK client and throws a "try again" ServerError. Otherwise, it manually follows the redirect via retryWith.
- **AppProcessor** — Handles `bitwarden://sso-cookie-vendor` deep links, parses cookies from query parameters (excluding the d param), and delivers them to `ServerCommunicationConfigAPIService.cookiesAcquired`.

[PM-29877]: https://bitwarden.atlassian.net/browse/PM-29877?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[PM-29878]: https://bitwarden.atlassian.net/browse/PM-29878?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenKit/Core/Platform/Services/API/Handlers/ResponseValidationHandler.swift
```diff
@@ -36,7 +36,11 @@ public final class ResponseValidationHandler: ResponseHandler {
     /// - Parameter response: The `HTTPResponse` that was received by the `HTTPClient`.
     /// - Returns: The original or modified `HTTPResponse`.
     ///
-    public func handle(_ response: inout HTTPResponse) async throws -> HTTPResponse {
+    public func handle(
+        _ response: inout HTTPResponse,
+        for request: HTTPRequest,
+        retryWith: ((HTTPRequest) async throws -> HTTPResponse)?,
```
KatherineInCode, line 36: 🤔 It doesn't look like we do anything with the `retryWith` parameter yet. Is this just setting up for a future PR? Should there be a TODO, then? Or am I just missing where it's used?

⛏️ The new parameter should also be documented.

BitwardenShared/Core/Platform/Extensions/BitwardenSdk+Platform.swift
```diff
@@ -81,6 +81,14 @@ extension BitwardenSdk.ServerCommunicationConfig: @retroactive Codable {
         case bootstrap
     }
 
+    /// Returns the config for SSO Cookie vendor, `nil` if ``bootstrap`` is not the expected type.
```
KatherineInCode, line 84: 🎨 "Type" feels a little weird to me here, since it's an enum case. But I don't have a particularly good alternative, unfortunately.

Networking/Sources/Networking/HTTPService.swift
```diff
@@ -125,13 +125,18 @@ public final class HTTPService: Sendable {
...
+    ///     handler from triggering a further redirect. This prevents infinite recursion when a
+    ///     redirect-following handler re-invokes `send`.
     /// - Returns: The http response received for the request.
     ///
-    ///
     public func send(
         _ httpRequest: HTTPRequest,
         validate: ((HTTPResponse) throws -> Void)? = nil,
         shouldRetryIfUnauthorized: Bool = false,
+        shouldFollowRedirect: Bool = true,
```
KatherineInCode, line 137: 🎨 I think these parameters should be reorganized to be `_ httpRequest`, `shoulFollowRedirect`, `shouldRetryIfUnautherized`, `validate`. This would put optional properties after the non-optional property, alphabetize the non-closure optional properties, and put the closure at the end so we can take advantage of trailing closure syntax.

Networking/Sources/Networking/HTTPService.swift
```diff
@@ -185,22 +208,35 @@ public final class HTTPService: Sendable {
...
+        do {
+            for handler in requestHandlers {
+                httpRequest = try await handler.handle(&httpRequest)
+            }
+
+            if let tokenProvider {
+                try await httpRequest.headers["Authorization"] = "Bearer \(tokenProvider.getToken())"
+            }
+        } catch {
+            let asdfasfd = error
```
KatherineInCode: 🎨 Accidentally left in for debugging purposes?

BitwardenKit/Core/Platform/Services/API/Handlers/ResponseValidationHandler.swift
fedemkr: It's being used in the new `SSO...ResponseHandler` https://github.com/bitwarden/ios/pull/2384/changes#diff-932f9234662302411985dd724ee6b7c49d5e29e7a1d9e8945915e56340792cb3.
I'll update the docs.

BitwardenShared/Core/Platform/Extensions/BitwardenSdk+Platform.swift
fedemkr: That's how it's named [server side](https://github.com/bitwarden/server/blob/main/src/Api/Models/Response/ConfigResponseModel.cs#L106).

Networking/Sources/Networking/HTTPService.swift
fedemkr: 🤦 Nice catch! I'll remove that.

BitwardenKit/Core/Platform/Services/API/Handlers/ResponseValidationHandler.swift
fedemkr: Hmm not sure why we're adding the DocC in the `ResponseValidationHanlder->handle` function if that has the docs in the `ResponseHandler` protocol. I'll just remove the docs from here to align with other parts of the app.

fedemkr:
Close in favor of #2401 to handle `main` merge easier.

PR #2399, opened 2026-03-03
[PM-32221] chore: Add appcontext to crashlytics

LRNcardozoWDF:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-32221

## 📔 Objective

Add new value to logs to crashlytics in order to better identify the source of the error, `main app` vs `extension` .


BitwardenShared/UI/Platform/Application/AppContext.swift
```diff
@@ -9,3 +9,15 @@ public enum AppContext: Equatable {
...
+    /// A safe string representation of the current app context.
+    var appContextName: String {
+        switch self {
+        case .appExtension:
+            "App Extension"
+        case .mainApp:
+            "Main App"
+        }
+    }
+}
```
fedemkr, line 22: 🎨 You can use `CustomStringConvertible` protocol conformance instead which is the standard way to add a human description to enums. So if you use `String(describing: value)` or interpolation `"\(value)"` it will use the `description` implementation of the protocol automatically.
⛏️ No need to add "safe" as there's no special treatment to get the `string`.

PR #2401, opened 2026-03-03
[PM-29877] feat: Handle acquire cookies and 302 redirection for SSO cookies vendor flow

fedemkr:
## 🎟️ Tracking

[PM-29877](https://bitwarden.atlassian.net/browse/PM-29877)
[PM-29878](https://bitwarden.atlassian.net/browse/PM-29878)

## 📔 Objective

Implements the acquire-cookies flow for the SSO Cookie Vendor bootstrap type. This covers the networking layer, HTTP handlers, hostname resolution, and deep link handling — everything except the UI.

- **HTTPService** — Added `shouldFollowRedirect` parameter to `end(_:)`. When enabled, a retryWith closure is passed to response handlers; when disabled (on retried calls), `nil` is passed to prevent infinite recursion.
- **NoRedirectSessionDelegate** — New URLSessionTaskDelegate that surfaces 302 responses to the handler chain instead of letting URLSession auto-follow them. Only 302s are blocked; 301, 303, etc. follow normally.
- **SSOCookieVendorRequestHandler** — New RequestHandler that injects stored SSO cookies as Cookie headers into outgoing requests when the server's bootstrap config is `.ssoCookieVendor.
- **SSOCookieVendorResponseHandler** — New ResponseHandler that intercepts 302 responses. If the host needs a cookie bootstrap, it calls acquireCookie on the SDK client and throws a "try again" ServerError. Otherwise, it manually follows the redirect via retryWith.
- **AppProcessor** — Handles `bitwarden://sso-cookie-vendor` deep links, parses cookies from query parameters (excluding the d param), and delivers them to `ServerCommunicationConfigAPIService.cookiesAcquired`.

[PM-29877]: https://bitwarden.atlassian.net/browse/PM-29877?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[PM-29878]: https://bitwarden.atlassian.net/browse/PM-29878?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/Core/Platform/Services/ServerCommunicationConfigClientSingletonTests.swift
+++ b/BitwardenShared/Core/Platform/Services/ServerCommunicationConfigClientSingletonTests.swift
@@ -170,7 +170,7 @@ class ServerCommunicationConfigClientSingletonTests: BitwardenTestCase {
                 SsoCookieVendorConfig(
                     idpLoginUrl: "https://idp.example.com",
                     cookieName: "sso_cookie",
-                    cookieDomain: ".example.com",
+                    cookieDomain: "example.com",
                     cookieValue: cookieValue,
                 ),
             ),
@@ -188,7 +188,7 @@ class ServerCommunicationConfigClientSingletonTests: BitwardenTestCase {
                     SsoCookieVendorConfig(
                         idpLoginUrl: "https://idp.example.com",
                         cookieName: "sso_cookie",
-                        cookieDomain: ".example.com",
+                        cookieDomain: "example.com",
                         cookieValue: [AcquiredCookie(name: "cookie", value: "stored_value")],
                     ),
                 ),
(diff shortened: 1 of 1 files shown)


Networking/Sources/Networking/HTTPService.swift
```diff
@@ -185,22 +208,35 @@ public final class HTTPService: Sendable {
...
+            for handler in requestHandlers {
+                httpRequest = try await handler.handle(&httpRequest)
+            }
+
+            if let tokenProvider {
+                try await httpRequest.headers["Authorization"] = "Bearer \(tokenProvider.getToken())"
+            }
+        } catch {
+            let asdfasfd = error
+            throw error
```
matt-livefront: ⛏️ Is this and the do...catch block still needed?

fedemkr: Missed the 🍒 pick on that one, nice catch!

Issue #2030, opened 2025-10-10
[BWA-198] Authenticator -> Password Manager on iOS not working


aa6819:
### Steps To Reproduce

1. Hold the TOTP item in Authenticator
2. choose "copy to bitwarden Vault"

### Expected Result

until last releases, both PM and Authenticator [(PM ver 2025.9.1 (2605) and Auth ver 2025.9.1 (216)] this awesome function worked properly in iOS (Iphone), and it is super handy while i am thinking to subscribe soon for PM - .
But since last apps the copy function doesn’t work anymore. Either adding a new item in PM or adding the TOTP calculator to an existing item only get to create a new one or opening the exiting one in edit mode.

Thank you to check on this, most likely a bug introduced when adding new features.. 
I am thinking now that the bug may be in the Password Manager and not the Authenticator...

best,

alberto

### Actual Result

TOTP calculator codes are not copied into the PM vault.

### Build Version

2025.9.1 (216)

### What server are you connecting to?

EU

### Issue Tracking Info

- [x] I understand that work is tracked outside of Github. A PR will be linked to this issue should one be opened to address it, but Bitwarden doesn't use fields like "assigned", "milestone", or "project" to track progress.

pamperer562580892423:
Another user here. -- If I understand you correctly, both ways don't work for you, right now?!

So, to "copy" a code from the authenticator app into the password manager/vault doesn't work - and a new TOTP code from the password manager/vault doesn't show up in the authenticator app also, though the "sync" is activated?

If that's the case, you could add that other part to the "Actual Result", too.

aa6819:
hello pamperer.. 

sync is on and works properly, or did.. i can't test it now until i am able to first copy a code to PM..
for how i get it, the sync shows in the Authenticator the TOPT codes that are in the PM, and this works good still
the problem is when its created a new code in Authenticator and trying to pass it to the PM..
if the thread is wrong, or theres a better one (bug may be most likely in PM), please move it if you can.. or tell me how to do it,  this is my first post and totally a newbie :-)

pamperer562580892423:
@aa6819 Thanks for the clarification. It's all fine for me!

ghost:
I can confirm what alberto has written. I originally posted this on the community forum, but I missed the answer to post this on Github.

Just in case my original post is here

https://community.bitwarden.com/t/authenticator-password-manager-on-ios-not-working/89711?u=iammike

NovaSilentium:
Hi there,

[...]

PR #1930, opened 2025-09-12
[PM-24371] Fix cipher out-of-date error after adding attachment

matt-livefront:
## 🎟️ Tracking

[PM-24371](https://bitwarden.atlassian.net/browse/PM-24371)

## 📔 Objective

Adding or deleting an attachment for a cipher now updates the revision date on the cipher. If you add or delete an attachment while editing a cipher and then attempt to save the cipher, the app was showing a cipher out-of-date error because the edit view wasn't listening for cipher updates. This change adds a subscription for monitoring if the cipher changes to the AddEditItemProcessor to ensure the cipher in memory is up-to-date.

## 📸 Screenshots

| Before | After |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/a008a466-e650-447d-bb70-773ea853dddb" /> | <video src="https://github.com/user-attachments/assets/78cdf38e-c38a-43d7-ba94-a9f042a1da74" /> | 

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

[PM-24371]: https://bitwarden.atlassian.net/browse/PM-24371?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2410, opened 2026-03-05
[BWA-198] fix: Retain TOTP code when streaming updates on item edit in PM

KatherineInCode:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/BWA-198

## 📔 Objective

This fixes the issue laid out in [2030](https://github.com/bitwarden/ios/issues/2030), where when trying to copy a TOTP key into PM was opening up the edit screen without the TOTP copied over. This ultimately is because of a change made in #1930  where the edit screen streamed cipher details over again, which overwrote the imported TOTP key. This maintains the login information on the add/edit screen now, so that metadata such as update time can change, but the imported TOTP key will not.

### DIFF
--- a/AuthenticatorBridgeKit/AuthenticatorBridgeItemDataView.swift
+++ b/AuthenticatorBridgeKit/AuthenticatorBridgeItemDataView.swift
@@ -3,7 +3,7 @@ import Foundation
 /// A struct for storing **unencrypted** information about items that are shared between the Bitwarden
 /// and Authenticator apps.
 ///
-public struct AuthenticatorBridgeItemDataView: Codable, Equatable {
+public struct AuthenticatorBridgeItemDataView: Codable, Equatable, Sendable {
     // MARK: Properties
 
     /// The domain of the Bitwarden account that owns this item. (e.g. https://vault.bitwarden.com)
--- a/AuthenticatorBridgeKit/AuthenticatorBridgeItemService.swift
+++ b/AuthenticatorBridgeKit/AuthenticatorBridgeItemService.swift
@@ -144,10 +144,11 @@ public class DefaultAuthenticatorBridgeItemService: AuthenticatorBridgeItemServi
     /// - Parameter userId: the id of the user for which to fetch items.
     ///
     public func fetchAllForUserId(_ userId: String) async throws -> [AuthenticatorBridgeItemDataView] {
-        let fetchRequest = AuthenticatorBridgeItemData.fetchByUserIdRequest(userId: userId)
-        let result = try dataStore.backgroundContext.fetch(fetchRequest)
-        let encryptedItems = result.compactMap { data in
-            data.model
+        let context = dataStore.backgroundContext
+        let encryptedItems = try await context.perform {
+            let fetchRequest = AuthenticatorBridgeItemData.fetchByUserIdRequest(userId: userId)
+            let result = try context.fetch(fetchRequest)
+            return result.compactMap(\.model)
         }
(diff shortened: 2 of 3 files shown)


BitwardenShared/UI/Vault/VaultItem/AddEditItem/AddEditItemProcessor.swift
```diff
@@ -770,13 +770,19 @@ final class AddEditItemProcessor: StateProcessor<// swiftlint:disable:this type_
...
         guard let cipherId = state.cipher.id else { return }
         do {
             for try await cipherView in try await services.vaultRepository.cipherDetailsPublisher(id: cipherId) {
                 guard let cipherView else { continue }
-                state.update(from: cipherView)
+                state.update(
+                    from: cipherView,
+                    overrideLoginItemState: state.loginState,
+                )
             }
```
fedemkr, line 788: 🤔 I'm not quite sure if it's possible but wouldn't we have data "corruption" in the next edge case scenario:
- We have a login item with `username: test1` locally on PM.
- We import the item from BWA which adds the TOTP data. However, "at the same time" the item has been updated in another client to be `username: test2` and PM synced it.
- Therefore when streaming the details and overriding the login state, it will indeed preserve the BWA state but will omit the change of other login details. So if user saves this edit, `username` will go back to be `test1` which would be wrong IMO.
  - Moreover, the cipher's `revisionDate` will be ok as that is not replaced so server won't throw an error on this case of "missyncing".

So should we only care about overriding the TOTP data instead of the whole login state object? Should we check if the cipher's `revisionDate` changed from the item BWA has to the one is streamed here to prevent this situation?
I know it would be very rare to happen but raising it just in case.

KatherineInCode: That would cause a data race, yes, though I figured that would be rare enough (and fall under the purview of "don't edit the same object from two different places at the same time") that it wasn't worth handling as a special case.

The reason I did the whole login state compared to just TOTP data was because the API already had the `overrideLoginItemState`, and it seemed reasonable to keep that rather than create another version of `update` specifically for TOTP data.

Both of those are things I'm not super firm, on, so if we'd rather go the other way, I can build out a thing that only updates TOTP data and/or more rigorously checks to make sure it only updates if the revision date hasn't been updated elsewhere (though would both really be necessary? Probably not).

After considering it, I think just having it update TOTP data is the better way to go, though it maintains the potential data race problem, but I think if someone on their phone goes "yes this is the one" we should honor that.

fedemkr: I would go for safest and minimum data exposed to the race condition. If only TOTP data can be updated from BWA then I'd only replace that. We could ask Product for feedback as well on this to see if it's an accepted risk even though I also agree it'd very unlikely to happen.

KatherineInCode: Yeah, after thinking about it more, reducing the scope to just the TOTP code seems reasonable. I'll circle back on that.

ro-kue:
Thanks for merging this. Do you know when this will be available on the AppStore?

KatherineInCode:
@ro-kue It should get picked up in the next release or two, so I would anticipate sometime in April.

ro-kue:
> @ro-kue It should get picked up in the next release or two, so I would anticipate sometime in April.

@KatherineInCode Thanks a lot!

PR #2411, opened 2026-03-06
[PM-31470] bug: Show migrate personal vault on unlock

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-31470
https://bitwarden.atlassian.net/browse/PM-33143

## 📔 Objective

Fix bug where in some instances the app extension was not prompting the user for personal vault migration.
Changed the alert in the extension to look similar to the one in the main app and add a button to "**Continue on Bitwarden**".
Address the issue by checking on vault unlock instead of vault autofill appearing.

Tested scenarios:
- Authenticator CTA
- Authenticator Settings
- Share action Autofill
- Keyboard Autofill
- Send file
- Send text
- OTP QRCode read
- Shortcuts

## 📸 Screenshots

image

### DIFF
--- a/BitwardenShared/UI/Platform/Application/AppCoordinatorTests.swift
+++ b/BitwardenShared/UI/Platform/Application/AppCoordinatorTests.swift
@@ -651,6 +651,9 @@ class AppCoordinatorTests: BitwardenTestCase { // swiftlint:disable:this type_bo
     func test_navigateTo_migrateToMyItems() {
         rootNavigator.rootViewController = MockUIViewController()
 
+        // First authenticate to set up the childCoordinator (TabRoute).
+        subject.didCompleteAuth(rehydratableTarget: nil)
+
         let task = Task {
             subject.navigate(to: .migrateToMyItems(organizationId: "org-123"))
         }
@@ -675,6 +678,10 @@ class AppCoordinatorTests: BitwardenTestCase { // swiftlint:disable:this type_bo
             services: services,
         )
         rootNavigator.rootViewController = MockUIViewController()
+
+        // First authenticate to set up the childCoordinator (VaultRoute).
+        appExtensionDelegate.authCompletionRoute = .vault(.autofillList)
+        subject.didCompleteAuth(rehydratableTarget: nil)
 
         let task = Task {
             subject.navigate(to: .migrateToMyItems(organizationId: "org-123"))
--- a/BitwardenShared/UI/Platform/Application/AppProcessorTests.swift
+++ b/BitwardenShared/UI/Platform/Application/AppProcessorTests.swift
@@ -691,6 +691,19 @@ class AppProcessorTests: BitwardenTestCase { // swiftlint:disable:this type_body
         XCTAssertEqual(coordinator.events, [.setAuthCompletionRoute(.tab(.settings(.accountSecurity)))])
     }
 
+    /// `openUrl(_:)` handles receiving a bitwarden deep link and setting an auth completion route on the
+    /// coordinator if the user's vault is unlocked but a vault migration is required.
+    @MainActor
+    func test_openUrl_bitwardenAccountSecurity_vaultUnlockedMigrationRequired() async throws {
+        let account = Account.fixture()
+        stateService.activeAccount = .fixture()
+        vaultTimeoutService.isClientLocked[account.profile.userId] = false
+        syncService.organizationIdRequiringVaultMigrationResult = .success("org-123")
+
+        await subject.openUrl(.bitwardenAccountSecurity)
+        XCTAssertEqual(coordinator.events, [.setAuthCompletionRoute(.tab(.settings(.accountSecurity)))])
+    }
+
     /// `openUrl(_:)` handles receiving a bitwarden Authenticator new item deep link with the vault unlocked and an
     /// invalid item is found. It shows a generic error alert and does not produce a route.
     @MainActor


BitwardenShared/Core/Vault/Services/SyncService.swift
```diff
@@ -77,6 +67,17 @@ protocol SyncService: AnyObject {
     ///
     /// - Returns: A bool indicating if the user needs a sync or not.
     func needsSync(for userId: String, onlyCheckLocalData: Bool) async throws -> Bool
+
+    /// Checks if the user needs to migrate their personal vault items to an organization.
+    ///
+    /// The user needs to migrate if:
+    /// - The feature flag is enabled
```
fedemkr, line 74: ⛏️ I wouldn't add the comment for the requirement of the feature flag so we don't have to remind clearing it when the FF is removed. However, if you want to keep it just add a reference to the FF by using double backticks so it references it, thus when we search for the FF this appears in the comment and we can delete it.
```swift
``FeatureFlag.migrateMyVaultToMyItems``
```

BitwardenShared/Core/Vault/Services/SyncService.swift
```diff
@@ -275,6 +266,40 @@ class DefaultSyncService: SyncService {
...
+        else {
+            return nil
+        }
+
+        guard try await cipherService.hasPersonalCiphers() else {
+            return nil
+        }
+
+        return organizationId
+    }
```
fedemkr: ⛏️ No DocC comments are necessary as they're already defined in the protocol.
⛏️ Reorder in the file as this is not a private function.
🎨 Add tests for this function.

BitwardenShared/UI/Platform/Application/AppCoordinator.swift
```diff
@@ -393,13 +394,21 @@ extension AppCoordinator: AuthCoordinatorDelegate {
...
-            guard let route = appExtensionDelegate.authCompletionRoute else { return }
-            navigate(to: route)
+            // Check for pending vault migration after setting up the vault coordinator.
+            Task {
+                if let organizationId = try? await services.syncService.organizationIdRequiringVaultMigration() {
+                    await MainActor.run {
+                        navigate(to: .migrateToMyItems(organizationId: organizationId, isExtension: true))
+                    }
+                }
+            }
```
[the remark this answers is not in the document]
fedemkr: Agree

BitwardenShared/UI/Platform/Application/AppCoordinator.swift
```diff
@@ -419,6 +428,15 @@ extension AppCoordinator: AuthCoordinatorDelegate {
...
             }
+
+            // Check for pending vault migration after unlock.
+            Task {
+                if let organizationId = try? await services.syncService.organizationIdRequiringVaultMigration() {
+                    await MainActor.run {
+                        navigate(to: .migrateToMyItems(organizationId: organizationId))
+                    }
+                }
+            }
```
fedemkr: 🤔 Could this be done after the switch/case as it's the same for the extension and the main app cases but changing the `isExtension` flag? So you can add this logic once and use `isExtension: appContext == .appExtension` for the parameter.

BitwardenShared/UI/Vault/VaultItem/VaultItemCoordinator.swift
```diff
@@ -400,17 +401,24 @@ class VaultItemCoordinator: NSObject, Coordinator, HasStackNavigator { // swiftl
...
             coordinator: asAnyCoordinator(),
             delegate: delegate,
             services: services,
-            state: MigrateToMyItemsState(organizationId: organizationId),
+            state: MigrateToMyItemsState(
+                isExtension: isExtension,
+                organizationId: organizationId,
+                page: isExtension ? .extensionPrompt : .transfer,
+            ),
         )
```
fedemkr, line 421: 🤔 Instead of passing around the `isExtension` parameter through the navigation, I think it would be easier to inject `AppContextHelper` in this coordinator and check it here:
```swift
page: appContextHelper.appContext == .appExtension ? .extensionPrompt : .transfer,
```

BitwardenShared/UI/Platform/Application/AppCoordinator.swift
```diff
@@ -287,11 +287,7 @@ class AppCoordinator: Coordinator, HasRootNavigator {
...
     ///
     private func showMigrateToMyItems(organizationId: String) {
-        // Make sure that the user is authenticated and not currently viewing the migrate to my items view.
-        // In the main app, childCoordinator is TabRoute. In extensions, it's VaultRoute.
-        guard childCoordinator is AnyCoordinator<TabRoute, Void>
-            || childCoordinator is AnyCoordinator<VaultRoute, AuthAction>
-        else { return }
+        // Make sure that the user is not currently viewing the migrate to my items view.
         let currentView = rootNavigator?.rootViewController?.topmostViewController()
         guard !(currentView is UIHostingController<MigrateToMyItemsView>) else { return }
```
fedemkr, line 299: 🤔 Not quite sure why is it necessary to remove the child coordinator checks.

andrebispo5: I removed them while trying to debug something and forgot to add them back in.

BitwardenShared/UI/Platform/Application/AppProcessor.swift
```diff
@@ -442,7 +442,8 @@ extension AppProcessor {
     private func checkIfLockedAndPerformNavigation(route: AppRoute) async {
         if let userId = try? await services.stateService.getActiveAccountId(),
            !services.vaultTimeoutService.isLocked(userId: userId),
-           await (try? services.vaultTimeoutService.hasPassedSessionTimeout(userId: userId)) == false {
+           await (try? services.vaultTimeoutService.hasPassedSessionTimeout(userId: userId)) == false,
+           await (try? services.syncService.organizationIdRequiringVaultMigration()) == nil {
```
andrebispo5, line 446: Added this guard to avoid app opening from external sources and showing dialogs on top of the migration one. Example is the camera scan of a TOTP QRCode.

fedemkr: 🤔  Hmmm I'm not sure if this is behaving as you want for all cases or how events are handled. Because in here if you are in a vault with "Never" session timeout action and it requires migration then you are telling this to go and set a completion route for auth which won't happen as there's nothing to unlock given the "Never" timeout. Or am I missing something?
Also how does this behave with CXP import?

PR #2004, opened 2025-09-29
[PM-26293] Remove duplicate strings

mKoonrad:
## 📔 Objective

I removed duplicated strings.

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

PR #2412, opened 2026-03-06
[PM-26293] tool: Add scripts to delete duplicate Localizable.strings entries

KatherineInCode:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-26293

## 📔 Objective

This is inspired by, and solves, the issue brought up in #2004, in addition to making it easier to avoid that same sort of issue in the future.

This introduces a modular python script that can be used to "fix" our Localizable.strings files; long-term, it may make sense to run this as a pre-commit hook as well. Currently, the one module introduced handles duplicate entries in the strings files; I intend in a future PR to include a module for unused entries, and then potentially a module for alphabetization, among other things. These all live under the `Scripts/fix-localizable-strings` directory.

As well, there is a `Scripts/fix-localizable-strings.sh` script that invokes the Python script against the various localizable files. As the modules expand, it can also handle doing those. This would be what the pre-commit hook would run.

Finally, there is a `Scripts/test-fix-localizable-strings.sh` that is an easy invocation of running the tests for the fix-localizable-strings Python script (which currently just tests the duplication work. In the future, we may consider adding the running of these to CI.

One last thing: one of the duplicated strings (`"SendUpdated"`) had two different strings for the key. I ended up choosing the one that matches what I could find [in Figma](https://www.figma.com/design/cKUiM5EqMChDk9bEuzffBZ/Dev-Ready--Bitwarden-iOS?node-id=20492-40893&t=SJDilsVag7pkFiY2-4).


Scripts/fix-localizable-strings.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+set -euo pipefail
```
matt-livefront, line 9: 🎨 Is it worth documenting at the top of the file here what this script does? I've found that to be helpful when looking at other scripts that we have.

Scripts/test-fix-localizable-strings.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+set -euo pipefail
```
matt-livefront, line 3: 🎨 Same thing here, worth documenting what this does?

Scripts/fix-localizable-strings.sh
KatherineInCode: Good call. Updated.

PR #2413, opened 2026-03-06
[PM-33252] fix: Fix update profile KDF values from User decryption options on sync

fedemkr:
## 🎟️ Tracking

[PM-33252](https://bitwarden.atlassian.net/browse/PM-33252)

## 📔 Objective

Fix update profile KDF values from User decryption options on sync so KDF upgrade checks are done with latest KDF values from the server. This was causing an alert to upgrade KDF to display in a loop as the iOS app was never updating the old KDF values when the server got updated from another client.

[PM-33252]: https://bitwarden.atlassian.net/browse/PM-33252?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/Core/Platform/Services/StateService.swift
+++ b/BitwardenShared/Core/Platform/Services/StateService.swift
@@ -1869,6 +1869,12 @@ actor DefaultStateService: StateService, ActiveAccountStateProvider, ConfigState
                 trustedDeviceOption: nil,
             )
         userDecryptionOptions.masterPasswordUnlock = masterPasswordUnlock
+
+        profile.kdfIterations = masterPasswordUnlock.kdf.iterations
+        profile.kdfType = masterPasswordUnlock.kdf.kdfType
+        profile.kdfMemory = masterPasswordUnlock.kdf.memory
+        profile.kdfParallelism = masterPasswordUnlock.kdf.parallelism
+
         profile.userDecryptionOptions = userDecryptionOptions
         state.accounts[userId]?.profile = profile
         appSettingsStore.state = state
--- a/BitwardenShared/Core/Auth/Services/ChangeKdfService.swift
+++ b/BitwardenShared/Core/Auth/Services/ChangeKdfService.swift
@@ -98,7 +98,7 @@ class DefaultChangeKdfService: ChangeKdfService {
 
             // If the account needs a KDF update, sync the user's account and check again before
             // proceeding to ensure the local data is up-to-date.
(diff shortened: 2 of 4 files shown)


BitwardenShared/Core/Platform/Services/StateService.swift
```diff
@@ -1869,6 +1869,12 @@ actor DefaultStateService: StateService, ActiveAccountStateProvider, ConfigState
                 trustedDeviceOption: nil,
             )
         userDecryptionOptions.masterPasswordUnlock = masterPasswordUnlock
+
+        profile.kdfIterations = masterPasswordUnlock.kdf.iterations
+        profile.kdfType = masterPasswordUnlock.kdf.kdfType
+        profile.kdfMemory = masterPasswordUnlock.kdf.memory
+        profile.kdfParallelism = masterPasswordUnlock.kdf.parallelism
```
matt-livefront, line 1876: 🤔 I see why this is needed in this case, but I thought one of the motivations for separating `masterPasswordUnlock` from the profile's KDF values was that they could diverge in the future, where you could have separate KDF settings for authentication vs master password unlock. If that was the case, would we still want to overwrite the profile's KDF here?

Alternatively, could we use `masterPasswordUnlock`'s KDF to detect if a KDF upgrade is needed, if it exists? Is one approach preferred over the other?

quexten: TBH, we should not prepare for that here, but do it deliberately when such a separation actually happens, on all clients (mobile and non-mobile).

From KM's perspective, there *could* be three values:
- The KDF used to authenticate as part of master-password authenticator
- The KDF used to decrypt the user-key
- The KDF shown in the UI as user-preference

All three are currently the same, on the database, and also in `clients` state. It is likely that KM would move these to SDK managed state before we ever split them at this point.

Though, I agree, using `masterPasswordUnlock` would work as well. It's just important that `masterPasswordUnlock` is overwritten after updating the KDF *and* on sync in that case.

quexten (approved):
From my perspective, this fix seems good.

BitwardenShared/Core/Platform/Services/StateService.swift
matt-livefront: @quexten Thanks, that makes sense to me.

PR #2427, opened 2026-03-10
[PM-31722] fix: Prevent WatchService error when switching to locked account

matt-livefront:
## 🎟️ Tracking

[PM-31722](https://bitwarden.atlassian.net/browse/PM-31722)

## 📔 Objective

This fixes an error that occurs if the app attempts to sync a user’s TOTP codes to the watch app and the vault is still locked. `WatchService` wasn't previously checking if the vault was locked prior to syncing so it would attempt to sync TOTP codes while the vault was locked, resulting in an error. This updates the behavior to only attempt a sync if the vault is unlocked.

> Error: Decrypt(BitwardenSdk.DecryptError.Crypto(message: "Missing Key for Id: User"))

https://console.firebase.google.com/u/1/project/bitwarden-c2b35/crashlytics/app/ios:com.8bit.bitwarden/issues/05a803bddaa321694058212b0fc7f482?time=7d&sessionEventKey=805b537d27ca49dfb6b25d98af010e83_2194072842563009389

[PM-31722]: https://bitwarden.atlassian.net/browse/PM-31722?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/Core/Platform/Services/WatchService.swift
+++ b/BitwardenShared/Core/Platform/Services/WatchService.swift
@@ -198,17 +198,35 @@ class DefaultWatchService: NSObject, WatchService {
         // and recreated each time.
         syncCiphersTask?.cancel()
         syncCiphersTask = Task {
-            let userId = try? await stateService.getActiveAccountId()
-            let connectToWatch = try? await stateService.getConnectToWatch()
-            let lastUserShouldConnect = await stateService.getLastUserShouldConnectToWatch()
+            let userId: String?
+            do {
+                userId = try await stateService.getActiveAccountId()
+            } catch StateServiceError.noActiveAccount {
+                userId = nil
+            } catch {
+                userId = nil
+                errorReporter.log(error: error)
+            }
+
+            let shouldConnect: Bool
+            do {
+                shouldConnect = try await stateService.getConnectToWatch()
+            } catch StateServiceError.noActiveAccount {
+                shouldConnect = await stateService.getLastUserShouldConnectToWatch()
+            } catch {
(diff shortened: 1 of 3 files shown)


BitwardenShared/Core/Platform/Services/WatchService.swift
```diff
@@ -179,6 +190,30 @@ class DefaultWatchService: NSObject, WatchService {
...
+    /// Fetch the current account state and sync with the watch.
+    ///
+    private func syncWithWatch() {
+        // This method will be called whenever the connect value, vault lock state, or account
+        // value changes, so the task listening to the cipher updates will need to be cancelled
+        // and recreated each time.
+        syncCiphersTask?.cancel()
+        syncCiphersTask = Task {
+            let userId = try? await stateService.getActiveAccountId()
+            let connectToWatch = try? await stateService.getConnectToWatch()
```
fedemkr: 🤔 Should we log the error and assign `nil` to these if an error is thrown? So we have tracking of the service error.

fedemkr (approved):
Looks good Matt, nice catch!
Just a small question, but I'm good as it's now if you feel it's necessary to log there.

BitwardenShared/Core/Platform/Services/WatchService.swift
matt-livefront: Yep, good idea. I've added error logging for these.

PR #2428, opened 2026-03-10
[PM-33498] fix: Create a safePresent method to avoid VC presentation races

KatherineInCode:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33498

## 📔 Objective

This adjusts our `UIViewController.present()` methods to call an underlying `safePresent()` method. This "safe present" first checks to see if the view controller we want to present on is already presenting, and if that presented view controller is currently in the middle of dismissing, and if so, it waits a bit and then re-checks. This should avoid situations where we try to present right after dismissing something, and it not presenting because the currently-dismissing view blocks it.

This came up as part of PM-33311 while trying to fix some issues with how the `VaultItemSelection*` classes handle presentation and dismissal, particularly around the iOS 26 profile switcher added in #1906. There was an issue with the following flow:

- User is in account A in PM
- User goes to BWA, selects an item, and selects "copy to Bitwarden"
- Once the `VaultItemSelectionView` appears, they tap on the profile switcher, and switch profiles
- This was making the entire `VaultItemSelectionView` disappear and not reappear, because trying to re-present it was failing because the old one still wasn't fully dismissed.

This thus solves that problem by making sure presenting the new `VaultItemSelectionView` with the new profile happens only after the previous one dismissed.

This is potentially far-reaching in the app because it's changing how all `present()` calls work. That *does* make me a little nervous. However, I think it will ultimately be a good change that may avoid some additional issues down the line if we have similar sorts of flows when changing profiles.

### DIFF
--- a/BitwardenKit/UI/Platform/Application/Extensions/UIViewControllerTests.swift
+++ b/BitwardenKit/UI/Platform/Application/Extensions/UIViewControllerTests.swift
@@ -1,3 +1,4 @@
+import BitwardenKitMocks
 import XCTest
 
 @testable import BitwardenKit
@@ -7,7 +8,7 @@ class UIViewControllerTests: BitwardenTestCase {
 
     /// `safePresent` presents the view controller immediately when nothing is currently presented.
     func test_safePresent_presentsWhenNoPresentedViewController() {
-        let subject = TrackingViewController()
+        let subject = MockUIViewController()
         let viewController = UIViewController()
         var completionCalled = false
 
@@ -20,8 +21,8 @@ class UIViewControllerTests: BitwardenTestCase {
 
     /// `safePresent` defers presentation while a dismiss is in progress and retries until it succeeds.
     func test_safePresent_retriesAndPresentsAfterDismissCompletes() {
-        let subject = TrackingViewController()
-        subject.mockPresentedViewController = BeingDismissedViewController()
+        let subject = MockUIViewController()
+        subject.presentedView = BeingDismissedViewController()
         let viewController = UIViewController()
         var completionCalled = false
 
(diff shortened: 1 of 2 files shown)


KatherineInCode:
Unsurprisingly, I suppose, Claude's local review suggested fixes for both of the caveats I mentioned. I'll add those before moving this out of draft.

BitwardenKit/UI/Platform/Application/Extensions/UIViewController.swift
```diff
@@ -36,13 +37,60 @@ public extension UIViewController {
...
+    ///   - viewController: The view controller to present.
+    ///   - animated: Whether the transition should be animated.
+    ///   - remainingAttempts: The number of retry attempts remaining before giving up. Defaults to `10`.
+    ///   - completion: A closure to call on completion. Note: if the retry limit is exceeded, the
+    ///     presentation is dropped and this closure will not be called.
+    ///
+    internal func safePresent(
+        _ viewController: UIViewController,
+        animated: Bool,
+        remainingAttempts: Int = 10,
```
fedemkr: 🤔 I think that waiting 3 seconds (remainingAttempts * deadline Δ) may be a lot as a default for the maximum. I feel like it should fail and throw the warning at 1 second or 1,5 seconds for a navigation and use more than that on very specific cases.
What do you think? Was your case taking this long and that's the reason behind the 10 attempts?

KatherineInCode: In the simulator, it hits the retry 2 times when switching profiles, so ~0.6 seconds later. The 10 was just trying to give adequate time for things to potentially dismiss, and to have a circuit breaker; I'm not wedded to it at all, and can just reduce it down to 5. 1.5 seconds seems like a good amount to handle slower phones, as I would rather err on the side of "showing it, if delayed" if possible

PR #2429, opened 2026-03-10
[PM-33231] fix: Fix keyboard being shown on sync with browser from login screen

fedemkr:
## 🎟️ Tracking

[PM-33231](https://bitwarden.atlassian.net/browse/PM-33231)

## 📔 Objective

Fix keyboard being shown on sync with browser from login screen request.
The proposed fix here is just wait a bit to dismiss the keyboard so it has time to be shown first.

[PM-33231]: https://bitwarden.atlassian.net/browse/PM-33231?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/UI/Platform/Application/Extensions/View.swift
+++ b/BitwardenShared/UI/Platform/Application/Extensions/View.swift
@@ -175,6 +175,8 @@ private struct KeyboardDismissOnAppearView: UIViewRepresentable {
     func makeUIView(context: Context) -> UIView {
         let view = UIView(frame: .zero)
         view.backgroundColor = .clear
+
+        /// Wait a bit so the animation ends and also until the keyboard is actually shown to be dismissed.
         DispatchQueue.main.asyncAfter(deadline: .now() + 0.4) {
             view.window?.endEditing(true)
         }


BitwardenShared/UI/Platform/Application/Extensions/View.swift
```diff
@@ -175,8 +175,12 @@ private struct KeyboardDismissOnAppearView: UIViewRepresentable {
     func makeUIView(context: Context) -> UIView {
         let view = UIView(frame: .zero)
         view.backgroundColor = .clear
-        DispatchQueue.main.async {
-            view.window?.endEditing(true)
+        Task {
```
KatherineInCode: 🤔 Why a Task that then runs back on `MainActor` instead of `DispatchQueue.main.asyncAfter`?

fedemkr: Always forget that exists, I'll update it thanks Katherine!

BitwardenShared/UI/Platform/Application/Extensions/View.swift
```diff
@@ -175,7 +175,9 @@ private struct KeyboardDismissOnAppearView: UIViewRepresentable {
     func makeUIView(context: Context) -> UIView {
         let view = UIView(frame: .zero)
         view.backgroundColor = .clear
-        DispatchQueue.main.async {
+
+        /// Wait a bit so the animation ends and also until the keyboard is actually shown to be dismissed.
+        DispatchQueue.main.asyncAfter(deadline: .now() + 0.4) {
```
KatherineInCode, line 180: 🤔 I find myself wondering if the `0.4` can be pulled out into a global constant of some sort, so we can have some sort of consistency with "how long we wait for dismissal" (particularly considering also my #2428 that would potentially use it also. I don't see it as blocking this PR, but maybe something we should consider in the future.

fedemkr: Yes, makes sense. If we find the same wait to be added in other places I feel it would be good to extract it into a constant. Just in case in here it's a bit more than VC presentation/dismissal as we need to wait for the keyboard being shown so I wouldn't reuse it. In your case it's 0.3, and IMO if we can wait less time we should do so as it's a better UX. Therefore for now I'd incline to leave it as is given that this is in an extension function of `View`.

PR #2435, opened 2026-03-12
[PM-33136] fix: Centralize TOTP key error handling to reduce Crashlytics noise

matt-livefront:
## 🎟️ Tracking

[PM-33136](https://bitwarden.atlassian.net/browse/PM-33136)

## 📔 Objective

The SDK will throw user-correctable errors (`MissingSecret`, `InvalidOtpauth`) when generating TOTP codes for ciphers with invalid or incomplete TOTP keys. These were previously being logged to Crashlytics, but since they aren't errors that we can fix, they just add noise.

Since TOTP codes are generated from many places across the app, the easiest way to prevent these user-correctable errors from being logged to Crashlytics is to separate user-correctable errors from other errors. `TOTPKeyError` was created for the user-correctable errors and `TOTPServiceError` remains for any other TOTP generation errors. `TOTPKeyError` is a `NonLoggableError` so it won't be logged to Crashlytics, but will still show up in flight recorder logs if it occurs. All of the SDK errors are intercepted in `VaultClientService` and transformed to these error types. This allows us to use a uniform `catch { errorReporter.log(error: error) }` pattern throughout the app without having to inspect at each TOTP generation callsite to determine if the error should be logged or not.

[PM-33136]: https://bitwarden.atlassian.net/browse/PM-33136?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/AuthenticatorShared/Core/Vault/Repositories/AuthenticatorItemRepositoryTests.swift
+++ b/AuthenticatorShared/Core/Vault/Repositories/AuthenticatorItemRepositoryTests.swift
@@ -1,5 +1,6 @@
 import AuthenticatorBridgeKit
 import AuthenticatorBridgeKitMocks
+import BitwardenKit
 import BitwardenKitMocks
 import BitwardenResources
 import InlineSnapshotTesting
--- a/AuthenticatorShared/Core/Vault/Services/TOTP/TOTPServiceTests.swift
+++ b/AuthenticatorShared/Core/Vault/Services/TOTP/TOTPServiceTests.swift
@@ -1,3 +1,4 @@
+import BitwardenKit
 import BitwardenKitMocks
 import XCTest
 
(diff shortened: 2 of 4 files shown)


BitwardenShared/Core/Vault/Services/TOTP/TOTPKeyError.swift
```diff
@@ -0,0 +1,15 @@
+import BitwardenKit
+
+/// Errors related to an invalid or missing TOTP key, representing user-correctable
+/// data issues. Conforms to `NonLoggableError` so these are not sent to Crashlytics.
+///
+enum TOTPKeyError: NonLoggableError, Equatable {
```
KatherineInCode: 🤔 Would it make sense to move this to `BitwardenKit`, since eventually it would be good to centralize our TOTP-handling logic there anyways?

matt-livefront: Yeah, I think moving this and `TOTPServiceError` would make sense.

PR #2436, opened 2026-03-12
[PM-33370] fix: Fix [un]archive flow on missing cipher key case

fedemkr:
## 🎟️ Tracking

[PM-33370](https://bitwarden.atlassian.net/browse/PM-33370)

## 📔 Objective

When a cipher was missing its `Key` and the user wanted to archive it, the cipher's archived date was being updated before the cipher mission key update to server, therefore the cipher was being updated to the server with the archive change which was causing the actual server call for archive to fail given that the cipher on server had already been changed with an archived date given the previous endpoint call for missing key update.

Additionally, flows for cipher key handling have been extracted to a new `CipherEncryptionMediator` file so it's easier to test and we reduce duplication on tests for the same behavior on different operations.

[PM-33370]: https://bitwarden.atlassian.net/browse/PM-33370?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/Core/Vault/Helpers/CipherEncryptionMediator.swift
+++ b/BitwardenShared/Core/Vault/Helpers/CipherEncryptionMediator.swift
@@ -11,6 +11,7 @@ protocol CipherEncryptionMediator { // sourcery: AutoMockable
     func encryptAndUpdateCipher(_ cipherView: CipherView) async throws -> Cipher
 
     /// Sets the delegate to use.
+    /// - Parameter delegate: The delegate to use.
     func setDelegate(_ delegate: CipherEncryptionMediatorDelegate)
 
     /// Checks if `cipherView` lacks a cipher key and if so tries to update it with the server


BitwardenShared/Core/Vault/Helpers/CipherEncryptionMediator.swift
```diff
@@ -0,0 +1,100 @@
...
+protocol CipherEncryptionMediator { // sourcery: AutoMockable
+    /// Encrypts the cipher. If the cipher was migrated by the SDK (e.g. added a cipher key), the
+    /// cipher will be updated locally and on the server.
+    ///
+    /// - Parameter cipherView: The cipher to encrypt.
+    /// - Returns: The encrypted cipher.
+    ///
+    func encryptAndUpdateCipher(_ cipherView: CipherView) async throws -> Cipher
+
+    /// Sets the delegate to use.
```
KatherineInCode, line 13: ⛏️ Nominally, this should also have a parameter list in the documentation comments

KatherineInCode (approved):

PR #2442, opened 2026-03-13
[PM-33355] feat: Sync when push notification policy changed is received

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33355

## 📔 Objective

This pull request introduces support for a new notification type, `policyChanged`, and ensures that the app responds appropriately by triggering a sync when such a notification is received. Additionally, it fixes a typo in a localization key and updates related code to use the corrected key. The changes also include a new test to verify the sync behavior for `policyChanged` notifications.

Support for new notification type:

* Added `policyChanged` as a new case in the `NotificationType` enum, allowing the app to recognize policy change notifications.
* Updated `DefaultNotificationService` to trigger a sync (`fetchSync`) when a `policyChanged` notification is received, ensuring the app stays up-to-date with policy changes.
* Added a unit test in `NotificationServiceTests.swift` to verify that receiving a `policyChanged` notification triggers a sync.

Localization and notification content fixes:

* Fixed a typo in the localization key by adding `ConfirmLogInAttemptForX` and updating code to use this corrected key instead of the misspelled one.

### DIFF
--- a/BitwardenShared/Core/Platform/Services/NotificationServiceTests.swift
+++ b/BitwardenShared/Core/Platform/Services/NotificationServiceTests.swift
@@ -412,7 +412,6 @@ class NotificationServiceTests: BitwardenTestCase { // swiftlint:disable:this ty
     /// `messageReceived(_:notificationDismissed:notificationTapped:)` triggers a sync when a
     /// policy changed notification is received.
     func test_messageReceived_policyChanged() async throws {
-        // Set up the mock data.
         stateService.setIsAuthenticated()
         appSettingsStore.appId = "10"
         let message: [AnyHashable: Any] = [
@@ -422,10 +421,8 @@ class NotificationServiceTests: BitwardenTestCase { // swiftlint:disable:this ty
             ],
         ]
 
-        // Test.
         await subject.messageReceived(message, notificationDismissed: nil, notificationTapped: nil)
 
-        // Confirm the results.
         XCTAssertTrue(syncService.didFetchSync)
     }
 


BitwardenShared/Core/Platform/Services/NotificationServiceTests.swift
```diff
@@ -409,6 +409,26 @@ class NotificationServiceTests: BitwardenTestCase { // swiftlint:disable:this ty
...
+                "type": NotificationType.policyChanged.rawValue,
+                "payload": "anything",
+            ],
+        ]
+
+        // Test.
+        await subject.messageReceived(message, notificationDismissed: nil, notificationTapped: nil)
+
+        // Confirm the results.
+        XCTAssertTrue(syncService.didFetchSync)
```
fedemkr: ⛏️ Remove comments as IMO they are not adding any value to the code.

BitwardenShared/Core/Platform/Services/NotificationService.swift
```diff
@@ -266,6 +266,8 @@ class DefaultNotificationService: NotificationService {
                 // No action necessary, since the LoginWithDeviceProcessor already checks for updates
                 // every few seconds.
                 break
+            case .policyChanged:
+                try await syncService.fetchSync(forceSync: false)
```
fedemkr, line 270: 🤔 Should this be forced? Not sure about the answer, perhaps something for Product on the urgency on having the policies applied as I'm not sure whether the `accountRevisionDate` is updated when a policy changes; if it's not then the sync may not be done depending on the last time the sync was performed.

andrebispo5: I tested it and the revision seems to be updated that is why I didn't asked for the forced here 👍

fedemkr: Great, thanks for checking it up Andre!

PR #2444, opened 2026-03-13
[PM-33569] feat: Add notification service extension to update auth request notifications

matt-livefront:
## 🎟️ Tracking

[PM-33569](https://bitwarden.atlassian.net/browse/PM-33569)

## 📔 Objective

We're switching from background push notifications to alert push notifications for "login with device" auth requests. This adds a UNNotificationServiceExtension that intercepts those alert notifications and updates the body with the requesting user's email before delivery.

- Adds a new BitwardenNotificationExtension app extension target
- Adds `DefaultNotificationExtensionHelper` in BitwardenShared which reads account state and updates the notification title/body for auth request notifications

The backend sends:
> Title: Login requested
> Message: Confirm login attempt

The notification extension updates this to:
> Title: Login request (localized based on app language)
> Message: Confirm login attempt for <email> (localized and email added based on the user's ID)

This handles updating the content of the push notification. There will be a separate PR to handle processing this notification in the app when it's tapped.

[PM-33569]: https://bitwarden.atlassian.net/browse/PM-33569?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenShared/Core/Platform/Services/NotificationExtensionHelper.swift
```diff
@@ -0,0 +1,121 @@
...
+    ///
+    /// - Parameter userInfo: The notification's `userInfo` dictionary.
+    /// - Returns: The decoded `PushNotificationData`, or `nil` if the `"data"` key is absent.
+    /// - Throws: An error if JSON serialization or decoding fails.
+    ///
+    private func decodePushNotificationData(from userInfo: [AnyHashable: Any]) throws -> PushNotificationData? {
+        guard let messageContent = userInfo["data"] as? [AnyHashable: Any] else { return nil }
+        let jsonData = try JSONSerialization.data(withJSONObject: messageContent)
+        return try JSONDecoder().decode(PushNotificationData.self, from: jsonData)
+    }
```
fedemkr: 🤔 Could this be extracted to a common place to be shared between this and `NotificationService`?

https://github.com/bitwarden/ios/blob/ac65c415e8ab5e9456c6a16598ac8c3aac782657/BitwardenShared/Core/Platform/Services/NotificationService.swift#L313-L317

BitwardenShared/Core/Platform/Services/NotificationExtensionHelper.swift
```diff
@@ -0,0 +1,121 @@
...
+
+// MARK: - DefaultNotificationExtensionHelper
+
+/// The default implementation of `NotificationExtensionHelper`.
+///
+public class DefaultNotificationExtensionHelper: NotificationExtensionHelper {
+    // MARK: Properties
+
+    /// The store used to read persisted app settings.
+    private let appSettingsStore: AppSettingsStore
```
fedemkr, line 30: 🤔 Shouldn't we use some service to access the store instead of direct access from here as we do in other parts?

BitwardenShared/Core/Platform/Services/NotificationExtensionHelper.swift
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    /// Initializes a `DefaultNotificationExtensionHelper` using the shared app group `UserDefaults`
+    /// and an `OSLogErrorReporter`.
+    ///
+    public convenience init() {
+        let userDefaults = UserDefaults(suiteName: Bundle.main.groupIdentifier)!
+        self.init(
+            appSettingsStore: DefaultAppSettingsStore(userDefaults: userDefaults),
+            errorReporter: OSLogErrorReporter(),
```
fedemkr, line 58: ⚠️ I think the `ErrorReporterFactory` should be used here instead of always using the `OSLogErrorReporter` for prod as well unless there's a problem with Crashlytics here.

fedemkr (commented):
Looks good, just some minor comments.

BitwardenShared/Core/Platform/Services/NotificationExtensionHelper.swift
matt-livefront: I was trying to keep this helper as lightweight as possible and only initialize what was needed. But maybe that's being too conservative. If we introduce `StateService`, do you think I should initialize `StateService` directly or use the full `ServiceContainer`?

BitwardenShared/Core/Platform/Services/NotificationExtensionHelper.swift
matt-livefront: `ErrorReporterFactory` is only available in the BItwarden target. I will test this again, but I thought we had issues using Crashlytics in the extensions. Maybe something we can spin off into a separate ticket?

fedemkr: Yes, you're right sorry; it's ok like this.

BitwardenShared/Core/Platform/Services/NotificationExtensionHelper.swift
fedemkr: Good point, I think we can leave it as is for now as it's pretty simple. Then if we need more access to other properties we could introduce `StateService` or some other lighter service in between.

BitwardenShared/Core/Platform/Services/NotificationExtensionHelper.swift
matt-livefront: Good idea, I moved this into a `PushNotificationData` initializer.

fedemkr (dismissed):

vvolkgang:
@matt-livefront Triggered runs to test the new provisioning profiles:

1. BWPM - https://github.com/bitwarden/ios/actions/runs/23442727040
2. BWA (double checking everything is ok) - https://github.com/bitwarden/ios/actions/runs/23442739178

edit: both passed!

KatherineInCode (approved):

PR #2450, opened 2026-03-13
[PM-33576] llm: Add orchestration command, agent, and skill index to CLAUDE.md

SaintPatrck:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33576

## 📔 Objective

Adds the final orchestration layer: a `/work-on-ios` command that sequences all skills end-to-end, an autonomous agent for hands-off execution, and a skill/command index in CLAUDE.md.

**What changed:**

- **`commands/work-on-ios.md`** — 7-phase guided workflow with user confirmation between each phase: implement → test → build & verify → self-review → commit → code review → create draft PR. Invokes the appropriate skill at each phase. Structured to match the Android `work-on-android.md` template.

- **`commands/plan-ios-work.md`** — 3-phase planning command: ingest requirements (Jira or description) → refine with gap analysis → design implementation approach. Plans are presented in-context rather than saved to disk.

- **`agents/ios-implementer/AGENT.md`** — Opus model, blue. Runs `/work-on-ios` as primary workflow with a self-review protocol at each phase transition, auto-approving when criteria are met. Includes decision-making framework and explicit skill tool declarations. Aligned with the Android `android-implementer` agent structure.

- **CLAUDE.md** — Adds a Skills & Commands index table: 10 skills with trigger phrases, and 2 commands with usage descriptions.

This is the final commit in the 7-PR series (PM-33570 through PM-33576) that brings the iOS `.claude/` configuration to full parity with the Android template.

### DIFF
--- a/.claude/commands/work-on-ios.md
+++ b/.claude/commands/work-on-ios.md
@@ -1,77 +1,66 @@
 ---
-description: Execute a complete Bitwarden iOS implementation workflow from ticket to PR. Loads or creates a plan, implements, tests, builds, runs preflight, commits, and opens a PR with labels.
-argument-hint: <PM-XXXXX ticket ID or task description>
+description: Guided iOS development workflow through all lifecycle phases
+argument-hint: <task description, plan, or Jira ticket reference>
 ---
 
-# Work on iOS: $ARGUMENTS
+# iOS Development Workflow
 
-Full implementation workflow with user confirmation at each phase boundary.
+You are guiding the developer through a complete iOS development lifecycle for the Bitwarden iOS project. The task to work on is:
 
-## Phase 1: Load or Create Plan
+**Task**: $ARGUMENTS
 
-Check for an existing plan at `.claude/outputs/plans/`:
-- **If found**: Load it and present a summary to the user
-- **If not found**: Invoke `/plan-ios-work $ARGUMENTS` to create one
+## Workflow Phases
 
-**Confirm**: "Plan loaded/created. Ready to begin implementation? (y/n)"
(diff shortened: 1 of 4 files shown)


.claude/commands/work-on-ios.md
```diff
@@ -0,0 +1,77 @@
...
+argument-hint: <PM-XXXXX ticket ID or task description>
+---
+
+# Work on iOS: $ARGUMENTS
+
+Full implementation workflow with user confirmation at each phase boundary.
+
+## Phase 1: Load or Create Plan
+
+Check for an existing plan at `.claude/outputs/plans/`:
```
KatherineInCode: 🤔 We don't have a `.gitignore` of some sort set up against this directory, which means, are we anticipating now that we'll be committing Claude's generated plans into the repo?

SaintPatrck: No. This shouldn't reference any specific file locations yet. However, I do suggest making a global `~/.gitignore_global` entry to a common place for Claude to safely drop its outputs. As you can see, I use `.claude/outputs/`. 

Once these tools are converted to plugins, they'll have their own output locations outside of any repo/project context. 🚀

SaintPatrck: P.S.

I removed the references and realigned with the Android counterpart. Thanks for calling this out.

KatherineInCode: My just base-level plan-mode documents have been going in `~/.claude/plans` for the time being, which *is* a mite inconvenient. Probably something for me to consider doing some more reconfiguration around 🤔

SaintPatrck:
Thanks @KatherineInCode

PR #2458, opened 2026-03-16
[PM-33311] fix: Fix ViewController interactions in VaultCoordinator

KatherineInCode:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33311

## 📔 Objective

This cleans up some interactions between view controllers in `VaultCoordinator` while also keeping the `VaultItemSelectionProcessor` from dismissing the view when the search changes.

With the latter, the `dismiss()` was added probably blithely because the pre-26 profile switcher dismisses on search change, but the search bar is available pre-26 when the profile switcher is open; it no longer is.

Furthermore, the changes in `VaultCoordinator` cover some other new interactions with the post-26 profile switcher, by first making sure we only dismiss the profile switcher if it's displayed (instead of it *and* the `VaultItemSelectionView`) and also in general making sure when we go to display the `VaultItemSelectionView` we don't have one hanging around from before the user changing profiles.

### DIFF
--- a/BitwardenShared/UI/Vault/Vault/VaultCoordinator.swift
+++ b/BitwardenShared/UI/Vault/Vault/VaultCoordinator.swift
@@ -242,8 +242,7 @@ final class VaultCoordinator: Coordinator, HasStackNavigator { // swiftlint:disa
             // If we're presenting a more complicated stack of view controllers (in particular, this could happen
             // if the user goes to the change profile sheet in the VaultItemSelection modal) then we only want to
             // dismiss the presented one, not the full stack.
-            if let stackNavigatorAsViewController = stackNavigator as? UIViewController,
-               let presentedViewController = stackNavigatorAsViewController.presentedViewController {
+            if let presentedViewController = stackNavigator?.rootViewController?.presentedViewController {
                 presentedViewController.dismiss(animated: UI.animated)
             } else {
                 stackNavigator?.dismiss()


BitwardenShared/UI/Vault/Vault/VaultCoordinator.swift
```diff
@@ -239,7 +239,15 @@ final class VaultCoordinator: Coordinator, HasStackNavigator { // swiftlint:disa
                 }
             }
         case .dismiss:
-            stackNavigator?.dismiss()
+            // If we're presenting a more complicated stack of view controllers (in particular, this could happen
+            // if the user goes to the change profile sheet in the VaultItemSelection modal) then we only want to
+            // dismiss the presented one, not the full stack.
+            if let stackNavigatorAsViewController = stackNavigator as? UIViewController,
+               let presentedViewController = stackNavigatorAsViewController.presentedViewController {
```
matt-livefront: ⛏️ You can use `stackNavigator.rootViewController` to get the view controller without having to cast.

BitwardenShared/UI/Vault/Vault/VaultCoordinator.swift
```diff
@@ -464,7 +472,13 @@ final class VaultCoordinator: Coordinator, HasStackNavigator { // swiftlint:disa
...
         )
 
-        stackNavigator?.present(VaultItemSelectionView(store: Store(processor: processor)))
+        // When we display the Vault Item Selection, we want to make sure we're not already displaying something else.
+        // In particular, this could be another Vault Item Selection screen, if the user is changing profiles.
+        // If the `stackNavigator` isn't displaying anything, the `dismiss()` will be a no-op, and it will simply
+        // present as expected.
+        stackNavigator?.dismiss {
+            self.stackNavigator?.present(VaultItemSelectionView(store: Store(processor: processor)))
+        }
```
matt-livefront, line 480: 🤔 Good catch! I wonder if this should apply to any other deep link type navigations that open in the app? Would it make sense to apply this to anything that goes through `AppProcessor.checkIfLockedAndPerformNavigation(route:)`? It looks like we also similarly present the import CXF screen from there.

KatherineInCode: I think that makes sense to at least investigate, but also that it's outside of the scope of this issue. I've opened [PM-33795](https://bitwarden.atlassian.net/browse/PM-33795) so we don't lose track of that.

PR #2460, opened 2026-03-18
llm: Update CODEOWNERS for Claude related files

SaintPatrck:
## 🎟️ Tracking

Internal Claude tooling improvements

## 📔 Objective

Added @bitwarden/team-ios as code owners for Claude related files. The majority of these files will contain content specific to the iOS project so it makes sense for @bitwarden/team-ios to be owners of these files as well.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -42,6 +42,6 @@
 **/docker-compose.yml @bitwarden/team-appsec @bitwarden/dept-bre
 
 # Claude related files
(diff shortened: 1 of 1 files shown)


michaltk (commented):
## Questions

1. **Workflow ownership** — should `@bitwarden/team-ios` also be added to the `respond.yml` and `review-code.yml` lines, or are those intentionally kept under `team-ai-sme` only?

PR #2003, opened 2025-09-29
[PM-26292] chore: Remove unused strings

mKoonrad:
## 📔 Objective

I removed many unused strings. I checked manually each strings. Some of strings was related to Android app.

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

PR #2464, opened 2026-03-18
[PM-26292] tool: Add script to delete unused Localizable.strings entries

KatherineInCode:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-26292

## 📔 Objective

This builds on #2412 by adding a module that finds unused strings and deletes them. I also did some refactoring of the strings file management into a separate python module.

I also ran the script, and it found 308 unused keys, many of which were identified in #2003. I'm not entirely sure why it's so many, especially of what seem like basic strings, though my guess is that some are used on Android, and a lot of them are detritus from MAUI or earlier that just never got cleaned up.

As noted in #2412, eventually we could have running these scripts be part of the CI process in some way, but currently these are still manual runs.


Scripts/fix-localizable-strings/main.py
```diff
@@ -45,6 +51,40 @@ def cmd_delete_duplicates(args: argparse.Namespace) -> None:
         print(f"    {key}")
 
 
+def cmd_delete_unused(args: argparse.Namespace) -> None:
+    if args.dry_run:
+        with open(args.strings, encoding="utf-8") as f:
+            content = f.read()
+        import os
```
vvolkgang: 🎨 suggestion - moving inline imports to the top of the scripts, makes it easier to know in advance what the script is using. There's another in the `test_delete_unused_strings.py‎` file (`import shutil`).

Scripts/fix-localizable-strings/delete_unused_strings.py
```diff
@@ -0,0 +1,129 @@
...
+# Matches any `Localizations.identifier` reference in Swift source, including
+# cases where the identifier is on the next line (e.g. `Localizations\n    .foo`).
+_LOCALIZATIONS_RE = re.compile(r'Localizations\s*\.([a-zA-Z_][a-zA-Z0-9_]*)')
+
+# Matches any character that is not valid in a Swift identifier.
+_NON_IDENTIFIER_RE = re.compile(r'[^a-zA-Z0-9_]')
+
+
+def _normalize_key(key: str) -> str:
+    """Normalize a ``.strings`` key for comparison against a SwiftGen identifier.
```
vvolkgang, line 25: 💭 Very personal take - given these scripts aren't for public distribution, these docstrings are blowing up line count without adding much, method name and the 3 comments above the regexps were enough for me at least.

vvolkgang (changes requested):
Left some comments but nothing major here, we're just waiting for the `Localizable.strings` file to be updated following the merge of https://github.com/bitwarden/ios/pull/2003

Scripts/fix-localizable-strings/delete_unused_strings.py
KatherineInCode: Our general MO in the codebase is to err on the side of documenting everything, even if it's private and small, so I'd prefer to keep with that trend. I can see the arguments either way, but in something like this, even if it's a pretty-niche script not for public distribution, it is still public, so I'd err on the side of over-documenting over under-documenting, especially since it calls out some of the caveats and logic and implications more than the regex alone does.

vvolkgang (approved):

PR #2465, opened 2026-03-18
[PM-27595] feat: Improve autofill onboarding

fedemkr:
## 🎟️ Tracking

[PM-27595](https://bitwarden.atlassian.net/browse/PM-27595)

## 📔 Objective

Improve autofill onboarding so it uses `ASSettingsHelper.requestToTurnOnCredentialProviderExtension` on iOS 18+ phones.

A `ASSettingsHelperProxy` and `ASSettingsMediator` have been added to deal with `ASSettingsHelper` stuff and be able to be mocked.
Added `AutofillStateService` as a protocol bucket for `StateService` functions related to Autofill.

[PM-27595]: https://bitwarden.atlassian.net/browse/PM-27595?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/UI/Platform/Settings/Settings/AutoFill/AutoFillProcessor.swift
+++ b/BitwardenShared/UI/Platform/Settings/Settings/AutoFill/AutoFillProcessor.swift
@@ -148,8 +148,8 @@ final class AutoFillProcessor: StateProcessor<AutoFillState, AutoFillAction, Aut
         }
 
         do {
-            let isOS = try await services.asSettingsMediator.requestToTurnOnCredentialProviderExtension()
-            guard isOS else { return }
+            let isOn = try await services.asSettingsMediator.requestToTurnOnCredentialProviderExtension()
+            guard isOn else { return }
 
             await dismissSetUpAutofillActionCard()
             state.toast = Toast(title: Localizations.autofillEnabled)


BitwardenShared/Core/Platform/Services/StateService.swift
```diff
@@ -224,6 +224,11 @@ protocol StateService: AnyObject {
     ///
     func getIntroCarouselShown() async -> Bool
 
+    /// Gets the time of the last request to turn on credential provider.
+    ///
+    /// - Returns: The last time a request to turn on credential provider was done.
+    func getLastRequestToTurnOnCredentialProvider() async -> Date?
```
KatherineInCode: 🤔 Does it make sense for this to be exposed to the entire application through `StateService`, or would it make more sense to be segregated into a smaller interface—either something new, or one of the ones we already have?

BitwardenShared/UI/Autofill/Utilities/ASSettingsHelperProxyTests.swift
```diff
@@ -0,0 +1,87 @@
...
+    @Test
+    @available(iOS 18, *)
+    func requestToTurnOnCredentialProviderExtension_canRequestWhenNoPriorRequest() async {
+        // No prior request date means `canRequest` should be true.
+        #expect(stateService.lastRequestToTurnOnCredentialProvider == nil)
+
+        // We can't call the actual ASSettingsHelper API in unit tests, but we can verify
+        // the proxy saves the timestamp after a successful request attempt. Since the
+        // system API will fail in a test environment, we only verify the `.cantRequest`
+        // branch is not taken.
```
matt-livefront: 🎨 Could we mock `ASSettingsHelper` for these tests? I was thinking that was what `ASSettingsHelperProxy` was for, but it's actually adding additional logic on top of `ASSettingsHelper`.

BitwardenShared/UI/Platform/Settings/Settings/AutoFill/AutoFillProcessor.swift
```diff
@@ -186,3 +214,18 @@ final class AutoFillProcessor: StateProcessor<AutoFillState, AutoFillAction, Aut
         })
     }
 }
+
+// MARK: - PasswordAutoFillProcessorDelegate
+
+/// A delegate to notify when autofill was successfully enabled from the password autofill screen.
+///
+protocol PasswordAutoFillProcessorDelegate: AnyObject {
```
matt-livefront: 🎨 Should this live alongside the processor in `PasswordAutofillProcessor.swift`? I think that's the convention we've used elsewhere.

BitwardenShared/UI/Autofill/Utilities/ASSettingsHelperProxy.swift
```diff
@@ -0,0 +1,85 @@
...
+    @available(iOS 17.0, *)
+    func openVerificationCodeAppSettings() async throws
+
+    /// Call this method from your containing app to request to turn on a contained Credential Provider Extension.
+    /// If the extension is not currently enabled, a prompt will be shown to allow it to be turned on.
+    /// Returns ``CredentialProviderExtensionRequestResult/cantRequest`` if the required wait time has not elapsed
+    /// since the last request, or ``CredentialProviderExtensionRequestResult/turnedOn(_:)`` with whether the
+    /// credential provider is enabled.
+    @available(iOS 18.0, *)
+    func requestToTurnOnCredentialProviderExtension() async -> CredentialProviderExtensionRequestResult
```
matt-livefront: 🎨 Would it make sense to throw an error here in the "can't request" state? Then you could just return a bool instead of the result type. Does that simplify the processors/tests? IMO it feels more "Swifty", but I'm curious what it feels like if you implement it.

BitwardenShared/Core/Platform/Services/StateService.swift
fedemkr: Yes, I've split this into `AutofillStateService`. I was thinking we could have "general" state service protocols as needed for each major group (auth, autofill, platform, vault...) when we can't put the functions into a more specific state service like the user session or biometrics' state services.
I think with that we could have these high level buckets for state service's things which would better segregate the big `StateService` until a more specific bucket is needed.
What do you think? cc: @matt-livefront

BitwardenShared/UI/Autofill/Utilities/ASSettingsHelperProxyTests.swift
fedemkr: I've split this into `ASSettingsHelperProxy` and `ASSettingsMediator` where the proxy is now acting like that so we can mock it and the mediator has the additional logic.

BitwardenShared/Core/Platform/Services/StateService.swift
KatherineInCode: That seems like a good direction to go to at least start breaking up the monolith.

BitwardenShared/UI/Platform/Settings/Settings/AutoFill/AutoFillProcessor.swift
```diff
@@ -123,15 +130,36 @@ final class AutoFillProcessor: StateProcessor<AutoFillState, AutoFillAction, Aut
...
 
+    /// Sets up as credential provider for autofill.
+    private func setUpAutofill() async {
+        guard #available(iOS 18.0, *) else {
+            coordinator.navigate(to: .passwordAutoFill)
+            return
+        }
+
+        do {
+            let isOS = try await services.asSettingsMediator.requestToTurnOnCredentialProviderExtension()
```
KatherineInCode: 🤔 I'm blanking on what OS would mean in this context. I think the logic is fine, I'm just not fully sure how to interpret the (short-lived) variable name.

fedemkr: 🤦 Typo, I meant `isOn`. Fixed.

PR #2467, opened 2026-03-19
[PM-32401] fix: Prevent TOTP autofill for non-premium accounts

matt-livefront:
## 🎟️ Tracking

[PM-32401](https://bitwarden.atlassian.net/browse/PM-32401)

## 📔 Objective

This fixes an issue where non-premium accounts were able to use TOTP autofill.

- Adds an authorization check prior to autofilling OTP codes.
- Only creates `ASOneTimeCodeCredentialIdentity` for the credential identity store if authorized.

[PM-32401]: https://bitwarden.atlassian.net/browse/PM-32401?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/Core/Autofill/Services/AutofillCredentialServiceTests.swift
+++ b/BitwardenShared/Core/Autofill/Services/AutofillCredentialServiceTests.swift
@@ -694,8 +694,10 @@ class AutofillCredentialServiceTests: BitwardenTestCase { // swiftlint:disable:t
 
     /// `provideFido2Credential(for:autofillCredentialServiceDelegate:fido2UserVerificationMediatorDelegate:)`
     /// succeeds with device auth key.
-    @available(iOS 18.0, *)
     func test_provideFido2Credential_succeeds_deviceAuthKey() async throws {
+        guard #available(iOS 18.0, *) else {
+            throw XCTSkip("Skipped on iOS < 18.0")
+        }
         stateService.activeAccount = .fixture()
         configService.featureFlagsBool[.deviceAuthKey] = true
         let passkeyIdentity = ASPasskeyCredentialIdentity.fixture(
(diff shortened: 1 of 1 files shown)


BitwardenShared/Core/Autofill/Services/AutofillCredentialServiceTests.swift
```diff
@@ -919,6 +919,47 @@ class AutofillCredentialServiceTests: BitwardenTestCase { // swiftlint:disable:t
         }
     }
 
+    /// `provideOTPCredential(for:autofillCredentialServiceDelegate:repromptPasswordValidated:)`
+    /// succeeds when the user is authorized to use TOTP.
+    @available(iOS 18.0, *)
+    func test_provideOTPCredential_totpAuthorized() async throws {
```
fedemkr: 🎨 On XCTest the `@available` is not respected for the test runs (it does on new Swift Testing Framework), so if one runs this on iOS 17 simulator it will try to run the test and it will fail. I'd add a guard at the beginning as with other similar tests like in `ASPasskeyCredentialRequestExtensionsTests`.

```suggestion
    func test_provideOTPCredential_totpAuthorized() async throws {
        guard #available(iOS 18.0, *) else {
            throw XCTSkip("Skipped on iOS < 18.0")
        }
```
Same applies to other tests here.

BitwardenShared/Core/Vault/Services/TOTP/TOTPService.swift
```diff
@@ -56,25 +61,24 @@ struct DefaultTOTPService: TOTPService {
...
             throw TOTPKeyError.invalidKeyFormat
         }
 
         return TOTPKeyModel(authenticatorKey: key)
     }
+
+    func isTotpAuthorized(for cipher: CipherView) async -> Bool {
+        let accountHasPremium = await stateService.doesActiveAccountHavePremium()
+        return cipher.organizationUseTotp || accountHasPremium
+    }
```
fedemkr, line 83: ♻️  Perhaps we could create a tech debt task to update other places where this logic is used like `DefaultVaultRepository` or `DefaultVaultListPreparedDataBuilder`. What do you think?

fedemkr (commented):
Looks good, just one ⛏️

BitwardenShared/Core/Autofill/Services/AutofillCredentialServiceTests.swift
matt-livefront: Good catch, updated!

BitwardenShared/Core/Vault/Services/TOTP/TOTPService.swift
matt-livefront: Yep, great idea, I added a ticket: https://bitwarden.atlassian.net/browse/PM-34016.

PR #2468, opened 2026-03-19
[PM-32662] feat: Improve item transfer UI

fedemkr:
## 🎟️ Tracking

[PM-32662](https://bitwarden.atlassian.net/browse/PM-32662)

## 📔 Objective

Improve item transfer UI to change "Learn more" link buttons and "Decline and leave" button.

## 📸 Screenshots

Item transfer
Are you sure you want to leave

[PM-32662]: https://bitwarden.atlassian.net/browse/PM-32662?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/UI/Vault/MigrateToMyItems/MigrateToMyItemsView.swift
+++ b/BitwardenShared/UI/Vault/MigrateToMyItems/MigrateToMyItemsView.swift
@@ -71,6 +71,8 @@ struct MigrateToMyItemsView: View {
                 .styleGuide(.body, weight: .semibold)
                 .foregroundColor(SharedAsset.Colors.textSecondary.swiftUIColor)
                 .tint(SharedAsset.Colors.buttonOutlinedForeground.swiftUIColor)
+                // we need this moved up a bit to look like it's continuing
+                // from the previous "message" paragraph without much space between them.
                 .padding(.top, -8)
             }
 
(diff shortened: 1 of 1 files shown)


BitwardenShared/UI/Vault/MigrateToMyItems/MigrateToMyItemsView.swift
```diff
@@ -62,26 +62,29 @@ struct MigrateToMyItemsView: View {
...
+            ) {
+                Text(LocalizedStringKey(
+                    Localizations.learnMoreLink(
+                        ExternalLinksConstants.transferOwnership,
+                    ),
+                ))
+                .styleGuide(.body, weight: .semibold)
+                .foregroundColor(SharedAsset.Colors.textSecondary.swiftUIColor)
+                .tint(SharedAsset.Colors.buttonOutlinedForeground.swiftUIColor)
+                .padding(.top, -8)
```
KatherineInCode, line 76: 🤔 The other places where we do `.padding(.top, -8)`, we have a comment to explain why; do we want to do that here, too?

fedemkr: It's a bit redundant as the reason is just to move it up a bit so it's closer to the message paragraph, but added a comment in there with that just in case.

PR #2470, opened 2026-03-20
[PM-33908] chore: Consolidate KeychainService to BitwardenKit

KatherineInCode:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33908

## 📔 Objective

This consolidates `KeychainService` into `BitwardenKit`, preferring the `BitwardenShared` variant over the variants in `AuthenticatorShared` and `AuthenticatorBridgeKit`. This should help us remain more consistent about how we handle keychain things across the board.

This also, as a result, means we no longer throw a -25300 on a delete (where we can't find the item) with the other two variants, which should fix some of our issues with them.


BitwardenKit/Core/Platform/Services/KeychainServiceTests.swift
```diff
@@ -61,16 +57,4 @@ class KeychainServiceErrorTests: BitwardenTestCase {
         )
         XCTAssertEqual(accessControl, expected)
     }
-
-    /// `getter:errorUserInfo` gets the appropriate user info based on the error case.
-    func test_errorUserInfo() {
```
KatherineInCode, line 66: Deleted because it [already exists](https://github.com/bitwarden/ios/blob/main/BitwardenKit/Core/Platform/Models/Enum/KeychainServiceErrorTests.swift)

BitwardenKit/Core/Platform/Services/Mocks/MockKeychainService.swift
```diff
@@ -0,0 +1,154 @@
+import BitwardenKit
+import Foundation
+
+/// A mock implementation of `KeychainService` for testing purposes.
+///
+/// This mock allows you to control the behavior and results of keychain operations,
+/// and inspect the parameters passed to each method call.
+public class MockKeychainService {
```
fedemkr, line 8: ♻️ Given the refactor in this PR, is it possible to use Sourcery directly for the Keychain Service or do we need a custom mock for something in particular? I'm fine with this for now but if we could use Sourcery it'd be great.

fedemkr (commented):
Looks great, just one thought.

BitwardenKit/Core/Platform/Services/Mocks/MockKeychainService.swift
KatherineInCode: I actually started down the Sourcery route, but ran into there being a fair number of tests that would need updating, and decided to pull back on it for now, especially as I plan on doing additional work around the keychain. I'm definitely wanting to drive towards that, though, it just might be a little further down the road.

fedemkr: Sounds good.

PR #2488, opened 2026-03-25
[PM-33981] feat: Add device management localizations

andrebispo5:
Depends on: https://github.com/bitwarden/ios/pull/2487

## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33981

## 📔 Objective

Add localized strings for the device management feature including device types, activity status labels, and UI text.


BitwardenResources/Localizations/en.lproj/Localizable.strings
```diff
@@ -1330,3 +1330,30 @@
 "LearnMoreAboutAutofill" = "Learn more about autofill";
 "TurnOnNow" = "Turn on now";
 "AcceptTransfer" = "Accept transfer";
+"ManageDevices" = "Manage devices";
+"NoDevicesFound" = "No devices found";
+"CurrentSession" = "Current session";
+"PendingRequest" = "Pending request";
+"PendingRequests" = "%1$d Pending requests";
```
fedemkr: ⚠️ Localizations that need pluralization based on a parameter should be added in `Localizable.stringsdict` now so they're handled correctly on all languages.

BitwardenResources/Localizations/en.lproj/Localizable.strings
```diff
@@ -1330,3 +1330,30 @@
...
+"Trusted" = "Trusted";
+"RecentlyActive" = "Recently active: %1$@";
+"RecentlyActiveLabel" = "Recently active:";
+"FirstLogin" = "First login: %1$@";
+"FirstLoginLabel" = "First login:";
```
fedemkr, line 1179: ⛏️ I would use directly "Colon" instead of "Label" as suffix. What do you think @KatherineInCode ?

BitwardenResources/Localizations/en.lproj/Localizable.strings
```diff
@@ -1330,3 +1330,30 @@
 "LearnMoreAboutAutofill" = "Learn more about autofill";
 "TurnOnNow" = "Turn on now";
 "AcceptTransfer" = "Accept transfer";
+"ManageDevices" = "Manage devices";
+"NoDevicesFound" = "No devices found";
+"CurrentSession" = "Current session";
+"PendingRequest" = "Pending request";
+"PendingRequests" = "%1$d Pending requests";
+"Trusted" = "Trusted";
+"RecentlyActive" = "Recently active: %1$@";
```
fedemkr: 🎨 This has parameter so I'd change it to `"RecentlyActiveX"`

BitwardenResources/Localizations/en.lproj/Localizable.strings
fedemkr: 🤔 Not following why you have two first login localizations? Couldn't you have just the one with the parameter and if the parameter is empty then it's just not shown? Or is there a distinction so the "space" is not shown? Same applies to `RecentlyActive`

BitwardenResources/Localizations/en.lproj/Localizable.strings
```diff
@@ -1330,3 +1330,30 @@
...
+"Today" = "Today";
+"ThisWeek" = "This week";
+"LastWeek" = "Last week";
+"ThisMonth" = "This month";
+"OverThirtyDaysAgo" = "Over 30 days ago";
+"UnknownDevice" = "Unknown device";
+"Unknown" = "Unknown";
+"RemoveDevice" = "Remove device";
+"AreYouSureYouWantToRemoveDevice" = "Are you sure you want to remove %1$@?";
+"AreYouSureYouWantToLogOutDevice" = "Are you sure you want to log out %1$@?";
```
fedemkr: ⛏️ We often use "X" "Y" "Z" for parameters in the key (if you want you can add a comment to indicate that it represents a device):
```suggestion
"AreYouSureYouWantToRemoveX" = "Are you sure you want to remove %1$@?";
"AreYouSureYouWantToLogOutX" = "Are you sure you want to log out %1$@?";
```

BitwardenResources/Localizations/en.lproj/Localizable.strings
KatherineInCode: I'm in the category of wondering if there's a better way to do this in general, as I agree with the later comment of not following why there's two of them. But if we *do* have to have two, then I agree with `Colon` over `Label`.

BitwardenResources/Localizations/en.lproj/Localizable.strings
```diff
@@ -1042,6 +1042,30 @@
...
+"CurrentSession" = "Current session";
+"PendingRequest" = "Pending request";
+"Trusted" = "Trusted";
+"RecentlyActiveX" = "Recently active: %1$@";
+"FirstLoginX" = "First login: %1$@";
+"Today" = "Today";
+"ThisWeek" = "This week";
+"LastWeek" = "Last week";
+"ThisMonth" = "This month";
+"OverThirtyDaysAgo" = "Over 30 days ago";
```
KatherineInCode: 🤔 Would it make sense to just pluralize this as a `OverXDaysAgo` property, to give us flexibility in the future?

BitwardenResources/Localizations/en.lproj/Localizable.stringsdict
```diff
@@ -19,6 +19,23 @@
...
+		<key>NSStringLocalizedFormatKey</key>
+		<string>%#@PendingRequests@</string>
+		<key>PendingRequests</key>
+		<dict>
+			<key>NSStringFormatSpecTypeKey</key>
+			<string>NSStringPluralRuleType</string>
+			<key>NSStringFormatValueTypeKey</key>
+			<string>d</string>
+			<key>one</key>
+			<string>%d Pending request</string>
```
KatherineInCode: 🤔 Why is `Pending` capitalized here? If we're following sentence casing, the number would be capitalized (if spelled out), and it'd be a lowercased `pending`

andrebispo5: Actually I was looking into the stacked PRs and it seems that this strings is not in use anymore, I'll remove it.

BitwardenResources/Localizations/en.lproj/Localizable.strings
andrebispo5: I'm not sure that would give us much flexibility since if we later want the plural "this week" is "these weeks", etc. And for the key naming it would be "Over 1 day ago" -> "Over 30 days ago".

fedemkr: I'm split on this one, I mean if we want the parameter then this one in particular should be moved to `Localizable.stringsdict` so pluralization can be handled correctly depending on the number of days without problems.
However, I also think we could apply YAGNI here as maybe we won't need any other X number of days so adding the complexity of the parameter translation may not be necessarily at all for a long time and we'd reduce translation effort.

andrebispo5: let's leave it as is for the time being, if this change in the future and needs expansion we will convert it.

PR #2492, opened 2026-03-25
[PM-34106] chore: Consolidate KeychainItem into BitwardenKit via protocol

KatherineInCode:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34106

## 📔 Objective

This expands what was the `KeychainStorageKeyPossessing` protocol to include other properties from `KeychainItem`, and then renames it to `KeychainItem` while also renaming the former `KeychainItem` classes to `BitwardenKeychainItem` and `AuthenticatorKeychainItem`. This will make it easier to do further refactoring of our keychain code by making it easier to consolidate more things into `BitwardenKit`.

I considered turning this into a struct with extensions with static members, like `FeatureFlag`, but because many `KeychainItem` enum cases take an associated value (usually of the user ID), those would have to become static functions instead of static variables, and I wasn't particularly fond of that. However, I do think if we're good with the static functions, we could take that route instead, though I feel like it's six of one and half dozen of another.

### DIFF
--- a/AuthenticatorShared/Core/Auth/Services/KeychainRepository.swift
+++ b/AuthenticatorShared/Core/Auth/Services/KeychainRepository.swift
@@ -200,7 +200,7 @@ class DefaultKeychainRepository: KeychainRepository {
     ///
     func setValue(_ value: String, for item: AuthenticatorKeychainItem) async throws {
         let accessControl = try keychainService.accessControl(
-            protection: kSecAttrAccessibleWhenUnlockedThisDeviceOnly,
+            protection: item.protection,
             for: item.accessControlFlags ?? [],
         )
         let baseQuery = await keychainQueryValues(for: item)


AuthenticatorBridgeKit/SharedKeychain/SharedKeychainStorage.swift
```diff
@@ -5,13 +5,17 @@ import Foundation
...
+public enum SharedKeychainItem: Equatable, Hashable, Sendable, KeychainItem {
     /// The keychain item for the authenticator encryption key.
     case authenticatorKey
 
     /// A date at which a BWPM account automatically logs out.
     case accountAutoLogout(userId: String)
 
+    public var accessControlFlags: SecAccessControlCreateFlags? { nil }
+
+    public var protection: CFTypeRef { kSecAttrAccessibleWhenUnlockedThisDeviceOnly }
```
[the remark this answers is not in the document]
KatherineInCode: This is a reasonable change. Made.

AuthenticatorShared/Core/Auth/Services/KeychainRepository.swift
```diff
@@ -196,10 +198,10 @@ class DefaultKeychainRepository: KeychainRepository {
     ///   - value: The value associated with the keychain item to set.
     ///   - item: The keychain item used to set the associated value.
     ///
-    func setValue(_ value: String, for item: KeychainItem) async throws {
+    func setValue(_ value: String, for item: AuthenticatorKeychainItem) async throws {
         let accessControl = try keychainService.accessControl(
             protection: kSecAttrAccessibleWhenUnlockedThisDeviceOnly,
```
matt-livefront: ❓ Does it make sense to use `item.protection` here now that it exists?
```suggestion
            protection: item.protection,
```

KatherineInCode: Good call. I was going to end up changing it to that when I pulled things into the façade, but I can just take care of that now, too.

PR #2495, opened 2026-03-25
[PM-250] fix: Update MP Hint Validation

morganzellers-bw:
## 🎟️ Tracking

[PM-250](https://bitwarden.atlassian.net/browse/PM-250])

## 📔 Objective

Fixes an issue where users are able to set their password hint to the same thing as their Master Password

## 📸 Screenshots

mpHint_Final

[PM-250]: https://bitwarden.atlassian.net/browse/PM-250?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -164,13 +164,6 @@ typos                                   # Spell check
 - Use `// MARK: -` sections to organize code within files
 - JIRA ticket required in TODO comments (enforced by `todo_without_jira` SwiftLint rule)
 
-### Pre-build Scripts
-Configured in `project-pm.yml`:
-- SwiftLint runs as post-compile script
-- SwiftFormat lint runs as post-compile script
-- Sourcery mock generation runs as pre-build phase
-- SwiftGen asset code generation runs as pre-build phase
-
 ## Patterns
 
 ### DO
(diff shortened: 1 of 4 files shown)


BitwardenShared/UI/Auth/CompleteRegistration/CompleteRegistrationProcessor.swift
```diff
@@ -169,6 +173,10 @@ class CompleteRegistrationProcessor: StateProcessor<
                 })
                 return
             }
+            guard !state.doesMasterPasswordMatchHint else {
```
andrebispo5: 🤔 is this needed here? `completeRegistration()` already does this check inside.

BitwardenShared/UI/Auth/Extensions/Alert+Auth.swift
```diff
@@ -247,6 +247,20 @@ extension Alert {
...
+    ///
+    static func masterPasswordAndHintMatchAlert() -> Alert {
+        Alert(
+            title: Localizations.anErrorHasOccurred,
+            message: Localizations.yourPasswordAndHintCannotBeTheSame,
+            alertActions: [
+                AlertAction(title: Localizations.continue, style: .default),
+            ],
+        )
+    }
```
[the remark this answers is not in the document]
andrebispo5: agreed 👍

BitwardenResources/Localizations/en.lproj/Localizable.strings
```diff
@@ -1330,3 +1330,4 @@
 "LearnMoreAboutAutofill" = "Learn more about autofill";
 "TurnOnNow" = "Turn on now";
 "AcceptTransfer" = "Accept transfer";
+"YourPasswordAndHintCannotBeTheSame" = "Your password and hint cannot be the same. Please choose a different hint.";
```
fedemkr: ⛏️ Use the full value for the key or add `DescriptionLong` suffix. In this case I think it's not that long so you can use the full value.
```suggestion
"YourPasswordAndHintCannotBeTheSamePleaseChooseADifferentHint" = "Your password and hint cannot be the same. Please choose a different hint.";
```

BitwardenShared/UI/Auth/CompleteRegistration/CompleteRegistrationProcessor.swift
```diff
@@ -34,6 +34,9 @@ enum CompleteRegistrationError: Error {
 
     /// The environment urls to complete the registration are empty
     case preAuthUrlsEmpty
+
+    /// The password and password hint match
+    case passwordAndHintMatch
```
fedemkr: ⛏️ Order alphabetically as the others.

BitwardenShared/UI/Auth/CompleteRegistration/CompleteRegistrationProcessor.swift
```diff
@@ -293,6 +305,8 @@ class CompleteRegistrationProcessor: StateProcessor<
                 title: Localizations.anErrorHasOccurred,
                 message: Localizations.theRegionForTheGivenEmailCouldNotBeLoaded,
             ))
+        case .passwordAndHintMatch:
+            coordinator.showAlert(.passwordMatchesHint)
```
fedemkr: ⛏️ Order alphabetically in the switch.

BitwardenShared/UI/Auth/CompleteRegistration/CompleteRegistrationProcessor.swift
morganzellers-bw: No it isn't, will remove 👍

BitwardenShared/UI/Auth/CompleteRegistration/CompleteRegistrationState.swift
```diff
@@ -36,6 +36,10 @@ struct CompleteRegistrationState: Equatable, Sendable {
         return passwordStrengthScore < 3
     }
 
+    var doesMasterPasswordMatchHint: Bool {
+        passwordText == passwordHintText
+    }
```
fedemkr, line 43: 🤔 Should the `passwordHintText` be trimmed of whitespaces for an additional comparison? This for the case the user adds a space by mistake in the hint field putting the same MP.
```suggestion
    var doesMasterPasswordMatchHint: Bool {
        passwordText == passwordHintText || passwordText == passwordHintText.trimmingCharacters(in: .whitespacesAndNewlines)
    }
```

BitwardenShared/UI/Auth/CompleteRegistration/CompleteRegistrationView+ViewInspectorTests.swift
```diff
@@ -66,6 +66,25 @@ class CompleteRegistrationViewTests: BitwardenTestCase {
...
+        try button.tap()
+
+        waitFor(!processor.effects.isEmpty)
+        XCTAssertEqual(processor.effects.last, .completeRegistration)
+
+        // Checking Alert
+        let alert = Alert.passwordMatchesHint
+        XCTAssertEqual(alert.message, Localizations.yourPasswordAndHintCannotBeTheSame)
+    }
+
```
fedemkr: 🎨 Typically we name the function in ViewInspector tests with the component to do the action, in this case the continue button, so it would be something like:
`func test_continue_tap_passwordMatchesHint() throws`
⚠️ In this case you're testing two different things, one is what happens when tapping the continue button and then the message of the alert:
- What happens when tapping the `continue` button is already covered by `test_createAccountButton_tap`
- The specific message of the alert is already covered by `AlertAuthTests::test_masterPasswordAndHintMatchAlert`. Moreover, a ViewInspector test shouldn't test logic of the processor, just that the processor got the `action`/`effect` after the user made the action, in this case tapping.
- Therefore this test is not needed and you can remove it.

BitwardenShared/UI/Auth/CompleteRegistration/CompleteRegistrationProcessor.swift
```diff
@@ -223,6 +231,10 @@ class CompleteRegistrationProcessor: StateProcessor<
                 throw CompleteRegistrationError.passwordsDontMatch
             }
 
+            guard !state.doesMasterPasswordMatchHint else {
+                throw CompleteRegistrationError.passwordAndHintMatch
+            }
+
```
fedemkr, line 233: ⛏️ Add tests

BitwardenShared/UI/Platform/Application/Extensions/Alert+Extensions.swift
```diff
@@ -66,6 +66,18 @@ extension Alert {
...
+    static var passwordMatchesHint: Alert {
+        Alert(
+            title: Localizations.anErrorHasOccurred,
+            message: Localizations.yourPasswordAndHintCannotBeTheSame,
+            alertActions: [
+                AlertAction(title: Localizations.continue, style: .default),
+            ],
+        )
+    }
+
```
fedemkr: 🤔 Does the button need to be "Continue"? I feel a bit counterintuitive to "continue" an action that is an error, I'd ok with "Ok". If so, I'd directly use `defaultAlert` (in [Alert+Error.swift](https://github.com/bitwarden/ios/blob/main/BitwardenKit/UI/Platform/Application/Utilities/Alert/Alert%2BError.swift#L36)) as it's used in `preAuthUrlsEmpty` [here](https://github.com/bitwarden/ios/pull/2495/changes#diff-b34245366c260a36dea288ddf0fbb1bb28842923dddfb857ee3b60a5bb55af1eR304).

BitwardenResources/Localizations/en.lproj/Localizable.strings
morganzellers-bw: Sounds good, thanks Fede 👍

BitwardenShared/UI/Auth/CompleteRegistration/CompleteRegistrationState.swift
morganzellers-bw: I like it!

BitwardenShared/UI/Auth/CompleteRegistration/CompleteRegistrationView+ViewInspectorTests.swift
morganzellers-bw: Okay that makes sense, I'll add a test in [CompleteRegistrationProcessor.swift](https://github.com/bitwarden/ios/pull/2495/files/3698700b26e76e92aec838fce7aa0b40d60799bb#diff-b34245366c260a36dea288ddf0fbb1bb28842923dddfb857ee3b60a5bb55af1e) to cover the comparison logic.

BitwardenShared/UI/Platform/Application/Extensions/Alert+Extensions.swift
morganzellers-bw: Ah yeah I'll do that so that we don't even need to add a new alert, thanks for the tip! @fedemkr

BitwardenShared/UI/Auth/CompleteRegistration/CompleteRegistrationState.swift
```diff
@@ -36,6 +36,10 @@ struct CompleteRegistrationState: Equatable, Sendable {
         return passwordStrengthScore < 3
     }
 
+    var doesMasterPasswordMatchHint: Bool {
+        passwordText == passwordHintText || passwordText == passwordHintText.trimmingCharacters(in: .whitespacesAndNewlines)
+    }
```
fedemkr, line 43: ⛏️ Can you add tests for this in the `CompleteRegistrationStateTests`? Although I see it hasn't been created yet, not sure why as there are other computed properties that should have been tested but well if you could create it and add tests for this one that would be great.

morganzellers-bw: @fedemkr I can do that - I'll try and cover the other computed properties too.

BitwardenShared/UI/Auth/CompleteRegistration/CompleteRegistrationStateTests.swift
```diff
@@ -0,0 +1,121 @@
+import XCTest
+
+@testable import BitwardenShared
+
+class CompleteRegistrationStateTests: BitwardenTestCase {
+    func test_doesMasterPasswordMatchHint_whenMatching_returnsTrue() {
+        var subject = CompleteRegistrationState(
```
fedemkr, line 8: ⛏️ All tests should have DocC comments explaining what they test. Check other test files as examples..

PR #2498, opened 2026-03-26
[PM-33569] fix: Remove pending login request if it no longer exists

matt-livefront:
## 🎟️ Tracking

[PM-33569](https://bitwarden.atlassian.net/browse/PM-33569)

## 📔 Objective

When a login with device push notification is received by the app, the app will attempt to prompt the user immediately.

It's possible that it can't be shown immediately (app is backgrounded and killed before the user returns or the user's vault is locked when the push notification is received). For this case, the login with device request is persisted until the next vault unlock, when we attempt to show it to the user. In the meantime, it's possible that the request has expired and been removed from the server, resulting in a 404.

When this occurred, an error was logged to the console/flight recorder, but the persisted request was never deleted. So the app would continue to try and fetch the request on each vault unlock.

This catches the error and deletes the request so the app doesn't continue to try fetching it.

[PM-33569]: https://bitwarden.atlassian.net/browse/PM-33569?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/Core/Platform/Services/NotificationService.swift
+++ b/BitwardenShared/Core/Platform/Services/NotificationService.swift
@@ -78,7 +78,7 @@ protocol NotificationServiceDelegate: AnyObject {
 
 /// The default implementation of `NotificationService`.
 ///
-class DefaultNotificationService: NotificationService {
+class DefaultNotificationService: NotificationService { // swiftlint:disable:this type_body_length
     // MARK: Properties
 
     /// The delegate to handle login request actions originating from notifications.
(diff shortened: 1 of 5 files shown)


BitwardenShared/Core/Platform/Services/NotificationService.swift
```diff
@@ -376,7 +374,7 @@ class DefaultNotificationService: NotificationService {
             // Create an in-app banner notification to tell the user about the login request.
             let content = UNMutableNotificationContent()
             content.title = Localizations.logInRequested
-            content.body = Localizations.confimLogInAttempForX(loginSourceEmail)
+            content.body = Localizations.confirmLogInAttemptForX(loginSourceEmail)
```
andrebispo5, line 377: nice 👍

PR #2183, opened 2025-12-05
[PM-28540] Add BitwardenSharedMocks target

matt-livefront:
## 🎟️ Tracking

[PM-28540](https://bitwarden.atlassian.net/browse/PM-28540)

## 📔 Objective

Adds a `BitwardenSharedMocks` target containing the mocks and fixtures used within BitwardenShared.

This adds consistency with how we've used `BitwardenKitMocks` and `AuthenticatorBridgeKitMocks` and will make it easier to implement Xcodegen templates (https://github.com/bitwarden/ios/pull/2153).

`BitwardenSharedMocks` contains a lot of fixtures and mocks for `internal` types within BitwardenShared. Rather than make these types `public`, I've used testable imports of the new `BitwardenSharedMocks` target in tests.

## ⏰ Reminders before review

- Contributor guidelines followed
- All formatters and local linters executed and passed
- Written new unit and / or integration tests where applicable
- Protected functional changes with optionality (feature flags)
- Used internationalization (i18n) for all UI strings
- CI builds passed
- Communicated to DevOps any deployment requirements
- Updated any necessary documentation (Confluence, contributing docs) or informed the documentation team

## 🦮 Reviewer guidelines

- 👍 (`:+1:`) or similar for great changes
- 📝 (`:memo:`) or ℹ️ (`:information_source:`) for notes or general info
- ❓ (`:question:`) for questions
- 🤔 (`:thinking:`) or 💭 (`:thought_balloon:`) for more open inquiry that's not quite a confirmed issue and could potentially benefit from discussion
- 🎨 (`:art:`) for suggestions / improvements
- ❌ (`:x:`) or ⚠️ (`:warning:`) for more significant problems or concerns needing attention
- 🌱 (`:seedling:`) or ♻️ (`:recycle:`) for future improvements or indications of technical debt
- ⛏ (`:pick:`) for minor or nitpick changes

[PM-28540]: https://bitwarden.atlassian.net/browse/PM-28540?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2514, opened 2026-04-01
[PM-25831] chore: Remove cxp-import-mobile feature flag

matt-livefront:
## 🎟️ Tracking

[PM-25831](https://bitwarden.atlassian.net/browse/PM-25831)

## 📔 Objective

Removes the `cxp-import-mobile` feature flag.

I also removed the remaining `#if SUPPORTS_CXP` macros which were previously removed in https://github.com/bitwarden/ios/pull/1964. 

[PM-25831]: https://bitwarden.atlassian.net/browse/PM-25831?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/Core/Tools/Repositories/ExportCXFCiphersRepositoryTests.swift
+++ b/BitwardenShared/Core/Tools/Repositories/ExportCXFCiphersRepositoryTests.swift
@@ -1,4 +1,3 @@
-#if SUPPORTS_CXP
 import AuthenticationServices
 import BitwardenKitMocks
 import InlineSnapshotTesting
@@ -6,6 +5,7 @@ import TestHelpers
 import XCTest
 
 @testable import BitwardenShared
+@testable import BitwardenSharedMocks
 
 // MARK: - ExportCXFCiphersRepositoryTests
 
@@ -117,7 +117,7 @@ class ExportCXFCiphersRepositoryTests: BitwardenTestCase {
         XCTAssertEqual(result.count, 2)
         XCTAssertEqual(result[0].id, "1")
         XCTAssertEqual(result[1].id, "2")
-        XCTAssertEqual(exportVaultService.fetchAllCiphersIncludeArchived, false)
+        XCTAssertEqual(exportVaultService.fetchAllCiphersToExportIncludeArchived, false)
     }
 
     /// `getAllCiphersToExportCXF()` throws when fetching ciphers throws.
@@ -235,5 +235,3 @@ class ExportCXFCiphersRepositoryTests: BitwardenTestCase {
         }
     }
 }
-
-#endif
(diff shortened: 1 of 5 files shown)


BitwardenShared/Core/Tools/Repositories/ExportCXFCiphersRepositoryTests.swift
```diff
@@ -1,11 +1,11 @@
-#if SUPPORTS_CXP
 import AuthenticationServices
 import BitwardenKitMocks
 import InlineSnapshotTesting
 import TestHelpers
 import XCTest
 
 @testable import BitwardenShared
+@testable import BitwardenSharedMocks
```
KatherineInCode, line 8: 🤔 There shouldn't be anything in the `Mocks` that's not public; you should be able to just straight import it, instead of a `@testable import`, I think?

BitwardenShared/Core/Tools/Repositories/ExportCXFCiphersRepositoryTests.swift
```diff
@@ -117,7 +117,7 @@ class ExportCXFCiphersRepositoryTests: BitwardenTestCase {
         XCTAssertEqual(result.count, 2)
         XCTAssertEqual(result[0].id, "1")
         XCTAssertEqual(result[1].id, "2")
-        XCTAssertEqual(exportVaultService.fetchAllCiphersIncludeArchived, false)
+        XCTAssertEqual(exportVaultService.fetchAllCiphersToExportIncludeArchived, false)
```
KatherineInCode, line 120: 🤔 Maybe I'm missing something, but why did this change?

matt-livefront: This whole file was gated behind `SUPPORTS_CXP`, which is no longer defined. So the file wasn't being compiled. `fetchAllCiphersIncludeArchived` must have been renamed at some point, so it had to be updated once I removed the `#if SUPPORTS_CXP` check.

BitwardenShared/Core/Tools/Repositories/ExportCXFCiphersRepositoryTests.swift
matt-livefront: I think we have fixtures in `BitwardenSharedMocks` which aren't public. I was getting an error for using `Account.fixture()`.

When I added the `BitwardenSharedMocks` target (https://github.com/bitwarden/ios/pull/2183), there were a lot of fixtures and mocks for internal types within `BitwardenShared`. It didn't feel right to make all of those public just for the tests so there are testable imports of `BitwardenSharedMocks`.

KatherineInCode: Ah, okay, that makes sense

PR #2523, opened 2026-04-06
[PM-33849] feat: Add Premium Upgrade view

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33849

## 📔 Objective

Adds the Premium Upgrade view that allows users to upgrade to a premium plan via Stripe checkout.

### Changes
- **New PremiumUpgrade feature** (`BitwardenShared/UI/Billing/PremiumUpgrade/`):
  - `PremiumUpgradeView` - Displays premium pricing, features list, and upgrade button
  - `PremiumUpgradeProcessor` - Handles upgrade button tap by creating a Stripe checkout session
  - `PremiumUpgradeCoordinator` / `PremiumUpgradeModule` - Navigation management
  - `PremiumUpgradeAction`, `PremiumUpgradeEffect`, `PremiumUpgradeRoute`, `PremiumUpgradeState`
- **Vault integration**: Added `.premiumUpgrade` route and `.upgradeToPremium` action to VaultList
- **Testing**: Added `BillingAPIService` mock support to `ServiceContainer`
- **Localization**: Added 8 new strings for premium upgrade UI

## 📸 Screenshots
New
Simulator Screenshot - iPhone 17 Pro - 2026-04-06 at 17 42 19


BitwardenShared/Core/Platform/Services/ServiceContainer.swift
```diff
@@ -27,6 +27,9 @@ public class ServiceContainer: Services { // swiftlint:disable:this type_body_le
     /// The service used by the application to make API requests.
     let apiService: APIService
 
+    /// An optional override for the billing API service, used for testing.
+    var billingAPIServiceOverride: BillingAPIService?
+
```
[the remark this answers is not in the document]
andrebispo5: Thanks forgot to delete this.

BitwardenResources/Localizations/en.lproj/Localizable.strings
```diff
@@ -1350,3 +1350,11 @@
 "TurnOnNow" = "Turn on now";
 "AcceptTransfer" = "Accept transfer";
 "YourPasswordAndHintCannotBeTheSamePleaseChooseADifferentHint" = "Your password and hint cannot be the same. Please choose a different hint.";
+"PerMonth" = "/ month";
```
matt-livefront, line 1037: 🎨 Would it be better to localize the full string with a placeholder if the order needs to change on translation?

```suggestion
"XPerMonth" = "%1$@ / month";
```

BitwardenShared/Core/Billing/Services/BillingService.swift
```diff
@@ -0,0 +1,46 @@
...
+        self.billingAPIService = billingAPIService
+    }
+
+    // MARK: Methods
+
+    func createCheckoutSession() async throws -> URL {
+        let response = try await billingAPIService.createCheckoutSession()
+        let url = response.checkoutSessionUrl
+        guard url.scheme == "https" else {
+            throw BillingError.invalidCheckoutUrl
```
matt-livefront, line 44: ❓ Just curious why this check is needed? Is this more of a safeguard or is there a meaningful reason behind this? Maybe worth a comment for future context.

BitwardenShared/Core/Platform/Services/ServiceContainer.swift
```diff
@@ -60,6 +60,9 @@ public class ServiceContainer: Services { // swiftlint:disable:this type_body_le
     /// The repository to manage biometric unlock policies and access controls the user.
     let biometricsRepository: BiometricsRepository
 
+    /// The service used by the application to manage billing operations.
+    let billingService: BillingService
```
matt-livefront, line 61: ⛏️ The ordering of this should be before `biometricsRepository`.

BitwardenShared/Core/Billing/Services/BillingServiceTests.swift
```diff
@@ -0,0 +1,82 @@
+import TestHelpers
+import XCTest
```
matt-livefront: 🎨 We've been trying to use Swift Testing for new tests because they are significantly faster. Do you want to try and convert the tests in this PR? I'm not sure about the view tests, but the service, processor and coordinators should be easy to convert.

BitwardenShared/UI/Billing/PremiumUpgrade/PremiumUpgradeProcessor.swift
```diff
@@ -0,0 +1,82 @@
...
+    /// Creates a checkout session by calling the billing service.
+    ///
+    private func createCheckoutSession() async {
+        do {
+            state.isLoading = true
+            let url = try await services.billingService.createCheckoutSession()
+            state.isLoading = false
+            state.checkoutURL = url
+        } catch {
+            state.isLoading = false
```
matt-livefront, line 77: 🎨 You could also use a defer block to toggle `isLoading` after async work completes or throws:


```suggestion
            state.isLoading = true
            defer { state.isLoading = false }
            let url = try await services.billingService.createCheckoutSession()
            state.checkoutURL = url
        } catch {
```

BitwardenShared/UI/Billing/PremiumUpgrade/PremiumUpgradeProcessorTests.swift
```diff
@@ -0,0 +1,113 @@
...
+        try await waitForAsync { self.coordinator.errorAlertsShown.count == 1 }
+        XCTAssertEqual(coordinator.errorAlertsShown.first as? BillingError, .unableToOpenCheckout)
+    }
+
+    /// `perform(_:)` with `.upgradeNowTapped` shows an error when the service returns an invalid URL error.
+    @MainActor
+    func test_perform_upgradeNowTapped_invalidUrl() async throws {
+        billingService.createCheckoutSessionThrowableError = BillingError.invalidCheckoutUrl
+
+        await subject.perform(.upgradeNowTapped)
```
matt-livefront, line 97: ⛏️ Could you group this with the other `test_perform_upgradeNowTapped` tests?

BitwardenShared/UI/Billing/PremiumUpgrade/PremiumUpgradeView.swift
```diff
@@ -0,0 +1,169 @@
...
+
+            upgradeButton
+                .padding(.bottom, 12)
+
+            stripeFooter
+        }
+        .padding(.horizontal, 16)
+        .padding(.top, 24)
+        .frame(maxWidth: .infinity, maxHeight: .infinity, alignment: .top)
+        .background(Color(asset: SharedAsset.Colors.backgroundPrimary))
```
matt-livefront: 🎨 Should this be in a scroll view so that it can scroll if necessary for smaller devices, landscape, or large type? You can use the `.scrollView` modifier to set the background color and default padding.

BitwardenShared/UI/Billing/PremiumUpgrade/PremiumUpgradeView.swift
```diff
@@ -0,0 +1,169 @@
...
+            Text(Localizations.unlockMoreAdvancedFeaturesWithPremiumPlan)
+                .styleGuide(.body)
+                .foregroundColor(Color(asset: SharedAsset.Colors.textSecondary))
+                .padding(.bottom, 16)
+
+            featuresList
+        }
+        .padding(16)
+        .background(Color(asset: SharedAsset.Colors.backgroundSecondary))
+        .clipShape(RoundedRectangle(cornerRadius: 12))
```
matt-livefront, line 71: 🎨 It'd be better to use a `ContentBlock` here which will provide the background, clip shape, and dividers.

I think that should work for this use case, and you won't have to manually insert the dividers in `featureRow`. This should simplify some of the view and helps to keep things consistent if/when we need to update the UI across the app.

BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessor.swift
```diff
@@ -212,6 +207,15 @@ final class VaultListProcessor: StateProcessor<
 extension VaultListProcessor {
     // MARK: Private Methods
 
+    /// Handles the app review prompt shown action by updating state and clearing review data.
+    private func handleAppReviewPromptShown() {
```
matt-livefront, line 211: ⛏️ Can you alphabetize this?

BitwardenShared/UI/Vault/Vault/VaultRoute.swift
```diff
@@ -77,6 +77,9 @@ public enum VaultRoute: Equatable, Hashable {
     /// A route to the vault item selection screen.
     case vaultItemSelection(TOTPKeyModel)
 
+    /// A route to the premium upgrade view.
+    case premiumUpgrade
```
matt-livefront: ⛏️ Can you alphabetize this?

BitwardenShared/Core/Billing/Services/BillingService.swift
andrebispo5: Just a safeguard since we are opening a url coming from a service call.

BitwardenShared/UI/Billing/PremiumUpgrade/PremiumUpgradeProcessor.swift
andrebispo5: actually I need to see how this is going to pan out on end to end testing, because I believe i'll need to keep the loading on even after this async func. We'll keep it as is for now.

matt-livefront (dismissed):

matt-livefront (approved):

PR #2527, opened 2026-04-09
[BWA-99] feat: Add next TOTP code preview to BWA

KatherineInCode:
## 🎟️ Tracking
[BWA-99](https://bitwarden.atlassian.net/browse/BWA-99)

## 📔 Objective
When the current TOTP code has 10 or fewer seconds remaining, the Authenticator app can now display the upcoming code below the current one. This preview is gated behind a new "Show next code" toggle in Settings > Appearance (defaults to off), so users can opt in to the behavior.

## 📷 Screenshots

Simulator Screenshot - iPhone 17 Pro - 2026-04-09 at 14 19 52

Simulator Screenshot - iPhone 17 Pro - 2026-04-13 at 14 01 37

[BWA-99]: https://bitwarden.atlassian.net/browse/BWA-99?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


morganzellers-bw (dismissed):

AuthenticatorShared/Core/Platform/Services/TOTPItemDisplayStateService.swift
```diff
@@ -0,0 +1,42 @@
+import Combine
+
+// MARK: - TOTPItemDisplayStateService
+
+/// A protocol for a service that manages how TOTP items are displayed in the item list.
+///
+protocol TOTPItemDisplayStateService: AnyObject { // sourcery: AutoMockable
+
```
fedemkr: ⛏️ Remove empty line (I think this is considered as a warning):
```suggestion
```

AuthenticatorShared/UI/Vault/Views/ItemListItemRow/ItemListItemRowView.swift
```diff
@@ -100,14 +106,43 @@ struct ItemListItemRowView: View {
...
+        }
+    }
+
+    /// The HStack showing the TOTP countdown timer and code(s).
+    @ViewBuilder
+    private func totpCountdownRow(
+        model: TOTPCodeModel,
+        nextCode: TOTPCodeModel?,
+        remaining: Int,
+    ) -> some View {
```
fedemkr: ⛏️ Update docs parameters.

AuthenticatorShared/UI/Vault/Views/ItemListItemRow/ItemListItemRowView.swift
```diff
@@ -100,14 +106,43 @@ struct ItemListItemRowView: View {
...
+        TimelineView(.periodic(from: .now, by: 0.1)) { _ in
+            totpCountdownRow(
+                model: model,
+                nextCode: nextCode,
+                remaining: TOTPExpirationCalculator.remainingSeconds(
+                    for: timeProvider.presentTime,
+                    using: Int(model.period),
+                ),
+            )
+        }
```
fedemkr: 🤔 Why do we need the `TimelineView` if we already have `TOTPCountdownTimerView` to make the periodic updates and calculations? Is it to calculate the `remaining` parameter in there?
I think this could be doing work twice, this is redrawing the content every 0.1 seconds but `TOTPCountdownTimerView` will also update the display time every 0.1 seconds, so will it be redrawn twice per ~0.1 sec?
Are there any drawbacks on using `TOTPCountdownTimerView` for the "next code" as well? instead of the `TimelineView`?

AuthenticatorShared/UI/Vault/Views/ItemListItemRow/ItemListItemRowView.swift
```diff
@@ -100,14 +106,43 @@ struct ItemListItemRowView: View {
...
+        model: TOTPCodeModel,
+        nextCode: TOTPCodeModel?,
+        remaining: Int,
+    ) -> some View {
+        HStack(spacing: 8) {
+            VStack(alignment: .trailing, spacing: 0) {
+                Text(model.displayCode)
+                    .styleGuide(.bodyMonospaced, weight: .regular, monoSpacedDigit: true)
+                    .foregroundColor(Asset.Colors.textPrimary.swiftUIColor)
+                if remaining <= Constants.nextTOTPCodePreviewThreshold, let nextCode, store.state.showNextTOTPCode {
```
fedemkr: 🤔 I think we could have this logic in `state` if possible or is moving "remaining" to state/processor very cumbersome?

AuthenticatorShared/UI/Vault/Views/ItemListItemRow/ItemListItemRowView.swift
KatherineInCode: Good call on catching the duplication. There should be a way to reduce that.

AuthenticatorShared/UI/Vault/Views/ItemListItemRow/ItemListItemRowView.swift
KatherineInCode: The state currently doesn't do any sort of handling of timing; everything about what to display based on time remaining happens in the view layer. Instead of pulling that into the `state`, I'd rather find a way to use the same `TOTPCountdownTimer` logic to control both of them, in line with the previous comment to get rid of the `TimelineView`

fedemkr (approved):

PR #2540, opened 2026-04-13
[PM-31836] fix: Prevent double-resume crash on Fido2 credential continuations

matt-livefront:
## 🎟️ Tracking

[PM-31836](https://bitwarden.atlassian.net/browse/PM-31836)

## 📔 Objective

Fixes a continuation double-resume crash if an error occurs while adding a passkey.

> Crash: Fatal error: SWIFT TASK CONTINUATION MISUSE: checkUserAndPickCredentialForCreation(options:newCredential:) tried to resume its continuation more than once, throwing MakeCredential(BitwardenSdk.MakeCredentialError.Other(message: "make_credential error: Ctap2(Vendor(VendorError(241)))"))!

When saving a new passkey, the SDK calls `checkUserAndPickCredentialForCreation`/`pickCredentialForAuthentication`, which suspends on a continuation waiting for the user to pick a cipher. Once the user picks, the continuation is resumed with `.success`. The SDK then proceeds and, in some cases (e.g. API error due to enterprise personal ownership policy violation), fails. The catch block in `handleFido2CredentialCreation`/`handleFido2CredentialAutofill` then attempted to resume the already-completed continuation with `.failure`, triggering Swift's fatal continuation misuse error.

This fixes the issue by removing the potential second resume call, adding an error alert, and clearing the continuations after they have been resumed.

[PM-31836]: https://bitwarden.atlassian.net/browse/PM-31836?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/UI/Vault/Vault/AutofillList/VaultAutofillListProcessor.swift
+++ b/BitwardenShared/UI/Vault/Vault/AutofillList/VaultAutofillListProcessor.swift
@@ -637,8 +637,8 @@ extension VaultAutofillListProcessor {
 
             autofillAppExtensionDelegate.completeAssertionRequest(assertionCredential: assertionCredential)
         } catch {
-            services.fido2UserInterfaceHelper.pickedCredentialForAuthentication(result: .failure(error))
             services.errorReporter.log(error: error)
+            await coordinator.showErrorAlert(error: error)
         }
     }
 
@@ -696,8 +696,8 @@ extension VaultAutofillListProcessor {
                 return
             }
 
-            services.fido2UserInterfaceHelper.pickedCredentialForCreation(result: .failure(error))
             services.errorReporter.log(error: error)
+            await coordinator.showErrorAlert(error: error)
         }
     }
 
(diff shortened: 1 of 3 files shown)


andrebispo5 (approved):
👍

BitwardenShared/UI/Vault/Vault/AutofillList/VaultAutofillListProcessor.swift
```diff
@@ -637,8 +637,8 @@ extension VaultAutofillListProcessor {
 
             autofillAppExtensionDelegate.completeAssertionRequest(assertionCredential: assertionCredential)
         } catch {
-            services.fido2UserInterfaceHelper.pickedCredentialForAuthentication(result: .failure(error))
```
fedemkr, line 640: 🤔 @matt-livefront If there is an error early in the process of `provideFido2Credential` then the continuation won't end in failure, so it will keep waiting indefinitely or am I missing something? I know it will be replaced by the new continuation once `pickCredentialForAuthentication` is called again but not sure if that could end up in some kind of race condition as the previous one was not completed.

matt-livefront: If there was an error early in the process of `provideFido2Credential`, wouldn't that occur before the continuation was created? So then there'd be no continuation to resume, right?

PR #2543, opened 2026-04-14
[PM-31836] fix: Handle no default collection case when saving Fido2 credential as new login

matt-livefront:
## 🎟️ Tracking

[PM-31836](https://bitwarden.atlassian.net/browse/PM-31836)

## 📔 Objective

Fixes a case where a passkey would fail to save if the personal ownership disabled policy is enabled and there's no default collection (My Items is disabled). In this case, the passkey is being saved to the organization, but the user must select a collection to save the passkey to, or saving will fail.

This case can be detected, and instead of saving the passkey automatically, the add item screen is shown so that a collection can be selected.

## 📸 Screenshots

| Before | After |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/0e887b1f-937e-4722-bed9-1d213bf966b3" /> | <video src="https://github.com/user-attachments/assets/aa310d6b-71e6-4573-8271-79088e2cd5ca" /> | 

[PM-31836]: https://bitwarden.atlassian.net/browse/PM-31836?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/Core/Vault/Helpers/CipherOwnershipHelper.swift
+++ b/BitwardenShared/Core/Vault/Helpers/CipherOwnershipHelper.swift
@@ -5,11 +5,11 @@ import BitwardenSdk
 
 /// Errors thrown by `CipherOwnershipHelper`.
 enum CipherOwnershipHelperError: Error, Equatable {
-    /// No eligible organization is available for cipher ownership when personal ownership is disabled.
-    case noEligibleOrganization
-
     /// No default collection is available for the organization when personal ownership is disabled.
     case noDefaultCollection
+
+    /// No eligible organization is available for cipher ownership when personal ownership is disabled.
+    case noEligibleOrganization
 }
 
 // MARK: - CipherOwnershipHelper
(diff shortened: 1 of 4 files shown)


BitwardenShared/Core/Vault/Helpers/CipherOwnershipHelper.swift
```diff
@@ -7,6 +7,9 @@ import BitwardenSdk
 enum CipherOwnershipHelperError: Error, Equatable {
     /// No eligible organization is available for cipher ownership when personal ownership is disabled.
     case noEligibleOrganization
+
+    /// No default collection is available for the organization when personal ownership is disabled.
+    case noDefaultCollection
```
andrebispo5, line 9: 🎨 Alphabetise

KatherineInCode (approved):

Issue #2528, opened 2026-04-09
[BWA-245] Bitwarden Auth still uses old color palette here and there


speedy1804:
### Steps To Reproduce

1. Install the bitwarden authenticator app
2. enable dark mode
3. navigate through different screens

### Expected Result

Have the same color(palette) throughout the whole app. (Match design with the bitwarden password manager app)

### Actual Result

There are some leftovers from the native app release in 2024. After that, bitwarden had several redesigns but at some points, the authenticator app still looks like the 2024 version of bitwarden.

### Screenshots or Videos

<img width="1194" height="834" alt="Image" src="https://github.com/user-attachments/assets/6e9b50f5-5b7b-4356-a612-6ca8cdc352c8" />
<img width="1194" height="834" alt="Image" src="https://github.com/user-attachments/assets/2b557d5b-90ed-43ef-8c63-651c344af652" />
<img width="1194" height="834" alt="Image" src="https://github.com/user-attachments/assets/2aae8f31-0eb2-477f-89e5-b9cd6f5f22a7" />
<img width="1194" height="834" alt="Image" src="https://github.com/user-attachments/assets/2c4ae3c4-baec-4557-b693-149ad1322d53" />

### Additional Context

The grey doesn’t fit in bitwarden’s current design language. Also I believe it is not supposed to still be in the app.

### Build Version

2026.3.1

### What server are you connecting to?

US

### Environment Details

tested on:
iPhone 16 Pro, running latest iOS 26.4.1
iPad Pro 11“ with M2 Chip, running iPadOS 26.4

### Issue Tracking Info

- [x] I understand that work is tracked outside of Github. A PR will be linked to this issue should one be opened to address it, but Bitwarden doesn't use fields like "assigned", "milestone", or "project" to track progress.

speedy1804:
© Bitwarden Inc. 2015–2026

📝 Authenticator 2026.3.1 (545)
📦 Bundle: com.bitwarden.authenticator
📱 Device: iPad14,3
🍏 System: iPadOS 26.4
🦀 SDK: 2.0.0-4735-26e2b10
🌩️ Server: 2026.3.2 @ US
🧱 Commit: bitwarden/ios/release/2026.3-rc45@216866bde2e09054f4083035da0b8e83242975ea
💻 Build Source: bitwarden/ios/actions/runs/23357101095/attempts/1

BBGhub:
Hi there,

Thank you for your report! 

[...]

PR #2537, opened 2026-04-13
[PM-35040] fix: Prevent Authenticator lock overlay from being dismissed

KatherineInCode:
## 🎟️ Tracking

[PM-35040](https://bitwarden.atlassian.net/browse/PM-35040)

## 📔 Objective

Fixes [VULN-512](https://bitwarden.atlassian.net/browse/VULN-512) (CVSS 6.8): the Authenticator app's lock overlay could be interactively swipe-dismissed after a failed or cancelled Face ID prompt, bypassing authentication entirely.

**Root cause**: `AppCoordinator.showAuth()` presented the lock navigation controller without `isModalInPresentation = true`, allowing iOS to interactively dismiss it. The Password Manager app already had this protection. Fix sets `isModalInPresentation = true` and `modalPresentationStyle = .fullScreen` on the auth navigation controller.

[PM-35040]: https://bitwarden.atlassian.net/browse/PM-35040?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[VULN-512]: https://bitwarden.atlassian.net/browse/VULN-512?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2551, opened 2026-04-16
[PM-35251] fix: Adjust presentation style to avoid Tab Bar issues in BWA

KatherineInCode:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35251

## 📔 Objective

This adjusts how the BWA auth screen presents to be `.overFullScreen` instead of `.fullScreen`, which fixes the color issue with the tab bar. The ultimate reason for this is that `.fullScreen` doesn't properly render things about the tab bar, causing issues with rendering on dismissal; with `.overFullScreen`, the system can render the tab bar appropriately before the dismissal.


matt-livefront (approved):

speedy1804:
Hi, does this pr fix #2528 ?

KatherineInCode:
@speedy1804 No; there's a broader effort to be done towards unifying components and colors between the apps. This is handling a particular bug that came up as a result of #2537

speedy1804:
@KatherineInCode oh okay, thanks for clarifying! 👍

PR #2507, opened 2026-03-30
[PM-34009] chore: Create KeychainServiceFacade

KatherineInCode:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34009

## 📔 Objective

This consolidates logic that was spread between `BitwardenShared.KeychainRepository`, `AuthenticatorShared.KeychainRepository`, and `AuthenticatorBridgeKit.KeychainStorage` into one `KeychainServiceFacade` object, and updates the `KeychainRepository` objects accordingly. In the process, I also tried to write new tests in Swift Testing, and convert a lot of our bespoke mock objects into AutoMockable, and therefore also needed to update tests.

This effectively updates the architecture from this:
```mermaid
flowchart TD
    BSKRAT[getAccessToken]
    BSKRNL[getNeverLock]
    BSKRG[getValue]

    subgraph BitwardenShared.KeychainRepository
    direction TD
    BSKRAT-->BSKRG
    BSKRNL-->BSKRG
    end

    ASKRSK[getSecretKey]
    ASKRUAK[getUserAccessKey]
    ASKRG[getValue]
    subgraph AuthenticatorShared.KeychainRepository
    direction TD
    ASKRSK-->ASKRG
    ASKRUAK-->ASKRG
    end
    
    subgraph AuthenticatorBridgeKit.KeychainRepository
    direction TD
    ABKAK[getAuthenticatorKey]
    ABKALT[getAutoLogoutTime]
    end
    subgraph AuthenticatorBridgeKit.KeychainStorage
    ABKG[getValue]
    ABKAK-->ABKG
    ABKALT-->ABKG
    end

    subgraph BitwardenKit.KeychainService
    BKKSS[search]
    BSKRG-->BKKSS
    ASKRG-->BKKSS
    ABKG-->BKKSS
    end
```

To this:
```mermaid
flowchart TD
    subgraph BitwardenShared.KeychainRepository
    direction TD
    BSKRAT[getAccessToken]
    BSKRNL[getNeverLock]
    end

    subgraph AuthenticatorShared.KeychainRepository
    direction TD
    ASKRSK[getSecretKey]
    ASKRUAK[getUserAccessKey]
    end
    
    subgraph AuthenticatorBridgeKit.KeychainRepository
    direction TD
    ABKAK[getAuthenticatorKey]
    ABKALT[getAutoLogoutTime]
    end

    subgraph BitwardenKit.KeychainServiceFacade
    BKKSWG[getValue]
[...]

PR #2553, opened 2026-04-16
[PM-35285] fix: Prevent StateService.isAuthenticated(userId:) from throwing if token doesn't exist in keychain

matt-livefront:
## 🎟️ Tracking

[PM-35285](https://bitwarden.atlassian.net/browse/PM-35285)

## 📔 Objective

The keychain changes in #2507 changed the behavior of which error is thrown if an item isn't found in the keychain. `StateService.isAuthenticated(userId:)` was catching `KeychainServiceError.osStatusError(errSecItemNotFound)`, but not `KeychainServiceError.keyNotFound`. This resulted in soft-logged out accounts appearing as if they were still logged in. When attempting to log back into a soft-logged out account, the app would ask if you want to switch to the already logged in account which would land you on an empty vault unlock screen.

I also removed the catch block from `AuthService.getPendingAdminLoginRequest(userId:)` since the error is already caught in the `KeychainRepository`.

https://github.com/bitwarden/ios/blob/c8ba32b6280bfb3aae5570b5630206765ee5cbf3/BitwardenShared/Core/Auth/Services/KeychainRepository.swift#L392-L399

## 📸 Screenshots

| Before | After |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/b92c8a88-ad6a-4ea2-8037-ce200ab25773" /> | <video src="https://github.com/user-attachments/assets/4ab84175-7888-4c5d-9fd2-599e632b9444" /> | 

[PM-35285]: https://bitwarden.atlassian.net/browse/PM-35285?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/Core/Auth/Services/AuthService.swift
+++ b/BitwardenShared/Core/Auth/Services/AuthService.swift
@@ -446,15 +446,11 @@ class DefaultAuthService: AuthService { // swiftlint:disable:this type_body_leng
 
     func getPendingAdminLoginRequest(userId: String?) async throws -> PendingAdminLoginRequest? {
         let activeUserId = try await stateService.getAccountIdOrActiveId(userId: userId)
-        do {
-            if let jsonString = try await keychainRepository.getPendingAdminLoginRequest(userId: activeUserId),
-               let jsonData = jsonString.data(using: .utf8) {
-                return try JSONDecoder().decode(PendingAdminLoginRequest.self, from: jsonData)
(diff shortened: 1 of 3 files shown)


andrebispo5 (approved):
👍

PR #2555, opened 2026-04-17
[PM-35100] feat: Load subscription data into premium plan screen

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35100

## 📔 Objective

- Load subscription data from BillingService and display it on the premium plan screen
- Introduce PremiumSubscription domain model to decouple the UI from the API response
- Add PlanCadenceType label for billing cadence display (monthly/annually)
- Derive all display labels (billing amount, storage cost, discount, estimated tax, dates) as computed properties in PremiumPlanState
- Show loading overlay while fetching subscription data
- Hide the plan section in settings for self-hosted users
- Add missing localization strings (PremiumPlanNotAvailable, Days, PerYear, EstimatedTax)

## 📸 Screenshots

https://github.com/user-attachments/assets/ded62cc1-c7ae-44e0-a3ef-2e6272eafd6f


BitwardenShared/Core/Billing/Services/BillingService.swift
```diff
@@ -62,7 +62,66 @@ class DefaultBillingService: BillingService {
...
+            return cost * discount.value / 100
+        }
+    }
+
+    /// Maps a subscription API response to a `PremiumSubscription` domain model.
+    ///
+    /// - Parameter response: The API response model.
+    /// - Returns: A flattened `PremiumSubscription`.
+    ///
+    private func mapSubscription(_ response: BitwardenSubscriptionResponseModel) -> PremiumSubscription {
```
matt-livefront: 🎨 Swift tends to prefer using initializers instead of functions when mapping between types. We've used this pattern elsewhere in the app too. Does it make sense to move this to an initializer on `PremiumSubscription`? I like that this keeps the logic closer to the thing being initialized, makes it reusable, and usually easier to test.

BitwardenShared/Core/Billing/Services/BillingService.swift
```diff
@@ -62,7 +62,66 @@ class DefaultBillingService: BillingService {
...
+        let seatDiscount = discountAmount(seats?.discount, on: seatsCost)
+        let storageDiscount = discountAmount(storage?.discount, on: storageCost)
+        let cartDiscount = discountAmount(response.cart.discount, on: itemTotal)
+
+        let status: PremiumPlanStatus = switch response.status {
+        case "canceled": .canceled
+        case "past_due": .pastDue
+        case "unpaid": .updatePayment
+        default: .active
+        }
```
matt-livefront: 🎨 Does it make sense to use `PremiumPlanStatus` or create a similar Codable enum for decoding `response.status` from the API, rather than decoding to a string and then comparing the strings here?

BitwardenShared/UI/Billing/PremiumPlan/PremiumPlanState.swift
```diff
@@ -7,53 +7,100 @@ import Foundation
...
+        guard let gracePeriod = subscription?.gracePeriod else { return "" }
+        return "\(gracePeriod)"
+    }
+
+    /// The next charge amount with currency, formatted for display (e.g. "$24.35 USD").
+    var nextChargeAmount: String {
+        guard let subscription, subscription.nextCharge != nil else { return "" }
+        let total = max(
+            0,
+            subscription.seatsCost + subscription.storageCost + subscription.estimatedTax - subscription.discount,
```
matt-livefront: 🎨 What do you think about calculating this in a computed property of `PremiumSubscription`? I think it helps to keep all of the calculations together, just in case any of this changes, it's local to the type rather than being spread out in the app.

BitwardenShared/UI/Billing/PremiumPlan/PremiumPlanStateTests.swift
```diff
@@ -1,47 +1,159 @@
...
 // MARK: - PremiumPlanStateTests
 
 struct PremiumPlanStateTests {
+    // MARK: Properties
+
+    /// A date used for testing: April 2, 2026 at 12:00 UTC.
+    private let testDate = Date(timeIntervalSince1970: 1_775_304_000)
+
+    /// A helper to build a subscription with the given parameters.
+    private func makeSubscription(
```
matt-livefront: 🎨 It might be worth turning this into a fixture, similar to other types in the app. `PremiumPlanProcessorTests` also initializes a few `PremiumSubscription` so a fixture with default values might help to simplify using this type across tests and previews.

BitwardenShared/UI/Billing/PremiumPlan/PremiumPlanView.swift
```diff
@@ -49,18 +49,30 @@ struct PremiumPlanView: View {
...
+            if store.state.showStorageCost {
+                Divider()
+                billingRow(
+                    label: Localizations.storageCost,
+                    value: store.state.storageCostLabel,
+                    valueColor: Color(asset: SharedAsset.Colors.textPrimary),
+                )
+            }
+            if store.state.showDiscount {
+                Divider()
```
matt-livefront: ❓ Does it work to use a `ContentBlock` here instead of the `VStack`? I think you'd get the `Divider()`s for free instead of having to add each one.

BitwardenShared/UI/Billing/PremiumPlan/PremiumPlanView.swift
```diff
@@ -170,16 +182,30 @@ struct PremiumPlanView: View {
 // MARK: - Previews
 
 #if DEBUG
+private extension PremiumSubscription {
+    static let previewActive = PremiumSubscription(
```
matt-livefront, line 181: 🎨 This could be in the fixtures extension too for reuse.

BitwardenResources/Localizations/en.lproj/Localizable.strings
```diff
@@ -1043,13 +1043,16 @@
 "BillingAmount" = "Billing amount";
 "StorageCost" = "Storage cost";
 "Discount" = "Discount";
+"EstimatedTax" = "Estimated tax";
 "ManagePlan" = "Manage plan";
 "CancelPremium" = "Cancel Premium";
 "YourNextChargeIsForXDueOnY" = "Your next charge is for **%1$@**, due on **%2$@**.";
 "WeCouldNotProcessYourPaymentUpdateYourPaymentMethodDescriptionLong" = "We could not process your payment. Update your payment method before your subscription ends on **%1$@**.";
-"YouHaveAGracePeriodOfXFromYourSubscriptionDescriptionLong" = "You have a grace period of **%1$@** from your subscription expiration date. Please resolve the past due invoices by **%2$@**.";
+"YouHaveAGracePeriodOfXFromYourSubscriptionDescriptionLong" = "You have a grace period of **%1$@ days** from your subscription expiration date. Please resolve the past due invoices by **%2$@**.";
```
matt-livefront: ⛏️ Should the key be updated to include `Days` since the value was changed?

KatherineInCode: This feels like the sort of thing that probably should go in the stringsdict, if we're dealing with pluralization?

BitwardenShared/UI/Billing/PremiumPlan/PremiumPlanView.swift
andrebispo5: I don't think so, we cannot access the fixtures from here.

BitwardenResources/Localizations/en.lproj/Localizable.stringsdict
```diff
@@ -189,6 +189,23 @@
 			<string>%d weeks</string>
 		</dict>
 	</dict>
+	<!-- A message indicating the grace period for a past due subscription, with pluralized days and the date. -->
+	<key>YouHaveAGracePeriodOfXDaysFromYourSubscriptionDescriptionLong</key>
+	<dict>
+		<key>NSStringLocalizedFormatKey</key>
+		<string>You have a grace period of **%1$#@GracePeriodDays@** from your subscription expiration date. Please resolve the past due invoices by **%2$@**.</string>
```
KatherineInCode: 🤔 I guess if you don't specify a specific key for `%2$@`, it defaults to taking in a string?

andrebispo5: I'm not sure but it should have a value for it. Right now with the plurals it is showing like so: 
image

KatherineInCode: Neat!

fedemkr: 🤔 Shouldn't the keys be whole sentences so they can be pluralized correctly in some languages that have more rules? I saw that was the path taken with the other strings in this file.

BitwardenShared/Core/Billing/Models/Domain/PremiumSubscription.swift
```diff
@@ -0,0 +1,134 @@
...
+        let seatDiscount = Self.discountAmount(seats?.discount, on: seatsCost)
+        let storageDiscount = Self.discountAmount(storage?.discount, on: storageCost)
+        let cartDiscount = Self.discountAmount(response.cart.discount, on: itemTotal)
+
+        let status: PremiumPlanStatus = switch response.status {
+        case .active: .active
+        case .canceled: .canceled
+        case .pastDue: .pastDue
+        case .unpaid: .updatePayment
+        }
```
fedemkr: 🎨 IMO we should follow how we map other enums and just have an `init` in `PremiumPlanStatus` that takes the `SubscriptionStatus` to create it so we have the logic centralized there.

BitwardenShared/Core/Billing/Models/Enum/SubscriptionStatus.swift
```diff
@@ -0,0 +1,26 @@
...
+
+    /// The subscription is unpaid.
+    case unpaid
+
+    /// Decodes a `SubscriptionStatus` from the API, defaulting to `.active` for unknown values.
+    init(from decoder: Decoder) throws {
+        let container = try decoder.singleValueContainer()
+        let rawValue = try container.decode(String.self)
+        self = SubscriptionStatus(rawValue: rawValue) ?? .active
+    }
```
fedemkr: 🤔 Is it ok that `.active` is the default value when it's unknown? I mean if server adds a new status (let's say `paused`) which would be unknown to the app then this would take it as `.active` which could be confusing to the user as they may see something else on the web. Moreover, shouldn't we just add an `.unknown` value for ourselves so we could just display something on the UI to go to the web to check it out or update the app to the latest version?

BitwardenShared/Core/Billing/Services/BillingServiceTests.swift
```diff
@@ -139,31 +176,109 @@ struct BillingServiceTests {
...
+
+        let result = try await subject.getSubscription()
+
+        #expect(result.discount == 5)
+    }
+
+    /// `getSubscription()` maps past_due status correctly.
+    @Test
+    func getSubscription_pastDue() async throws {
+        billingAPIService.getSubscriptionReturnValue = BitwardenSubscriptionResponseModel(
```
fedemkr: ⛏️ As with other response models, it should have a `.fixture` to use here.

BitwardenShared/UI/Billing/PremiumPlan/PremiumPlanProcessorTests.swift
```diff
@@ -31,10 +38,114 @@ struct PremiumPlanProcessorTests {
 
     // MARK: Tests
 
-    /// `perform(_:)` with `.appeared` completes without error.
+    /// `perform(_:)` with `.appeared` loads subscription with canceled status.
     @Test
-    func perform_appeared() async {
+    func perform_appeared_canceled() async {
```
fedemkr: 🤔 Is it necessary to have this test and `perform_appeared_pastDue`? I mean the test that checks whether the subscription and status got assigned on the state is `perform_appeared_success`, so which additional scenarios are these tests adding?

BitwardenShared/UI/Billing/PremiumPlan/PremiumPlanState.swift
```diff
@@ -7,53 +7,90 @@ import Foundation
...
-    /// The billing amount label (e.g. "$1.65 / month").
-    var billingAmount: String = "$1.65 / month"
+    /// A cached formatter for US dollar currency strings.
+    private static let usdCurrencyFormatter: NumberFormatter = {
+        let formatter = NumberFormatter()
+        formatter.numberStyle = .currency
+        formatter.currencyCode = "USD"
+        formatter.locale = Locale(identifier: "en_US")
+        return formatter
+    }()
```
fedemkr: 🎨 I would put this formatter configuration in an extension so it's more direct to use and can be reuse in other places if necessary.
```swift
  extension NumberFormatter {
      // docs
      static let usdCurrency: NumberFormatter = {
          let formatter = NumberFormatter()
          formatter.numberStyle = .currency
          formatter.currencyCode = "USD"
          formatter.locale = Locale(identifier: "en_US")
          return formatter
      }()
  }
```
Then you don't need this property and can just call `NumberFormatter.usdCurrency`.

BitwardenShared/UI/Billing/PremiumPlan/PremiumPlanState.swift
```diff
@@ -7,53 +7,90 @@ import Foundation
...
-                "May 2, 2026",
+                subscriptionEndDate,
             )
         }
     }
 
+    /// The discount label (e.g. "-$0.10").
+    var discount: String {
+        guard let subscription, subscription.discount > 0 else { return "" }
+        return "-\(formatCurrency(subscription.discount))"
```
fedemkr: 🤔 Are we sure that all languages represent discounts with `-` prefix? Otherwise, that should be localized as well.

BitwardenShared/UI/Billing/PremiumPlan/PremiumPlanState.swift
```diff
@@ -63,4 +100,55 @@ struct PremiumPlanState: Equatable {
...
+
+    // MARK: Private Methods
+
+    /// Formats a decimal price as a US dollar currency string.
+    ///
+    /// - Parameter price: The price to format.
+    /// - Returns: A formatted currency string (e.g. "$1.65").
+    ///
+    private func formatCurrency(_ price: Decimal) -> String {
+        Self.usdCurrencyFormatter.string(from: price as NSDecimalNumber) ?? "$0.00"
```
fedemkr: ⚠️  I don't think `0.00` should be the default here as e.g. the user may see that they have 0.00 estimated taxes because of an error in the formatter. Shouldn't that return some other string like `unknown` or `error` or something more meaningful for the user?

BitwardenShared/UI/Billing/PremiumPlan/PremiumPlanState.swift
```diff
@@ -7,53 +7,90 @@ import Foundation
...
     var urlToOpen: URL?
 
     // MARK: Computed Properties
 
-    // TODO: PM-35100 Replace mock values with real data from the API.
+    /// The billing amount label (e.g. "$19.80 / year").
+    var billingAmount: String {
+        guard let subscription else { return "" }
+        return "\(formatCurrency(subscription.seatsCost)) \(subscription.cadence.label)"
+    }
```
fedemkr, line 30: 🎨 I think the whole string should be localized as there are some languages that format this differently like  Japanese.

BitwardenShared/UI/Billing/PremiumPlan/PremiumPlanState.swift
andrebispo5: yes we want it to have the `-` prefix I'll add it to the localized strings.

BitwardenShared/Core/Billing/Extensions/NumberFormatter+Currency.swift
```diff
@@ -0,0 +1,12 @@
+import Foundation
+
+extension NumberFormatter {
+    /// A formatter for US dollar currency strings (e.g. "$19.80").
+    static let usdCurrency: NumberFormatter = {
+        let formatter = NumberFormatter()
+        formatter.numberStyle = .currency
+        formatter.currencyCode = "USD"
+        formatter.locale = Locale(identifier: "en_US")
```
fedemkr: 🤔 Is it necessary to force locale to use `en_US` instead of localizing based off the current locale the user has?

BitwardenShared/UI/Billing/PremiumPlan/PremiumPlanState.swift
fedemkr: ⚠️ `USD` shouldn't be hardcoded here. There could be some inconsistencies on some languages as you may get `24,35 $ USD` which would be weird. So I would go to **just show code** which you can set in the `NumberFormatter` instead of both the symbol and the code. So you get something like `24.35 USD` or `USD 24.35` depending on the locale.
Additionally, there are some languages that do not display "USD" and go with a translation so I'd just leave the work for the `NumberFormatter` to apply the right localization to the code.
Check this with Product by I think this should be the way to go.

BitwardenShared/UI/Billing/PremiumPlan/PremiumPlanProcessor.swift
```diff
@@ -46,18 +48,86 @@ final class PremiumPlanProcessor: StateProcessor<
...
+                message: Localizations.youllContinueToHavePremiumAccessUntilX(state.nextChargeDate),
+                alertActions: [
+                    AlertAction(title: Localizations.cancelNow, style: .destructive) { [weak self] _ in
+                        guard let self else { return }
+                        await openPortalUrl()
+                    },
+                    AlertAction(title: Localizations.close, style: .cancel),
+                ],
+            ),
+        )
```
fedemkr, line 98: ⛏️ We often put this in an `Alert+Something` extension file; but I can see a point on this not being reused, so leave it up to you on whether to move it.

fedemkr (dismissed):

fedemkr (approved):

PR #2559, opened 2026-04-20
[PM-35329] fix: Show certificate validation alerts via coordinator to prevent dialog race condition

matt-livefront:
## 🎟️ Tracking

[PM-35329](https://bitwarden.atlassian.net/browse/PM-35329)

## 📔 Objective

Fixes a race condition in SelfHostedProcessor where validation errors (empty alias or password) during certificate import were silently dismissed. The SwiftUI alert binding sends dialogDismiss whenever its isPresented binding transitions to false, including when state.dialog changes type, causing a newly set .error dialog to be immediately cleared. This updates the processor to display alerts via the coordinator, which matches the pattern for alerts elsewhere.

[PM-35329]: https://bitwarden.atlassian.net/browse/PM-35329?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenKit/UI/Platform/Application/Utilities/InputValidator/Validators/EmptyInputValidator.swift
+++ b/BitwardenKit/UI/Platform/Application/Utilities/InputValidator/Validators/EmptyInputValidator.swift
@@ -22,7 +22,7 @@ public struct EmptyInputValidator: InputValidator {
 
     // MARK: InputValidator
 
-    public func validate(input: String?) throws {
+    public func validate(input: String?) throws(InputValidationError) {
         guard input?.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty == false else {
             throw InputValidationError(message: Localizations.validationFieldRequired(fieldName))
         }
--- a/BitwardenKit/UI/Platform/Application/Utilities/InputValidator/InputValidator.swift
+++ b/BitwardenKit/UI/Platform/Application/Utilities/InputValidator/InputValidator.swift
@@ -7,5 +7,5 @@ public protocol InputValidator {
     ///
     /// - Parameter input: The input to validate.
     ///
-    func validate(input: String?) throws
+    func validate(input: String?) throws(InputValidationError)
 }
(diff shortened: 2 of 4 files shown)


BitwardenKit/UI/Platform/Application/Utilities/InputValidator/Validators/EmptyInputValidator.swift
```diff
@@ -22,7 +22,7 @@ public struct EmptyInputValidator: InputValidator {
 
     // MARK: InputValidator
 
-    public func validate(input: String?) throws {
+    public func validate(input: String?) throws(InputValidationError) {
```
fedemkr, line 25: 🤔 Should we "override" the signature of the protocol function with the specific validation error here?
I mean the `InputValidator` protocol has the function defined as:
```swift
func validate(input: String?) throws
```
whereas here it's being modified to add the specific error `throws(InputValidationError)`.
Not sure if this can get us some trouble in the future 🤔 .
Perhaps we should also update the protocol?

matt-livefront: The implementation has the ability to throw a more specific error as `throws(InputValidationError)` is a subtype of plain `throws`. See more here: https://github.com/swiftlang/swift-evolution/blob/main/proposals/0413-typed-throws.md#protocol-conformance

That said, I think it does make sense to have the protocol throw the typed error in this case, so I can update that. Good callout!

fedemkr (approved):

BitwardenKit/UI/Platform/Application/Utilities/InputValidator/Validators/EmptyInputValidator.swift
fedemkr: Oh nice I wasn't aware of that, thanks!

PR #2569, opened 2026-04-23
[PM-35657] llm: Add converting-mocks-to-automockable skill

KatherineInCode:
## 🎟️ Tracking
[PM-35657](https://bitwarden.atlassian.net/browse/PM-35657)

## 📔 Objective
Adds a new `converting-mocks-to-automockable` skill that guides Claude through assessing whether a bespoke hand-written mock can be replaced with a Sourcery-generated one, performing the migration, and updating affected tests.

The skill was developed iteratively by running real conversions (`MockSyncService`, `MockSettingsDataStore`, `MockKeyConnectorService`) and capturing patterns discovered during each pass:
- Implicit-unwrap default pitfall (bespoke defaults vs. generated `nil`)
- Checking production code for indirect method calls that would crash without a default
- `BitwardenSharedMocks` import requirement for test files
- Multi-parameter vs. single-parameter capture naming (`ReceivedArguments` vs. `Received{Label}`)
- Mid-test reset pattern (`fooCalled = false` → `fooCallsCount = 0`)
- Handler closure translation (`fooHandler: (() -> Void)?` → `fooClosure: ((ArgTypes) async throws -> Void)?`)
- Xcodegen step required after deleting a standalone mock file

[PM-35657]: https://bitwarden.atlassian.net/browse/PM-35657?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.claude/skills/converting-mocks-to-automockable/SKILL.md
```diff
@@ -0,0 +1,301 @@
...
+
+# Converting Bespoke Mocks to AutoMockable
+
+This skill guides you through assessing whether a hand-written mock can be replaced with a Sourcery-generated one, performing the migration, and updating affected tests.
+
+## Step 1: Read and Assess the Candidate
+
+Read the bespoke mock file and locate its corresponding protocol. Understand what the mock currently does.
+
+### Good candidates for AutoMockable
```
fedemkr: 🤔 Does it work by only leaving the "Don't" patterns thus removing this section? I mean if the skill is just to migrate existing mocks then I would only clarify what things can't be handled by Sourcery so any other stuff can be tried to be migrated. Moreover, the skill would be slightly slimmer.

.claude/skills/converting-mocks-to-automockable/SKILL.md
```diff
@@ -0,0 +1,301 @@
...
+### Poor candidates — keep as bespoke
+
+These patterns require custom behavior that AutoMockable can't express:
+
+| Pattern | Why it can't be auto-generated |
+|---------|-------------------------------|
+| Mock executes passed-in closure parameters (e.g., runs a `() async throws -> T` argument) | Closures passed as arguments are non-escaping; Sourcery stubs these with `fatalError` |
+| Mock accumulates all calls into a combined array (e.g., `var alerts: [Alert]`) | AutoMockable stores last call's args, not a merged collection across overloads |
+| Methods with overloaded signatures that share state | Each overload gets a separate mock; shared logic must be bespoke |
+| Mock is a class/struct, not a protocol implementation | AutoMockable only works on protocols |
```
fedemkr: ❓ Do you have an example of this? I'm not sure which of our bespoke mocks would qualify for this last statement as I would think that all of them are based in protocol implementations.

theMickster:
@KatherineInCode looks like the workflows are hung. Will you please push an empty commit to the branch?
That should hopefully shake things loose. 😃

KatherineInCode:
@theMickster Looks like workflows might still be having some issues, which is to be expected. We might have to wait a bit still for it all to come back online.

.claude/skills/converting-mocks-to-automockable/SKILL.md
KatherineInCode: Claude can't find an example, and I can't think of anything, so I'll drop the line.

.claude/skills/converting-mocks-to-automockable/SKILL.md
KatherineInCode: Looks like it does, so I've had Claude re-generate that section to be a bit more clear in that regard

.claude/skills/converting-mocks-to-automockable/SKILL.md
```diff
@@ -0,0 +1,302 @@
...
+var deleteItemCalled = false
+var deleteItemError: Error?
+func deleteItem() async throws {
+    deleteItemCalled = true
+    if let error = deleteItemError { throw error }
+}
+
+// Generated equivalent
+mockService.deleteItemThrowableError = someErr  // nil by default (no throw)
+// deleteItemCalled still works (computed from deleteItemCallsCount)
```
fedemkr: ⛏️ Is it necessary to clarify this? Claude should be able to infer that from the generated code.

.claude/skills/converting-mocks-to-automockable/SKILL.md
```diff
@@ -0,0 +1,302 @@
...
+// Before
+#expect(mockService.doThingName == "hello")
+#expect(mockService.doThingCount == 3)
+#expect(mockService.doThingEnabled == true)
+
+// After
+#expect(mockService.doThingReceivedArguments?.name == "hello")
+#expect(mockService.doThingReceivedArguments?.count == 3)
+#expect(mockService.doThingReceivedArguments?.enabled == true)
+```
```
fedemkr: 🤔 I think we could shrink this to have a slimmer skill by just focusing on the particular lines that change without having to have an example per testing tech.
```suggestion
```swift
// Before
mockService.doThingName == "hello"
mockService.doThingCount == 3
mockService.doThingEnabled == true

// After
mockService.doThingReceivedArguments?.name == "hello"
mockService.doThingReceivedArguments?.count == 3
mockService.doThingReceivedArguments?.enabled == true
```
```

File: .claude/skills/converting-mocks-to-automockable/SKILL.md [hunk 5]
```diff
@@ -0,0 +1,302 @@
...
+```swift
+// Bespoke
+var isEnabledResult: Bool = false
+func isEnabled() -> Bool { isEnabledResult }
+
+// Generated equivalent
+mockService.isEnabledReturnValue = false  // note: implicitly unwrapped optional, crashes if not set
+```
+
+Always set `ReturnValue` before the test exercises that method — it's `T!`, not `T?`.
```
fedemkr: 🤔 Do we need to clarify that it can use the closure as well instead? `isEnabledClosure`

.claude/skills/converting-mocks-to-automockable/SKILL.md
```diff
@@ -0,0 +1,302 @@
...
+
+## Step 6: Update Tests
+
+With the generated mock's actual property names in hand, update each test that used the bespoke mock. Translate according to the mapping in Step 3.
+
+Common things to change:
+
+- Replace `Result`-based setup with `ReturnValue`/`ThrowableError`
+- Replace explicit `fooCalled = false` resets (not needed; generated mocks start fresh)
+- Replace bespoke parameter capture properties with `Received{Label}` (single param) or `ReceivedArguments` tuple (multiple params)
```
fedemkr, line 259: ⛏️ I think we can remove this as it's already explained in Step 3

.claude/skills/converting-mocks-to-automockable/SKILL.md
```diff
@@ -0,0 +1,302 @@
...
+If you see `cannot find type 'MockFoo' in scope` after deleting a bespoke mock, a missing import is almost always the cause. Check every file that previously had access to the bespoke mock via co-location.
+
+## Step 7: Verify
+
+Build and run the affected tests:
+
+```bash
+mint run swiftformat .
+mint run swiftlint
+```
```
fedemkr: 🤔 Not sure how the bash commands are related to "build and run affected tests", isn't it missing the build and run commands? or were you meaning to check format/lint and also build and run?

SaintPatrck (approved):

theMickster (approved):

PR #2574, opened 2026-04-24
[PM-35101] feat: Complete Stripe premium checkout flow

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-35101

## 📔 Objective

Implements the complete Stripe premium checkout flow for iOS:
- Adds `PremiumCheckoutStatus` and `SubscriptionStatus` enums
- Extends `BillingService` with checkout status publishing, `premiumCheckoutCanceled()`, and `premiumStatusChanged()`
- Adds billing alerts (payment not received, checkout failed, upgrade pending)
- Routes the `premiumStatusChanged` push notification through `BillingService` instead of a plain sync
- Handles the `bitwarden://premium-checkout-result` deep link in `AppProcessor`
- Subscribes to checkout status in `PremiumUpgradeProcessor` (loading overlay, cancel/retry alert, dismiss on confirm)
- Adds an "Upgraded to Premium" action card to the vault list and streams checkout status to update premium state

## 📸 Screenshots
IMG_6329 Large
IMG_6327 Large
IMG_6328 Large

https://github.com/user-attachments/assets/3ab41030-f3f2-4c1d-9caf-8283d4697448


BitwardenShared/Core/Billing/Services/BillingService.swift
```diff
@@ -82,4 +122,34 @@ class DefaultBillingService: BillingService {
         let response = try await billingAPIService.getSubscription()
         return PremiumSubscription(response: response)
     }
+
+    func premiumCheckoutCanceled() {
+        premiumCheckoutStatusSubject.send(.canceled)
+        premiumCheckoutStatusSubject.send(nil)
+    }
```
fedemkr, line 141: 🤔 Why do you need to send `nil` after canceled?

BitwardenShared/Core/Billing/Services/BillingService.swift
```diff
@@ -82,4 +122,34 @@ class DefaultBillingService: BillingService {
...
+
+    func premiumCheckoutStatusPublisher() -> AnyPublisher<PremiumCheckoutStatus, Never> {
+        premiumCheckoutStatusSubject
+            .compactMap(\.self)
+            .debounce(for: .milliseconds(300), scheduler: DispatchQueue.main)
```
fedemkr, line 146: 🤔 Are you adding the debounce in case several status are sent in a row and you only want to publish the last one?

BitwardenShared/Core/Billing/Services/BillingService.swift
```diff
@@ -82,4 +134,37 @@ class DefaultBillingService: BillingService {
...
+        do {
+            try await syncService.fetchSync(forceSync: true)
+        } catch {
+            errorReporter.log(error: error)
+        }
+        let hasPremium = await stateService.doesActiveAccountHavePremium()
+        premiumCheckoutStatusSubject.send(hasPremium ? .confirmed : .pending)
+        if hasPremium {
+            premiumCheckoutStatusSubject.send(nil)
+        }
```
fedemkr, line 168: 🤔 Why is it needed to send `nil` if you have already sent `confirmed`?

BitwardenShared/UI/Billing/PremiumUpgrade/PremiumUpgradeProcessor.swift
```diff
@@ -70,18 +79,48 @@ final class PremiumUpgradeProcessor: StateProcessor<
 
     // MARK: Private Methods
 
+    /// Subscribes to premium checkout status updates and reacts accordingly.
+    ///
+    private func subscribeToPremiumCheckoutStatus() {
+        premiumStatusChangedCancellable = services.billingService
+            .premiumCheckoutStatusPublisher()
+            .receive(on: DispatchQueue.main)
+            .sink { [weak self] status in
```
fedemkr, line 110: 🤔 Why aren't you using `for await` here instead of this sink subscription? Additionally, if the user goes back (away from this screen) after it's subscribed, is the subscription canceled or remains in memory?

BitwardenShared/UI/Platform/Application/AppProcessor.swift
```diff
@@ -431,6 +435,27 @@ extension AppProcessor {
         }
     }
 
+    /// Handles the premium checkout result deep link.
+    ///
+    /// - Parameter url: The URL to check.
+    /// - Returns: `true` if the URL was a premium checkout result and was handled.
+    ///
+    private func handlePremiumCheckoutResult(url: URL) async -> Bool {
+        guard url.scheme == "bitwarden", url.host == "premium-checkout-result" else {
```
fedemkr: 🎨 You can use `isBitwardenAppScheme` String extension for the scheme and I would move the host to the constants file.

BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessorTests.swift
```diff
@@ -680,6 +685,96 @@ class VaultListProcessorTests: BitwardenTestCase { // swiftlint:disable:this typ
...
+    func test_subscribeToPremiumCheckoutStatus_canceled() async throws {
+        let statusSubject = PassthroughSubject<PremiumCheckoutStatus, Never>()
+        billingService.premiumCheckoutStatusPublisherReturnValue = statusSubject.eraseToAnyPublisher()
+        subject.receive(.upgradeToPremium)
+        let routeCountBeforeSend = coordinator.routes.count
+
+        statusSubject.send(.canceled)
+
+        try await Task.sleep(nanoseconds: 100_000_000)
+        XCTAssertEqual(coordinator.routes.count, routeCountBeforeSend)
```
fedemkr: ⛏️ Use waitForAsync instead.

BitwardenShared/UI/Vault/Vault/VaultCoordinator.swift
```diff
@@ -3,6 +3,18 @@ import BitwardenResources
...
+
+/// A context object used to pass a completion handler through `coordinator.navigate(to: .dismiss)`.
+///
+class DismissCompletionContext: NSObject {
+    let completion: () -> Void
+
+    init(completion: @escaping () -> Void) {
+        self.completion = completion
+    }
+}
```
fedemkr: 🤔 Isn't this alike `DismissAction`? Or how is this different?

BitwardenShared/Core/Billing/Services/BillingService.swift
andrebispo5: yup exactly.

BitwardenShared/Core/Billing/Services/BillingService.swift
andrebispo5: `private let premiumCheckoutStatusSubject = CurrentValueSubject<PremiumCheckoutStatus?, Never>(nil)` saves the last value passed for new subscribers. when we get to the `.canceled` state we want to remove it from being propagated again so we set it at `nil`. We can also send them in sequence because the `.compactMap(\.self)` will clear the `nil` before the debounce.

BitwardenShared/Core/Billing/Services/BillingService.swift
andrebispo5: same explanation has above in this case `.confirmed` is an end state.

BitwardenShared/UI/Billing/PremiumUpgrade/PremiumUpgradeProcessor.swift
andrebispo5: Using this method is easier to cancel the subscription without having to explicitly handling tasks.

Regarding the cancellation:

> It will be cancelled and released:
>   1. premiumStatusChangedCancellable is a property on the processor                                                                                                                                                                                 
>   2. When the processor deallocates, that property deallocates with it                                                                                                                                                                              
>   3. AnyCancellable automatically calls cancel() on deallocation → subscription is torn down                                                                                                                                                        
>                                                                                                                                                                                                                                                     
>   The [weak self] is the key safety piece. Without it:                                                                                                                                                                                              
>                                                                                                                                                                                                                                                     
>   processor → cancellable → sink closure → strong self (processor)                                                                                                                                                                                  
>                                                                                                                                                                                                                                                     
>   That's a retain cycle — the processor would keep itself alive and never deallocate.                                                                                                                                                               
>                                                                                                                                                                                                                                                     
>   With [weak self]:                                                                                                                                                                                                                                 
>   processor → cancellable → sink closure → weak ref (processor)                                                                                                                                                                                     
>                                                                                                                                                                                                                                                     
>   No cycle. Processor can deallocate freely, taking the cancellable with it.

BitwardenShared/UI/Vault/Vault/VaultCoordinator.swift
andrebispo5: of course we can... 🤦

BitwardenShared/Core/Billing/Services/BillingService.swift
fedemkr: And can't you use [PassthroughSubject](https://developer.apple.com/documentation/combine/passthroughsubject) instead? which doesn't hold the last value and just propagates only to already subscribed subscribers.

BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessorTests.swift
```diff
@@ -712,18 +729,20 @@ class VaultListProcessorTests: BitwardenTestCase { // swiftlint:disable:this typ
 
         statusSubject.send(.confirmed)
 
-        try await waitForAsync { self.coordinator.contexts.last is DismissCompletionContext }
-        let context = try XCTUnwrap(coordinator.contexts.last as? DismissCompletionContext)
-        XCTAssertEqual(coordinator.routes.last, .dismiss)
-        context.completion()
+        try await waitForAsync {
+            guard case .dismiss(let action) = self.coordinator.routes.last else { return false }
```
fedemkr: ⛏️ We usually use `guard case let` instead of the `let` inside.

BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessorTests.swift
```diff
@@ -732,16 +751,18 @@ class VaultListProcessorTests: BitwardenTestCase { // swiftlint:disable:this typ
 
         statusSubject.send(.pending)
 
-        try await waitForAsync { self.coordinator.contexts.last is DismissCompletionContext }
-        let context = try XCTUnwrap(coordinator.contexts.last as? DismissCompletionContext)
-        XCTAssertEqual(coordinator.routes.last, .dismiss)
-        context.completion()
+        try await waitForAsync {
+            guard case .dismiss(let action) = self.coordinator.routes.last else { return false }
```
fedemkr: ⛏️ We usually use `guard case let` instead of the `let` inside.

BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessorTests.swift
```diff
@@ -750,10 +771,12 @@ class VaultListProcessorTests: BitwardenTestCase { // swiftlint:disable:this typ
 
         statusSubject.send(.syncing)
 
-        try await waitForAsync { self.coordinator.contexts.last is DismissCompletionContext }
-        let context = try XCTUnwrap(coordinator.contexts.last as? DismissCompletionContext)
-        XCTAssertEqual(coordinator.routes.last, .dismiss)
-        context.completion()
+        try await waitForAsync {
+            guard case .dismiss(let action) = self.coordinator.routes.last else { return false }
```
fedemkr: ⛏️ We usually use `guard case let` instead of the `let` inside.

fedemkr (commented):
Looks good, just some minor ⛏️

fedemkr (approved):

PR #2583, opened 2026-04-24
[PM-38507] feat: Add Binding<Bool> initializer to ExpandableHeaderView

SaintPatrck:
## 🎟️ Tracking

[PM-38507](https://bitwarden.atlassian.net/browse/PM-38507)

## 📔 Objective

`ExpandableHeaderView` stores its expand/collapse state in an internal `@State`, so the preference resets whenever the view is recreated. The upcoming BWPM vault-list section headers need that preference to survive app relaunch, which requires caller-owned storage.

This adds a second initializer accepting `isExpanded: Binding<Bool>` so callers can drive expansion from their own store. The existing `(title:count:content:)` initializer is unchanged, so Authenticator's call site is untouched.

[PM-38507]: https://bitwarden.atlassian.net/browse/PM-38507?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenKit/UI/Platform/Application/Views/ExpandableHeaderView+ViewInspectorTests.swift
+++ b/BitwardenKit/UI/Platform/Application/Views/ExpandableHeaderView+ViewInspectorTests.swift
@@ -1,0 +1,69 @@
+// swiftlint:disable:this file_name
+import SwiftUI
+import ViewInspector
+import ViewInspectorTestHelpers
+import XCTest
+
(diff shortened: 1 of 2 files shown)


morganzellers-bw (approved):

SaintPatrck:
Thanks @morganzellers-bw

PR #2586, opened 2026-04-27
[PM-31181] fix: Toast displays after view/edit cipher actions

fedemkr:
## 🎟️ Tracking
[PM-31181](https://bitwarden.atlassian.net/browse/PM-31181)

## 📔 Objective
When a user performs an action (archive, soft delete, permanent delete, restore, unarchive) on the view or edit cipher screen, no toast was displayed after the screen dismissed.

**Root cause**: All `DismissAction` completion blocks in `ViewItemProcessor` and `AddEditItemProcessor` captured `self` weakly via `[weak self]`. SwiftUI tears down the view hierarchy (and deallocates the processor) before UIKit fires the dismiss-animation completion, so `self` is nil by the time the block runs — meaning `CipherItemOperationDelegate` callbacks are never called and no toast appears. The triple-dot menu was unaffected because it calls the toast handler directly without a dismiss completion.

**Fix**: Capture `delegate` as a strong local `let` immediately before navigating to dismiss, so the `DismissAction` closure no longer needs `self` at all. No retain cycle is introduced — the processor does not own the `DismissAction`.

Additionally, the previously unused `performOperationAndDismiss` helper was corrected (changed `onDismiss` from `(ViewItemProcessor) -> Void` to `() -> Void`) and is now used by `permanentDeleteItem`, `softDeleteItem`, and `restoreItem` to remove duplicated loading-overlay boilerplate.

[PM-31181]: https://bitwarden.atlassian.net/browse/PM-31181?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenShared/UI/Vault/VaultItem/AddEditItem/AddEditItemProcessor.swift
```diff
@@ -282,12 +282,11 @@ final class AddEditItemProcessor: StateProcessor<// swiftlint:disable:this type_
...
     ///
     private func archiveItem() async {
+        let delegate = delegate
         await vaultItemActionHelper.archive(cipher: state.cipher) { [weak self] url in
             self?.state.url = url
         } completionHandler: { [weak self] in
-            self?.dismiss { [weak self] in
-                self?.delegate?.itemArchived()
-            }
+            self?.dismiss { delegate?.itemArchived() }
```
matt-livefront: 🎨 Instead of capturing delegate in a local variable, you can capture it in the closure's capture list. It will have the same effect, but I think it's slightly cleaner and conveys the intention better.

```suggestion
        await vaultItemActionHelper.archive(cipher: state.cipher) { [weak self] url in
            self?.state.url = url
        } completionHandler: { [weak self, delegate] in
            self?.dismiss { delegate?.itemArchived() }
```

matt-livefront (approved):

PR #2589, opened 2026-04-28
[PM-30129] chore: Remove `pm-19148-innovation-archive` feature flag

matt-livefront:
## 🎟️ Tracking

[PM-30129](https://bitwarden.atlassian.net/browse/PM-30129)

## 📔 Objective

Removes the `pm-19148-innovation-archive` feature flag.

[PM-30129]: https://bitwarden.atlassian.net/browse/PM-30129?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenShared/UI/Vault/VaultItem/CipherItemState.swift
```diff
@@ -282,7 +279,7 @@ struct CipherItemState: Equatable { // swiftlint:disable:this type_body_length
 
     /// Whether the item should be displayed as archived.
     var shouldDisplayAsArchived: Bool {
```
KatherineInCode, line 281: 🤔 I wonder if now that this is a passthrough we can just eliminate it?

KatherineInCode (commented):
Looks good, just one small thing

BitwardenShared/UI/Vault/VaultItem/CipherItemState.swift
matt-livefront: From a code standpoint, I agree that this isn't providing much value if it's just a passthrough. From the UI side though, I feel like `shouldDisplayAsArchived` is clearer than using `canBeUnarchived` for determining whether to display if the cipher is archived or not. Whether something is archived vs whether something can be unarchived are the same now, but they could diverge.

KatherineInCode: Reasonable

KatherineInCode (approved):

PR #2591, opened 2026-04-28
[PM-33161] feat: Local User Data Key updates

morganzellers-bw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33161

## 📔 Objective

Register against the LocalUserDataKeyState in clients.

## 📸 Screenshots

https://github.com/user-attachments/assets/6b254d01-f08f-44fb-be58-62f925413844


AuthenticatorShared/Core/Platform/Services/Stores/AppSettingsStore.swift
```diff
@@ -84,6 +84,13 @@ protocol AppSettingsStore: AnyObject {
     ///
     func lastActiveTime(userId: String) -> Date?
 
+    /// Gets the local user data key states for the user ID.
+    ///
+    /// - Parameter userId: The user ID associated with the key states.
+    /// - Returns: A dictionary mapping key ID to encrypted wrapped key string.
+    ///
+    func localUserDataKeyStates(userId: String) -> [String: String]?
```
matt-livefront: 🎨 Can you add tests for these changes?

BitwardenShared/Core/Platform/Services/StateService.swift
```diff
@@ -1853,6 +1853,7 @@ actor DefaultStateService: StateService, ActiveAccountStateProvider, ConfigState
         appSettingsStore.setLastSyncTime(nil, userId: knownUserId)
         appSettingsStore.setMasterPasswordHash(nil, userId: knownUserId)
         appSettingsStore.setPasswordGenerationOptions(nil, userId: knownUserId)
+        appSettingsStore.setLocalUserDataKeyStates(nil, userId: knownUserId)
```
matt-livefront: ⛏️ Looks like we aren't entirely consistent, but if you wouldn't mind, could you alphabetize this among the other `appSettingsStore.set*` calls here?

🎨 Could you also add a check that we clear this data in the tests?

BitwardenShared/Core/Vault/Repositories/SdkLocalUserDataKeyStateRepositoryTests.swift
```diff
@@ -0,0 +1,136 @@
+import BitwardenSdk
+import XCTest
```
matt-livefront: 🎨 Do we want to convert this to use Swift Testing since it's a new test?

BitwardenShared/Core/Vault/Repositories/SdkLocalUserDataKeyStateRepository.swift
```diff
@@ -0,0 +1,74 @@
...
+    func set(id: String, value: LocalUserDataKeyState) async throws {
+        var states = appSettingsStore.localUserDataKeyStates(userId: userId) ?? [:]
+        states[id] = value.wrappedKey
+        appSettingsStore.setLocalUserDataKeyStates(states, userId: userId)
+    }
+
+    func setBulk(values: [String: LocalUserDataKeyState]) async throws {
+        var states = appSettingsStore.localUserDataKeyStates(userId: userId) ?? [:]
+        for (id, state) in values {
+            states[id] = state.wrappedKey
```
matt-livefront: 🤔 One downside of just storing `wrappedKey` is that if the SDK were to add another property to `LocalUserDataKeyState`, we'd have to perform a migration of the existing stored data. I wonder if it's worth matching the SDK here and storing a dictionary of keys which map to a struct. We probably can't use `LocalUserDataKeyState` directly since it doesn't conform to Codable, but we could create a local mirror and persist that. That would give us a bit more flexibility in the future if needed.

BitwardenShared/Core/Vault/Repositories/SdkLocalUserDataKeyStateRepository.swift
```diff
@@ -0,0 +1,74 @@
...
+    init(appSettingsStore: AppSettingsStore, userId: String) {
+        self.appSettingsStore = appSettingsStore
+        self.userId = userId
+    }
+
+    // MARK: LocalUserDataKeyStateRepository
+
+    func get(id: String) async throws -> LocalUserDataKeyState? {
+        appSettingsStore.localUserDataKeyStates(userId: userId)?[id]
+            .map { LocalUserDataKeyState(wrappedKey: $0) }
```
matt-livefront: 🤔 Usually we wouldn't access `appSettingsStore` directly and instead go through `StateService`. This repository though looks very similar to `StateService` though if we were exposing this to the rest of the app as opposed to the SDK. @fedemkr / @KatherineInCode what do you think about this? Also, do we need to be storing these keys in the keychain?

KatherineInCode: I kinda feel like we should still have it go through a StateService, just to cut down on places where other things touch AppSettingsStore, for consistency. Though I also wonder if it would make sense to break out into a separate `SdkLocalStateService` protocol for interface segregation reasons (and likewise with the AppSettingsStore or KeychainService accordingly)

morganzellers-bw: I saw some items in `appSettingsStore`  that were also encrypted, but I too wondered about the keychain -  IIRC, the sdk keeps the key used to encrypt these keys, but keychain would be more safe

morganzellers-bw: Converted `SdkLocalUserDataKeyStateRepository` in both apps to use `StateService`  in commit [ca3c0e0](https://github.com/bitwarden/ios/pull/2591/commits/ca3c0e02aaed1b95b9d57de406c2995addd89f41) 👍

AuthenticatorShared/Core/Platform/Services/ClientService.swift
```diff
@@ -225,7 +235,26 @@ actor DefaultClientService: ClientService {
...
+        return newClient
+    }
+
+    /// Configures a new SDK client.
+    /// - Parameters:
+    ///   - client: The SDK client to configure.
+    ///   - userId: The user ID the SDK client instance belongs to.
+    func configureNewClient(_ client: BitwardenSdkClient, for userId: String) async {
+        client.platform().state().registerClientManagedRepositories(
+            repositories: sdkRepositoryFactory.makeCipherRepositories(userId: userId),
```
matt-livefront: 🤔 I wonder if I makes sense to make all of these changes to Authenticator, given that I think the only thing we use the SDK for in Authenticator is generating TOTP codes. Can you see if anything breaks in authenticator if we don't provide a `LocalUserDataKeyStateRepository`?

BitwardenShared/Core/Platform/Services/Stores/AppSettingsStore.swift
```diff
@@ -215,6 +215,13 @@ protocol AppSettingsStore: AnyObject {
     ///
     func lastSyncTime(userId: String) -> Date?
 
+    /// Gets the local user data key states for the user ID.
+    ///
+    /// - Parameter userId: The user ID associated with the key states.
+    /// - Returns: A dictionary mapping key ID to encrypted wrapped key string.
```
matt-livefront: ⛏️ Can you update this since the key ID maps to `UserKeyData`?

BitwardenShared/Core/Platform/Services/Stores/AppSettingsStore.swift
```diff
@@ -438,6 +445,14 @@ protocol AppSettingsStore: AnyObject {
     ///
     func setLastSyncTime(_ date: Date?, userId: String)
 
+    /// Sets the local user data key states for a user ID.
+    ///
+    /// - Parameters:
+    ///   - states: A dictionary mapping key ID to encrypted wrapped key string, or `nil` to clear.
```
matt-livefront: ⛏️ Can you update this since the key ID maps to `UserKeyData`?

BitwardenShared/Core/Platform/Services/StateService.swift
```diff
@@ -2052,6 +2102,38 @@ actor DefaultStateService: StateService, ActiveAccountStateProvider, ConfigState
         appSettingsStore.learnGeneratorActionCardStatus = status
     }
 
+    func removeLocalUserDataKeyState(id: String, userId: String) async {
+        var states = appSettingsStore.localUserDataKeyStates(userId: userId) ?? [:]
+        states.removeValue(forKey: id)
+        appSettingsStore.setLocalUserDataKeyStates(states.isEmpty ? nil : states, userId: userId)
```
matt-livefront: 🎨 You can use the `nilIfEmpty` helper to skip the ternary here:
```suggestion
        appSettingsStore.setLocalUserDataKeyStates(states.nilIfEmpty, userId: userId)
```

BitwardenShared/Core/Platform/Services/StateService.swift
```diff
@@ -2052,6 +2102,38 @@ actor DefaultStateService: StateService, ActiveAccountStateProvider, ConfigState
...
+    func removeAllLocalUserDataKeyStates(userId: String) async {
+        appSettingsStore.setLocalUserDataKeyStates(nil, userId: userId)
+    }
+
+    func removeBulkLocalUserDataKeyStates(keys: [String], userId: String) async {
+        var states = appSettingsStore.localUserDataKeyStates(userId: userId) ?? [:]
+        for key in keys {
+            states.removeValue(forKey: key)
+        }
+        appSettingsStore.setLocalUserDataKeyStates(states.isEmpty ? nil : states, userId: userId)
```
matt-livefront: 🎨 
```suggestion
        appSettingsStore.setLocalUserDataKeyStates(states.nilIfEmpty, userId: userId)
```

BitwardenShared/Core/Platform/Services/StateService.swift
```diff
@@ -244,6 +244,12 @@ protocol StateService: AnyObject {
     ///
     func getLearnGeneratorActionCardStatus() async -> AccountSetupProgress?
 
+    /// Gets the local user data keys for the user ID
+    ///
+    /// - Returns: A dictionary of user keys.
+    ///
+    func getLocalUserDataKeyStates(userId: String) async -> [String: UserKeyData]?
```
matt-livefront: 🎨 We've been trying to create more specific protocols that are used in place of the full `StateService` protocol, but ultimately implemented by `DefaultStateService`, just to limit scope of what other services have access to, rather than the full `StateService`. Does it make sense to put all of these new methods in a `LocalUserDataStateService` protocol? See `UserSessionStateService` as an example.

BitwardenShared/Core/Vault/Models/Data/UserKeyData.swift
```diff
@@ -0,0 +1,14 @@
+import BitwardenSdk
+import Foundation
+
+public struct UserKeyData: Codable, Equatable {
+    let wrappedKey: EncString
```
matt-livefront, line 10: ⛏️ Can you add docs for this?

BitwardenShared/Core/Platform/Services/TestHelpers/MockStateService.swift
```diff
@@ -58,6 +58,7 @@ class MockStateService: StateService, ActiveAccountStateProvider, AutofillStateS
     var isInitialSyncRequiredByUserId = [String: Bool]()
     var learnGeneratorActionCardStatus: AccountSetupProgress?
     var learnNewLoginActionCardStatus: AccountSetupProgress?
+    var localUserDataKeyStatesByUserId: [String: [String: UserKeyData]?] = [:]
```
matt-livefront: ❓ Does the dictionary here need to be an optional? Is there a distinction between storing `nil` vs the key not existing?
```suggestion
    var localUserDataKeyStatesByUserId: [String: [String: UserKeyData]] = [:]
```

AuthenticatorShared/Core/Platform/Services/ClientService.swift
morganzellers-bw: You know, that's a good point, we probably do not need them. I'll check on `sdlc/sdk-update`.

AuthenticatorShared/Core/Platform/Models/Domain/UserKeyData.swift
```diff
@@ -0,0 +1,14 @@
...
+    let wrappedKey: EncString
+
+    init(wrappedKey: EncString) {
+        self.wrappedKey = wrappedKey
+    }
+
+    init(localUserDataKeyState: LocalUserDataKeyState) {
+        self.init(wrappedKey: localUserDataKeyState.wrappedKey)
+    }
+}
```
fedemkr: ⛏️ Add DocC comments.

AuthenticatorShared/Core/Platform/Services/Stores/AppSettingsStore.swift
```diff
@@ -138,6 +145,14 @@ protocol AppSettingsStore: AnyObject {
     ///
     func setLastActiveTime(_ date: Date?, userId: String)
 
+    /// Sets the local user data key states for a user ID.
+    ///
+    /// - Parameters:
+    ///   - states: A dictionary mapping key ID to encrypted wrapped key string, or `nil` to clear.
+    ///   - userId: The user ID associated with the key states.
+    ///
+    func setLocalUserDataKeyStates(_ states: [String: UserKeyData]?, userId: String)
```
fedemkr: 🤔 Is it ok/safe to store the `UserKeyData` in the UserDefaults or should it be placed in the keychain? I see it's a wrapped `EncString` but double checking just in case.

AuthenticatorShared/Core/Platform/Services/TestHelpers/MockStateService.swift
```diff
@@ -76,6 +81,38 @@ class MockStateService: StateService {
...
+    func removeAllLocalUserDataKeyStates(userId: String) async {
+        localUserDataKeyStatesByUserId.updateValue(nil, forKey: userId)
+    }
+
+    func removeBulkLocalUserDataKeyStates(keys: [String], userId: String) async {
+        var current = (localUserDataKeyStatesByUserId[userId] ?? nil) ?? [:]
+        for key in keys {
+            current.removeValue(forKey: key)
+        }
+        localUserDataKeyStatesByUserId[userId] = current.isEmpty ? nil : current
```
fedemkr: ⛏️ You can use `nilIfEmpty` here.

AuthenticatorShared/Core/Platform/Services/TestHelpers/MockStateService.swift
```diff
@@ -76,6 +81,38 @@ class MockStateService: StateService {
         flightRecorderData = data
     }
 
+    func removeLocalUserDataKeyState(id: String, userId: String) async {
+        var current = (localUserDataKeyStatesByUserId[userId] ?? nil) ?? [:]
+        current.removeValue(forKey: id)
+        localUserDataKeyStatesByUserId[userId] = current.isEmpty ? nil : current
```
fedemkr: ⛏️  You can use `nilIfEmpty` here.

AuthenticatorShared/Core/Platform/Services/TestHelpers/MockClientBuilder.swift
fedemkr, line 1: 🤔 Could we move this to `BitwardenKit` so we have the same mock for the client builder both in PM and BWA? Or are they somehow different? I see the code is pretty much the same as in PM.

AuthenticatorShared/Core/Platform/Services/StateService.swift
```diff
@@ -240,6 +289,38 @@ actor DefaultStateService:
         appSettingsStore.flightRecorderData = data
     }
 
+    func removeLocalUserDataKeyState(id: String, userId: String) async {
+        var states = appSettingsStore.localUserDataKeyStates(userId: userId) ?? [:]
+        states.removeValue(forKey: id)
+        appSettingsStore.setLocalUserDataKeyStates(states.isEmpty ? nil : states, userId: userId)
```
fedemkr: ⛏️ You can use `nillIfEmpty`

fedemkr (commented):
🎨 I see that `SdkLocalUserDataKeyStateRepository` is the same or pretty much the same between PM and BWA implementations and the same happens with the `StateService` and `AppSettingsStore` parts needed for such repository.
Therefore in order to share more code between the apps, could we move the repository to `BitwardenKit` and also create some protocols for the `StateService` and `AppSettingsStore` it can use? e.g. `LocalUserDataKeyStateService` and `LocalUserDataKeyAppSettingsStore` that then each service/store of each app conforms to.
Moreover, the `BitwardenKit`'s `LocalUserDataKeyStateService ` protocol could have a default implementation that has a get-only property of `LocalUserDataKeyAppSettingsStore ` protocol so you can add the whole state service implementation for this feature there so it all lives in  `BitwardenKit`, thus then in the PM/BWA `StateService` all you have to do is conform to the protocol and return the implementation of `LocalUserDataKeyAppSettingsStore ` of each app without duplicating all this service implementation code.
What do you think?

AuthenticatorShared/Core/Platform/Services/ClientService.swift
morganzellers-bw: Yeah, BWA builds, works, & tests pass without `LocalUserDataKeyStateRepository` so I'll remove those changes.

morganzellers-bw:
> 🎨 I see that `SdkLocalUserDataKeyStateRepository` is the same or pretty much the same between PM and BWA implementations and the same happens with the `StateService` and `AppSettingsStore` parts needed for such repository. Therefore in order to share more code between the apps, could we move the repository to `BitwardenKit` and also create some protocols for the `StateService` and `AppSettingsStore` it can use? e.g. `LocalUserDataKeyStateService` and `LocalUserDataKeyAppSettingsStore` that then each service/store of each app conforms to. Moreover, the `BitwardenKit`'s `LocalUserDataKeyStateService ` protocol could have a default implementation that has a get-only property of `LocalUserDataKeyAppSettingsStore ` protocol so you can add the whole state service implementation for this feature there so it all lives in `BitwardenKit`, thus then in the PM/BWA `StateService` all you have to do is conform to the protocol and return the implementation of `LocalUserDataKeyAppSettingsStore ` of each app without duplicating all this service implementation code. What do you think?

That sounds great! I'm thinking BWA may not actually need these changes for LocalUserDataKey, but if it does I'll make the updates there too. 

Thanks for explaining this, because I had a feeling there was a way I could only make one version of these types and share between both apps.

BitwardenShared/Core/Platform/Services/StateService.swift
morganzellers-bw: Added `LocalUserDataStateService` in commit [89f300b](https://github.com/bitwarden/ios/pull/2591/commits/89f300bf8a75a10010378eabe0f5bc30cf22ff24) 👍

BitwardenShared/Core/Platform/Services/Stores/TestHelpers/MockAppSettingsStore.swift
```diff
@@ -6,7 +6,7 @@ import Foundation
 
 // swiftlint:disable file_length
 
-class MockAppSettingsStore: AppSettingsStore { // swiftlint:disable:this type_body_length
+class MockAppSettingsStore: AppSettingsStore, LocalUserDataKeyAppSettingsStore { // swiftlint:disable:this type_body_length line_length
```
KatherineInCode: 🤔 In practice, if something needs a `LocalUserDataKeyAppSettingsStore`, you should pass in an Automockable `MockLocalUserDataKeyAppSettingsStore` rather than a `MockAppSettingsStore`, so I don't think this needs to conform to the protocol.

BitwardenShared/Core/Platform/Services/LocalUserDataStateService.swift
```diff
@@ -0,0 +1,92 @@
+import Foundation
+
+/// A service that provides state management functionality for `UserKeyData`.
+protocol LocalUserDataStateService {
+    var localUserDataKeyStore: LocalUserDataKeyAppSettingsStore? { get }
```
matt-livefront: 🤔 This feels like an implementation detail that shouldn't be exposed by the protocol. It also gives any consumer of this protocol direct access to the store, bypassing the protocol methods altogether. Would it be better to implement the extension below as an extension on `DefaultStateService`, rather than the protocol itself?

morganzellers-bw: Yes, I realized I didn't need this in the protocol after taking a look at `ServerCommunicationConfigStateService` ; I didn't realize at first that the extension was on `DefaultStateService` there.

AuthenticatorShared/Core/Platform/Services/Stores/AppSettingsStore.swift
morganzellers-bw: Refactored to use the keychain just to be safe

morganzellers-bw:
Update:

Now stores user data keys in the keychain rather than User Defaults. 

Updated login video:
https://github.com/user-attachments/assets/3b4a69cc-c8c0-421d-8a0b-8b33125a4666

BitwardenShared/Core/Vault/Repositories/SdkLocalUserDataKeyStateRepositoryTests.swift
morganzellers-bw: Not sure how I missed these, but converting them now

BitwardenShared/Core/Auth/Services/KeychainRepository.swift
```diff
@@ -132,7 +139,7 @@ enum BitwardenKeychainItem: Equatable, KeychainItem {
 
 // MARK: - KeychainRepository
 
-protocol KeychainRepository: AnyObject, ServerCommunicationConfigKeychainRepository { // sourcery: AutoMockable
+protocol KeychainRepository: AnyObject, ServerCommunicationConfigKeychainRepository, LocalUserDataKeychainRepository { // sourcery: AutoMockable
```
fedemkr, line 143: ⛏️ I think this has a warning of line too long.

Sourcery/Templates/AutoMockable.stencil
```diff
@@ -101,9 +101,9 @@ import {{ import }}
...
     {% endif %}
     {% elif not method.parameters.count == 0 and not hasNonEscapingClosures %}
-    {% call accessLevel method.accessLevel %}{% call staticSpecifier method %}var {% call swiftifyMethodName method %}ReceivedArguments: ({% for param in method.parameters %}{{ param.name }}: {% if param.typeAttributes.escaping %}{% call existentialClosureTupleVariableTypeName param.typeName.unwrappedTypeName param.isVariadic false %}{% else %}{% call existentialClosureTupleVariableTypeName param.typeName param.isVariadic false %}{% endif %}{{ ', ' if not forloop.last }}{% endfor %})?
+    {% call accessLevel method.accessLevel %}{% call staticSpecifier method %}var {% call swiftifyMethodName method %}ReceivedArguments: ({% for param in method.parameters %}{{ param.name }}: {% if param.typeAttributes.escaping %}{% call existentialClosureTupleVariableTypeName param.typeName.unwrappedTypeName param.isVariadic true %}{% else %}{% call existentialClosureTupleVariableTypeName param.typeName param.isVariadic false %}{% endif %}{{ ', ' if not forloop.last }}{% endfor %})?
```
matt-livefront, line 104: ❓ What does this end up changing in the generated mocks?

BitwardenShared/Core/Vault/Repositories/SdkRepositoryFactory.swift
```diff
@@ -18,6 +16,8 @@ protocol SdkRepositoryFactory { // sourcery: AutoMockable
 struct DefaultSdkRepositoryFactory: SdkRepositoryFactory {
     // MARK: Properties
 
+    /// The service for managing account state.
+    private let stateService: LocalUserDataStateService
```
matt-livefront: ⛏️ Can you alphabetize this?

BitwardenShared/Core/Platform/Services/StateServiceTests.swift
```diff
@@ -1385,12 +1399,12 @@ class StateServiceTests: BitwardenTestCase { // swiftlint:disable:this type_body
...
 
-        keychainRepository.getAccessTokenThrowableError = KeychainServiceError.keyNotFound(BitwardenKeychainItem.accessToken(userId: "1"))
+        keychainRepository?.getAccessTokenThrowableError = KeychainServiceError.keyNotFound(BitwardenKeychainItem.accessToken(userId: "1"))
         var authenticationState = try await subject.isAuthenticated()
         XCTAssertFalse(authenticationState)
 
-        keychainRepository.getAccessTokenThrowableError = nil
-        keychainRepository.getAccessTokenReturnValue = "ACCESS_TOKEN"
+        keychainRepository?.getAccessTokenThrowableError = nil
+        keychainRepository?.getAccessTokenReturnValue = "ACCESS_TOKEN"
```
matt-livefront: ❓ Any reason these`keychainRepository` references use optional chaining now? It doesn't look like it's an optional.

BitwardenShared/Core/Auth/Services/LocalUserDataKeychainRepository.swift
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    func setLocalUserDataKeyStates(_ states: [String: UserKeyData]?, userId: String) async throws {
+        if let states {
+            try await keychainServiceFacade.setValue(
+                states,
+                for: BitwardenKeychainItem.localUserDataKeyStates(userId: userId),
+            )
+        } else {
+            try await keychainServiceFacade.deleteValue(
```
matt-livefront: 🤔 It looks like both clear and mutate operations are serialized, should set be as well? Maybe that causes issues though since set is called from the clear and mutate functions in the serializer... 🤔

BitwardenShared/Core/Platform/Services/StateServiceTests.swift
morganzellers-bw: ~~Yeah I'm not sure why either. Maybe I applied a fix on some stale errors or something. I also thought that I had removed these but maybe it was a different file.~~

Dug a little more:

The truly optional ones are the `keychainRepository` uses that are inside the closures, but it looks like I did a select next on `keychainRepository` when I saw the error that they needed the optional chaining, and all of the ones in the new tests got the `?`.

Since `keychainRepository` is an implicitly-unwrapped optional, it's optional at runtime when its value snapshot is taken to pass into the closure with `[keychainRepository]`.  So I'll update this to remove ?'s from the ones typed as `MockKeychainRepository!` outside of the closures

Sourcery/Templates/AutoMockable.stencil
morganzellers-bw: It was for an inout dictionary here, but I will double check it's actually required.

```
func mutateLocalUserDataKeyStates(
        userId: String,
        _ transform: @escaping @Sendable (inout [String: UserKeyData]) -> Void,
    ) async throws
```

BitwardenShared/Core/Auth/Utilities/LocalUserDataMutationSerializer.swift
```diff
@@ -0,0 +1,20 @@
+import Foundation
+
+/// Serializes concurrent mutations to local user data per user ID using task chaining.
+/// Each mutation awaits the previous one before reading from keychain, eliminating read/write race
+/// conditions that would otherwise occur across actor suspension points.
+actor LocalUserDataMutationSerializer {
+    private var pendingMutationsByUserId: [String: Task<Void, Error>] = [:]
```
fedemkr: ⛏️ Add docs

BitwardenShared/Core/Auth/Services/LocalUserDataKeychainRepository.swift
morganzellers-bw: What Claude specifically flagged as being unsafe about these operations before was the `read -> modify -> write` pattern, where we were `await`ing on both `keychainServiceFacade` calls to `read` and `write`.  Essentially allowing a gap for data to be overwritten.

So I believe the set doesn't need serialized because it just writes what it's given as one operation without there being a chance of something happening in the time between two `await` calls.

BitwardenShared/Core/Auth/Utilities/LocalUserDataMutationSerializer.swift
```diff
@@ -0,0 +1,20 @@
...
+
+    /// Serializes `operation` after any in-progress mutation for `userId`.
+    func serialize(userId: String, operation: @escaping @Sendable () async throws -> Void) async throws {
+        let previous = pendingMutationsByUserId[userId]
+        let task = Task<Void, Error> {
+            // ensure previous mutation is finished
+            _ = try? await previous?.value
```
fedemkr: 🤔 Shouldn't it log the error if any occur while waiting here or are all operations error handling done outside? Because here it will just swallow it.

BitwardenShared/Core/Auth/Utilities/LocalUserDataMutationSerializer.swift
```diff
@@ -0,0 +1,20 @@
+import Foundation
+
+/// Serializes concurrent mutations to local user data per user ID using task chaining.
+/// Each mutation awaits the previous one before reading from keychain, eliminating read/write race
+/// conditions that would otherwise occur across actor suspension points.
+actor LocalUserDataMutationSerializer {
```
fedemkr: 🎨 I don't see anything in here that is specific for `LocalUserData`. We could just call this `SerialWorker` or something like that. I would think we could be  able to benefit from this on other places as well.
⛏️ Also, I would use `enqueue` instead of `serialize` to avoid any confusion on the function name but not a very strong opinion.

Sourcery/Templates/AutoMockable.stencil
morganzellers-bw: Yeah looks like Sourcery got tripped trying to make the inout closure on the tuple 

### Differences in mock:

###### before change:
```swift
var mutateLocalUserDataKeyStatesReceivedArguments: (userId: String, transform: ([String: UserKeyData]) -> Void)?

func mutateLocalUserDataKeyStates(userId: String, _ transform: @Sendable @escaping (inout [String: UserKeyData]) -> Void) async throws {
    ...
    mutateLocalUserDataKeyStatesReceivedArguments = (userId: userId, transform: transform)
    ...
}
```

Error message:
```
Cannot assign value of type 

'(userId: String, transform: @Sendable (inout [String : UserKeyData]) -> Void)' 
to type 
'(userId: String, transform: ([String : UserKeyData]) -> Void)'
```

###### after change:
```swift
var mutateLocalUserDataKeyStatesReceivedArguments: (userId: String, transform: (inout [String: UserKeyData]) -> Void)?

func mutateLocalUserDataKeyStates(userId: String, _ transform: @Sendable @escaping (inout [String: UserKeyData]) -> Void) async throws {
    ...
    mutateLocalUserDataKeyStatesReceivedArguments = (userId: userId, transform: transform)
    ...
}
```

BitwardenShared/Core/Auth/Utilities/LocalUserDataMutationSerializer.swift
morganzellers-bw: Yeah this is essentially just a simple queue made with an `actor` so I'm all for making it super generic! I like `enqueue` as well.

BitwardenShared/Core/Auth/Services/LocalUserDataKeychainRepository.swift
matt-livefront: Could you still run into an issue if the set happens between the read -> modify -> write operations of the other methods?

- Thread 1: calls `mutateLocalUserDataKeyStates` and reads the data (A).
- Thread 2: calls `setLocalUserDataKeyStates(_:userId:)` which writes new data (B).
- Thread 1: mutates the original data (A) and writes it back.

Now the changes from `setLocalUserDataKeyStates` have been lost. if `setLocalUserDataKeyStates` was also serialized, it would happen after Thread 1 wrote the data back.

Looking at it more though, `setLocalUserDataKeyStates` isn't called from outside the keychain. I think you could remove that from the protocol and leave it as a private helper which would remove the issue.

morganzellers-bw: Oh yeah that would overwrite too; even better to just make it private.

BitwardenShared/Core/Auth/Utilities/LocalUserDataMutationSerializer.swift
morganzellers-bw: All of the callers will receive their error when their task is the "current" one. 

So for example if `previous` would throw, it does get swallowed here, but that is okay because when it was the current task, or `operation()`, it's caller got the error thrown to it then.

Additionally, swallowing here ensures `operation()` is always run when `previous` finishes even if `previous` throws.

matt-livefront (approved):
Looks good!

BitwardenShared/Core/Auth/Utilities/SerialWorker.swift
```diff
@@ -0,0 +1,21 @@
+import Foundation
+
+/// Serializes concurrent operations per user ID using task chaining.
+/// Each operation awaits the previous one before running, eliminating race
+/// conditions that would otherwise occur across actor suspension points such as `await`.
+actor SerialWorker {
```
fedemkr, line 6: ♻️ @morganzellers-bw thanks for changing this. Also I think we could move this directly to `BitwardenKit` as it's something we could use anywhere and it's not specific for `Auth`.

morganzellers-bw: I like it; I've got a TODO this week to make a PR to move it over 👍

morganzellers-bw: @fedemkr I added PR #2669 to do this 👍

PR #2599, opened 2026-04-30
[PM-28227] fix: Keep keyboard alive in autofillText mode via SwiftUI @FocusState anchor

### LINKED ISSUE #2138: [PM-28227] Long press autofill in iOS closes randomly
### Steps To Reproduce

1. Go to any text field to autofill, for example with a 2fa code;
2. Long-press on the text field;
3. Select autofill-password
4. When Bitwarden opens after a few seconds it closes automatically, although I'm looking for the correct entry

### Expected Result

It must not close.

### Actual Result

It closes after a few seconds. Not always, it rarely stays open properly. I can't understand if there's a precise reason why it can stay open.

### Screenshots or Videos

https://github.com/user-attachments/assets/b7a99e8d-173a-4c33-baf1-5f25f59e17bd

### Additional Context

_No response_

### Build Version

2025.10.1 (2691)

EU

### Self-host Server Version

_No response_

### Environment Details

iPhone 16
iOS 18.7.2

### Issue Tracking Info

fedemkr:
## 🎟️ Tracking
[PM-28227](https://bitwarden.atlassian.net/browse/PM-28227)

## 📔 Objective
Fixes the `autofillText` mode keyboard being dismissed (and the extension closing ~5 seconds later) when the user cancels an inline vault search.

In `autofillText` mode the extension is shown as a keyboard panel (InputUI) backed by a Remote Text Input (RTI) session. Any text field that becomes first responder via UIKit's explicit `becomeFirstResponder()` causes the RTI system to tag the session with `delayEndInputSession:YES` (`fromBecomeFirstResponder:1`), starting a ~5-second countdown that tears down the keyboard panel and dismisses the extension when it fires.

The previous approach used a UIKit `KeyboardAnchorTextField` and called `becomeFirstResponder()` on it when the search Cancel button was tapped — triggering exactly that timer. This PR replaces that path with a hidden SwiftUI `TextField` bound to a `@FocusState` enum (`FocusableField.anchor / .search`). SwiftUI's internal focus machinery uses `fromBecomeFirstResponder:0` → `delayEndInputSession:NO`, so every focus handoff (including Cancel → anchor) is a clean RTI session continuation with no countdown timer.

As a secondary fix, `.searchable` / `UISearchController` is replaced with a custom inline search bar in `autofillText` mode, since `UISearchController` also calls explicit `becomeFirstResponder()` on resign.

## 📸 Screenshots
N/A — keyboard extension panel behaviour, no visual UI change.

[PM-28227]: https://bitwarden.atlassian.net/browse/PM-28227?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenAutoFillExtension/CredentialProviderViewController.swift
```diff
@@ -470,26 +524,46 @@ extension CredentialProviderViewController: RootNavigator {
...
-            fromViewController.removeFromParent()
-        }
+// MARK: - KeyboardAnchorTextField
+
+/// A zero-frame text field used in `autofillText` mode to seed the initial RTI session before the
+/// SwiftUI view hierarchy is ready.
+///
+/// `useKeyboard` is computed as `allowsSystemInputView && !hasCustomInputView && responderRequiresKeyboard`.
+/// Any non-nil `inputView` makes `!hasCustomInputView = 0` → `useKeyboard = 0` → `delayEndInputSession:YES`
+/// on every delegate transition, firing `endRemoteTextInputSessionWithID` after ~5 seconds.
```
matt-livefront: ❓ Is this still a valid comment? I don't see `useKeyboard` declared anywhere? Some of the comments throughout this PR seem overly verbose, do you think these are all helpful for future context?

BitwardenAutoFillExtension/CredentialProviderViewController.swift
```diff
@@ -291,9 +316,34 @@ class CredentialProviderViewController: ASCredentialProviderViewController {
...
     override func prepareInterfaceForUserChoosingTextToInsert() {
+        // Create the anchor directly here — `self.context` is nil at this point so
+        // `installKeyboardAnchor`'s context guard would return early. This is the only
+        // synchronous call site where the view is in the window, no children exist, and
+        // nothing else can steal focus. `initializeApp` launches an async Task, so anchoring
+        // inside it would be too late.
+        let anchor = KeyboardAnchorTextField(frame: .zero)
+        view.addSubview(anchor)
+        keyboardAnchor = anchor
+        anchor.becomeFirstResponder()
```
matt-livefront: 🤔 This is very similar to `installKeyboardAnchor()`. I see why it's duplicated from the comment, but I wonder if we could simplify by:

1. Keep this logic here, which creates and installs `KeyboardAnchorTextField`.
2. In `show(child:)` the keyboard anchor should already exist at that point if you're in the autofill text context. So could we replace `installKeyboardAnchor()` with `keyboardAnchor?.becomeFirstResponder()` which will only cause it to become first responder if it has already been created in this flow here?
3. Remove `installKeyboardAnchor()`.

BitwardenAutoFillExtension/CredentialProviderViewController.swift
```diff
@@ -291,9 +316,34 @@ class CredentialProviderViewController: ASCredentialProviderViewController {
...
+
+        // Safety net: reclaim anchor when keyboard starts hiding and anchor isn't FR.
+        NotificationCenter.default.addObserver(
+            self,
+            selector: #selector(keyboardWillHide),
+            name: UIResponder.keyboardWillHideNotification,
+            object: nil,
+        )
```
matt-livefront: 🤔 Could we put this and `keyboardWillHide()` in `KeyboardAnchorTextField`? That would help keep this isolated to `KeyboardAnchorTextField` and hopefully easier to remove if the root issue is fixed.

matt-livefront (approved):

PR #2607, opened 2026-05-01
[PM-36251] feat: Navigate archive premium upsell through in-app upgrade flow

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-36251

## 📔 Objective

Wires the archive unavailable upsell CTA to the same in-app premium upgrade flow used by the vault list banner, instead of always opening the web vault URL.

- Renames `shouldShowPremiumUpgradeBanner` → `isPremiumUpgradeEligible` to better reflect that it checks user eligibility, not banner visibility.
- Extracts `isPremiumUpgradeBannerDismissed` as a separate `StateService` method so banner dismissal is a distinct concern.
- Adds `isInAppUpgradeAvailable()` and `navigateToPremiumUpgrade()` to `VaultListProcessor`, encapsulating the routing decision (in-app upgrade flow vs. web vault fallback) based on the feature flag and US storefront.
- The banner respects dismissal; the archive CTA bypasses it so a dismissed banner does not block the user from upgrading via the archive entry point.


BitwardenShared/Core/Platform/Services/StateService.swift
```diff
@@ -394,6 +394,19 @@ protocol StateService: AnyObject {
...
     func isInitialSyncRequired(userId: String?) async -> Bool
 
+    /// Returns whether the premium upgrade banner has been permanently dismissed by the user.
+    ///
+    /// - Returns: `true` if the user has dismissed the banner.
+    ///
+    func isPremiumUpgradeBannerDismissed() async -> Bool
+
+    /// Returns whether the user meets the eligibility criteria for the premium upgrade
+    /// (free account and 7+ days old).
```
fedemkr: ⛏️ I wouldn't say which are the requirements here in the protocol as if they change you need to remember updating this comment as well. Also the caller doesn't need to know which are the requirements just what the function does and returns.
So for easier maintainability I'd just remove this line
```suggestion
```

BitwardenShared/Core/Platform/Services/StateService.swift
```diff
@@ -1829,6 +1836,23 @@ actor DefaultStateService: StateService, ActiveAccountStateProvider, ConfigState
         }
     }
 
+    func isPremiumUpgradeBannerDismissed() async -> Bool {
+        await ((try? getPremiumUpgradeBannerDismissed()) ?? false)
```
fedemkr: ⛏️ I think you can drop the outer parenthesis here:
```suggestion
        await (try? getPremiumUpgradeBannerDismissed()) ?? false
```
🤔 Also, I think you should log the error if `getPremiumUpgradeBannerDismissed` throws and then return `false` instead of of using `try?` here. Or is it fine to ignore the errors on this function?

BitwardenShared/Core/Platform/Services/StateService.swift
```diff
@@ -1829,6 +1836,23 @@ actor DefaultStateService: StateService, ActiveAccountStateProvider, ConfigState
...
+    }
+
+    func isPremiumUpgradeEligible() async -> Bool {
+        guard await !doesActiveAccountHavePremium() else { return false }
+
+        // Check account age >= 7 days
+        guard let account = try? await getActiveAccount(),
```
fedemkr: 🤔 Shouldn't an error be logged if `getActiveAccount` throws or is it safe to ignore here?

BitwardenShared/Core/Platform/Services/StateService.swift
```diff
@@ -1829,6 +1836,23 @@ actor DefaultStateService: StateService, ActiveAccountStateProvider, ConfigState
...
+              let creationDate = account.profile.creationDate else { return false }
+        guard timeProvider.timeSince(creationDate) >= Constants.premiumUpgradeBannerAccountAge else {
+            return false
+        }
+
+        return true
```
fedemkr: 🎨 Can't you just return the operation result here?
```suggestion
        return timeProvider.timeSince(creationDate) >= Constants.premiumUpgradeBannerAccountAge
```

BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessor.swift
```diff
@@ -562,6 +557,33 @@ extension VaultListProcessor {
...
+    /// Returns `true` when the in-app premium upgrade path is available for the active user.
+    /// Requires the feature flag to be enabled, the user to be eligible (free account, 7+ days old),
+    /// the vault to contain the minimum number of required items, and the user to be on the US storefront.
+    /// Does not check banner dismissal — callers that need that check must do so separately.
+    ///
+    private func isInAppUpgradeAvailable() async -> Bool {
+        guard await services.configService.getFeatureFlag(.premiumUpgradePath) else { return false }
+        guard await services.stateService.isPremiumUpgradeEligible() else { return false }
+        guard (try? await services.vaultRepository
+            .hasMinimumCipherCount(Constants.minimumPremiumUpgradeBannerCipherCount)) ?? false
```
fedemkr: 🤔 Shouldn't this be logging the error when `hasMinimumCipherCount` throws instead of ignoring it?

BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessor.swift
```diff
@@ -562,6 +557,33 @@ extension VaultListProcessor {
...
+        else { return false }
+        guard await services.storefrontService.isUSStorefront() else { return false }
```
fedemkr: 🎨 IMO this `services.storefrontService.isUSStorefront` should be checked before `hasMinimumCipherCount` or even `services.stateService.isPremiumUpgradeEligible` as it should be faster and consume less resources than the others, as it's just a cached value.

BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessor.swift
```diff
@@ -562,6 +557,33 @@ extension VaultListProcessor {
...
+    /// Navigates to the premium upgrade flow. Uses the in-app upgrade path when available;
+    /// otherwise opens the web vault upgrade URL as a fallback.
+    ///
+    private func navigateToPremiumUpgrade() async {
+        if await isInAppUpgradeAvailable() {
+            subscribeToPremiumCheckoutStatus()
+            coordinator.navigate(to: .premiumUpgrade)
+        } else {
+            state.url = services.environmentService.upgradeToPremiumURL
+        }
```
fedemkr: ⛏️ I'd use a guard here to avoid the nesting blocks and be more "Swifty":
```suggestion
        guard await isInAppUpgradeAvailable() else {
            state.url = services.environmentService.upgradeToPremiumURL
            return
        }
        subscribeToPremiumCheckoutStatus()
        coordinator.navigate(to: .premiumUpgrade)
```

BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessor.swift
```diff
@@ -261,15 +261,9 @@ extension VaultListProcessor {
...
-            let hasEnoughItems = await (try? services.vaultRepository
-                .hasMinimumCipherCount(Constants.minimumPremiumUpgradeBannerCipherCount)) ?? false
-            let isUSStorefront = await services.storefrontService.isUSStorefront()
-            state.shouldShowPremiumUpgradeActionCard = shouldShow && hasEnoughItems && isUSStorefront
-        } else {
-            state.shouldShowPremiumUpgradeActionCard = false
-        }
+        let isBannerDismissed = await services.stateService.isPremiumUpgradeBannerDismissed()
+        let isUpgradeAvailable = await isInAppUpgradeAvailable()
+        state.shouldShowPremiumUpgradeActionCard = !isBannerDismissed && isUpgradeAvailable
```
fedemkr: 🎨 If the banner has already been dismissed then it's wasteful to check for `isInAppUpgradeAvailable` as that value doesn't really matter later. So I would rewrite this with short-circuit as:
```swift
        let isBannerDismissed = await services.stateService.isPremiumUpgradeBannerDismissed()
        guard !isBannerDismissed else {
            state.shouldShowPremiumUpgradeActionCard = false
            return
        }
        state.shouldShowPremiumUpgradeActionCard = await isInAppUpgradeAvailable()
```
or like this:
```swift
        let isBannerDismissed = await services.stateService.isPremiumUpgradeBannerDismissed()
        state.shouldShowPremiumUpgradeActionCard = !isBannerDismissed && (await isInAppUpgradeAvailable())
```

BitwardenShared/Core/Platform/Services/StateService.swift
andrebispo5: it will fail in `doesActiveAccountHavePremium` the call above and return false

BitwardenShared/Core/Billing/Repositories/TestHelpers/MockBillingRepository.swift
```diff
@@ -0,0 +1,11 @@
+@testable import BitwardenShared
+
+class MockBillingRepository: BillingRepository {
```
KatherineInCode: 🤔 `BillingRepository` is already marked as `AutoMockable`; we shouldn't need a manual mock, right?

BitwardenShared/Core/Billing/Repositories/BillingRepository.swift
```diff
@@ -0,0 +1,88 @@
+import BitwardenKit
+import Foundation
+
+// MARK: - BillingRepository
+
+/// A protocol for a `BillingRepository` which manages access to billing-related data
+/// needed by the UI layer.
+///
+protocol BillingRepository: AnyObject { // sourcery: AutoMockable
```
KatherineInCode: 🤔 Why is it necessary for this to conform to `AnyObject`?

BitwardenShared/Core/Billing/Repositories/BillingRepositoryTests.swift
```diff
@@ -0,0 +1,129 @@
+import BitwardenKitMocks
+import TestHelpers
+import XCTest
+
+@testable import BitwardenShared
+@testable import BitwardenSharedMocks
+
+class BillingRepositoryTests: BitwardenTestCase {
```
KatherineInCode: 🎨 Ideally new tests are Swift Testing, instead of XCTest. The skill in the repo should be able to convert it for you fairly easily

BitwardenShared/Core/Platform/Services/StateService.swift
```diff
@@ -394,6 +394,18 @@ protocol StateService: AnyObject {
     ///
     func isInitialSyncRequired(userId: String?) async -> Bool
 
+    /// Returns whether the premium upgrade banner has been permanently dismissed by the user.
+    ///
+    /// - Returns: `true` if the user has dismissed the banner.
+    ///
+    func isPremiumUpgradeBannerDismissed() async -> Bool
```
KatherineInCode: 🤔 I have to wonder if it would make sense to actually put these in a `BillingStateService` for some sort of interface segregation, like we have with the `ConfigStateService`, `BiometricsStateService`, `UserSessionStateService`, etc.

BitwardenShared/Core/Billing/Repositories/BillingRepository.swift
```diff
@@ -0,0 +1,88 @@
...
+///
+protocol BillingRepository { // sourcery: AutoMockable
+    /// Returns `true` when the in-app premium upgrade path is available for the active user.
+    ///
+    /// Checks (in order):
+    /// 1. The `.premiumUpgradePath` feature flag is enabled.
+    /// 2. The App Store storefront is in the United States.
+    /// 3. The user is eligible for a premium upgrade (not already premium and account is 7+ days old).
+    /// 4. The vault contains at least the minimum cipher count.
+    ///
```
fedemkr: ⛏️ Remove what this checks inside. Unless it's strictly necessary this shouldn't be added to protocol docs as the caller should only care in this case if the in-app premium upgrade path is available and not how billing internally checks that.
👍 The callout for the banner dismissal is correct as it's something that callers may need to know for their logic.
```suggestion
```

BitwardenShared/Core/Billing/Repositories/BillingRepository.swift
```diff
@@ -0,0 +1,88 @@
...
+        self.storefrontService = storefrontService
+        self.vaultRepository = vaultRepository
+    }
+
+    // MARK: Methods
+
+    func isInAppUpgradeAvailable() async -> Bool {
+        guard await configService.getFeatureFlag(.premiumUpgradePath) else { return false }
+        guard await storefrontService.isUSStorefront() else { return false }
+        guard await stateService.isPremiumUpgradeEligible() else { return false }
```
fedemkr: ⛏️ You can use comma separated conditions here with one `guard`:
```suggestion
        guard await configService.getFeatureFlag(.premiumUpgradePath),
            storefrontService.isUSStorefront(),
            stateService.isPremiumUpgradeEligible() 
        else { 
            return false 
        }
```

BitwardenShared/Core/Platform/Services/StateService.swift
```diff
@@ -1396,7 +1390,7 @@ enum StateServiceError: LocalizedError {
 
 /// A default implementation of `StateService`.
 ///
-actor DefaultStateService: StateService, ActiveAccountStateProvider, ConfigStateService, FlightRecorderStateService, LanguageStateService { // swiftlint:disable:this type_body_length line_length
+actor DefaultStateService: StateService, ActiveAccountStateProvider, BillingStateService, ConfigStateService, FlightRecorderStateService, LanguageStateService { // swiftlint:disable:this type_body_length line_length
```
fedemkr: 🎨 This file is gigantic, could you move this to `BillingStateService` as an extension so the implementation is there? You can check `ServerCommunicationConfigStateService` as an example.
Also move its tests.

BitwardenShared/UI/Vault/Helpers/VaultItemMoreOptionsHelper.swift
```diff
@@ -130,7 +131,14 @@ class DefaultVaultItemMoreOptionsHelper: VaultItemMoreOptionsHelper {
...
+                    Task { [weak self] in
+                        guard let self else { return }
+                        if await self.services.billingRepository.isInAppUpgradeAvailable() {
+                            self.coordinator.navigate(to: .premiumUpgrade)
+                        } else {
+                            handleOpenURL(self.services.environmentService.upgradeToPremiumURL)
+                        }
+                    }
                 }),
             )
```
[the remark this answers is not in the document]
fedemkr: @andrebispo5 have you checked this?

BitwardenShared/UI/Vault/Vault/VaultGroup/VaultGroupProcessor.swift
```diff
@@ -209,6 +218,62 @@ final class VaultGroupProcessor: StateProcessor<
...
+                        to: .dismiss(DismissAction { [weak self] in
+                            guard let self else { return }
+                            coordinator.showLoadingOverlay(
+                                LoadingOverlayState(title: Localizations.confirmingYourUpgrade),
+                            )
+                        }),
+                    )
+                }
+            }
+    }
```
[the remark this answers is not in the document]
andrebispo5: This navigation will change in a future PR, the amount of work to refactor it is not worth it for now.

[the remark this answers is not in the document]
andrebispo5: This navigation will change in a future PR, the amount of work to refactor it is not worth it for now.

fedemkr (approved):
Looks good 🎉

PR #2617, opened 2026-05-06
[PM-36604] test: Fix flaky WatchServiceTests

KatherineInCode:
## 🎟️ Tracking
[PM-36604](https://bitwarden.atlassian.net/browse/PM-36604)

## 📔 Objective
Fixes three flaky tests in `WatchServiceTests`:

1. `syncWithWatch_shouldNotConnect_sendsNeedSetup` — captures the application context inside block 2's closure rather than reading the shared mock property afterwards. A still-running `syncCiphersTask` from the first sync could overwrite `updateApplicationContextReceivedApplicationContext` between the `resume()` call and the assertion.
2. `syncWithWatch_vaultLocked_skipsCipherSync` — replaces `Task.sleep(10_000_000)` with a deterministic barrier sync: after the locked event, a `shouldConnect=false` event triggers a `.needSetup` response (which short-circuits before `decryptCiphers`) and provides a reliable signal that the locked sync was fully processed with no extra decryption.
3. `syncWithWatch_vaultUnlocked_syncsAfterUnlock` — replaces `Task.sleep(10_000_000)` + post-sleep assertion with an in-closure assertion on `updateApplicationContextCallsCount`. The locked sync returns early at the `isVaultLocked` guard without calling `updateApplicationContext`, so asserting `callsCount == 1` inside the unlock callback is both correct and deterministic.

[PM-36604]: https://bitwarden.atlassian.net/browse/PM-36604?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenShared/Core/Platform/Services/WatchServiceTests.swift
```diff
@@ -12,6 +12,10 @@ import WatchConnectivity
 
 // MARK: - WatchServiceTests
 
+/// Tests create `DefaultWatchService` instances that run background tasks on the cooperative thread
+/// pool. Running tests in parallel causes thread contention that can push async chains past the
+/// CI 1-second per-test time limit, so the suite is serialized.
+@Suite(.serialized)
```
matt-livefront: 🤔 Is there enough contention that making these test serialized reduces their overall test time? When I run the tests just in this file, serializing the tests takes almost twice as long. I wonder if the 1 second per-test time limit makes sense with Swift Testing, since running the tests in parallel is going to increase individual test times, but hopefully reduce overall test times.

KatherineInCode: I think that's a very good question, and may obviate this issue (and some others I've seen) altogether. I think if we pushed that time limit up to 2 or 3 seconds, it probably would greatly reduce our failures, *and* as we migrate more to Swift Testing, we might save on overall time despite that 🤔 

@fedemkr @vvolkgang do you have any thoughts on this?

fedemkr: Could we change the limit only for Swift Testing? 
Nonetheless, I would always try to go for parallel testing as much as possible, and as last resort serialize them as it has performance drawbacks.

KatherineInCode: Since it's passed in [as a parameter](https://github.com/bitwarden/ios/blob/main/.github/workflows/test.yml#L161) to the `xcodebuild test` command, I don't think we can have different limits for Swift Testing vs XCTest 😞 

I agree on parallel testing as much as possible

morganzellers-bw (approved):

matt-livefront (approved):

PR #2628, opened 2026-05-10
[PM-10548] fix: Biometric auth setting being reset on logout.

morganzellers-bw:
## 🎟️ Tracking

There are a few tickets covering this bug, both on iOS and Android; I've linked below the two I assigned to myself to track the fix, all others are linked to these as related items.

[PM-17704](https://bitwarden.atlassian.net/browse/PM-17704)

[PM-10548](https://bitwarden.atlassian.net/browse/PM-10548)

## 📔 Objective

Users' biometric unlock preference is being cleared on logout. When that user logs back in, there's no preference to dictate whether we should restore the biometric auth key value in the keychain or not.

These changes update the logout flow to leave the biometric unlock preference in place in UserDefaults on logout. Then, this is checked on login and if true, the auth key is added back to the keychain on login.

## 📸 Screenshots

No biometric setting on simulator for me to do a true test, but I reproduced on production iOS version **2026.4.0 (3082).** Hoping a reviewer with a test device can test.

[PM-17704]: https://bitwarden.atlassian.net/browse/PM-17704?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[PM-10548]: https://bitwarden.atlassian.net/browse/PM-10548?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenKit/Core/Auth/Services/Biometrics/BiometricsRepository.swift
```diff
@@ -128,6 +146,23 @@ public class DefaultBiometricsRepository: BiometricsRepository {
...
+            try await stateService.setBiometricAuthenticationEnabled(true, userId: userId)
+        case .denied,
+             .noBiometrics,
+             .notEnrolled,
+             .unknownError:
+            try await stateService.setBiometricAuthenticationEnabled(false, userId: userId)
+        case .lockedOut,
+             .notDetermined:
+            break
+        }
```
fedemkr: 🤔 If lockedOut or not determined, shouldn't we just fail-close and disable biometrics? or why would this do nothing on these cases?

morganzellers-bw: For `.lockedOut`, I was thinking that the operating system checks would sufficiently resolve this state if we just did nothing. To hit this case, the user has been locked out of biometric auth _after_ unlocking the device, and at this point, they have already unlocked their vault using a non-biometric method. So it doesn't feel right to turn off their biometric auth setting in app just because they're temporarily locked out of using it by the OS.

For `.notDetermined`, we guard that the biometric unlock status is enabled in the calling function [here](https://github.com/bitwarden/ios/pull/2628/changes/BASE..ff94000ab9ccdee2051d1c2b746acbc410dbd081#diff-cb8846480d62aaf308bd0e3a3cc1f018f98ec0031d052faf961fe8e90982fc1bR1282). 

And I _think_ the only way this state can happen past that guard is if a user restores from an iCloud backup, but has not been prompted for the biometric auth permission yet. In that case, we shouldn't turn off the setting they've set, they'll just hit the bio auth prompt the next time they unlock.

BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
```diff
@@ -1273,6 +1273,20 @@ extension DefaultAuthRepository: AuthRepository {
...
 
+    /// Restores the biometric unlock keychain entry after a successful vault unlock when the
+    /// biometric preference is enabled. The preference survives logout but the keychain entry
+    /// is cleared; this rewrites it so biometric unlock works without re-enabling in settings.
+    ///
+    private func configureBiometricUnlockIfNeeded() async {
+        do {
+            guard try await biometricsRepository.getBiometricUnlockStatus().isEnabled else { return }
+            let authKey = try await clientService.crypto().getUserEncryptionKey()
+            try await biometricsRepository.restoreBiometricUnlockKey(authKey: authKey)
```
matt-livefront, line 1285: ⚠️ This is going to get called on each unlock, even if the unlock key already exists for the user. And because it's updating a key in the keychain protected by biometrics, you see the Face ID prompt twice on unlock. Could we check to see if the key exists before generating and saving it?

BitwardenKit/Core/Auth/Services/Biometrics/BiometricsRepository.swift
```diff
@@ -52,6 +52,16 @@ public protocol BiometricsRepository: AnyObject { // sourcery: AutoMockable
...
+
+    /// Restores the biometric unlock key after a trusted vault unlock (e.g., master password unlock),
+    /// without prompting for biometric authentication. If biometrics is no longer available on the device,
+    /// the stored preference is cleared.
+    ///
+    /// - Parameters:
+    ///   - authKey: The user auth key to store in the keychain.
+    ///   - userId: The user ID for the user to restore biometric unlock. Defaults to the active user if nil.
+    ///
+    func restoreBiometricUnlockKey(authKey: String, userId: String?) async throws
```
matt-livefront, line 61: ⛏️ Could you alphabetize this (and the two instances further down)?

BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
morganzellers-bw: Definitely, good catch

BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
```diff
@@ -1273,6 +1273,20 @@ extension DefaultAuthRepository: AuthRepository {
         return try await stateService.getActiveAccountId()
     }
 
+    /// Restores the biometric unlock keychain entry after a successful vault unlock when the
+    /// biometric preference is enabled. The preference survives logout but the keychain entry
+    /// is cleared; this rewrites it so biometric unlock works without re-enabling in settings.
+    ///
+    private func configureBiometricUnlockIfNeeded() async {
+        do {
+            guard try await biometricsRepository.getBiometricUnlockStatus().isEnabled else { return }
```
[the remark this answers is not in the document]
morganzellers-bw: If we decide to no-op on `.lockedOut`, this will fix:

```swift
private func configureBiometricUnlockIfNeeded() async {
    do {
        ...
    } catch BiometricsServiceError.biometryLocked {
      // Lockout is transient; do nothing and let the user retry later.
   } catch {
        errorReporter.log(error: error)
    }
}
```

BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
morganzellers-bw: @matt-livefront I added:

```swift
/// Returns whether a value exists for the given keychain item without triggering biometric
/// authentication.
///
/// Two query options work together to make this safe:
/// - `kSecUseAuthenticationContext` with `interactionNotAllowed = true`: suppresses the Face ID
///   prompt. When a biometric-protected item is found, iOS returns `errSecInteractionNotAllowed`
///   instead of prompting, which this method catches and treats as "item exists".
/// - `kSecReturnAttributes: true`: causes a successful match to return a non-nil attribute
///   dictionary, confirming existence without reading the secret. Without any `kSecReturn*` key, `SecItemCopyMatching`
///   returns `errSecSuccess` with a nil result even when an item exists.
///
/// - Parameter item: The keychain item to check.
/// - Returns: `true` if the item exists in the keychain, `false` otherwise.
///
func containsValue(for item: any KeychainItem) async -> Bool
```

to `KeychainServiceFacade` [here](https://github.com/bitwarden/ios/pull/2628/changes#diff-d10c525e61b1c70d95e8e5ee22f0eb60ed5140d0aba8f341805d26e0f71e512cR260). 
It allows for a no-prompt check for the biometric key value from AuthRepository [here](https://github.com/bitwarden/ios/pull/2628/changes#diff-cb8846480d62aaf308bd0e3a3cc1f018f98ec0031d052faf961fe8e90982fc1bR1280).

BitwardenKit/Core/Auth/Services/Biometrics/BiometricsRepository.swift
fedemkr: Perfect, sounds good thanks for checking it Morgan!

BitwardenKit/Core/Platform/Services/KeychainServiceFacade.swift
```diff
@@ -33,6 +34,22 @@ public protocol KeychainServiceFacade { // sourcery: AutoMockable
...
+    /// Returns whether a value exists for the given keychain item without triggering biometric
+    /// authentication.
+    ///
+    /// Two query options work together to make this safe:
+    /// - `kSecUseAuthenticationContext` with `interactionNotAllowed = true`: suppresses the Face ID
+    ///   prompt. When a biometric-protected item is found, iOS returns `errSecInteractionNotAllowed`
+    ///   instead of prompting, which this method catches and treats as "item exists".
+    /// - `kSecReturnAttributes: true`: causes a successful match to return a non-nil attribute
+    ///   dictionary, confirming existence without reading the secret. Without any `kSecReturn*` key, `SecItemCopyMatching`
+    ///   returns `errSecSuccess` with a nil result even when an item exists.
```
fedemkr: ⛏️ I believe the protocol shouldn't need to clarify all this as it's referring to the implementation of `DefaultKeychainServiceFacade`. If there were a different implementation or the e.g. the mock, won't necessarily have the same details. So I think we could just say here that this suppresses the Face ID prompt and then move all the details to the function implementation where it's specifically done.

BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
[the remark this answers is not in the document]
matt-livefront: I think it's worth going the no-op route to avoid logging an error here.

BitwardenKit/Core/Platform/Services/KeychainServiceFacade.swift
morganzellers-bw: Ah yeah great point, these are all implementation details haha. I'll update, thanks Fede!

BitwardenKit/Core/Platform/Services/KeychainServiceFacade.swift
```diff
@@ -221,6 +229,41 @@ public class DefaultKeychainServiceFacade: KeychainServiceFacade {
 
     // MARK: Value
 
+
```
matt-livefront, line 231: ⛏️ Extra newline?

BitwardenKit/Core/Platform/Services/KeychainServiceFacade.swift
```diff
@@ -221,6 +229,41 @@ public class DefaultKeychainServiceFacade: KeychainServiceFacade {
 
     // MARK: Value
 
+    
+    /// Returns whether a value exists for the given keychain item without triggering a FaceID prompt.
+    ///
+    ///   Two query options work together to make this safe:
+    /// - `kSecUseAuthenticationContext` with `interactionNotAllowed = true`: suppresses the Face ID
```
matt-livefront: ⛏️ Should this be aligned with the bullets below it?

```suggestion
    /// Two query options work together to make this safe:
    /// - `kSecUseAuthenticationContext` with `interactionNotAllowed = true`: suppresses the Face ID
```

matt-livefront (approved):

PR #2633, opened 2026-05-11
[PM-35295] fix: Prevent iOS autofill from clearing BitwardenTextField via local state buffer

### LINKED ISSUE #2554: [PM-35295] Username field is not accepting keyboard suggestions
### Steps To Reproduce

1. Go to App
2. Add new Login item
3. Click on username field to enter username
4. There should be suggestion in keyboard for your default email
5. Click on suggestion, it won't work

### Expected Result

When a user clicks it should be filled with suggested username

### Actual Result

When a user clicks it is not getting filled with suggested username

### Screenshots or Videos

https://github.com/user-attachments/assets/f436cc51-cf42-4413-be87-a4a545fef63b

### Additional Context

It is very annoying as most of the items its going to be user's default email, and typing that long email correctly is not good experience 

### Build Version

2026.4.0 (3082)

US

### Self-host Server Version

_No response_

### Environment Details

iPhone 13
iOS 18.2.1

### Issue Tracking Info

matt-livefront:
## 🎟️ Tracking

[PM-35295](https://bitwarden.atlassian.net/browse/PM-35295)

## 📔 Objective

Fixes https://github.com/bitwarden/ios/issues/2554

Fixes a bug where tapping an iOS keyboard autofill suggestion (e.g., from the QuickType bar) would clear a `BitwardenTextField` instead of populating it.

iOS autofill's post-fill verification phase seems to read the text field value after writing. Because the native `TextField`/`SecureField` was backed directly by an external binding, the SwiftUI binding re-evaluation that occurred during autofill's write cycle was seen as a rejection signal, causing autofill to revert the field to its pre-fill state (empty).

This fixes it by introducing a `@State private var localText` buffer that decouples the native text field from the external binding. Two `onChange` handlers keep them in sync:
- localText → text (user input flows outward)
- text → localText (external/programmatic updates flow inward, with a guard to prevent mutual-update loops)

## 📸 Screenshots

| Before | After |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/f34ef88c-6f45-4c73-8e6c-4b97591955f0" /> | <video src="https://github.com/user-attachments/assets/50cdcf0d-1ddc-42ac-b9e5-a20cb299308a" /> | 

[PM-35295]: https://bitwarden.atlassian.net/browse/PM-35295?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenKit/UI/Platform/Application/Views/BitwardenTextField.swift
+++ b/BitwardenKit/UI/Platform/Application/Views/BitwardenTextField.swift
@@ -30,6 +30,10 @@ public struct BitwardenTextField<FooterContent: View, TrailingContent: View>: Vi
     /// post-fill verification doesn't see a binding-driven re-evaluation as a rejection.
     @State private var localText: String = ""
 
+    /// Tracks every value sent outward so that store echoes arriving out of order can be
+    /// identified and dropped rather than overwriting localText with a stale value.
+    @State private var pendingEchoes: [String: Int] = [:]
+
     /// Whether the placeholder text should be shown in the text field.
     private var showPlaceholder: Bool {
         !isFocused && localText.isEmpty
@@ -87,9 +91,15 @@ public struct BitwardenTextField<FooterContent: View, TrailingContent: View>: Vi
         .clipShape(RoundedRectangle(cornerRadius: 8))
         .accessibilityElement(children: .contain)
         .onChange(of: localText) { newValue in
+            guard text != newValue else { return }
+            pendingEchoes[newValue, default: 0] += 1
             text = newValue
         }
         .onChange(of: text) { newValue in
+            if let count = pendingEchoes[newValue], count > 0 {
+                pendingEchoes[newValue] = count == 1 ? nil : count - 1
+                return
+            }
             guard newValue != localText else { return }
             localText = newValue
         }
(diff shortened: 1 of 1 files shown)


morganzellers-bw (approved):
Only thought I had was whether a ViewInspector regression test for this would be useful?   Otherwise, looks great and I'd love to hear how you found the cause of this 😆

matt-livefront:
> Only thought I had was whether a ViewInspector regression test for this would be useful? Otherwise, looks great and I'd love to hear how you found the cause of this 😆

@morganzellers-bw That would be useful, but I'm not sure if it's possible. The issue relies on how the system autofills the field from keyboard suggestions, but that's outside of ViewInspector's reach. ViewInspector can write to the field and then we can confirm that the binding was updated, but we already have tests around that. As for finding it, this was mostly a Claude 🤖 suggestion. I was seeing that the processor would first receive the autofilled value followed by an empty string. If we avoided the roundtrip to the processor by using a local `@State` variable, then the autofill worked successfully.

fedemkr (approved):

PR #2635, opened 2026-05-12
[PM-31723] fix: Unlock vault error when missing login data

LRNcardozoWDF:
## 🎟️ Tracking

[PM-31723](https://bitwarden.atlassian.net/browse/PM-31723)

## 📔 Objective

Fixes unlock vault when the `masterPasswordUnlock` is nil, by calling `fetchSync`. This happens for users on self-hosted server older than v2025.11.0. 

[PM-31723]: https://bitwarden.atlassian.net/browse/PM-31723?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
```diff
@@ -1038,7 +1044,16 @@ extension DefaultAuthRepository: AuthRepository {
...
+        var account = try await stateService.getActiveAccount()
+
+        if account.profile.userDecryptionOptions?.masterPasswordUnlock == nil {
+            do {
+                try await syncService.fetchSync(forceSync: true)
+                account = try await stateService.getActiveAccount()
+            } catch {
+                errorReporter.log(error: error)
+            }
+        }
```
fedemkr, line 1130: 🤔 In case there's an error on `fetchSync` or `getActiveAccount`, is it ok to keep throwing the same error message to the user as before? I see that errors are swallowed and just reported to the logger in this case now.
🎨 Additionally, could we add a little comment on why this is needed here? It could help when maintaining this and whether at some point we can remove this code. Also for Claude to know why this is here.

LRNcardozoWDF: I thought so, if there's an error on either the `masterPasswordUnlock` would be `nil` and the code would fail as before, but on second thought it might not be the best option. Thank you

BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
```diff
@@ -1047,12 +1047,11 @@ extension DefaultAuthRepository: AuthRepository {
...
         if account.profile.userDecryptionOptions?.masterPasswordUnlock == nil {
-            do {
-                try await syncService.fetchSync(forceSync: true)
-                account = try await stateService.getActiveAccount()
-            } catch {
-                errorReporter.log(error: error)
-            }
+            // masterPasswordUnlock can be missing if the account data was stored before the server
+            // added it. Force a sync to refresh it.
+            // TODO: PM-37535
```
fedemkr: ⛏️ Could you explain what's the remaining TODO here in one sentence? So we have a bit more context on what's missing without going to the ticket.

fedemkr (approved):

PR #2638, opened 2026-05-12
[PM-37072] chore: Consolidate MockAuthClient with support for MockRegistrationClientProtocol

matt-livefront:
## 🎟️ Tracking

[PM-37072](https://bitwarden.atlassian.net/browse/PM-37072)

## 📔 Objective

This consolidates `MockAuthClient` to BitwardenKit and adds a `MockRegistrationClientProtocol`.

The current `MockAuthClient` conforms to `AuthClientProtocol`. When the SDK builds protocols for services, the protocol methods return the concrete type instead of a protocol type (e.g. `AuthClientProtocol` returns `RegistrationClient` instead of `RegistrationClientProtocol`). This makes it difficult to use SDK's protocols directly when they return other types which need to be mocked. As a workaround, this adds an `AuthClientService` protocol which returns a `RegistrationClientProtocol`, and then `AuthClient` can conform to that. This allows for mocking `RegistrationClient` without too much overhead. The `AuthClientService` protocol just needs to be used in place of `AuthClientProtocol`.

[PM-37072]: https://bitwarden.atlassian.net/browse/PM-37072?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenKit/Core/Auth/Services/SDK/Mocks/MockAuthClient.swift
```diff
@@ -0,0 +1,193 @@
+import BitwardenKit
+import BitwardenSdk
+import BitwardenSdkMocks
+import Foundation
+
+public class MockAuthClient: AuthClientService {
```
fedemkr, line 6: 🤔 Could we use `Sourcery` for this? And nonetheless, shouldn't this be named `MockAuthClientService`?

matt-livefront: We could, yes, this PR was already getting large so I was trying to limit the amount of changes. I have these changes on another branch, so ok if I do this as a follow-up PR?

fedemkr: Sure no problem

fedemkr (approved):

PR #2641, opened 2026-05-13
[QA-1865] Removing Rate the app alert for Beta builds

ifernandezdiaz:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/QA-1865

## 📔 Objective

We found the Rate the App alert was causing some false negatives during our nightly e2e runs. This PR aims to prevent this alert from being displayed in beta builds only.


BitwardenShared/Core/Platform/Services/ReviewPromptService.swift
```diff
@@ -49,16 +52,19 @@ class DefaultReviewPromptService: ReviewPromptService {
     /// - Parameters:
     ///   - appVersion: The current app version.
     ///   - identityStore: The service used to manage the credentials available for AutoFill suggestions.
+    ///   - isBetaBuild: Whether the running build is a beta build.
     ///   - stateService: The service used by the application to manage account state.
     ///
     init(
         appVersion: String,
         identityStore: CredentialIdentityStore = ASCredentialIdentityStore.shared,
+        isBetaBuild: Bool,
```
KatherineInCode: 🎨 It's unfortunate there doesn't seem to be an easy build flag (like there is with `DEBUG`) we could key off of. That said, I think it would be better if here it passed in an `any AppInfoService`, and we constrained the logic of whether or not it was a beta build to that object. It's a little more complicated (particularly in the test) but fits our patterns a bit better, I think.

BitwardenShared/Core/Platform/Services/ServiceContainer.swift
```diff
@@ -755,6 +755,7 @@ public class ServiceContainer: Services { // swiftlint:disable:this type_body_le
 
         let reviewPromptService = DefaultReviewPromptService(
             appVersion: Bundle.main.appVersion,
+            isBetaBuild: Bundle.main.bundleIdentifier?.hasSuffix(".beta") == true,
```
KatherineInCode: 🎨 Like with the comment above, I feel like this logic should be constrained to the `AppInfoService`, especially because it already has a local `Bundle` to work with. Then we can also test the logic itself, and we have a clear place to reference it if we encounter other things we want to key off of being a beta build. And then, in the future, if we *do* introduce a `BETA` build flag, the logic for managing it is centralized.

BitwardenKit/Core/Platform/Services/AppInfoServiceTests.swift
```diff
@@ -310,6 +310,24 @@ class AppInfoServiceTests: BitwardenTestCase { // swiftlint:disable:this type_bo
         )
     }
 
+    /// `isBetaBuild` is `true` when the bundle identifier ends with `.beta`.
+    func test_isBetaBuild_true() {
```
KatherineInCode: ⛏️ These should be alphabetized with the rest of the file

BitwardenShared/Core/Platform/Services/ReviewPromptService.swift
```diff
@@ -47,18 +50,21 @@ class DefaultReviewPromptService: ReviewPromptService {
...
+        appInfoService: AppInfoService,
         appVersion: String,
         identityStore: CredentialIdentityStore = ASCredentialIdentityStore.shared,
         stateService: StateService,
     ) {
+        self.appInfoService = appInfoService
         self.appVersion = appVersion
-        self.stateService = stateService
         self.identityStore = identityStore
+        self.stateService = stateService
```
KatherineInCode, line 67: 👍🏻 for alphabetizing

BitwardenShared/Core/Platform/Services/ReviewPromptServiceTests.swift
```diff
@@ -46,6 +51,19 @@ class ReviewPromptServiceTests: BitwardenTestCase {
         XCTAssertTrue(stateService.reviewPromptData?.userActions.isEmpty ?? false)
     }
 
+    /// `isEligibleForReviewPrompt()` returns false on beta builds even when all other criteria are met.
+    func test_isEligibleForReviewPrompt_betaBuild() async {
```
KatherineInCode: 🤔 Nominally this should be alphabetized with the other tests, but also is the first thing to worry about in line. I'm of two minds about it.

BitwardenKit/Core/Platform/Services/AppInfoServiceTests.swift
```diff
@@ -310,6 +310,24 @@ class AppInfoServiceTests: BitwardenTestCase { // swiftlint:disable:this type_bo
         )
     }
 
+    /// `isBetaBuild` is `false` for production bundle identifiers.
+    func test_isBetaBuild_false() {
```
KatherineInCode, line 322: ⛏️ Alphabetize

ifernandezdiaz: Oh, my bad!
Fixed.

KatherineInCode (approved):

PR #2645, opened 2026-05-13
[PM-37253] feat: Extract PremiumUpgradeHelper to centralise premium upgrade navigation

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37253

## 📔 Objective

Introduces `DefaultPremiumUpgradeHelper` in `UI/Billing/` to eliminate the duplicated premium upgrade navigation pattern across five processors (`VaultListProcessor`, `VaultGroupProcessor`, `VaultItemSelectionProcessor`, `ViewItemProcessor`, `AddEditItemProcessor`).

Also fixes `DefaultVaultItemActionHelper.archive()` to route through `handleNavigateToPremiumUpgrade` instead of always opening the web vault URL, and adds a `.premiumUpgrade` route to `VaultItemRoute` / `VaultItemCoordinator` so the view/edit-item screens can present the billing upgrade screen.

**Key changes:**
- New `PremiumUpgradeHelper` protocol + `DefaultPremiumUpgradeHelper` in `UI/Billing/` with `navigateToPremiumUpgrade(onConfirmed:)` and `startInAppPremiumUpgrade(onConfirmed:)`
- `DefaultVaultItemActionHelper.archive()`: replaces `handleOpenURL` with `handleNavigateToPremiumUpgrade` so the archive unavailable alert uses the in-app flow
- `VaultItemRoute` + `VaultItemCoordinator`: adds `.premiumUpgrade` route for view/edit-item screens
- All five processors delegate to the helper via a `lazy var premiumUpgradeHelper`; per-processor `subscribeToPremiumCheckoutStatus` and `premiumStatusChangedCancellable` removed
- `PremiumUpgradeHelperTests` covers the full availability-check, subscribe, confirmed, pending, and syncing paths in one place

### DIFF
--- a/BitwardenShared/UI/Billing/PremiumUpgradeHelper.swift
+++ b/BitwardenShared/UI/Billing/PremiumUpgradeHelper.swift
@@ -3,6 +3,24 @@ import BitwardenResources
 import Combine
 import Foundation
 
+// MARK: - PremiumUpgradeRoute
+
+/// A route type that supports premium upgrade navigation.
+///
+protocol PremiumUpgradeRoute {
+    /// The route to the premium upgrade screen.
+    static var premiumUpgrade: Self { get }
+
+    /// The route to dismiss the current screen with an optional action.
+    ///
+    /// - Parameter action: The action to perform on dismiss.
+    ///
+    static func dismiss(_ action: DismissAction?) -> Self
+}
+
+extension VaultItemRoute: PremiumUpgradeRoute {}
+extension VaultRoute: PremiumUpgradeRoute {}
+
 // MARK: - PremiumUpgradeHelper
 
 /// A helper that centralizes the premium upgrade navigation flow.
@@ -35,7 +53,7 @@ extension PremiumUpgradeHelper {
 /// The default implementation of `PremiumUpgradeHelper`.
 ///
 @MainActor
-class DefaultPremiumUpgradeHelper: PremiumUpgradeHelper {
+class DefaultPremiumUpgradeHelper<Route: PremiumUpgradeRoute, Event>: PremiumUpgradeHelper {
     // MARK: Types
 
     typealias Services = HasBillingRepository
@@ -47,14 +65,8 @@ class DefaultPremiumUpgradeHelper: PremiumUpgradeHelper {
     /// A cancellable for the premium checkout status subscription.
     private var premiumStatusChangedCancellable: AnyCancellable?
 
-    /// A closure called to hide the loading overlay.
-    private let hideLoadingOverlay: () -> Void
-
-    /// A closure called to navigate to a dismiss route.
-    private let navigateToDismiss: (DismissAction) -> Void
-
-    /// A closure called to navigate to the premium upgrade screen.
-    private let navigateToPremiumRoute: () -> Void
+    /// The coordinator used for navigation.
+    private let coordinator: any Coordinator<Route, Event>
 
     /// An optional closure called inside the pending dismiss action before showing the upgrade
     /// pending alert. Use to dismiss action cards or perform other per-screen cleanup.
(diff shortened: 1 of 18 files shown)


BitwardenShared/UI/Vault/Vault/VaultGroup/VaultGroupProcessor.swift
```diff
@@ -38,8 +38,28 @@ final class VaultGroupProcessor: StateProcessor<// swiftlint:disable:this type_b
...
+        showAlert: { [weak self] alert in
+            self?.coordinator.showAlert(alert)
+        },
+        hideLoadingOverlay: { [weak self] in
+            self?.coordinator.hideLoadingOverlay()
+        },
+        onPendingDismiss: { [weak self] in
+            Task { await self?.dismissPremiumUpgradeActionCard() }
+        },
+    )
```
matt-livefront, line 49: 🎨 This is cool! I think if you wanted to take it step further, you could simplify some of this by passing the coordinator into the helper. 

Here's what Claude had to say about this:

> ```
> protocol PremiumUpgradeRoute {
>    static var premiumUpgrade: Self { get }
>    static func dismiss(_ action: DismissAction?) -> Self
>}
>extension VaultRoute: PremiumUpgradeRoute {}
>extension VaultItemRoute: PremiumUpgradeRoute {}
>```
>Both VaultItemRoute and VaultRoute already have exactly these cases, so conformance is just a declaration. DefaultPremiumUpgradeHelper becomes generic:
> ```
> class DefaultPremiumUpgradeHelper<Route: PremiumUpgradeRoute, Event>: PremiumUpgradeHelper {
>    private weak var coordinator: (any Coordinator<Route, Event>)?
>    private let setURL: (URL) -> Void
>    private let onPendingDismiss: (() -> Void)?
>}
>```
>
>This drops from 5 closures to 1 required closure (setURL) + the coordinator + the optional onPendingDismiss. The call sites get much leaner.
>
>Tradeoff: Generics on DefaultPremiumUpgradeHelper add a small layer of complexity, and Swift's any Coordinator<Route, Event> existentials require both type params to be known at the call site — but since processors already have a concrete coordinator type, this is fine.

BitwardenShared/UI/Billing/PremiumUpgradeHelper.swift
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    func startInAppPremiumUpgrade(onConfirmed: (() async -> Void)? = nil) {
+        subscribeToPremiumCheckoutStatus(onConfirmed: onConfirmed)
+        navigateToPremiumRoute()
+    }
+
+    // MARK: Private Methods
+
+    private func subscribeToPremiumCheckoutStatus(onConfirmed: (() async -> Void)?) {
```
matt-livefront, line 125: ⛏️ Can you add docs for this?

BitwardenShared/UI/Vault/VaultItem/VaultItemActionHelper.swift
```diff
@@ -64,14 +63,15 @@ class DefaultVaultItemActionHelper: VaultItemActionHelper {
...
     ) async {
         guard await services.vaultRepository.doesActiveAccountHavePremium() else {
             await coordinator.showAlert(
-                Alert.archiveUnavailable(action: { [weak self] in
-                    guard let self else { return }
-                    handleOpenURL(services.environmentService.upgradeToPremiumURL)
+                Alert.archiveUnavailable(action: {
+                    Task { @MainActor in
+                        await handleNavigateToPremiumUpgrade()
+                    }
```
matt-livefront: 🎨 Alert actions support async calls, so I think if you change the Alert builder to allow an asynchronous closure:

```
    static func archiveUnavailable(
        action: @escaping () async -> Void,
    ) -> Alert {
        let preferredAction = AlertAction(title: Localizations.upgradeToPremium, style: .default) { _ in
            await action()
        }
```

Then you can skip creating a Task here:
```suggestion
                Alert.archiveUnavailable(action: {
                    await handleNavigateToPremiumUpgrade()
                }
```

BitwardenShared/UI/Vault/VaultItem/VaultItemActionHelperTests.swift
```diff
@@ -65,11 +59,13 @@ struct VaultItemActionHelperTests {
         #expect(!completionCalled)
 
         try await alert.tapAction(title: Localizations.upgradeToPremium)
-        #expect(openedURL == URL(string: "https://example.com/someURLToUpgradeToPremium"))
+        // The alert action fires a Task; yield to let it complete.
+        await Task.yield()
```
matt-livefront: 🎨 If you make the async action change I mentioned, then you shouldn't need this yield, since `await alert.tapAction()` will wait until the action has finished the async calls.

BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessor.swift
```diff
@@ -593,53 +613,13 @@ extension VaultListProcessor {
...
-                    // PremiumUpgradeProcessor shows the loading overlay on the upgrade screen.
-                    break
-                }
-            }
+        await premiumUpgradeHelper.navigateToPremiumUpgrade(onConfirmed: { [weak self] in
+            guard let self else { return }
+            await refreshVault(syncWithPeriodicCheck: false)
+            state.hasPremium = await services.stateService.doesActiveAccountHavePremium()
+            state.shouldShowPremiumUpgradeActionCard = false
+            state.shouldShowUpgradedToPremiumActionCard = true
```
fedemkr: 🤔 Not sure I follow, if you need to check whether the account has premium to set `hasPremium` then shouldn't `state.shouldShowPremiumUpgradeActionCard` and `state.shouldShowUpgradedToPremiumActionCard` be based on whether the user has premium?

andrebispo5: Yes in a sense checking for premium is a formality to be fair, when confirmed is fired it already checked if the account has premium. But I'll tie the variables to the `state.hasPremium`

matt-livefront: Should this here also be tied to `state.hasPremium`? I looks like just the other case was updated.

andrebispo5: didn't noticed, extracted the completion handler to share between both calls

matt-livefront (approved):

PR #2648, opened 2026-05-14
[PM-27238] feat: Support v2 encryption on key connector signups

matt-livefront:
## 🎟️ Tracking

[PM-27238](https://bitwarden.atlassian.net/browse/PM-27238)

## 📔 Objective

Enables support for V2 encryption on key connector signups. Most of the existing logic that was implemented in the app, including API requests, has been moved into the SDK (via the `postKeysForKeyConnectorRegistration(keyConnectorUrl:ssoOrgIdentifier:)` function).

These changes are behind the `enable-account-encryption-v2-key-connector-registration` feature flag.

[PM-27238]: https://bitwarden.atlassian.net/browse/PM-27238?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
```diff
@@ -632,10 +637,14 @@ extension DefaultAuthRepository: AuthRepository {
     }
 
     func convertNewUserToKeyConnector(keyConnectorURL: URL, orgIdentifier: String) async throws {
-        try await keyConnectorService.convertNewUserToKeyConnector(
+        let conversionResult = try await keyConnectorService.convertNewUserToKeyConnector(
             keyConnectorUrl: keyConnectorURL,
             orgIdentifier: orgIdentifier,
         )
+        try await unlockVault(method: .keyConnector(
```
quexten, line 655: Thought: Can we use the newer `keyConnectorUrl` unlock method? https://github.com/bitwarden/sdk-internal/blob/e13c4fcc75a608c075b1a2d9e8fc1530a12bc36c/crates/bitwarden-core/src/key_management/crypto.rs#L168

The `keyConnector` method will be deprecated at some point so may as well switch to it in any new places.

BitwardenShared/Core/Platform/Models/Domain/AccountEncryptionKeys.swift
```diff
@@ -32,4 +34,38 @@ extension AccountEncryptionKeys {
...
+    ///
+    /// - Parameters:
+    ///   - accountCryptographicState: The SDK's wrapped account cryptographic state.
+    ///   - encryptedUserKey: An optional encrypted user key to store alongside the cryptographic state.
+    ///
+    init(accountCryptographicState: WrappedAccountCryptographicState, encryptedUserKey: String? = nil) {
+        switch accountCryptographicState {
+        case let .v1(privateKey):
+            self.init(
+                accountKeys: nil,
```
quexten: AccountKeys should not be nil. It should be populated with a PrivateKeysResponseModel with with signature key pair and security state set to nil, but the publicKeyEncryptionKeyPair present.

BitwardenShared/Core/Platform/Models/Domain/AccountEncryptionKeys.swift
```diff
@@ -32,4 +34,38 @@ extension AccountEncryptionKeys {
...
+                encryptedPrivateKey: privateKey,
+                encryptedUserKey: encryptedUserKey,
+            )
+        case let .v2(privateKey, signedPublicKey, signingKey, securityState):
+            self.init(
+                accountKeys: PrivateKeysResponseModel(
+                    publicKeyEncryptionKeyPair: PublicKeyEncryptionKeyPairResponseModel(
+                        publicKey: "", // Not returned by SDK at registration; will populate on next sync.
```
quexten: This feels like it has a high risk of introducing bugs. Can we not just store the accountCryptographicState? That's what the SDK takes for init, and KM expects clients to just store it instead of the response model.

BitwardenShared/Core/Platform/Models/Domain/AccountEncryptionKeys.swift
```diff
@@ -32,4 +34,38 @@ extension AccountEncryptionKeys {
...
+                        signedPublicKey: signedPublicKey,
+                        wrappedPrivateKey: privateKey,
+                    ),
+                    signatureKeyPair: SignatureKeyPairResponseModel(
+                        wrappedSigningKey: signingKey,
+                        verifyingKey: "", // Not returned by SDK at registration; will populate on next sync.
```
quexten: Same comment as above, this seems like a high risk.

BitwardenShared/Core/Platform/Models/Domain/AccountEncryptionKeys.swift
```diff
@@ -32,4 +34,38 @@ extension AccountEncryptionKeys {
...
         )
     }
+
+    /// Initializes an `AccountEncryptionKeys` from a `WrappedAccountCryptographicState`.
+    ///
+    /// - Parameters:
+    ///   - accountCryptographicState: The SDK's wrapped account cryptographic state.
+    ///   - encryptedUserKey: An optional encrypted user key to store alongside the cryptographic state.
+    ///
+    init(accountCryptographicState: WrappedAccountCryptographicState, encryptedUserKey: String? = nil) {
```
quexten: I don't quite understand what we are doing here. We should just store the WrappedAccountCryptographicState, not a response model. On sync, the response model should get converted to WrappedAccountCryptographicState.

quexten (changes requested):
@matt-livefront It's fine if that's out of scope for this PR, but the way iOS is storing the account cryptographic state as a custom model which is different from the public API surface of the SDK is weird and seems brittle / likely to introduce bugs. Is it not possible to just store the SDK's model, or even a string serialized version of it?

Specifically, mobile *should not care* about the internals of that object or individually set fields on it (i guess with the exception of sync, though we should work on moving that part to the SDK too).

matt-livefront:
> @matt-livefront It's fine if that's out of scope for this PR, but the way iOS is storing the account cryptographic state as a custom model which is different from the public API surface of the SDK is weird and seems brittle / likely to introduce bugs. Is it not possible to just store the SDK's model, or even a string serialized version of it?
> 
> Specifically, mobile _should not care_ about the internals of that object or individually set fields on it (i guess with the exception of sync, though we should work on moving that part to the SDK too).

@quexten Thanks for your feedback! In Swift, the easiest way to support persistence is via [Codable](https://developer.apple.com/documentation/swift/codable). The Swift bindings from the SDK don't include this, and it's a fair bit of extra work for us to implement it manually on our side. So I suspect that's how we ended up in this position; our response models are already Codable (for decoding JSON) so it's straightforward to persist those and then transform into the SDK types as needed. But you make a good point, do you know if the SDK could include the Codable conformance? It seems like it could just be a configuration option, but maybe there's more to it. https://mozilla.github.io/uniffi-rs/latest/swift/configuration.html#:~:text=generate_codable_conformance

BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
matt-livefront: Yes, good callout. I saw that this was introduced, but thought there may be a separate ticket for switching to this. I'll check and if not, I'll create one.

BitwardenShared/Core/Platform/Models/Domain/AccountEncryptionKeys.swift
matt-livefront: Do you know where that could come from in this key connector registration flow? It isn't returned in `KeyConnectorRegistrationResult` when calling `postKeysForKeyConnectorRegistration`? Given that the SDK is generating V2 encryption keys, is it possible to end up in this `v1` case? Maybe the answer is, as you suggested, to persist `WrappedAccountCryptographicState` directly, but I was trying to make this work with we we already have in place to avoid additional changes.

BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
matt-livefront: ~~Would it be better to unlock with the decrypted key here returned by `KeyConnectorRegistrationResult`? That would avoid needing to fetch the key from the API.~~

I remember why I did it the way that I did. Using `keyConnector` here supports both the V1 and V2 flows, without having to fetch the key from key connector again. Once we remove the key connector V2 feature flag, we can update this here to use the decrypted key that's returned from `KeyConnectorRegistrationResult`.

quexten:
@matt-livefront I don't see a reason not to enable codable on the SDK. We want to keep the amount mobile implements to a minimum, so if this reduces it I support it. I'm opening a PR to enable it for crypto / core, which should cover anything on the key-management front.

quexten:
@matt-livefront Merged the PR, codable should now be available. Please note I did not test the UNIFFI build locally. I have a separate PR adding swift integration tests, after which we could add those tests into there. Until then, please LMK if there are any issues on this front.

BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
quexten: Makes sense!

BitwardenShared/Core/Platform/Models/Domain/AccountEncryptionKeys.swift
quexten: As mentioned, persisting the `WrappedAccountCryptographicState` is the correct solution.

IF needed, then you can make an accountKeys model in the same way you make the V2 model, only leave out the security state and signature key pair, and signed public key.

BitwardenShared/Core/Platform/Services/KeyConnectorService.swift
```diff
@@ -111,30 +119,56 @@ class DefaultKeyConnectorService {
...
+
+            try await stateService.setAccountEncryptionKeys(
+                AccountEncryptionKeys(
+                    cryptographicState: .v1(privateKey: keyConnectorResponse.keys.private),
+                    encryptedUserKey: keyConnectorResponse.encryptedUserKey,
+                ),
+            )
+            return KeyConnectorConversionResult(
+                encryptedUserKey: keyConnectorResponse.encryptedUserKey,
+                masterKey: keyConnectorResponse.masterKey,
```
fedemkr, line 152: 🤔  Is it safe to assume the `masterKey` won't change on server nor add anything? I see that previously the `masterKey` was fetched from server again before unlocking the vault https://github.com/bitwarden/ios/blob/main/BitwardenShared/Core/Auth/Repositories/AuthRepository.swift#L1093

matt-livefront: Good eye, I think the second fetch that we previously had was used out of convenience since the `unlockVaultWithKeyConnectorKey(keyConnectorURL:orgIdentifier:)` method already existed. The server doesn't change the key, so it's not necessary to re-fetch it.

fedemkr (approved):

BitwardenShared/Core/Platform/Services/KeyConnectorService.swift
fedemkr: Great, thanks Matt!

quexten (approved):

PR #2652, opened 2026-05-15
[PM-37604] tool: Add fix-ellipsis subcommand to fix-localizable-strings

KatherineInCode:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37604

## 📔 Objective

Adds a `fix-ellipsis` subcommand to the `fix-localizable-strings` script toolchain. The new module detects and converts three-dot sequences (`...`) in Localizable.strings values to the proper Unicode ellipsis character (`…`), matching existing style in our strings files.

The subcommand is wired into `fix-localizable-strings.sh` and runs against all English `.strings` files. Also fixes the one existing offender found in `BitwardenResources/Localizations/en.lproj/Localizable.strings` (`LoadingSubscription`).

### DIFF
--- a/Scripts/fix-localizable-strings/fix_ellipsis.py
+++ b/Scripts/fix-localizable-strings/fix_ellipsis.py
@@ -1,0 +1,77 @@
+"""
+fix_ellipsis
+
+Converts literal three-dot sequences (...) to the Unicode ellipsis character (…)
+in the values of Localizable.strings entries.
+"""
+
+import re
+
+# Matches a complete key/value entry line, capturing key and value separately.
+_ENTRY_VALUE_RE = re.compile(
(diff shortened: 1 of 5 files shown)


Scripts/fix-localizable-strings/fix_ellipsis.py
fedemkr, line 1: 🤔 Why do we need to split in key/value pairs to do this script? Can't we just find/replace all three dots strings for unicode ellipsis character directly?

KatherineInCode: So as not to disrupt the keys, if for some reason we have `...` in them. In practice I doubt it will matter, but I still like holding to the principle of the matter that the values are the things that matter in terms of display and "correct" characters

fedemkr (approved):

PR #2658, opened 2026-05-16
[PM-37654] fix: Route vault icon downloads through mTLS-aware URLSession

### LINKED ISSUE #2621: Vault item icons fail to load on self-hosted instance with mTLS (Client Certificate not sent during icon requests)
### Steps To Reproduce

1. Open the Bitwarden iOS app and configure it to connect to a self-hosted instance protected by mTLS.
2. Select a valid mTLS client certificate to allow the initial connection.
3. Log in to the Bitwarden instance using standard credentials (email and password).
4. Observe the vault list: none of the vault item/website icons are displayed.
5. Attempt to use the iOS Autofill feature; observe that icons are also missing in the Autofill UI.
6. Review the security/WAF logs (e.g., Cloudflare) enforcing the mTLS requirement. Observe a large number of rejected requests targeting the icon endpoints due to a missing or invalid client certificate.

Image

Image

### Expected Result

Because the app successfully utilizes the provided mTLS certificate for the primary API connection and authentication, it should also attach this same mTLS client certificate to the network requests responsible for downloading vault icons (both in the main app and the Autofill extension).

### Actual Result

The app fails to attach the selected mTLS client certificate when requesting vault icons. Consequently, the requests are blocked by the server's security rules, and the icons fail to load in both the main app and the Autofill UI.

### Screenshots or Videos

As seen the app shows not a single icon.

Image

### Additional Context

_No response_

### Build Version

2026.4.1 (3168)

Self-host

### Self-host Server Version

_No response_

### Environment Details

Device: iPad Pro, 11" (3. Generation)
iPad OS: 26.4.2

### Issue Tracking Info

jalenfran:
## 🎟️ Tracking

https://github.com/bitwarden/ios/issues/2621

## 📔 Objective

SwiftUI's `AsyncImage` loads through `URLSession.shared`, which never receives
client-certificate challenges. On self-hosted instances protected by mTLS, icon
requests were silently rejected while API requests succeeded.

Adds `CipherIconImageLoader`, a singleton `URLSession` delegate that forwards
client-certificate challenges to `ClientCertificateService` — the same service
used by `CertificateHTTPClient` for API requests. Replaces `AsyncImage` with
`CipherIconAsyncImage`, which routes through this loader.


BitwardenShared/Core/Platform/Services/CipherIconImageLoader.swift
```diff
@@ -0,0 +1,77 @@
...
+            }
+            return UIImage(data: data)
+        } catch {
+            return nil
+        }
+    }
+
+    // MARK: Private
+
+    private func handle(
```
matt-livefront, line 79: ⛏️ Would you mind adding docs to these properties and methods consistent with other classes?

BitwardenShared/Core/Platform/Services/CipherIconImageLoaderTests.swift
```diff
@@ -0,0 +1,113 @@
...
+    init() {
+        certificateService = MockClientCertificateService()
+        subject = CipherIconImageLoader()
+        subject.configure(certificateService: certificateService)
+    }
+
+    // MARK: Tests
+
+    @Test
+    func taskDidReceiveChallenge_nonClientCertificate_performsDefaultHandling() async {
```
matt-livefront, line 31: ⛏️ Can you add docs to these tests, similar to other tests?

BitwardenShared/Core/Platform/Services/CipherIconImageLoader.swift
```diff
@@ -0,0 +1,77 @@
...
+        _ challenge: URLAuthenticationChallenge,
+    ) async -> (URLSession.AuthChallengeDisposition, URLCredential?) {
+        switch challenge.protectionSpace.authenticationMethod {
+        case NSURLAuthenticationMethodClientCertificate:
+            guard let identity = await certificateService?.getClientCertificateIdentity() else {
```
matt-livefront, line 84: 🎨 Since `certificateService` needs to be set after initialization, this would end up silently failing if `configure(certificateService:)` was never called. It might be a good idea to also pass `errorReporter` into this type and use that to log an error if `certificateService` is `nil`.

You can use `BitwardenError.generalError`, something like this:
```
errorReporter.log(
    error: BitwardenError.generalError(
        type: "CipherIconImageLoader",
        message: "`configure(certificateService:)` must be called before loading icons.",
    )
)
```

matt-livefront (commented):
Thanks @jalenfran! These changes look good, I just have a few minor comments.

matt-livefront (approved):

matt-livefront:
Thanks @jalenfran!

PR #2661, opened 2026-05-18
[PM-23301] feat: Share user session key across process boundaries

fedemkr:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-23301

## 📔 Objective

Stores the decrypted user encryption key in the keychain after a successful vault unlock, so that app extensions (AutoFill, Action, Share) can auto-unlock the vault without requiring user interaction, as long as the active session timeout policy permits it.

**Key changes:**

- **`BitwardenKeychainItem.userSessionKey`** — new keychain item stored with `kSecAttrAccessibleWhenUnlockedThisDeviceOnly`, cleared on logout alongside other per-user items.
- **`SessionTimeoutValue.allowsUserSessionKeySharing`** — returns `true` for all timed timeout values; `false` for `.immediately`, `.onAppRestart`, and `.never` (which uses its own dedicated keychain path).
- **`AuthRepository.unlockVaultWithSessionKey()`** — reads the session key from the keychain and unlocks the vault. Returns `false` (not an error) if no key is stored; deletes the stale key and rethrows on a decryption failure.
- **Key lifecycle management** — the session key is written after every successful unlock, and deleted on: manual lock, session timeout, and when the user changes to a timeout value that excludes sharing.
- **`AuthRoute.completeWithUserSessionKey` / `AuthRouter+Redirects`** — new redirect branch that auto-unlocks via the session key when the vault is locked but a valid key exists, bypassing the unlock screen entirely.
- **`NotificationCenterService.willResignActivePublisher()`** — new publisher for `UIApplication.willResignActiveNotification`; `AppProcessor` subscribes to it to record the last-active timestamp, covering the foreground → app-switcher-kill path where `didEnterBackground` never fires.


speedy1804:
Hi @fedemkr @morganzellers-bw ,
I’m just curious and wanted to ask if and when development on this feature will continue. I feel this is a feature the app really needs, but I notice that nothing has happened since this pull request was created in may. Are other things taking priority right now?
I’d really appreciate a reply!
Best regards, Speedy

fedemkr:
Hi @speedy1804, thanks for the follow-up! Yes, this is still on our radar. We're going through some internal review before moving forward, so other priorities have shifted the timeline a bit. No firm ETA yet, but it hasn't been abandoned.

speedy1804:
@fedemkr alright, thanks for clarification! 👍

PR #2665, opened 2026-05-19
[PM-37746] Add fallback to local sdk package

LRNcardozoWDF:
## 🎟️ Tracking

[PM-37746](https://bitwarden.atlassian.net/browse/PM-37746)

## 📔 Objective

Fall back to a local sdk-internal checkout and locally built sdk package when "LOCAL_SDK=true bootstrap".

## 📸 Screenshots

image

image

[PM-37746]: https://bitwarden.atlassian.net/browse/PM-37746?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Scripts/generate-mocks.sh
```diff
@@ -43,16 +44,31 @@ fi
...
+    if [ -z "$BITWARDEN_SDK_PATH" ]; then
+        # Last resort: local sdk-internal checkout at the well-known sibling path.
+        # SRCROOT is set by Xcode in build phases; derive from script location for standalone runs.
+        _script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+        _repo_root="$(dirname "$_script_dir")"
+        _local_sdk_path="${SRCROOT:-$_repo_root}/../sdk-internal/crates/bitwarden-uniffi/swift"
+        if [ -d "$_local_sdk_path" ]; then
+            BITWARDEN_SDK_PATH="$(cd "$_local_sdk_path" && pwd)"
+        fi
     fi
```
fedemkr, line 66: 🤔 Is this needed? I mean wouldn't this end up being the exact same as the `BITWARDEN_SDK_PATH` path passed in `local-idk.yml -> settings`? Or which potential path case would this cover?

LRNcardozoWDF: Nice catch! You're right, lines 61-69 can be removed. Thank you

Scripts/generate-mocks.sh
fedemkr: ⛏️ I think this error message should be updated as `SourcePackages/` path is only for non-local flow.

fedemkr (approved):

PR #2667, opened 2026-05-19
[PM-7854] feat: Hide Send tab when disableSend policy is active

fedemkr:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-7854

## 📔 Objective

When the `disableSend` (`PolicyType.disableSend`) policy is enabled by an organization administrator, the Send tab is now completely removed from the tab bar rather than remaining visible with actions disabled.

**Changes:**

- **`TabCoordinator`** — Injects `PolicyService` and checks `policyAppliesToUser(.disableSend)` at startup and on every organization stream event, calling `updateTabs(isSendEnabled:)` to add or remove the Send tab reactively without requiring a restart.
- **`BitwardenTabBarController.navigator(for:)`** — Fixed to use a dictionary lookup instead of an array-index lookup, so `navigator(for: .generator(...))` returns the correct navigator when the Send tab is absent and the visual positions have shifted.

## 📸 Screenshots

No send tab


BitwardenShared/UI/Platform/Application/Views/BitwardenTabBarControllerTests.swift
```diff
@@ -98,6 +98,35 @@ class BitwardenTabBarControllerTests: BitwardenTestCase {
         XCTAssertIdentical(testController, viewController)
     }
 
+    /// `navigator(for:)` with a `TabRoute` uses the stored dictionary and returns the correct
+    /// navigator even when the Send tab is absent (non-contiguous indices).
+    @MainActor
+    func test_navigatorFor_tabRoute_sendHidden() throws {
```
KatherineInCode: 🤔 With how small this test file is, could we potentially just convert it to Swift Testing at this time? Or would that be too much, do you think?

BitwardenShared/UI/Platform/Tabs/TabCoordinator.swift
```diff
@@ -126,57 +151,70 @@ final class TabCoordinator: Coordinator, HasTabNavigator {
 
         rootNavigator.show(child: tabNavigator)
 
-        let vaultNavigator = module.makeNavigationController()
-        vaultNavigator.navigationBar.prefersLargeTitles = false
-        vaultNavigator.navigationBar.accessibilityIdentifier = "MainHeaderBar"
+        let vaultNav = module.makeNavigationController()
```
KatherineInCode, line 154: 🤔 It's probably overkill, but it might be nice to break out the other tabs' navigation controller initialization into helper methods, as well?

KatherineInCode (commented):
Looks good, just some thoughts

BitwardenShared/UI/Platform/Tabs/TabCoordinator.swift
fedemkr: I did it for the Send because it's used in more than one place but here it's just simple repetitive logic for each navigation; I think we wouldn't gain much.

BitwardenShared/UI/Platform/Application/Views/BitwardenTabBarControllerTests.swift
fedemkr: Done

BitwardenShared/UI/Platform/Tabs/TabCoordinator.swift
KatherineInCode: Yeah, it was just a thought. I think it's fine as-is

KatherineInCode (approved):

PR #2671, opened 2026-05-20
[PM-37697] feat: Show "Continue to web app?" alert when tapping Manage plan

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37697

## 📔 Objective

When tapping "Manage plan" on the Premium Plan screen, instead of opening Stripe in the browser, a "Continue to web app?" confirmation alert is shown. Tapping Continue opens the Bitwarden web app subscription page (`https://vault.bitwarden.com/#/settings/subscription`).

## 📸 Screenshots

https://github.com/user-attachments/assets/1e6afc51-10e1-4027-a7bd-55dbf50a106b

### DIFF
--- a/BitwardenShared/Core/Platform/Utilities/ExternalLinksConstants.swift
+++ b/BitwardenShared/Core/Platform/Utilities/ExternalLinksConstants.swift
@@ -32,6 +32,9 @@ extension ExternalLinksConstants {
     /// A link to Bitwarden's help page for learning more about Premium features.
     static let learnMoreAboutPremium = URL(string: "https://bitwarden.com/help/password-manager-plans/")!
 
+    /// A link to the subscription management page in the Bitwarden web app.
+    static let manageSubscription = URL(string: "https://vault.bitwarden.com/#/settings/subscription")!
+
     /// A link to the password options within the passwords section of the settings menu.
     static let passwordOptions = URL(string: "App-prefs:PASSWORDS&path=PASSWORD_OPTIONS")!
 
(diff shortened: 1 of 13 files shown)


BitwardenShared/Core/Platform/Utilities/ExternalLinksConstants.swift
```diff
@@ -32,6 +32,9 @@ extension ExternalLinksConstants {
     /// A link to Bitwarden's help page for learning more about Premium features.
     static let learnMoreAboutPremium = URL(string: "https://bitwarden.com/help/password-manager-plans/")!
 
+    /// A link to the subscription management page in the Bitwarden web app.
+    static let manageSubscription = URL(string: "https://vault.bitwarden.com/#/settings/subscription")!
```
fedemkr: ⚠️ This should be based on the environment's web vault URL and not hardcoded to `vault.bitwarden.com`.

fedemkr (changes requested):

BitwardenShared/Core/Platform/Utilities/ExternalLinksConstants.swift
andrebispo5: In this particular case it's not really a problem since this feature is only available to Cloud Us users. But I'll make the change.

fedemkr: Thanks Andre, yes I fell it's good to have this kind of things already covered if we ever change this to be available elsewhere.

fedemkr (approved):

PR #2674, opened 2026-05-20
[PM-37811] llm: Add audit build warnings command

fedemkr:
## 🎟️ Tracking

[PM-37811](https://bitwarden.atlassian.net/browse/PM-37811)

## 📔 Objective

Adds an LLM command `audit-build-warnings` to check build warnings, categorize them and try to fix them.

[PM-37811]: https://bitwarden.atlassian.net/browse/PM-37811?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.claude/commands/audit-build-warnings.md
```diff
@@ -0,0 +1,136 @@
...
+
+Accept free-form input, or "none" / "proceed as-is".
+
+Record the final exclusion list.
+
+## Phase 5: Fix
+
+Apply fixes in this exact order:
+
+### Step 1 — SwiftFormat auto-fix (ALWAYS first, BEFORE reading any files)
```
matt-livefront: 🤔 Does it make sense to do this before anything else? Anything that could be fixed by SwiftFormat/SwiftLint seems pretty trivial to fix and wouldn't need to be included in the report/considered for exclusions. Or is useful to have these in the report?

❓ Swiftlint also has a way to fix warnings (`mint run swiftlint lint --fix`). Should we also include that?

.claude/commands/audit-build-warnings.md
```diff
@@ -0,0 +1,136 @@
...
+```bash
+DEVICE=$(tr -d '\n' < .test-simulator-device-name)
+OS=$(tr -d '\n' < .test-simulator-ios-version)
+```
+
+## Phase 1: Build & Capture
+
+Run the build and capture all output to a temp file:
+```bash
+xcodebuild build \
```
matt-livefront: ❓ Does it make sense to use `build-for-testing` to capture any build warnings from the tests?

.claude/commands/audit-build-warnings.md
fedemkr: I think that's a great idea! I'll make the change.

.claude/commands/audit-build-warnings.md
fedemkr: Yes, missed that!

.claude/commands/audit-build-warnings.md
fedemkr: Done and just in case this is why swift format is run twice:
SwiftLint's `--fix` mode auto-corrects a narrower set of rules than SwiftFormat — things like replacing unused closure parameters with `_`, removing redundant `let _ =` discards, and deleting superfluous disable comments. These are just as mechanical as SwiftFormat's changes and require no judgment, so they belong here rather than in the interactive fix phase. SwiftFormat runs a second time after SwiftLint because SwiftLint's corrections can introduce minor formatting drift — for example, it may insert or remove whitespace in ways that technically violate a SwiftFormat rule. The second SwiftFormat pass catches that drift and guarantees the codebase is in a clean, consistent state before the build runs.

matt-livefront (dismissed):

.claude/commands/audit-build-warnings.md
```diff
@@ -0,0 +1,138 @@
...
+### ⚪ Default Skips (propose to user unless they override)
+These categories are typically intentional or in third-party code — default to skipping them:
+- **Vendored / third-party code**: any file under paths like `BitwardenWatchShared/MessagePack/`, `**/Vendor/`, or other vendored directories
+- **Deprecated API warnings** where the replacement requires non-trivial SDK changes (flag these for the user to decide)
+- **Always-true / always-false casts** (`'as' test is always true`) — may be intentional defensive guards
+- **Protocol near-match** warnings in third-party code
+
+## Phase 3: Report
+
+Write a temporary markdown report to `/tmp/build_warnings_report.md` (NOT to `Docs/` or any tracked directory — this is an ephemeral audit artifact).
```
KatherineInCode, line 75: 🤔 Even though I as the user shouldn't be required to open the file, I still would like it to be in an easy-to-access location; I [instruct Claude](https://github.com/KatherineInCode/Dotfiles/blob/main/dotfiles/.claude/CLAUDE.md#output-files) to put all of those in the `.claude/output` directory, which is similarly ephemeral and untracked. I'm not sure what Claude would generate given those two strings of instructions, but it would be nice if we had some way of managing "if the user has a place they want it, use that; otherwise put it here" sorts of instructions

fedemkr: As far as I see `.claude/output` is not included in `.gitignore` so currently it's being tracked.
I think we had a discussion to leave `output` tracked in case we wanted to share requirements/plans in the repo.
We can revisit that discussion or create a `.claude/tmp` folder to put these outputs. What do you think?

KatherineInCode: Ah, right, I have that untracked in [my global gitignore](https://github.com/KatherineInCode/Dotfiles/blob/main/dotfiles/.gitignore#L44) 😅 

This feels like the sort of thing @bitwarden/team-ai-sme should weigh in on, and see if we can get some sort of consistent standard across all the Bitwarden repos 🤔 

But I very much do want there to be some `.claude/[something]` folder in each repo that's ephemeral and untracked, for this sort of thing. I'm just very ambivalent on what it is, so long as it's the standard.

fedemkr: Hi there, just raising this up @bitwarden/team-ai-sme as some time has passed in case you've missed the notification and/or new standards are already in place for this that we may need to add here.

.claude/commands/audit-build-warnings.md
SaintPatrck, line 1: ⚠️ Let's move this to `.claude/skills/audit-build-warnings/SKILL.md`. Skills have replaced slash commands, and encapsulating it in a dedicated directory will keep things better organized if/when additional files like scripts or reference examples are added. It also gives us a place to put outputs and easily `.gitignore` them.

.claude/commands/audit-build-warnings.md
```diff
@@ -0,0 +1,138 @@
+---
+description: Build the Bitwarden iOS app, capture all compiler and lint warnings, categorize them into Swift 6 concurrency vs actionable, and interactively fix the actionable ones.
+argument-hint: "[scheme] — defaults to Bitwarden"
```
SaintPatrck, line 3: ⛏️ Would be helpful to give examples of the support arguments.

.claude/commands/audit-build-warnings.md
```diff
@@ -0,0 +1,138 @@
...
+- If `$ARGUMENTS` is provided, use it as the scheme name (e.g., `Authenticator`, `TestHarness`).
+- Otherwise default to `Bitwarden`.
+
+Read simulator config — always read these files, never hardcode values:
+```bash
+DEVICE=$(tr -d '\n' < .test-simulator-device-name)
+OS=$(tr -d '\n' < .test-simulator-ios-version)
+```
+
+## Phase 0: Pre-process (auto-fix before building)
```
SaintPatrck, line 20: ⛏️ Can we start with Phase 1?

.claude/commands/audit-build-warnings.md
```diff
@@ -0,0 +1,138 @@
...
+
+Run all deterministic auto-fixers *before* the build so their warnings never appear in the report.
```
SaintPatrck, line 22: 🌱 🎨 Most of Phase 0 is executing deterministic bash commands. Maybe as a followup, consider wrapping these individual commands into a single script that prints what Claude needs to continue?

My main reason for asking is so that we can add it to `allowed-tools` in the front matter.

```yaml
allowed-tools: Bash(scripts/audit-build-warnings-prep.sh), Read, Grep
```

.claude/commands/audit-build-warnings.md
SaintPatrck: Since this is a project-level skill, I think you can take whatever approach works best for the team. This is more important to standardize in the context of plugin outputs because they're not tied to a particular project so we can point ephemeral docs to a location that isn't tracked by default, like `${CLAUDE_SKILL_DIR}/reports/*` or `${CLAUDE_PLUGIN_DATA}/<skill-name/*` which resolve to something like `~/.claude/plugins/path/to/marketplace/`.

If you want them outside of the repo and don't want to add `.gitignore`, `tmp/` has become the defacto location. Otherwise, I see no harm in adding a `.gitignore` entry to a project-level location you all decide on.

SaintPatrck (changes requested):
Looks good. One blocking request; relocate this to the `skills` dir. The other comments are nitpicks and suggestions I leave up to your discretion.

.claude/commands/audit-build-warnings.md
fedemkr: Roger, moved it to `.claude/tmp` folder and git ignored it.

.claude/commands/audit-build-warnings.md
fedemkr: I think you mean that the Phases starts from 1 instead of 0, right? If that's the case, that's done.

.claude/skills/audit-build-warnings/SKILL.md
```diff
@@ -0,0 +1,138 @@
...
+
+Read simulator config — always read these files, never hardcode values:
+```bash
+DEVICE=$(tr -d '\n' < .test-simulator-device-name)
+OS=$(tr -d '\n' < .test-simulator-ios-version)
+```
+
+## Phase 0: Pre-process (auto-fix before building)
+
+Run all deterministic auto-fixers *before* the build so their warnings never appear in the report.
```
fedemkr: Great suggestion, done and added the next as well given that's the core run of the tool.
```
  - Bash(mint run swiftlint:*)
  - Bash(mint run swiftformat:*)
```

SaintPatrck (dismissed):

.claude/skills/audit-build-warnings/SKILL.md
```diff
@@ -0,0 +1,150 @@
+---
+name: audit-build-warnings
+description: Build the Bitwarden iOS app, capture all compiler and lint warnings, categorize them into Swift 6 concurrency vs actionable, and interactively fix the actionable ones. Use when asked to "audit build warnings", "fix warnings", "clean up warnings", "build warnings", or to reduce compiler noise before a PR.
+allowed-tools:
+  - Bash(Scripts/audit-build-warnings-prep.sh*)
+  - Bash(mint run swiftlint:*)
+  - Bash(mint run swiftformat:*)
+  - Read
+  - Grep
```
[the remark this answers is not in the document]
SaintPatrck: Incorrect.

SaintPatrck (dismissed):
LGTM. https://github.com/bitwarden/ios/pull/2674#discussion_r3766839079 may be valid, but seems like an edge case. 🤷

.claude/skills/audit-build-warnings/SKILL.md
```diff
@@ -0,0 +1,150 @@
...
+
+Run the build for testing and capture all output to a temp file:
+```bash
+mkdir -p .claude/tmp
+xcodebuild build-for-testing \
+  -project <SCHEME>.xcodeproj \
+  -scheme <SCHEME> \
+  -destination "platform=iOS Simulator,name=$DEVICE,OS=$OS" \
+  -configuration Debug \
+  2>&1 | tee .claude/tmp/bitwarden_build_output.txt
```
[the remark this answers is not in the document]
fedemkr: This has already been resolved.

SaintPatrck (approved):

matt-livefront (approved):

PR #2677, opened 2026-05-20
[PM-37819] refactor: Persist WrappedAccountCryptographicState directly

matt-livefront:
## 🎟️ Tracking

[PM-37819](https://bitwarden.atlassian.net/browse/PM-37819)

## 📔 Objective

Previously, we stored the user's account keys model and private key and constructed a `WrappedAccountCryptographicState` from these when unlocking the vault.

This worked fine when we always received these from the API. Now that the user registration methods are moving into the SDK, the SDK returns `WrappedAccountCryptographicState` at registration time, which needs to be persisted. Converting this back into a local model introduces unnecessary translation layers when we could just persist `WrappedAccountCryptographicState` directly.

This updates the app to construct a `WrappedAccountCryptographicState` from the API which is then persisted and used for vault unlock. When the registration methods are moved to the SDK, we'll be able to persist the returned `WrappedAccountCryptographicState` directly as well.

[PM-37819]: https://bitwarden.atlassian.net/browse/PM-37819?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenShared/Core/Platform/Services/StateServiceTests.swift
```diff
@@ -1658,17 +1621,13 @@ class StateServiceTests: BitwardenTestCase { // swiftlint:disable:this type_body
         let account = Account.fixture(profile: .fixture(userId: "1"))
         await subject.addAccount(account)
         try await subject.setAccountEncryptionKeys(AccountEncryptionKeys(
-            accountKeys: .fixture(
-                publicKeyEncryptionKeyPair: .fixture(wrappedPrivateKey: "1:WRAPPED_PRIVATE_KEY"),
-            ),
-            encryptedPrivateKey: "1:PRIVATE_KEY",
+            cryptographicState: .v1(privateKey: "1:PRIVATE_KEY"),
             encryptedUserKey: "1:USER_KEY",
```
quexten, line 1625: Out of scope here but most likely you can drop encryptedUserKey? We don't use this terminology in the SDK or `clients` - there it is the `masterKeyWrappedUserKey`. However, this is already included as part of the higher level `MasterPasswordUnlockData`. Mobile should eventually just persist this struct, which is also included as part of the sync.

quexten (dismissed):
Sweet! Looks good. I wonder whether we can also already drop the separate tracking of `privateKey` in some places. (At least I don't see a reason why we can't, which would simplify yet a bit further).

BitwardenShared/Core/Platform/Services/StateServiceTests.swift
matt-livefront: We do persist `MasterPasswordUnlockData`. It looks `encryptedUserKey` is used to store the user's key connector encrypted user key (`keyConnectorKeyWrappedUserKey`) - would you suggest just storing that as it's own thing?

We also have this logic when logging in with another device, can that be replaced with `masterKeyWrappedUserKey`? Given the master password hash check, I think that would only be used if the user has a master password (and hence a `masterKeyWrappedUserKey`)?
https://github.com/bitwarden/ios/blob/639b358deb574c04bf6d4f8f3a4b0b97c829064d/BitwardenShared/Core/Auth/Repositories/AuthRepository.swift#L974-L976

matt-livefront:
> I wonder whether we can also already drop the separate tracking of `privateKey` in some places. (At least I don't see a reason why we can't, which would simplify yet a bit further).

@quexten Thanks! I'll look into this. I think we still have this as a fallback to handle backwards compatibility with older self-hosted servers, but maybe it's been long enough that we could remove it.

BitwardenShared/Core/Platform/Services/StateServiceTests.swift
quexten: > would you suggest just storing that as it's own thing?
KM is slowly separating out the key-connector-key-wrapped-user-key from the master-key-wrapped-user-key. Yes, storing it separately makes sense from my perspective. We don't do this on the sync yet, but at some point we will.

However, that said, you don't need to store it at all. The keyConnectorWrappedUserKey is only ever used during initial *login* and never afterwards, and so it doesn't need to be persisted. Essentially, you should:
- Auth via SSO
  - This returns back the key-connector-key-wrapped-user-key
- Unlock SDK with the `keyConnectorUrl` unlock method:
  - https://github.com/bitwarden/sdk-internal/blob/5aae79b3e972129ebb106155f8de7bb76d748f33/crates/bitwarden-core/src/key_management/crypto.rs#L168
  - Takes the key connector url and the keyconnectorkeywrappeduserkey

Then, you have your unlocked SDK instance without persisting anything. Note, that the above unlock method was only recently added. In the medium to long term, KM would like to consolidate the unlock logic that currently auth has to maintain into the SDK, so that auth can drop the logic, and this would be a step towards it, but we did not do a close investigation of what's necessary on mobiles side.


> We also have this logic when logging in with another device, can that be replaced with masterKeyWrappedUserKey? Given the master password hash check, I think that would only be used if the user has a master password (and hence a masterKeyWrappedUserKey)?

The masterKey auth request method is actually no longer supported. Feel free to drop the code as it is unused.

matt-livefront: Ah, that's a great point, thanks! I will create a task for those changes.

BitwardenShared/Core/Platform/Services/MigrationService.swift
```diff
@@ -254,6 +255,39 @@ class DefaultMigrationService {
...
+        for (accountId, _) in state.accounts {
+            let privateKeyStorageKey = "bwPreferencesStorage:encPrivateKey_\(accountId)"
+            guard let privateKey = appGroupUserDefaults.string(forKey: privateKeyStorageKey) else { continue }
+
+            let accountKeysStorageKey = "bwPreferencesStorage:accountKeys_\(accountId)"
+            let accountKeys: PrivateKeysResponseModel? = {
+                guard let string = appGroupUserDefaults.string(forKey: accountKeysStorageKey),
+                      let data = string.data(using: .utf8)
+                else { return nil }
+                return try? JSONDecoder().decode(PrivateKeysResponseModel.self, from: data)
```
fedemkr: 🤔 Would it be useful to log the error here if thrown?

fedemkr (approved):
Looks good on my end. Approving as I think the other comments will be addressed on a future PR.

PR #2681, opened 2026-05-21
[PM-37569] chore: Remove cipher key encryption feature flag

morganzellers-bw:
## 🎟️ Tracking

[PM-37569](https://bitwarden.atlassian.net/browse/PM-37569)

## 📔 Objective

Remove any references to the `enableCipherKeyEncryption` feature flag
`cipher-key-encryption`

## 📸 Screenshots

[PM-37569]: https://bitwarden.atlassian.net/browse/PM-37569?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenShared/Core/Platform/Services/ClientService.swift
```diff
@@ -311,14 +311,7 @@ actor DefaultClientService: ClientService {
...
 
-            let cipherKeyEncryptionFlagEnabled: Bool = await configService.getFeatureFlag(
-                .cipherKeyEncryption,
-            )
-            let enableCipherKeyEncryption = cipherKeyEncryptionFlagEnabled && config.supportsCipherKeyEncryption()
-
-            try await client.platform().loadFlags([
-                FeatureFlag.enableCipherKeyEncryption.rawValue: enableCipherKeyEncryption,
-            ])
+            try await client.platform().loadFlags([:])
```
matt-livefront: 🤔 I could be wrong, but I think we still need the server version check (if the server doesn't support cipher key encryption, it won't save the key that is now encrypting your cipher). I understood this ticket to just remove the LD flag (`FeatureFlag.cipherKeyEncryption`).

morganzellers-bw: Ah okay, I think you're right. I'll update this to just pass 
`FeatureFlag.enableCipherKeyEncryption.rawValue: config.supportsCipherKeyEncryption(),`
to `loadFlags()`

fedemkr: We usually should only support up to 3 major versions client/server side and many more than that have passed since  `cipherKeyEncryptionMinServerVersion -> 2024.2.0` so it may actually be safe to be removed as well. However, @morganzellers-bw  double check with KM team to see if that would be ok.

matt-livefront: @fedemkr What happens if we encrypt a cipher with a cipher key but the server doesn't yet support cipher key encryption? Won't it drop the cipher key when the cipher is added/updated since it doesn't know about it? And then you wouldn't be able to decrypt that cipher?

fedemkr: Indeed, but that would only happen if you have a "very" old self-hosted server without it being updated which goes beyond what we support https://bitwarden.com/help/bitwarden-software-release-support/. That's why I said to check with KM team in case this is an exception that should be kept in place for the reason you mentioned.

morganzellers-bw: I pinged @mzieniukbw to take a look here when they have a chance 👍

mzieniukbw: @bitwarden/team-vault-dev I think this question is for you.

gbubemismith: @morganzellers-bw, this should be okay to remove

shane-melton: There's no exception as far as I'm aware, Vault had just left it alone while we focused on re-enabling cipher key encryption. Now that its enabled for everyone and flags are being removed, we should be safe to remove this flag as well and clean up the tech debt.

morganzellers-bw: Thanks @shane-melton & @gbubemismith!

fedemkr: 🤔 Is cipher key encryption enabled by default when not passing the flag?

morganzellers-bw: I see, we still want to pass the flag and value, it can just be hardcoded as true now?

morganzellers-bw: @fedemkr It looks like the SDK defaults this value to false if not set, so I'll keep it in here hardcoded 👍 

[crates/bitwarden-core/src/client/flags.rs:41-45](https://github.com/bitwarden/sdk-internal/blob/2a05085049bf161ad0bfd2b28184c602a4c6db1e/crates/bitwarden-core/src/client/flags.rs)

```rust
fn test_load_empty_map() {
    let map = std::collections::HashMap::new();
    let flags = Flags::load_from_map(map);
    assert!(!flags.enable_cipher_key_encryption);  // false
}
```

fedemkr (approved):

BitwardenShared/Core/Platform/Services/ClientService.swift
fedemkr: Great thanks for checking.

PR #2682, opened 2026-05-21
llm: Document line wrapping style in CLAUDE.md

KatherineInCode:
## 🎟️ Tracking

Internal CLAUDE.md maintenance — no Jira ticket.

## 📔 Objective

Documents the preferred line wrapping style for Swift code in CLAUDE.md: use Xcode's ⌃M "Format to Multiple Lines" style (opening delimiter on the first line, each argument on its own indented line with trailing comma, closing delimiter on its own line).

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -85,6 +85,20 @@ mint run swiftlint                      # Lint
 typos                                   # Spell check
 ```
 
+### Line Wrapping
+When a call exceeds the line length limit, use Xcode's ⌃M "Format to Multiple Lines" style — opening delimiter stays on the first line, each argument on its own indented line with a trailing comma, closing delimiter on its own line:
+```swift
+// ✅ Correct
+someFunction(
+    argumentOne: valueOne,
+    argumentTwo: valueTwo,
+)
+
+// ❌ Wrong — aligning to opening delimiter
+someFunction(argumentOne: valueOne,
+             argumentTwo: valueTwo)
+```
+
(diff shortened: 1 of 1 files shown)


SaintPatrck (approved):

.claude/CLAUDE.md
```diff
@@ -85,6 +85,20 @@ mint run swiftlint                      # Lint
 typos                                   # Spell check
 ```
 
+### Line Wrapping
+When a call exceeds the line length limit, use Xcode's ⌃M "Format to Multiple Lines" style — opening delimiter stays on the first line, each argument on its own indented line with a trailing comma, closing delimiter on its own line:
```
fedemkr, line 89: 🤔  Given that this is the `CLAUDE.md` file, does it add value for Claude to say "use Xcode's ⌃M "Format to Multiple Lines" style" instead of "use this style" as it's explicitly explained afterwards? Haven't tried myself locally but perhaps Claude tries to find what "Xcode's format to multiple lines" means instead of just applying what you explain afterwards potentially wasting resources.

KatherineInCode: In the usage that led to this, it did not appear to look it up at that time (aside from, I think, looking up what it meant when I told it to format in that manner). However, I thought it valuable to note that in case it comes across a case not covered by the statement, so that it *could* go look up how it performs. That is, it's a case of "here is the general rule, here is a common concrete appellation of the rule".

fedemkr (approved):

PR #2691, opened 2026-05-22
[PM-24114] fix: Prevent noActiveAccount errors when viewing pre-login Settings

matt-livefront:
## 🎟️ Tracking

[PM-24114](https://bitwarden.atlassian.net/browse/PM-24114)

## 📔 Objective

Fixes a few `noActiveAccount` errors that end up being logged to Crashlytics when viewing pre-login Settings:

- The `badgeUpdateTask` doesn't need to be created for pre-login since badgers are only shown after logged in.
-  The premium check can be skipped because plan information isn't shown in pre-login.

[PM-24114]: https://bitwarden.atlassian.net/browse/PM-24114?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


fedemkr (approved):
Nice catch!

PR #2692, opened 2026-05-22
[PM-37961] feat: Handle policiesNew and organizationsNew

fedemkr:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37961

## 📔 Objective

Adds two new optional fields to the sync response models to support server-side policy/org accepted-state filtering (PM-34145):

- **`SyncResponseModel.policiesNew`** — a new optional array of `PolicyResponseModel` that the server may return instead of (or alongside) the legacy `policies` field.
- **`ProfileResponseModel.organizationsNew`** — a new optional array of `ProfileOrganizationResponseModel` that the server may return alongside the legacy `organizations` field.

Both fields are decoded from the sync response JSON and stored **independently** in separate data stores:

- `response.policies` → `PolicyData` CoreData entity (unchanged, existing flow)
- `response.policiesNew ?? []` → `PolicyNewData` CoreData entity (new — shares the `PolicyData` Swift class via `representedClassName`, distinguished at the CoreData level by entity name)

A `replacePoliciesNew` method is added to `PolicyDataStore` and `PolicyService` to handle the new entity. `DataStore.deleteDataForUser` is updated to clean up `PolicyNewData` rows on user logout.

Also includes the `policiesInAcceptedState` feature flag (`pm-34145-policies-in-accepted-state`) needed by the upcoming routing logic.


BitwardenShared/Core/Vault/Services/SyncService.swift
```diff
@@ -458,6 +458,7 @@ extension DefaultSyncService {
         try await sendService.replaceSends(response.sends, userId: userId)
         try await settingsService.replaceEquivalentDomains(response.domains, userId: userId)
         try await policyService.replacePolicies(response.policies, userId: userId)
+        try await policyService.replacePoliciesNew(response.policiesNew ?? [], userId: userId)
```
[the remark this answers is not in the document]
fedemkr: Updated title

BitwardenShared/Core/Vault/Models/Domain/Policy.swift
```diff
@@ -43,6 +43,7 @@ extension Policy {
         }
         self.init(responseModel: model)
     }
+
```
matt-livefront: ⛏️ Extra newline?
```suggestion
```

BitwardenShared/Core/Vault/Models/Data/PolicyData.swift
```diff
@@ -69,3 +69,52 @@ extension PolicyData {
...
+        _ policies: [PolicyResponseModel],
+        userId: String,
+    ) -> NSBatchInsertRequest {
+        var index = 0
+        return NSBatchInsertRequest(entityName: policiesNewEntityName) { (object: NSManagedObject) -> Bool in
+            guard index < policies.count else { return true }
+            defer { index += 1 }
+            (object as? PolicyData)?.update(with: policies[index], userId: userId)
+            return false
+        }
```
matt-livefront: 🎨 Is it worth adding an optional overload to `ManagedObject.batchInsertRequest()` that accepts a custom entity name? It seems like that's the primary difference here. Then this could be updated to:

```
try batchInsertRequest(entityName: policiesNewEntityName, objects: policies) { policyData, policy in
    policyData.update(with: policy, userId: userId)
}
```

BitwardenKit/Core/Platform/Models/Data/ManagedObject.swift
```diff
@@ -14,6 +14,31 @@ public extension ManagedObject where Self: NSManagedObject {
...
+    ///   - objects: The objects (or objects that can be converted to managed objects) to insert.
+    ///   - handler: A handler that is called for each object to set the properties on the
+    ///     `NSManagedObject` to insert.
+    /// - Returns: A `NSBatchInsertRequest` for batch inserting an array of objects.
+    ///
+    static func batchInsertRequest<T>(
+        entityName: String,
+        objects: [T],
+        handler: @escaping (Self, T) -> Void,
+    ) -> NSBatchInsertRequest {
```
matt-livefront, line 30: 🤔 Does it make sense to make this throwing similar to `batchInsertRequest(objects:handler:)`? I was picturing this method containing the implementation of the existing `batchInsertRequest` (with the entity name applied) and then that method could call this with `batchInsertRequest(entityName: entityName ...)` so they would share the same implementation.

fedemkr: I started this like that and then refactored it given that I didn't need to throw anything here. However, if you feel it could be better reused that way I can make the change.

matt-livefront: That makes sense, and this implementation is pretty straightforward. Having it not throw if errors aren't possible is a better experience as well. We can always refactor it if we end up needing a throwing variant.

matt-livefront (approved):

PR #2693, opened 2026-05-22
[PM-37962] feat: Decode providerOrganizations and surface isProviderUser

fedemkr:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37962

## 📔 Objective

Adds provider-relationship awareness to the organization models decoded during sync.

**New model:**
- `ProfileProviderOrganizationResponseModel` — lightweight model decoded from `profile.providerOrganizations` in the sync response; currently carries only an `id`.

**Extended models:**
- `ProfileResponseModel` — adds optional `providerOrganizations: [ProfileProviderOrganizationResponseModel]?` (alphabetical position: after `privateKey`).
- `ProfileOrganizationResponseModel` — adds optional `isProviderUser: Bool?` to indicate the user's provider relationship to this org.
- `Organization` (domain model) — mirrors `isProviderUser` from the response model.

**Sync wiring:**
Replaces the direct `response.profile?.organizations` lookup in `DefaultSyncService.fetchSync` with a `resolveEffectiveOrganizations(from:)` private helper that:
1. Prefers `organizationsNew` over legacy `organizations` when present.
2. Coalesces `isProviderUser = true` onto any org whose ID appears in `providerOrganizations`.

This PR builds on top of #2692 (PM-37961).


BitwardenShared/Core/Vault/Services/SyncService.swift
```diff
@@ -477,6 +477,30 @@ extension DefaultSyncService {
...
     }
 
+    /// Resolves the effective list of organizations from a sync response.
+    ///
+    /// Prefers `organizationsNew` (accepted-state members) over the legacy `organizations` list,
+    /// and coalesces `isProviderUser` by checking membership in `providerOrganizations`.
+    ///
+    /// - Parameter response: The sync response from the server.
+    /// - Returns: The effective organizations, or `nil` when neither list is present.
+    private func resolveEffectiveOrganizations(
```
matt-livefront: ⛏️ Could this be organized with the other private methods down below?

🎨 Alternatively, we could move this out of `SyncService` too, maybe into `SyncResponseModel`?

BitwardenShared/Core/Vault/Models/Response/ProfileResponseModelTests.swift
```diff
@@ -0,0 +1,101 @@
+import XCTest
```
matt-livefront: 🎨 Want to switch these new tests to use Swift Testing?

BitwardenShared/Core/Vault/Services/SyncServiceTests.swift
```diff
@@ -1364,6 +1364,98 @@ class SyncServiceTests: BitwardenTestCase { // swiftlint:disable:this type_body_
         XCTAssertTrue(policyService.replacePoliciesNewPolicies.isEmpty)
     }
 
+    // MARK: - organizationsNew fallback tests
+
+    /// `fetchSync()` passes `organizationsNew` to `replaceOrganizations` when the field is present,
+    /// ignoring the legacy `profile.organizations`.
+    func test_fetchSync_organizationsNew_newOverridesLegacyOrganizations() async throws {
```
matt-livefront, line 1373: 🎨 These tests overlap with the ones in `ProfileResponseModelTests`. Does it make sense to pair these down and rely on most of the tests in `ProfileResponseModelTests`?

BitwardenShared/Core/Vault/Services/API/Fixtures/APITestData+Vault.swift
```diff
@@ -15,6 +15,8 @@ public extension APITestData {
     static let syncWithCiphersCollections = loadFromJsonBundle(resource: "syncWithCiphersCollections")
     static let syncWithDomains = loadFromJsonBundle(resource: "syncWithDomains")
     static let syncWithNewAndLegacyFields = loadFromJsonBundle(resource: "syncWithNewAndLegacyFields")
+    static let syncWithOrganizationsNew = loadFromJsonBundle(resource: "syncWithOrganizationsNew")
+    static let syncWithProviderOrganization = loadFromJsonBundle(resource: "syncWithProviderOrganization")
```
KatherineInCode: ⛏️ Alphabetization

BitwardenShared/Core/Vault/Services/SyncServiceTests.swift
fedemkr: I think we should keep both as even though the model is already testing the logic, how do we make sure that the service is actually calling the model for such logic and is ok? It's a bit redundant but not sure how we'd assert that instead.

matt-livefront: Ok, makes sense.

matt-livefront (dismissed):

KatherineInCode (dismissed):

matt-livefront (approved):

PR #2694, opened 2026-05-22
[PM-37995] chore: Add pre-commit hook to lint changed Swift files

matt-livefront:
## 🎟️ Tracking

[PM-37995](https://bitwarden.atlassian.net/browse/PM-37995)

## 📔 Objective

Extends the existing pre-commit hook to run SwiftFormat and SwiftLint on staged Swift files before each commit, catching formatting and lint violations locally before they reach CI.

- Adds `Scripts/lint-changes.sh` — runs SwiftFormat (format check or fix) and SwiftLint (lint or fix + re-check) on the set of staged Swift files, detected via `$GIT_INDEX_FILE`
- Can be run standalone with `./Scripts/lint-changes.sh` (checks modified files) or `./Scripts/lint-changes.sh --fix` (auto-fixes what it can, then reports any remaining violations that require manual fixes)
- Wires the script into `Scripts/pre-commit` after the existing spell check step
- Uses `--diff-filter=ACMR` to cover renamed files (not just added/modified/copied)

[PM-37995]: https://bitwarden.atlassian.net/browse/PM-37995?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Scripts/lint-changes.sh
+++ b/Scripts/lint-changes.sh
@@ -16,14 +16,9 @@ if [ "${1:-}" == "--fix" ]; then
     FIX_MODE=true
 fi
 
-# Determine if we're running as a pre-commit hook
-if [ -n "${GIT_INDEX_FILE:-}" ]; then
-    # Running as pre-commit hook - check staged files
-    SWIFT_FILES=$(git diff --cached --name-only --diff-filter=ACMR | grep "\.swift$" || true)
-else
-    # Running standalone - check all modified files (staged and unstaged)
-    SWIFT_FILES=$(git diff --name-only --diff-filter=ACMR HEAD | grep "\.swift$" || true)
-fi
+# Staged files only when running as pre-commit hook, all changed files when standalone
+DIFF_ARGS=$( [ -n "${GIT_INDEX_FILE:-}" ] && echo "--cached" || echo "HEAD" )
+SWIFT_FILES=$(git diff --name-only $DIFF_ARGS --diff-filter=ACMR | grep "\.swift$" || true)
 
 # Exit early if no Swift files changed
 if [ -z "$SWIFT_FILES" ]; then
--- a/Scripts/lint-changes.sh
+++ b/Scripts/lint-changes.sh
@@ -1,0 +1,110 @@
+#!/bin/bash
+#
+# Script to run SwiftLint and SwiftFormat on changed Swift files
+# Can be used standalone or as a git pre-commit hook
+# Usage: ./Scripts/lint-changes.sh [--fix]
+
+set -euo pipefail
+
+# Track failures
+SWIFTFORMAT_FAILED=0
+SWIFTLINT_FAILED=0
+
+# Parse command line arguments
+FIX_MODE=false
+if [ "${1:-}" == "--fix" ]; then
+    FIX_MODE=true
+fi
+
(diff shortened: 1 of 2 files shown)


Scripts/lint-changes.sh
```diff
@@ -0,0 +1,110 @@
...
+fi
+
+# Determine if we're running as a pre-commit hook
+if [ -n "${GIT_INDEX_FILE:-}" ]; then
+    # Running as pre-commit hook - check staged files
+    SWIFT_FILES=$(git diff --cached --name-only --diff-filter=ACMR | grep "\.swift$" || true)
+else
+    # Running standalone - check all modified files (staged and unstaged)
+    SWIFT_FILES=$(git diff --name-only --diff-filter=ACMR HEAD | grep "\.swift$" || true)
+fi
```
fedemkr: ⛏️ I think this can be written shorter as:
```bash
# Running as pre-commit hook - check staged files
CACHED=$( [ -n "${GIT_INDEX_FILE:-}" ] && echo "--cached" || echo "HEAD" )
SWIFT_FILES=$(git diff --name-only $CACHED --diff-filter=ACMR | grep "\.swift$" || true)
```

Dunno if there could be a better name for _CACHED_ 🤔

Scripts/pre-commit
```diff
@@ -1,9 +1,9 @@
 #!/bin/bash
 #
-# Pre-commit hook that runs spell check on staged files
+# Pre-commit hook that runs spell check and linting on staged files
 # This hook is set up automatically by Scripts/bootstrap.sh
 
-PATH=/usr/local/bin:$PATH
+PATH=/opt/homebrew/bin:/usr/local/bin:$PATH
```
fedemkr, line 6: 🤔 Why is it needed to be changed to homebrew now?

Scripts/lint-changes.sh
```diff
@@ -0,0 +1,110 @@
...
+        echo "❌ SwiftFormat failed"
+        SWIFTFORMAT_FAILED=1
+    fi
+else
+    # Check formatting without modifying files
+    if ! mint run swiftformat --lint "${FILE_ARRAY[@]}"; then
+        echo "❌ SwiftFormat found formatting issues"
+        SWIFTFORMAT_FAILED=1
+    fi
+fi
```
fedemkr: 🤔 Could we move this into a function and call it again after Swift lint? As there are sometimes that swift lint introduces some wrong formatting that should be caught by swift format again.

Scripts/pre-commit
matt-livefront: If the pre-commit script is ran outside of your terminal (e.g. if you're using a git GUI), homebrew may not exist in the path and the mint commands will fail.

Scripts/lint-changes.sh
matt-livefront: Yep, good suggestion!

fedemkr (approved):

Scripts/lint-changes.sh
fedemkr: Awesome, I like `DIFF_ARGS` much better 👏

PR #2699, opened 2026-05-24
[PM-38024] fix: Retain cycle in DefaultTOTPExpirationManager

### LINKED ISSUE #2698: [PM-38023] Memory leak: DefaultTOTPExpirationManager Timer retain cycle prevents deinit
## Summary

`DefaultTOTPExpirationManager.init` schedules a repeating `Timer` whose block captures `self` strongly. This forms a retain cycle (`Manager → updateTimer → Timer → block → self`) that prevents `deinit` from running, which means `cleanup()` (the method that invalidates the Timer) is never called. Every instance ever constructed stays alive for the process lifetime.

## How this was discovered

I'm running an ongoing iOS memory-leak hunt, write-up of the methodology and first episode here: [Hunting iOS Memory Leaks — S1E1](https://www.amanjeet.me/hunting-ios-memory-leaks-s1e1/).

The initial audit surfaced only leaks in unit tests (uses macOS `leaks(1)`). 

A subsequent agent-based audit of production patterns then revealed this `DefaultTOTPExpirationManager` issue, which is "abandoned memory" rather than a strict leak, reachable from a process root (the main RunLoop, which owns the Timer), that apparently is invisible to the leaks tool. 

I learned leaks tool only considers a leak when objects are not reachable from the root of the graph, so this is a different type of leak where its abandoned sitting in heap, still connected to root run loop though.

## The bug

`BitwardenShared/UI/Vault/Utilities/TOTPExpirationManager.swift` (lines 56–74):

```swift
init(
    timeProvider: any TimeProvider,
    onExpiration: (([VaultListItem]) -> Void)?,
) {
    self.timeProvider = timeProvider
    self.onExpiration = onExpiration
    updateTimer = Timer.scheduledTimer(
        withTimeInterval: 0.25,
        repeats: true,
        block: { _ in
            self.checkForExpirations()   // ← strong self capture
        },
    )
}

deinit {
    cleanup()   // ← never called, because of the cycle
}
```

The cycle:
- `Manager` holds `updateTimer` (strong, stored property)
- `Timer` holds its block (strong, Foundation behavior)
- block captures `self` strongly (default Swift closure capture)
- back to `Manager`

ARC requires every edge in a cycle to be strong for the cycle to leak — all three are. `deinit` never runs; `cleanup()` is never invoked.

## Impact

Per leaked instance:
- ~96-byte `DefaultTOTPExpirationManager`
- 64-byte `Timer` + its block
- `itemsByInterval` dictionary (sized by number of TOTP items the screen was tracking)
- `onExpiration` closure (captures whatever the caller passed — current call sites use `[weak self]`, so doesn't propagate further)
- `timeProvider`

Each leaked manager also continues running its scheduled timer for the rest of the process lifetime on stale data — wasted work that compounds with each accumulated instance.

Scope: every `VaultGroup` screen view, every Vault search interaction with TOTP-bearing items, every TOTP autofill flow creates at least one instance.

## Proving the leak with a unit test

Because the cycle is reachable from a process root, `leaks(1)`-based audits report no leak. 

The verification has to ask a different question, "is this object alive when it shouldn't be?", which a weak-reference XCTest can encode directly:

```swift
import BitwardenKit
import BitwardenKitMocks
import XCTest

@MainActor
func test_DefaultTOTPExpirationManager_deallocatesAfterRelease() async throws {
    weak var weakManager: DefaultTOTPExpirationManager?

    do {
        let manager = DefaultTOTPExpirationManager(
            timeProvider: MockTimeProvider(.currentTime),
            onExpiration: { _ in },
        )
        weakManager = manager
    }

    // Let any in-flight work settle so deinit (if it could fire) would have.
    try await Task.sleep(nanoseconds: 200_000_000)

    XCTAssertNil(
        weakManager,
        "Manager should deallocate after release; retain cycle prevents deinit.",
    )
}
```

Run on current `main`, this test fails with `XCTAssertNil failed: "BitwardenShared.DefaultTOTPExpirationManager"`, the manager is still alive 200 ms after the only strong reference was dropped. Run after either fix below, it passes.

I used a red green verification to verify if its fixed.

### Option A — Minimal: `[weak self]` in the Timer block

```diff
- block: { _ in
-     self.checkForExpirations()
- }
+ block: { [weak self] _ in
+     self?.checkForExpirations()
+ }
```

- One-line change, non-breaking, no callers affected.
- Breaks the cycle; `deinit` runs normally; `cleanup()` runs automatically.
- The existing `VaultGroupProcessor` workaround becomes defensive instead of load-bearing.

### Option B: Structural: explicit `start()`/`stop()` lifecycle

Move Timer scheduling out of `init` into a new `start()` method, paired with `stop()`. `init` becomes pure — no side effects.

- This is a design hygiene that makes caller aware about the lifecycle of timer
- More robust: `deinit`-based cleanup no longer load-bearing.
- This would also need updating the callers, and would have more surface area of change. 
- I can go with this option if you like, but I wanted to keep the behavioural change minimum.

## Recommendation

**Option A** as the immediate fix, one-line, non-breaking, ships the bugfix without any caller migration. `[weak self]` is the actual change that resolves the cycle; the structural refactor doesn't fix the bug by itself (verified by isolating the changes in the unit test).

I'll open a PR for Option A linked to this issue, including the weak-reference regression test above.

Option B is worth doing as a follow-up to make the design less fragile against future regressions, but is a separate architectural conversation. Happy to open a separate Discussion for it if there's interest.

amanjeetsingh150:
## 🎟️ Tracking

Closes #2698

## 📔 Objective

`DefaultTOTPExpirationManager.init` schedules a repeating `Timer` whose block captures `self` strongly. The cycle (`Manager → updateTimer → Timer → block → self`) prevents `deinit` from running, which means `cleanup()` (the method that invalidates the Timer) is never called. Every instance ever constructed stays alive for the process lifetime.

`VaultGroupProcessor.deinit` already calls `cleanup()` on `groupTotpExpirationManager` as a workaround for this exact issue, suggesting prior awareness. The workaround is missing for `searchTotpExpirationManager`, and stops the timer without releasing the manager instance itself — symptom, not cause.

Full bug analysis, impact, and rationale for choosing this fix over a structural refactor are in #2698.

### Fix

Add `[weak self]` to the Timer block:

```diff
- block: { _ in
-     self.checkForExpirations()
- }
+ block: { [weak self] _ in
+     self?.checkForExpirations()
+ }
```

Non-breaking, no caller migration. With the cycle broken, `deinit` runs normally and `cleanup()` invalidates the Timer automatically. The existing `VaultGroupProcessor` workaround becomes defensive rather than load-bearing.

### Regression test

Added `TOTPExpirationManagerLeakTests.test_deallocatesAfterCallerReleasesReference`. Uses a weak reference to assert the manager deallocates after its only strong reference is dropped — `leaks(1)`-based audits don't catch this class of bug because the Timer is held by the main RunLoop, keeping the cycle reachable from a process root. The weak-reference approach catches it cleanly.

Test fails on `main` with `XCTAssertNil failed: "BitwardenShared.DefaultTOTPExpirationManager"`; passes with the fix.

## 📸 Screenshots

n/a — no UI changes.

### DIFF
--- a/BitwardenShared/UI/Vault/Utilities/TOTPExpirationManagerLeakTests.swift
+++ b/BitwardenShared/UI/Vault/Utilities/TOTPExpirationManagerLeakTests.swift
@@ -1,0 +1,40 @@
+import BitwardenKit
+import BitwardenKitMocks
+import XCTest
+
+@testable import BitwardenShared
+
+class TOTPExpirationManagerLeakTests: BitwardenTestCase {
+    /// Regression test for the retain cycle in `DefaultTOTPExpirationManager`
+    /// where the Timer scheduled in `init` captured `self` strongly,
+    /// preventing `deinit` (and therefore `cleanup()`) from ever running.
+    ///
+    /// Uses a weak reference to assert the manager deallocates after its
+    /// only strong reference is dropped. `leaks(1)`-based audits don't
+    /// catch this class of bug because the Timer is held by the main
+    /// RunLoop, keeping everything reachable from a process root.
+    @MainActor
+    func test_deallocatesAfterCallerReleasesReference() async throws {
+        weak var weakManager: DefaultTOTPExpirationManager?
+
+        do {
+            let manager = DefaultTOTPExpirationManager(
+                timeProvider: MockTimeProvider(.currentTime),
+                onExpiration: { _ in },
+            )
+            weakManager = manager
+        }
+
+        // Let any in-flight work settle so deinit (if it could fire) would have.
+        try await Task.sleep(nanoseconds: 200_000_000)
+
+        XCTAssertNil(
+            weakManager,
+            """
+            DefaultTOTPExpirationManager should deallocate after the caller \
+            releases its reference; a retain cycle in the Timer's block \
+            prevents deinit from running.
+            """,
+        )
+    }
+}
--- a/BitwardenShared/UI/Vault/Utilities/TOTPExpirationManager.swift
+++ b/BitwardenShared/UI/Vault/Utilities/TOTPExpirationManager.swift
@@ -62,8 +62,8 @@ class DefaultTOTPExpirationManager: TOTPExpirationManager {
         updateTimer = Timer.scheduledTimer(
             withTimeInterval: 0.25,
             repeats: true,
-            block: { _ in
-                self.checkForExpirations()
+            block: { [weak self] _ in
+                self?.checkForExpirations()
             },
         )
     }


BitwardenShared/UI/Vault/Utilities/TOTPExpirationManagerLeakTests.swift
```diff
@@ -0,0 +1,40 @@
+import BitwardenKit
+import BitwardenKitMocks
+import XCTest
+
+@testable import BitwardenShared
+
+class TOTPExpirationManagerLeakTests: BitwardenTestCase {
```
KatherineInCode, line 7: 🎨 We are trying to write all new tests in Swift Testing. Can you please convert this?

BitwardenShared/UI/Vault/Utilities/TOTPExpirationManagerLeakTests.swift
```diff
@@ -0,0 +1,40 @@
...
+    /// Regression test for the retain cycle in `DefaultTOTPExpirationManager`
```
KatherineInCode, line 8: 🎨 This documentation comment is verbose, and not in line with our usual test documentation comments. If there's further context necessary, it can be a comment inside the test function, though I also think a simple test documentation comment will suffice for explaining what is being tested.

BitwardenShared/UI/Vault/Utilities/TOTPExpirationManagerLeakTests.swift
```diff
@@ -0,0 +1,40 @@
...
+        do {
+            let manager = DefaultTOTPExpirationManager(
+                timeProvider: MockTimeProvider(.currentTime),
+                onExpiration: { _ in },
+            )
+            weakManager = manager
+        }
+
+        // Let any in-flight work settle so deinit (if it could fire) would have.
+        try await Task.sleep(nanoseconds: 200_000_000)
```
KatherineInCode, line 29: 🤔 Is this sleep really necessary? If so, does it have to be 0.2 seconds? As well, it appears that elsewhere in our codebase we document the ns->s conversion, can you please match that convention?

BitwardenShared/UI/Vault/Utilities/TOTPExpirationManagerLeakTests.swift
```diff
@@ -0,0 +1,40 @@
...
+
+        XCTAssertNil(
+            weakManager,
+            """
+            DefaultTOTPExpirationManager should deallocate after the caller \
```
KatherineInCode, line 34: 🎨 This feels overly verbose to me; most of the time we don't include a specific failure message, and when we do, it's a terse explanation of what failed

KatherineInCode (changes requested):
Thank you for your contribution! This is a nice thing to catch!

In addition to my inline comments, I have two more general ones:

- Something went wrong with the CLA bot; can you please re-check and make sure that you've signed it and all that is squared away?
- We have two `DefaultTOTPExpirationManager` classes that duplicate this code, but this fixes the retain cycle in only one of them. Can you please replicate this and/or find a way to consolidate that object into `BitwardenKit`?

Thanks!

PR #2708, opened 2026-05-25
[PM-38097] feat: Match local URI domain by port in iOS

xrvk:
## 🎟️ Tracking

- https://github.com/bitwarden/clients/pull/20838
- https://community.bitwarden.com/t/uri-match-detection-for-ports/54661
- https://bitwarden.atlassian.net/browse/PM-38097
-

## 📔 Objective

When multiple local services share the same IP or localhost host but run on different ports, domain matching can select the wrong login. This change adds iOS parity with the clients fix by requiring matching ports for domain matching only when both URIs are localhost or IP addresses.

Changes:
- Updated `DefaultCipherMatchingHelper` domain matching to reject localhost/IP matches when ports differ.
- Kept regular domain matching behavior unchanged so non-local domains remain port-agnostic for `.domain`.
- Added focused `CipherMatchingHelperTests` coverage for local IP and localhost port cases, plus a regression test that regular domains still match across different ports.

## 📸 Screenshots

N/A - no UI changes.


morganzellers-bw:
Hi @xrvk! 

These changes look good and I approved the workflows to run. I'll keep an eye on this one and let you know if I need any updates; thanks for the contribution! 👍

morganzellers-bw (approved):

PR #2710, opened 2026-05-26
[PM-37719] fix: Fix VoiceOver announcing "Star Star" in upcoming charge description

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37719

## 📔 Objective

Fix VoiceOver announcing "Star Star" before and after the dollar amount and date in the upcoming charge description. The localization strings use `**` markdown for bold formatting, which renders correctly via `Text(LocalizedStringKey(...))` but was passed verbatim to the accessibility label — causing VoiceOver to read the asterisks aloud.

Strip `**` from `descriptionAccessibilityLabel` for all plan statuses (active, canceled, updatePayment, etc.) so VoiceOver announces the text cleanly without markdown noise.

### DIFF
--- a/BitwardenKit/Core/Platform/Extensions/StringExtensionsTests.swift
+++ b/BitwardenKit/Core/Platform/Extensions/StringExtensionsTests.swift
@@ -1,72 +1,54 @@
 import BitwardenKit
-import Combine
-import XCTest
+import Testing
 
-class StringExtensionsTests: BitwardenTestCase {
-    // MARK: Tests
+// MARK: - StringExtensionsTests
 
-    /// `fixURLIfNeeded()` returns the same string when it can be directly converted to `URL`.
-    func test_fixURLIfNeeded_sameStringWhenCanConvertToURL() {
-        let string = "https://bitwarden.com"
-        XCTAssertEqual(string.fixURLIfNeeded(), "https://bitwarden.com")
-    }
+struct StringExtensionsTests {
+    // MARK: Tests - fixURLIfNeeded
 
-    /// `fixURLIfNeeded()` returns the same string prefixed with "http://" when it can't be
-    /// directly converted to `URL` and it's an IP address.
-    func test_fixURLIfNeeded_httpWhenIPAddress() {
-        let string = "192.168.0.1:8080"
-        XCTAssertEqual(string.fixURLIfNeeded(), "http://192.168.0.1:8080")
-    }
-
-    /// `fixURLIfNeeded()` returns the same string prefixed with "https://" when it can't be
-    /// directly converted to `URL` and it's not an IP address.
-    func test_fixURLIfNeeded_httpsWhenNotIPAddress() {
-        let string = "ht tp://broken.com"
-        XCTAssertEqual(string.fixURLIfNeeded(), "https://ht tp://broken.com")
+    /// `fixURLIfNeeded()` returns the correct URL string for valid URLs, IP addresses, and broken URLs.
+    @Test(arguments: zip(
+        [
+            "https://bitwarden.com",
+            "192.168.0.1:8080",
+            "ht tp://broken.com",
+        ],
+        [
+            "https://bitwarden.com",
+            "http://192.168.0.1:8080",
+            "https://ht tp://broken.com",
+        ]
+    ))
+    func fixURLIfNeeded(input: String, expected: String) {
+        #expect(input.fixURLIfNeeded() == expected)
     }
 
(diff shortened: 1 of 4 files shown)


BitwardenShared/UI/Billing/PremiumPlan/PremiumPlanState.swift
```diff
@@ -37,11 +37,13 @@ struct PremiumPlanState: Equatable {
...
     var descriptionAccessibilityLabel: String {
-        guard planStatus == .active else { return descriptionText }
+        guard planStatus == .active else {
+            return descriptionText.replacingOccurrences(of: "**", with: "")
+        }
         return Localizations.yourNextChargeIsForXDueOnY(
             nextChargeAmountAccessibilityLabel,
             nextChargeDate,
-        )
+        ).replacingOccurrences(of: "**", with: "")
```
KatherineInCode: 🤔 Instead of arbitrarily removing two asterisks, would it make sense to create an extension on `String` that is just a general `removingMarkdown` sort of function?

Something like this, though this is AI-generated so you'd want to check it and make sure there's good tests around it:
```
extension String {
    func removingMarkdown() -> String {
        var text = self
        
        // Pattern to match bold (**), italic (* or _), and strikethrough (~~)
        let patterns = [
            "\\*\\*(.*?)\\*\\*", // Bold **
            "__(.*?)__",         // Bold __
            "\\*(.*?)\\*",       // Italic *
            "_(.*?)_",           // Italic _
            "~~(.*?)~~",         // Strikethrough
            "\\[(.*?)\\]\\(.*?\\)" // Links: [text](url) -> text
        ]
        
        for pattern in patterns {
            text = text.replacingOccurrences(
                of: pattern,
                with: "$1", // Keeps the first captured group (the text inside)
                options: .regularExpression
            )
        }
        return text
    }
}
```

Feels like that's something we might end up wanting over the long run, maybe?

andrebispo5: Sure that way we can reuse it 👍

fedemkr: Totally agree with Katherine, although we need to be careful on how we handle urls for VoiceOver as the user should know there's a link to tap in there.
Additionally, on the naming of the function could it be specific for VoiceOver? Then if we need that elsewhere to be used we can extract it but I think I'd prefer that for VoiceOver stuff we have functions with names dedicated to that so they are easy to be modified without changing behavior on other parts of the app.

andrebispo5: Done 👍

BitwardenKit/Core/Platform/Extensions/StringExtensionsTests.swift
```diff
@@ -24,4 +24,49 @@ class StringExtensionsTests: BitwardenTestCase {
         let string = "ht tp://broken.com"
         XCTAssertEqual(string.fixURLIfNeeded(), "https://ht tp://broken.com")
     }
+
+    // MARK: Tests - removingMarkdownForVoiceOver
```
KatherineInCode, line 26: 🤔 While you're here, do you think this test file could be converted to Swift Testing?

KatherineInCode (commented):
Looks good, just one small Boyscout Rule thing

BitwardenKit/Core/Platform/Extensions/StringExtensionsTests.swift
andrebispo5: done

BitwardenKit/Core/Platform/Extensions/StringExtensionsTests.swift
```diff
@@ -1,27 +1,54 @@
...
 
-    /// `fixURLIfNeeded()` returns the same string prefixed with "https://" when it can't be
-    /// directly converted to `URL` and it's not an IP address.
-    func test_fixURLIfNeeded_httpsWhenNotIPAddress() {
-        let string = "ht tp://broken.com"
-        XCTAssertEqual(string.fixURLIfNeeded(), "https://ht tp://broken.com")
+    // MARK: Tests - removingMarkdownForVoiceOver
+
+    /// `removingMarkdownForVoiceOver()` strips markdown syntax while preserving readable text.
+    @Test(arguments: zip(
```
KatherineInCode, line 29: 👍🏻 Love the use of `zip` for these

KatherineInCode (approved):

PR #2713, opened 2026-05-27
[PM-38147] feat: Convert premium checkout to ASWebAuthenticationSession

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38147
https://bitwarden.atlassian.net/browse/PM-37994

## 📔 Objective

Replaces the browser-based Stripe checkout flow with `ASWebAuthenticationSession`, keeping the user in-app and giving the processor direct access to the callback URL.

Removes the `openURL` / `state.checkoutURL` indirection from the view, the `handlePremiumCheckoutResult` deep link handler from `AppProcessor`, and the `unableToOpenCheckout` error case — all now handled inline by the session delegate.

## 📸 Screenshots

https://github.com/user-attachments/assets/1afc63b8-f0b4-402a-9617-bc15e3791f94

### DIFF
--- a/BitwardenShared/UI/Billing/BillingCoordinator.swift
+++ b/BitwardenShared/UI/Billing/BillingCoordinator.swift
@@ -21,9 +21,6 @@ class BillingCoordinator: NSObject, Coordinator, HasStackNavigator {
     /// Determines the close behavior of PremiumUpgradeComplete.
     private var isUpgradeAsModalRoot = false
 
-    /// The active web authentication session for the Stripe checkout flow.
-    private var webAuthSession: ASWebAuthenticationSession?
-
     /// Closure invoked after PremiumUpgradeComplete is dismissed.
     private var premiumUpgradeCompleteOnClose: (() -> Void)?
 
@@ -32,6 +29,9 @@ class BillingCoordinator: NSObject, Coordinator, HasStackNavigator {
 
     /// The stack navigator that is managed by this coordinator.
     private(set) weak var stackNavigator: StackNavigator?
+
+    /// The active web authentication session for the Stripe checkout flow.
+    private var webAuthSession: ASWebAuthenticationSession?
 
     // MARK: Initialization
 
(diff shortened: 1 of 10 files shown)


BitwardenShared/UI/Billing/BillingCoordinator.swift
```diff
@@ -20,6 +21,9 @@ class BillingCoordinator: Coordinator, HasStackNavigator {
     /// Determines the close behavior of PremiumUpgradeComplete.
     private var isUpgradeAsModalRoot = false
 
+    /// The active web authentication session for the Stripe checkout flow.
+    private var webAuthSession: ASWebAuthenticationSession?
```
matt-livefront: ⛏️ Can you alphabetize this?

BitwardenShared/UI/Billing/BillingCoordinator.swift
```diff
@@ -128,3 +133,56 @@ class BillingCoordinator: Coordinator, HasStackNavigator {
...
+            let session = ASWebAuthenticationSession(
+                url: url,
+                callbackURLScheme: services.billingService.checkoutCallbackUrlScheme,
+            ) { [weak self] callbackURL, _ in
+                self?.webAuthSession = nil
+                let outcome: CheckoutWebAuthSessionOutcome = callbackURL
+                    .map { .completed(callbackURL: $0) } ?? .canceled
+                self?.resumeAfterDismissal(continuation: continuation, outcome: outcome)
+            }
+            session.prefersEphemeralWebBrowserSession = false
```
matt-livefront: ❓ Are there any cookies that need to be shared with this web sessions for the Stripe checkout? Or could `prefersEphemeralWebBrowserSession` be set to `true`? Setting it to `true` would avoid the "Bitwarden wants to use stripe.com to sign in" alert if we don't need to share cookies.

BitwardenShared/UI/Billing/PremiumUpgrade/PremiumUpgradeProcessorTests.swift
```diff
@@ -9,6 +9,21 @@ import Testing
 @testable import BitwardenShared
 @testable import BitwardenSharedMocks
 
+// MARK: - MockPremiumUpgradeProcessorDelegate
+
+@MainActor
+class MockPremiumUpgradeProcessorDelegate: PremiumUpgradeProcessorDelegate {
```
matt-livefront: ⛏️ Instead of manually creating this mock, could you use Sourcery/AutoMockable to generate this?

BitwardenShared/UI/Billing/PremiumUpgrade/PremiumUpgradeProcessor.swift
```diff
@@ -3,6 +3,32 @@ import BitwardenResources
...
+
+/// The result of a Stripe checkout web authentication session.
+///
+enum CheckoutWebAuthSessionOutcome {
+    /// The user dismissed the session without completing checkout.
+    case canceled
+
+    /// The session completed and Stripe returned a callback URL.
+    case completed(callbackURL: URL)
+}
```
matt-livefront: 🤔 I feel like it might be easier and cleaner to use the `Result<URL, Error>` type here rather than something custom. You can still pass the URL back and if an error does occur (other than canceled), you'd have more context around the failure to log or show to the user.

matt-livefront (approved):

PR #2716, opened 2026-05-27
[PM-37967] feat: Show premium canceled status in plan screen

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37967

## 📔 Objective

When a premium user with a canceled, past-due, or payment-required subscription taps the Plan row in Settings, they are now navigated to the premium plan screen (which shows their subscription status) instead of the upgrade/checkout screen.

Previously, `hasPremium = false` for these users caused them to always land on the upgrade flow. The fix checks the subscription status via `getSubscription()` when `hasPremium` is false and uses the new `isTroubleState` property on `PremiumPlanStatus` to detect non-active subscriptions. A loading overlay is shown during the API call. The already-fetched subscription is threaded through to `PremiumPlanState` via `PremiumSubscriptionContext` so `PremiumPlanProcessor` can skip a redundant second fetch.

## 📸 Screenshots

https://github.com/user-attachments/assets/a80690e1-190d-4079-96e6-8540587f244c

### DIFF
--- a/BitwardenShared/UI/Billing/BillingCoordinator.swift
+++ b/BitwardenShared/UI/Billing/BillingCoordinator.swift
@@ -59,8 +59,8 @@ class BillingCoordinator: Coordinator, HasStackNavigator {
             }
         case .premiumUpgradeComplete:
             showPremiumUpgradeComplete()
-        case .premiumPlan:
-            showPremiumPlan(subscription: (context as? PremiumSubscriptionContext)?.subscription)
+        case let .premiumPlan(subscription):
+            showPremiumPlan(subscription: subscription)
         case .premiumUpgrade:
             showPremiumUpgrade()
         }
--- a/BitwardenShared/UI/Billing/BillingCoordinator.swift
+++ b/BitwardenShared/UI/Billing/BillingCoordinator.swift
@@ -60,7 +60,7 @@ class BillingCoordinator: Coordinator, HasStackNavigator {
         case .premiumUpgradeComplete:
             showPremiumUpgradeComplete()
         case .premiumPlan:
-            showPremiumPlan()
+            showPremiumPlan(subscription: (context as? PremiumSubscriptionContext)?.subscription)
         case .premiumUpgrade:
             showPremiumUpgrade()
         }
@@ -79,7 +79,7 @@ class BillingCoordinator: Coordinator, HasStackNavigator {
                 // Pop PremiumUpgradeView silently so back navigation from PremiumPlanView
                 // returns to Settings, not to the (now-irrelevant) upgrade screen.
                 self?.stackNavigator?.pop(animated: false)
-                self?.showPremiumPlan()
+                self?.showPremiumPlan(subscription: nil)
             }
         let processor = PremiumUpgradeCompleteProcessor(coordinator: asAnyCoordinator())
         let view = PremiumUpgradeCompleteView(store: Store(processor: processor))
@@ -88,11 +88,14 @@ class BillingCoordinator: Coordinator, HasStackNavigator {
 
     /// Shows the premium plan screen.
     ///
-    private func showPremiumPlan() {
+    /// - Parameter subscription: An already-fetched subscription; pass `nil` to let the plan screen fetch it.
+    ///
+    private func showPremiumPlan(subscription: PremiumSubscription?) {
+        let state = subscription.map(PremiumPlanState.init(subscription:)) ?? PremiumPlanState()
         let processor = PremiumPlanProcessor(
             coordinator: asAnyCoordinator(),
             services: services,
-            state: PremiumPlanState(),
+            state: state,
         )
         let view = PremiumPlanView(store: Store(processor: processor))
         let viewController = UIHostingController(rootView: view)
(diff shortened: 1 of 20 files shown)


BitwardenShared/UI/Billing/BillingCoordinator.swift
```diff
@@ -60,7 +60,7 @@ class BillingCoordinator: Coordinator, HasStackNavigator {
         case .premiumUpgradeComplete:
             showPremiumUpgradeComplete()
         case .premiumPlan:
-            showPremiumPlan()
+            showPremiumPlan(subscription: (context as? PremiumSubscriptionContext)?.subscription)
```
matt-livefront: 🤔 Does `PremiumSubscriptionContext` need to be passed as `context`? Why not pass `PremiumSubscription` as an associated value on the `premiumPlan` case?

BitwardenShared/UI/Platform/Settings/Settings/SettingsProcessor.swift
```diff
@@ -113,14 +116,34 @@ final class SettingsProcessor: StateProcessor<SettingsState, SettingsAction, Set
...
+            if subscription.status.isTroubleState {
+                coordinator.navigate(to: .premiumPlan, context: PremiumSubscriptionContext(subscription))
             } else {
                 coordinator.navigate(to: .premiumUpgrade)
             }
-        case .vaultPressed:
-            coordinator.navigate(to: .vault)
+        } catch {
+            services.errorReporter.log(error: error)
+            coordinator.navigate(to: .premiumUpgrade)
```
matt-livefront: ❓ Given that this is making an API request, is it ok to navigate to the premium upgrade screen if the API request fails? Or should an error be shown in that case?

BitwardenShared/UI/Billing/PremiumPlan/PremiumPlanProcessorTests.swift
```diff
@@ -78,6 +78,35 @@ struct PremiumPlanProcessorTests {
...
+            ),
+            state: PremiumPlanState(subscription: preloaded),
+        )
+
+        await subjectWithSubscription.perform(.appeared)
+
+        #expect(billingService.getPremiumPlanCallsCount == 1)
+        #expect(billingService.getSubscriptionCallsCount == 0)
+        #expect(subjectWithSubscription.state.planStatus == .canceled)
+        #expect(subjectWithSubscription.state.subscription == preloaded)
```
matt-livefront: 🎨 Could you update the state directly on the existing processor rather than building a new one?

```suggestion
        let preloaded = PremiumSubscription.fixture(status: .canceled)
        subject.state = PremiumPlanState(subscription: preloaded)

        await subject.perform(.appeared)

        #expect(billingService.getPremiumPlanCallsCount == 1)
        #expect(billingService.getSubscriptionCallsCount == 0)
        #expect(subject.state.planStatus == .canceled)
        #expect(subject.state.subscription == preloaded)
```

BitwardenShared/UI/Platform/Settings/SettingsCoordinator.swift
```diff
@@ -6,6 +6,19 @@ import SwiftUI
 
 // swiftlint:disable file_length
 
+// MARK: - PremiumSubscriptionContext
+
+/// A reference-type box used to pass a `PremiumSubscription` value via the coordinator context.
+///
+final class PremiumSubscriptionContext {
+    /// The subscription to pass.
+    let subscription: PremiumSubscription
```
matt-livefront: ⛏️ If possible, it'd be better to pass `PremiumSubscription` as an associated value on the route rather than as a `context` variable, which is primarily meant for delegates.

BitwardenShared/UI/Billing/BillingCoordinator.swift
andrebispo5: had to make PremiumSubscription public and hashable but I think it's worth it here.

BitwardenShared/UI/Platform/Settings/Settings/SettingsProcessor.swift
andrebispo5: well it is a double edge because when the user doesn't have any premium and never had, the `getSubscription` returns 404 and we need to show the user the premium upgrade view.

matt-livefront: Ok, we could separate out the 404 error from other errors though. It just depends on what Product wants. IMO showing an error for non-404 errors is probably better since you'll have more context and could share the error report.

Here's an example of throwing a specific error for 404 which can then be caught and handled separately in the processor:
https://github.com/bitwarden/ios/blob/d1bdb77a66c5c1aaa5320b2b4e1189f9f6899b1e/BitwardenShared/Core/Auth/Services/API/Auth/Requests/PendingLoginRequest.swift#L32-L36

andrebispo5: done :)

matt-livefront (dismissed):

LRNcardozoWDF (approved):

PR #2717, opened 2026-05-27
[PM-27235] feat: Support v2 encryption on JIT password signups

matt-livefront:
## 🎟️ Tracking

[PM-27235](https://bitwarden.atlassian.net/browse/PM-27235)

## 📔 Objective

Enables support for v2 encryption on JIT password account registration. Most of the existing logic that was implemented in the app, including API requests, has been moved into the SDK (via the `postKeysForJitPasswordRegistration(request:)` function).

These changes are behind the `enable-account-encryption-v2-jit-password-registration` feature flag.

[PM-27235]: https://bitwarden.atlassian.net/browse/PM-27235?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
+++ b/BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
@@ -878,7 +878,38 @@ extension DefaultAuthRepository: AuthRepository {
             requestUserKey = passwordResult.newKey
             requestKeys = nil
             cryptographicState = accountKeys.cryptographicState
+        } else if await configService.getFeatureFlag(.accountEncryptionV2JITPassword) {
+            // V2 JIT password path: SDK handles all server-side API calls internally.
+            let organizationKeys = try await organizationAPIService.getOrganizationKeys(
+                organizationId: organizationId,
+            )
+            let request = JitMasterPasswordRegistrationRequest(
+                orgId: organizationId,
+                orgPublicKey: organizationKeys.publicKey,
+                organizationSsoIdentifier: organizationIdentifier,
+                userId: account.profile.userId,
+                salt: email,
+                masterPassword: password,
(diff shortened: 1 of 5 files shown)


BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
```diff
@@ -878,7 +878,38 @@ extension DefaultAuthRepository: AuthRepository {
             requestUserKey = passwordResult.newKey
             requestKeys = nil
             cryptographicState = accountKeys.cryptographicState
+        } else if await configService.getFeatureFlag(.accountEncryptionV2JITPassword) {
```
fedemkr, line 881: 🤔 Could we extract a function for this? As this function is getting pretty big.

fedemkr (commented):
Looks good, just one small ⛏️

BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
matt-livefront: I had this thought too when I wrote this, but I ended up not extracting it because this method will become quite a bit cleaner once the feature flag is removed. At that point it will become `if <TDE user> else <JIT password user>`. Long term, it seems nice to have both of these branches in the function, especially if the TDE set password logic also moves into the SDK. That said, if you feel strongly about extracting it, I'm happy to do that.

fedemkr: Alright yeah I think for the future sounds good 👍

fedemkr (approved):

PR #2723, opened 2026-05-28
[PM-38200] fix: Migrate collections update to v2 endpoint

KatherineInCode:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38200

## 📔 Objective

Migrates the collections update call from the legacy `PUT /ciphers/{id}/collections`
endpoint to `PUT /ciphers/{id}/collections_v2`.

The `_v2` endpoint returns the updated cipher and an `unavailable` flag. If
`unavailable` is `true`, the user has lost access to the item (e.g. they removed
themselves from all Can Manage collections) and it is removed from local storage;
otherwise the returned cipher is upserted locally.


BitwardenShared/Core/Vault/Services/CipherService.swift
```diff
@@ -448,10 +448,24 @@ extension DefaultCipherService {
         let userId = try await stateService.getActiveAccountId()
 
         // Update the cipher collections in the backend.
-        try await cipherAPIService.updateCipherCollections(cipher)
+        let response = try await cipherAPIService.updateCipherCollections(cipher)
 
-        // Update the cipher collections in local storage.
-        try await cipherDataStore.upsertCipher(cipher, userId: userId)
+        if response.unavailable {
```
KatherineInCode: This matches the logic [in the clients repo](https://github.com/bitwarden/clients/blob/main/libs/common/src/vault/services/cipher.service.ts#L1225)

BitwardenShared/Core/Vault/Models/Response/UpdateCipherCollectionsResponseModel.swift
```diff
@@ -0,0 +1,15 @@
...
+///
+struct UpdateCipherCollectionsResponseModel: JSONResponse, Equatable {
+    // MARK: Properties
+
+    /// The updated cipher, or `nil` if the user no longer has access.
+    let cipher: CipherDetailsResponseModel?
+
+    /// Whether the cipher is now unavailable to the current user (i.e. they no longer have any
+    /// collection access to it after the update).
+    let unavailable: Bool
```
fedemkr: 🤔 Does `unavailable` represent the same as the `cipher` being `nil`?

BitwardenShared/Core/Vault/Services/API/Cipher/Requests/UpdateCipherCollectionsRequest.swift
```diff
@@ -27,16 +27,16 @@ struct UpdateCipherCollectionsRequest: Request {
     let method = HTTPMethod.put
 
     /// The URL path for this request.
-    var path: String { "/ciphers/\(id)/collections" }
+    var path: String { "/ciphers/\(id)/collections_v2" }
```
fedemkr, line 30: ❓ Is it safe to replace this directly or should we have a feature flag to choose which endpoint to use?

BitwardenShared/Core/Vault/Services/CipherService.swift
fedemkr: 🤔 Shouldn't this also include whether the cipher is nil on the response?

```suggestion
        if response.unavailable || response.cipher == nil {
```

BitwardenShared/Core/Vault/Services/CipherService.swift
```diff
@@ -448,10 +448,17 @@ extension DefaultCipherService {
...
+            // The user no longer has any collection access to the cipher after the update,
+            // so remove it from local storage.
+            try await cipherDataStore.deleteCipher(id: cipher.id ?? "", userId: userId)
```
fedemkr: 🎨 Could we ignore calling `deleteCipher` if the `id` is `nil` or throwing an error? I mean we already know the outcome of that without calling the function.

BitwardenShared/Core/Vault/Services/CipherService.swift
```diff
@@ -448,10 +448,17 @@ extension DefaultCipherService {
...
+        } else {
+            // Use the server-returned cipher if available; fall back to the local model.
+            let updatedCipher = response.cipher.map { Cipher(responseModel: $0) } ?? cipher
+            try await cipherDataStore.upsertCipher(updatedCipher, userId: userId)
```
fedemkr: 🤔 If it fall backs to the local model and you upsert it, are you changing anything locally (I think the collections but double checking just in case)?
Could we have problems with future edit/updates given its `revisionDate` may not be updated? E.g. if this goes through this part of the flow, then web makes an edit on this afterwards and mobile edits after that.

BitwardenShared/Core/Vault/Services/API/Cipher/Requests/UpdateCipherCollectionsRequest.swift
KatherineInCode: Per conversation in Slack, we should be able to replace this directly. The endpoint is already in prod and being used by web; we should be safe here.

BitwardenShared/Core/Vault/Models/Response/UpdateCipherCollectionsResponseModel.swift
KatherineInCode: It's been removed in favor of just looking at `cipher` being nil

BitwardenShared/Core/Vault/Services/CipherService.swift
KatherineInCode: I've changed it to throw an error if we get a `nil` `id`. In practice that should never happen, but I like covering the base just in case. Good catch!

BitwardenShared/Core/Vault/Services/CipherService.swift
KatherineInCode: I think the other simplifications I've done have covered this scenario more cleanly; can you check me on that, though?

fedemkr (approved):
Looks good 🚀

PR #2728, opened 2026-05-29
[PM-38148] feat: Add License to vault list and search

SaintPatrck:
## 🎟️ Tracking

- Jira: [PM-38148](https://bitwarden.atlassian.net/browse/PM-38148)

## 📔 Objective

Surfaces synced Driver's License ciphers in the vault list, the **Types** section, and search — display-only. Ciphers created on other clients appear with the ID-card icon under a localized **License** group, searchable by name and subtitle.

## 📸 Screenshots

| Figma | Actual |
|--------|--------|
|  |  |

[PM-38148]: https://bitwarden.atlassian.net/browse/PM-38148?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenShared/UI/Vault/Vault/VaultList/VaultListGroup.swift
```diff
@@ -76,6 +79,8 @@ extension VaultListGroup {
             Localizations.typeCard
         case let .collection(_, name, _):
             name
+        case .driversLicense:
+            Localizations.license
```
KatherineInCode: 🤔 I know this is already in the strings file, but most of the rest of these are `"Type[thing]"` for the localization key, which might be better? I could go either way, though, especially as "SSH Key" *doesn't* do that, and that's the newest type until these. I don't have a super strong opinion either way.

SaintPatrck: I have no objection to adding/using `typeLicense`. Good catch. We should probably update the sshKey case too. 🚀

KatherineInCode (dismissed):

BitwardenResources/Localizations/en.lproj/Localizable.strings
```diff
@@ -200,6 +200,7 @@
 "SecurityCode" = "Security code";
 "TypeCard" = "Card";
 "TypeIdentity" = "Identity";
+"TypeLicense" = "License";
```
fedemkr, line 203: 🎨 I think we can remove this as we already have `"License"` string int he file. I'm not quite sure why `Type...` exist here unless in the case where the translation of the actual type is different on the other context that may be used.

```suggestion
```

BitwardenShared/Core/Vault/Extensions/CipherExtensionsTests.swift
fedemkr, line 1: 🎨 ♻️ I think we could migrate this to Swift Testing which would make testing `belongsToGroup` easier with the parametrized tests.

BitwardenShared/Core/Vault/Extensions/CipherListViewExtensionsTests.swift
fedemkr, line 1: 🎨 ♻️ I think we could migrate this to Swift Testing for the same reason as `CipherExtensionsTests`.

KatherineInCode (approved):

fedemkr (commented):
Looks good, just a few comments

PR #2731, opened 2026-05-29
[PM-38360] feat: Add reusable DateFieldPicker component

SaintPatrck:
## 🎟️ Tracking

- Jira: PM-38360

## 📔 Objective

Adds a reusable `DateFieldPicker` SwiftUI component to **BitwardenKit** for optional calendar-date entry. The field renders as a collapsed row (title + chevron); a tap reveals a **native inline graphical calendar** (`DatePicker` with `.datePickerStyle(.graphical)`) directly beneath the row, and a Clear control resets the selection back to empty. When no date is set, the field shows only its title — matching the design comps.

The calendar and its reveal animation are entirely native. The thin wrapper exists only to add an optional `Date?` (empty state + Clear), which a native `DatePicker` cannot represent on its own, inside the Add/Edit forms' `ScrollView` — where the native `.compact` inline-expansion style isn't available. This mirrors the existing `BitwardenMenuField` precedent of wrapping a native control for that context.

Built once for its consumers, this unblocks date fields for the new item types: Driver's License (date of birth, issue date, expiration) and Passport (date of birth, issue date, expiration).

Also adds an ISO 8601 calendar-date (`yyyy-MM-dd`) conversion helper on `Date` for round-tripping the string-backed date fields, which rejects invalid calendar dates (e.g. Feb 30). The Clear control uses a new 24pt `circle-x24` icon asset per Figma.

> **Note — date handling is still in flux:** the Bitwarden SDK is expected to change these fields from `String?` to a strongly-typed `Date` (https://github.com/bitwarden/sdk-internal/pull/1154). Once it does, the ISO 8601 string conversion and the invalid-date handling here become unnecessary and can be removed.

**Tests:** 7 `DateFieldPicker` ViewInspector tests + 6 `Date` ISO 8601 unit tests (all passing); snapshot tests added, disabled per repo convention.

### Collapsed

| Figma | Actual |
|--------|--------|
|   |  | 

### Expanded

| Figma | Actual |
|--------|--------|
|  |  |


BitwardenKit/Core/Platform/Extensions/Date.swift
```diff
@@ -53,3 +73,20 @@ public extension Date {
...
 }
+
+private extension Date {
+    /// Builds a strict, locale-independent formatter for ISO 8601 calendar dates (`yyyy-MM-dd`).
+    ///
+    /// A fresh formatter is returned on each call to avoid sharing a non-`Sendable` instance across
+    /// concurrency domains. Uses the POSIX locale and UTC time zone so output and parsing are
+    /// deterministic regardless of the device locale, and disables lenient parsing so invalid dates
+    /// such as `2023-02-30` fail to parse rather than rolling over.
+    static func iso8601DateOnlyFormatter() -> DateFormatter {
```
KatherineInCode, line 136: 🤔 How is this different from the [built-in](https://developer.apple.com/documentation/foundation/iso8601dateformatter) `ISO8601DateFormatter`?

SaintPatrck:
@RishikaSG-28 Figma didn't explicitly state which picker should be used for the dates, but I'm assuming this barrel spinner is expected since it's date only. I noticed there is no way to "Clear" a selected date so I've added a small "Clear" button. Can you take a look at the screenshot and let me know what you think. 

@bitwarden/team-ios @matt-livefront I'm not super familiar with the iOS native date picker. Is the the right approach to add a "Clear" button, or am I missing something that would handle clearing natively?

KatherineInCode:
@SaintPatrck Having it be a popup rather than a pseudo-keyboard or inline feels weird to me. The Contacts app still does a pseudo-keyboard; Reminders and Calendar put the date selector inline with a calendar; what are some Apple apps that do the popup (as it's possible I'm just not thinking of any offhand that do)?

RishikaSG-28:
@SaintPatrck, Emily added the date picker when she designed new item types. I've tagged you in it, please have a look.

SaintPatrck:
@KatherineInCode @RishikaSG-28 I've updated the picker based on our conversation. LMK what you think about the updated screenshots.

BitwardenKit/UI/Platform/Application/Views/DateFieldPicker.swift
```diff
@@ -0,0 +1,242 @@
...
+    @ViewBuilder
+    private func headerRow() -> some View {
+        HStack(spacing: 8) {
+            Button {
+                toggleExpanded()
+            } label: {
+                labelContent()
+            }
+            .buttonStyle(.plain)
+            .accessibilityIdentifier("DateFieldHeaderButton")
```
fedemkr, line 182: 🤔 Does the user understand that they need to tap to select a different date from VoiceOver?

BitwardenKit/UI/Platform/Application/Views/DateFieldPicker.swift
```diff
@@ -0,0 +1,242 @@
...
+
+            if date != nil {
+                AccessoryButton(
+                    asset: SharedAsset.Icons.circleX24,
+                    accessibilityLabel: Localizations.clear,
```
fedemkr: ⛏️ I believe this should say "Clear {title of the header}" as in VoiceOver the user may not know if that clear is for the date or for something that follow this field.

BitwardenKit/UI/Platform/Application/Views/DateFieldPicker.swift
```diff
@@ -0,0 +1,242 @@
...
+        if let range {
+            DatePicker("", selection: selection(), in: range, displayedComponents: [.date])
+                .labelsHidden()
+                .datePickerStyle(.graphical)
+        } else {
+            DatePicker("", selection: selection(), displayedComponents: [.date])
+                .labelsHidden()
+                .datePickerStyle(.graphical)
+        }
+    }
```
fedemkr, line 140: ⚠️ :accessibility: When in VoiceOver, `.graphical` style is really problematic as it's very hard to navigate. I think we should hide graphical pickers from accessibility and provide wheel style for such case which is easier for VoiceOver users to use.

fedemkr (commented):
Looks good, just a few a11y concerns.

BitwardenKit/UI/Platform/Application/Views/DateFieldPicker.swift
SaintPatrck: @maxkpower @RishikaSG-28 thoughts on this? Makes sense, imo.

RishikaSG-28: That makes sense! If it's doable to add that condition then let's go for it!

morganzellers-bw:
Shows VoiceOver + automatically focusing on the picker after expanding with VoiceOver on.


https://github.com/user-attachments/assets/9ccf197c-a4bc-41bd-b3fe-2ce471cea2ae

BitwardenKit/UI/Platform/Application/Views/DateFieldPicker.swift
```diff
@@ -215,12 +220,23 @@ public struct DateFieldPicker: View {
...
+            isPickerFocused = shouldFocusPicker(isExpanded: isExpanded, voiceOverEnabled: voiceOverEnabled)
+        }
+    }
+
+    /// Determines whether the wheel-style picker should receive accessibility focus for the given
+    /// expanded/VoiceOver state. Takes both inputs explicitly (rather than reading the environment
+    /// property internally) so the decision logic can be unit tested without a real SwiftUI render pass.
+    func shouldFocusPicker(isExpanded: Bool, voiceOverEnabled: Bool) -> Bool {
+        isExpanded && voiceOverEnabled
     }
```
fedemkr: 🤔 Instead of using the function and assigning on each part that it changes, could we use `.onChange` instead?
```swift
.onChange(of: isExpanded) { newValue in
            isPickerFocused = newValue && voiceOverEnabled
        }
```

fedemkr (commented):
Looks good, just one small improvement and not sure if you've revised what Claude raised @morganzellers-bw

BitwardenKit/Core/Platform/Extensions/Date.swift
```diff
@@ -53,3 +73,20 @@ public extension Date {
...
+        let formatter = DateFormatter()
+        formatter.locale = Locale(identifier: "en_US_POSIX")
+        formatter.timeZone = TimeZone(identifier: "UTC")
+        formatter.dateFormat = "yyyy-MM-dd"
+        formatter.isLenient = false
+        return formatter
+    }
```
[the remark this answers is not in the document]
morganzellers-bw: Fixed in commit 788b5b8

BitwardenKit/UI/Platform/Application/Views/DateFieldPicker.swift
```diff
@@ -0,0 +1,281 @@
...
+    /// empty) and, on selection, commits the value. With the graphical calendar, a selection also
+    /// collapses the calendar; under VoiceOver the wheel stays expanded so continuous scrubbing does
+    /// not dismiss it on the first change — the user collapses it via the header instead.
+    ///
+    /// `date` is UTC-anchored (see `iso8601DateOnlyString`), but the `DatePicker` reads/writes
+    /// calendar days in the device's local time zone. The get/set here convert at that boundary so
+    /// the day the user sees selected, and the day they pick, always match the day that gets stored.
+    private func selection() -> Binding<Date> {
+        Binding(
+            get: { (date ?? defaultDate).asLocalCalendarDay() },
```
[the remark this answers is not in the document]
morganzellers-bw: Fixed in commit e14a1dc

BitwardenKit/UI/Platform/Application/Views/DateFieldPicker.swift
morganzellers-bw: My thinking for the function was that it make the logic of when to focus testable, but after your suggestion, I think it adds too much boilerplate just to end up testing pretty simple logic. This feels a lot cleaner to me also, thanks for the suggestion Fede! 😁

BitwardenKit/Core/Platform/Extensions/Date.swift
```diff
@@ -24,6 +24,15 @@ public extension Date {
         Date.iso8601DateOnlyFormatter().string(from: self)
     }
 
+    /// The date formatted as a long localized calendar-date string (e.g. `June 23, 2023`), pinned to
+    /// UTC so a UTC-anchored stored date reads back as the same calendar day regardless of device
+    /// time zone.
+    var longCalendarDateDisplay: String {
+        var style = Date.FormatStyle(date: .long, time: .omitted)
+        style.timeZone = TimeZone(identifier: "UTC")!
```
fedemkr: ⚠️ We shouldn't use force-unwrapping unless strictly necessary. If you can use a `guard` checking optional instead.

BitwardenKit/Core/Platform/Extensions/Date.swift
```diff
@@ -72,6 +81,32 @@ public extension Date {
...
+    /// local-time UI control (such as `DatePicker`) back into the UTC-anchored form used for storage
+    /// via `iso8601DateOnlyString`.
+    ///
+    /// - Parameter timeZone: The time zone whose calendar day `self` is expressed in. Defaults to
+    ///   the device's current time zone.
+    func asUTCCalendarDay(from timeZone: TimeZone = .current) -> Date {
+        let components = Date.gregorianCalendar(timeZone: timeZone)
+            .dateComponents([.year, .month, .day], from: self)
+        return Date(year: components.year!, month: components.month!, day: components.day!)
+    }
```
fedemkr, line 126: ⚠️ We shouldn't use force-unwrapping unless strictly necessary. If you can use a `guard` checking optional instead.

fedemkr (commented):
Looks good, just 2 minor improvements to align with the codebase.
I'd approve after that but would appreciate another look from anyone else on the team in case I've missed anything cc: @bitwarden/team-ios @matt-livefront

BitwardenKit/Core/Platform/Extensions/Date.swift
morganzellers-bw: Updated to guard & return `self` if the guard fails

BitwardenKit/Core/Platform/Extensions/Date.swift
morganzellers-bw: Updated to guard and `return formatted(style)` without setting the timezone if the guard fails.

morganzellers-bw:
> Looks good, just 2 minor improvements to align with the codebase. I'd approve after that but would appreciate another look from anyone else on the team in case I've missed anything cc: @bitwarden/team-ios @matt-livefront

👍 Thanks for the review Fede!

BitwardenKit/Core/Platform/Extensions/Date.swift
```diff
@@ -52,4 +84,62 @@ public extension Date {
...
+    /// interpreted as a UTC calendar day. Use this to seed a UI control (such as `DatePicker`) that
+    /// reads/writes calendar days in the device's local time zone, from a UTC-anchored stored date,
+    /// so the day shown to the user matches the day that was stored.
+    ///
+    /// - Parameter timeZone: The time zone whose local midnight the day is expressed in. Defaults
+    ///   to the device's current time zone.
+    func asLocalCalendarDay(in timeZone: TimeZone = .current) -> Date {
+        let components = Date.gregorianCalendar(timeZone: TimeZone(identifier: "UTC")!)
+            .dateComponents([.year, .month, .day], from: self)
+        return Date(year: components.year!, month: components.month!, day: components.day!, timeZone: timeZone)
```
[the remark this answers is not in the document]
morganzellers-bw: Fixed in commit 06bde9f

KatherineInCode (approved):

PR #2558, opened 2026-04-20
[PM-34898] feat: Self-hosted support for PremiumUpgradeView

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-34898

## 📔 Objective

When the user is on a self-hosted server, the Premium Upgrade screen now shows a different layout:

- **Info banner**: A dismissable `ActionCard` banner above the content card informing the user to manage their Premium subscription via the web vault on a computer.
- **Header**: Replaces the price section with a single headline-style title ("Unlock more advanced features with a Premium plan.").
- **Hidden controls**: The "Upgrade now" button and Stripe footer are hidden since self-hosted users cannot purchase through the app.

Additionally, `ActionCard` was updated to support an optional `title` (`String?`), allowing cards with only a message and no title.

## 📸 Screenshots

https://github.com/user-attachments/assets/19dce7f2-3abc-455f-826f-eb2ae64c6625

PR #2736, opened 2026-06-01
[PM-38394] feat: Update settings plan row visibility

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38394

## 📔 Objective

The settings plan row visibility was only checking the `premiumUpgradePath` feature flag and whether the user is self-hosted. This PR updates the logic to also require the user's App Store storefront to be US (`isUSStorefront`), matching the full set of conditions specified in the ticket:

- Not self-hosted
- Is US store
- Feature flag enabled

### DIFF
--- a/BitwardenShared/UI/Platform/Settings/Settings/SettingsProcessor.swift
+++ b/BitwardenShared/UI/Platform/Settings/Settings/SettingsProcessor.swift
@@ -25,6 +25,7 @@ final class SettingsProcessor: StateProcessor<SettingsState, SettingsAction, Set
         & HasConfigService
         & HasErrorReporter
         & HasStateService
+        & HasStorefrontService
         & HasVaultRepository
 
     // MARK: Private Properties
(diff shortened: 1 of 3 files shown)


BitwardenShared/UI/Platform/Settings/Settings/SettingsProcessor.swift
```diff
@@ -95,9 +96,11 @@ final class SettingsProcessor: StateProcessor<SettingsState, SettingsAction, Set
                 .getFeatureFlag(.premiumUpgradePath, defaultValue: false)
             let hasPremium = await services.vaultRepository.doesActiveAccountHavePremium()
             let isSelfHosted = await services.billingService.isSelfHosted()
+            let isUSStorefront = await services.storefrontService.isUSStorefront()
             state.hasPremium = hasPremium
-            state.showPlanRow = featureEnabled && !isSelfHosted
-            state.shouldShowUpgradedToPremiumActionCard = await services.billingService.shouldShowUpgradedToPremiumActionCard()
+            state.showPlanRow = featureEnabled && !isSelfHosted && isUSStorefront
```
fedemkr, line 101: 🤔 Maybe I'm missing something but in here the plan row is shown if it's not self hosted. However, [isInAppUpgradeAvailable](https://github.com/bitwarden/ios/blob/main/BitwardenShared/Core/Billing/Repositories/BillingRepository.swift#L68) which is what's used to display the upgrade plan card doesn't check whether it's self-hosted. 
Should that be updated as well? If so, could we centralize part of such logic in the `BillingRepository` to be shared with this as well?

andrebispo5: No the conditions are different for the card and for the plan row 👍

fedemkr: I find it a bit strange that the upgrade plan card would be shown in Self-hosted servers, shouldn't `isInAppUpgradeAvailable` have that check as well? Perhaps it's done in some other place I'm missing but which would be the scenario where a Self-hosted server would have an "upgrade plan" card?

andrebispo5: Here check the video https://github.com/bitwarden/ios/pull/2558

fedemkr (approved):

PR #2739, opened 2026-06-01
[PM-33025] ci: Speed up Test workflows

vvolkgang:
## 🎟️ Tracking

PM-33025

## 📔 Objective

Test workflows tune-up inbound! While at it, updated the branch name patterns too. In my testing, some of these changes didn't net obvious runtime gains but they may require merging to main first and analysing future runs. 

Performance improvements I confirmed had immediate gains:

1. Large runners - better hardware, everything runs faster
2. Remove postCompileScripts (SwiftLint, SwiftFormat lint) and postBuildScripts (Settings.bundle version bump) so they no longer run on every build, these should run as git hooks instead. Found them by looking at the .xcresult files where we can confirm these tasks were taking a big chunk and weren't parallelised 
image

3. Caching DerivedData
4. Split cache into separate restore/save steps so caches are only saved on push to main, similar to the android [setup-gradle action](https://github.com/gradle/actions/blob/main/docs/setup-gradle.md#using-the-cache-read-only) approach
    * PRs cache were reducing cache effectiveness by evicting `main` branch cache.
    * When key didn't match it would fallback to an invalid cache from a different PR.
    * Uploading cache takes time, by not caching PR updates we're saving 3-4mins per run. 

Improvements that need further testing:

1. Xcode parallelisation flags suggested by @fedemkr - looking at .xcresult we can see tasks are already being parallelised but, we'll have to test these flags in future PR runs 
2. Enabling Xcode build cache - based on https://bitrise.io/blog/post/lifting-the-hood-on-build-cache-for-xcode
3. Disabling Compiler Index - responsible for Autocomplete and rich navigation in Xcode, which we don't need in our CI environment https://stackoverflow.com/questions/71109718/what-is-the-compiler-index-store-enable-xcode-build-settings-default-value
4. set `IgnoreFileSystemDeviceInodeChanges` so Xcode treats restored DerivedData files as up-to-date - source: https://stackoverflow.com/questions/53753511/is-it-possible-to-copy-an-xcode-derived-data-cache

Changes that were tested but removed: 
1. Restore file timestamps via chetan/git-restore-mtime-action - didn't see an immediate runtime impact and I'm hoping `IgnoreFileSystemDeviceInodeChanges` is enough to deal with this. If not, we should look into using the original python script instead of introducing an action: https://github.com/MestreLion/git-tools/blob/main/git-restore-mtime
2. Caching homebrew dependencies - didn't save us much, we may introduce it later if needed.


.github/workflows/test-bwa.yml
```diff
@@ -165,6 +214,7 @@ jobs:
             -derivedDataPath build/DerivedData \
             -test-timeouts-enabled yes \
             -maximum-test-execution-time-allowance 2 \
+            -showBuildTimingSummary
```
withinfocus: ⛏️ Missing a slash here.

withinfocus (dismissed):
Early review, since I was analyzing this for Actions approvals:

- Keep `actions/cache` (combined) for Mint and SPM. They're stable, lockfile-keyed, small, and the combined action already avoids re-saving them. Converting them adds two extra steps each and abandons the familiar, post-phase-managed pattern for no benefit. The `save` steps the PR adds for them are gated `cache-hit != 'true'` and `ref == main` which is functionally close to what the combined action already did on its own, minus the PR-branch save (and PR-branch saves of a lockfile-hashed cache are cheap and rare anyway).
- Use the `restore` / `save` split only for `DerivedData`, where it's load-bearing.

Does `DerivedData` need to be keyed by `github.sha` at all though? The per-SHA key is what forces the manual save. If a more stable key were viable (it usually isn't for `DerivedData`, since it depends on all sources -- this is why incremental-build caches universally use prefix+save-control), the whole manual apparatus would dissolve.

.github/workflows/test-bwa.yml
vvolkgang: Lingering local change, pushed!

vvolkgang:
> * (and PR-branch saves of a lockfile-hashed cache are cheap and rare anyway).

Only changed those too because I found that wasn't the case, between renovate PRs, SDK updates and team PRs that modified these files a big chunk of those caches came from branches other than `main`, when the file hash didn't match and it used the fallback `macOS-spm-` key most often than not that would be a cache that didn't came from `main` too, and while it had a cache-hit it wasn't actually useful. Took me a while to figure why the profiler was still showing time spent there.  

>Does DerivedData need to be keyed by github.sha at all though? 

Any ideas? Currently it's trying to avoid the same issue we had with the other caches, save step `if:` ensures we only save in `main`, commit sha ensures it's unique per merge to `main` and when commit sha doesn't match it falls back to a relevant cache. When I tested caching incremental changes it didn't impact build times much, and with each DerivedData cache being >700MB, their upload consistently increased runtime. Not caching PR updates saved 3-4min per run overall, IIRC.

withinfocus:
> Only changed those too because I found that wasn't the case, between renovate PRs, SDK updates and team PRs that modified these files a big chunk of those caches came from branches other than `main`, when the file hash didn't match and it used the fallback `macOS-spm-` key most often than not that would be a cache that didn't came from `main` too, and while it had a cache-hit it wasn't actually useful. Took me a while to figure why the profiler was still showing time spent there.

Makes sense and is justified then.

> Any ideas? Currently it's trying to avoid the same issue we had with the other caches, save step `if:` ensures we only save in `main`, commit sha ensures it's unique per merge to `main` and when commit sha doesn't match it falls back to a relevant cache. When I tested caching incremental changes it didn't impact build times much, and with each DerivedData cache being >700MB, their upload consistently increased runtime. Not caching PR updates saved 3-4min per run overall, IIRC.

I vibed a bit on this one and alas, no, so you're on the right track most likely. Said vibe:

> Having thought about it more: yes, your scheme is the right idiom, and I'd argue it's actually *required*, not just convenient: GitHub Actions caches are immutable once written, so you can't update a fixed key like `…-deriveddata-main`; the second save just no-ops with "cache already exists". A rolling unique suffix is the only way to push a fresh snapshot on each merge, and the `restore-keys` prefix then always pulls the newest one. `github.sha` is a fine token for that (`github.run_id` would behave identically); the sha has the minor edge of tying the entry to exact commit content.
> 
> One refinement worth considering on the key *prefix*: fold the Xcode/toolchain version (and maybe `Package.resolved`) in, e.g. `${{ runner.os }}-deriveddata-testbwpm-${xcode}-${{ github.sha }}`. DerivedData restored across an Xcode bump or a major dependency change can carry stale incremental state, and since you're also setting `IgnoreFileSystemDeviceInodeChanges` (telling Xcode to trust restored files as up-to-date), the usual safety net is off. Pinning the prefix to the toolchain means a runner-image upgrade cleanly invalidates instead of silently restoring mismatched module caches.
> 
> Last thing, genuinely curious rather than blocking: once PR incremental saves are gone, what's the measured win from the DerivedData *restore* by itself, separate from the xlarge runner and the postCompileScript removal? You mentioned incremental caching didn't move build times much, so I want to make sure the 700MB entry is still earning its keep on the restore side.

vvolkgang:
> One refinement worth considering on the key _prefix_: fold the Xcode/toolchain version (and maybe `Package.resolved`) in, e.g. `${{ runner.os }}-deriveddata-testbwpm-${xcode}-${{ github.sha }}`. DerivedData restored across an Xcode bump or a major dependency change can carry stale incremental state, and since you're also setting `IgnoreFileSystemDeviceInodeChanges` (telling Xcode to trust restored files as up-to-date), the usual safety net is off. Pinning the prefix to the toolchain means a runner-image upgrade cleanly invalidates instead of silently restoring mismatched module caches.

@withinfocus yeah makes sense, added!

project-pm.yml
```diff
@@ -191,25 +191,6 @@ targets:
...
-          if [[ ! "$PATH" =~ "/opt/homebrew/bin" ]]; then
-            PATH="/opt/homebrew/bin:$PATH"
-          fi
-          mint run swiftformat --lint --lenient .
-        name: SwiftFormat Lint
-        basedOnDependencyAnalysis: false
-    postBuildScripts:
-      - path: Scripts/update_settings_version_number.sh
-        name: "Settings.bundle: Update Version Number"
-        basedOnDependencyAnalysis: false
```
fedemkr, line 212: 🤔 I don't see this being added anywhere else like pre-commit or some other place. Is it ok to be removed? Same applies to BWA.

vvolkgang: It was running post build (ie, wasn't useful for the actual build). Not entirely sure why we had this in the first place, perhaps tied to how the project was initially setup. Linting is already handled in the pre-commit script too! https://github.com/bitwarden/ios/blob/a5e63300f52d59953bb653aa77d28ffb18c10abb/Scripts/pre-commit#L23

matt-livefront: `Scripts/update_settings_version_number.sh` adds the version number to the app's Settings bundle. If this is removed, I think the version number displayed in iOS Settings app > Apps > Bitwarden will default back to `1.0.0 (0)`. If we don't need that though, we should also remove the script and this: https://github.com/bitwarden/ios/blob/ca0539167380046305726c9717cb1451f2a1dc50/Bitwarden/Application/Support/Settings.bundle/Root.plist#L9-L17

I personally like lint warnings to show up in Xcode, which is enabled by running the lint steps post build. If we remove those, it makes it a bit harder to fix lint warnings which can't be automatically fixed (some of the SwiftLint warnings). Would it make sense to just skip the lint script if running on CI?

vvolkgang: @matt-livefront this was running post-build, reckon it would only affect local incremental builds?

About keeping linting for local dev, I'll take a look!

matt-livefront: It should still affect all builds - it modifies the settings bundle in the build output path (`"${CODESIGNING_FOLDER_PATH}/Settings.bundle/Root.plist"`), which doesn't exist until after the build completes (or at least copy bundle resources completes). It does that so it isn't modifying the actual source in the repo itself. I think post-build still runs before code signing.

vvolkgang: @matt-livefront "You're absolutely right"! When I initially tested it I only looked at the "App Information" section which has the right info, and missed the "Authenticator Settings" section above it which defaults to 1.0.0 (1) with this change. I'm assuming the App Information section wasn't always there and now that we have it, we can remove this?

https://github.com/bitwarden/ios/blob/d2b4188f37dc4e5ef1a5bcbfb7f33063ca68a3cf/Bitwarden/Application/Support/Settings.bundle/Root.plist#L9-L18

image

matt-livefront: 🤔 Where do you see the "App Information" section? I only see Bitwarden/Authenticator Settings.

Screenshot 2026-06-30 at 2 29 38 PM

vvolkgang: @matt-livefront same screen. Confirmed App Infomation and App Marketplace Information sections are only visible when a device has a third-party App Store setup. :finnadie: Checked other apps and most aren't setting versions here, seems like an old pattern that isn't used anymore, went ahead and removed it.

project-pm.yml
```diff
@@ -191,25 +191,6 @@ targets:
           - $(SRCROOT)/Bitwarden/Application/Support/Settings.bundle/Acknowledgements.plist
           - $(SRCROOT)/Bitwarden/Application/Support/Settings.bundle/Acknowledgements.latest_results.txt
           - $(SRCROOT)/Bitwarden/Application/Support/Settings.bundle/Acknowledgements
```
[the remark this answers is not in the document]
vvolkgang: Version is set by our build workflows, this refers to a field we were adding to the Settings page. That has been removed, more info in this [thread](https://github.com/bitwarden/ios/pull/2739#discussion_r3703541689).

fedemkr (approved):

PR #2744, opened 2026-06-02
[PM-27138] feat: Add credit card input form to Test Harness

morganzellers-bw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-27138

## 📔 Objective

Add a credit card input form to the Test Harness app in order to test card autofill features.

## 📸 Screenshots

Simulator Screenshot - iPhone 17 Pro - 2026-06-02 at 15 06 18


TestHarnessShared/UI/Platform/Application/Support/Localizations/en.lproj/Localizable.strings
```diff
@@ -1,3 +1,17 @@
+"CardAutofillForm" = "Card Autofill Form";
+"CardAutofillFormDescription" = "Tap each field and select Bitwarden from the QuickType bar to autofill card details. Requires iOS 18+, Bitwarden PM as AutoFill provider, and a saved Card cipher.";
```
fedemkr: ⚠️ BWPM can't autofill Cards directly from the QuickType bar. You need to use the long press autofill menu on each field, and fill each one independently, see https://bitwarden.com/help/auto-fill-ios/#long-press-a-text-field

fedemkr (commented):
🤔 Why do you need iOS 17 available for all Card form related stuff?

TestHarnessShared/UI/Platform/Application/Support/Localizations/en.lproj/Localizable.strings
morganzellers-bw: Ah okay, yeah I'm seeing now the QuickType bar surfaces cards from Wallet/Safari. 

The iOS 17+ requirement is for these `textContentType` values:

```swift
.creditCardName
.creditCardExpirationMonth
.creditCardExpirationYear
.creditCardSecurityCode
```

fedemkr: Oh alright thanks for clarifying it 😄 . Yes, iOS' apps have that ability for autofill cards but not third-party providers yet. I hope the upcoming iOS version adds support for it as well.

TestHarnessShared/UI/Platform/Application/Support/Localizations/en.lproj/Localizable.strings
```diff
@@ -1,3 +1,17 @@
+"CardAutofillForm" = "Card Autofill Form";
+"CardAutofillFormDescription" = "Long press a field, then select Credit Card to autofill with Bitwarden. Requires Bitwarden PM as AutoFill provider and a saved Card cipher.";
```
fedemkr: 🤔 As far as I remember selecting "Credit Card" on the autofill long press context menu did nothing with the BWPM app's autofill extension. Has that changed automatically somehow? AFAIK it should go with the same flow as passwords, i.e. Long Press -> Autofill -> Password.
⛏️ In order to follow the guidelines on the other localizations, this should have the suffix `DescriptionLong`

TestHarnessShared/UI/Platform/Application/Support/Localizations/en.lproj/Localizable.strings
```diff
@@ -1,3 +1,17 @@
...
+"CardholderNameValue" = "Cardholder Name: %@";
+"CardNumber" = "Card Number";
+"CardNumberValue" = "Card Number: %@";
+"EnterCardDetailsAbove" = "Enter card details above";
+"ExpirationMonth" = "Expiration Month";
+"ExpirationMonthValue" = "Expiration Month: %@";
+"ExpirationYear" = "Expiration Year";
+"ExpirationYearValue" = "Expiration Year: %@";
+"SecurityCode" = "Security Code";
+"SecurityCodeValue" = "Security Code: %@";
```
fedemkr: To follow how we name keys in other places we should replace the part of the parameter with X, Y, Z depending on the number of the parameters, e.g.:
`CardholderNameValue` -> `CardholderNameX`

fedemkr: ⛏️  @morganzellers-bw I see these were changed to a format like `CardholderNameValueX` instead of `CardholderNameX`. I would follow the same as we do in the other apps, and the reason that we don't add "value" is that because that's not part of the string value as if for some reason in the future we'd have `"Cardholder name value: %@"` as a string, the key would have already been used by this one. So in order to avoid this potential problem we just copy the same string value as the key with PascalCase (in this case omitting the colon) and add the X, Y, Z depending on the parameters.

KatherineInCode: 🤔 Why are these things that take variables at all, rather than just re-using the other localizations in situ?

So instead of `View.text = Localizations.expirationYearX(year)`, it would be `View.text = "\(Localizations.expirationYear): \(year)"` or something to that effect?

(If we're concerned about order around colon for RTL languages and the like, I'd recommend a generic `XColonY` string)

morganzellers-bw: @KatherineInCode I like the approach of re-using the existing localizations, I'll make that update.

Thanks @fedemkr I understand, I'll make sure to keep the keys the same going forward!

fedemkr: @KatherineInCode We can't use localizations in situ for RTL + because languages may use different characters for colon separator as in Chinese where it looks similar but it's different because of spacing. 
I think `XColonY` is the right approach as well.

KatherineInCode: Agreed on `XColonY`

TestHarnessShared/UI/Autofill/SimpleLoginForm/SimpleLoginFormView.swift
```diff
@@ -52,11 +52,11 @@ struct SimpleLoginFormView: View {
                 if !store.state.username.isEmpty || !store.state.password.isEmpty {
                     VStack(alignment: .leading, spacing: 8) {
                         if !store.state.username.isEmpty {
-                            Text(Localizations.usernameValue(store.state.username))
+                            Text(Localizations.usernameValueX(store.state.username))
                                 .styleGuide(.body)
                         }
                         if !store.state.password.isEmpty {
-                            Text(Localizations.passwordValue(String(repeating: "•", count: store.state.password.count)))
+                            Text(Localizations.passwordValueX(store.state.password))
```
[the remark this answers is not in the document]
morganzellers-bw: This is intentional, this allows the user testing the app to see that the correct value is filled.

fedemkr (approved):

PR #2745, opened 2026-06-02
[PM-27134] feat: Add passkey registration view to Test Harness

morganzellers-bw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-27134

## 📔 Objective

Add a passkey creation screen to the Test Harness app.

## 📸 Screenshots

IMG_0001


TestHarnessShared/UI/Autofill/CreatePasskey/CreatePasskeyProcessor.swift
```diff
@@ -0,0 +1,219 @@
...
+    // MARK: Private
+
+    /// Orchestrates state transitions and calls `performRegistration`.
+    private func registerPasskey() async {
+        state.status = .inProgress
+        do {
+            try await performRegistration(state.rpId, state.userName, state.displayName)
+            state.status = .success
+        } catch {
+            state.status = .failure(error.localizedDescription)
```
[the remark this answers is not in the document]
morganzellers-bw: Fixed in [c0918cd](https://github.com/bitwarden/ios/pull/2745/commits/c0918cd8d93ef093270e5009a258862cbc3a3068)

TestHarnessShared/UI/Autofill/CreatePasskey/CreatePasskeyProcessor.swift
```diff
@@ -0,0 +1,221 @@
...
+        }
+    }
+
+    /// The real passkey registration implementation using `ASAuthorizationController`.
+    private func performPasskeyRegistration(
+        rpId: String,
+        userName: String,
+        displayName: String,
+    ) async throws {
+        guard #available(iOS 17, *) else {
```
fedemkr: 🎨 Given that passkeys in particular do not exist on iOS < 17 I think we can just mark the whole processor as available 17.

TestHarnessShared/UI/Autofill/CreatePasskey/CreatePasskeyProcessor.swift
```diff
@@ -0,0 +1,221 @@
...
+            throw PasskeyRegistrationError.notAvailable
+        }
+
+        let window = await delegate?.presentationAnchorForPasskeyRegistration() ?? UIWindow()
+
+        let provider = ASAuthorizationPlatformPublicKeyCredentialProvider(relyingPartyIdentifier: rpId)
+        let challenge = Data((0 ..< 32).map { _ in UInt8.random(in: 0 ... 255) })
```
fedemkr: ⚠️ Even though this could be fine, using `SecRandomCopyBytes` is preferred for the `challenge` as the implementation behind is more explicit and avoid future ambiguity if Swift's implementation ever change.
```
var challengeBytes = [UInt8](repeating: 0, count: 32)
SecRandomCopyBytes(kSecRandomDefault, challengeBytes.count, &challengeBytes)
let challenge = Data(challengeBytes)
```
or you can use `CryptoKit`:
```
import CryptoKit

let challenge = SymmetricKey(size: .bits256).withUnsafeBytes { Data($0) }
```

TestHarnessShared/UI/Autofill/CreatePasskey/CreatePasskeyProcessor.swift
```diff
@@ -0,0 +1,221 @@
...
+            controller.presentationContextProvider = handler
+            // Retain the handler for the lifetime of the controller sheet.
+            objc_setAssociatedObject(
+                controller,
+                &passkeyContinuationKey,
+                handler,
+                .OBJC_ASSOCIATION_RETAIN_NONATOMIC,
+            )
+            controller.performRequests()
+        }
```
fedemkr: 🎨 Would there be a way to avoid `objc_setAssociatedObject`? Perhaps having a helper that has the continuation which is passed around here?

TestHarnessShared/UI/Autofill/CreatePasskey/CreatePasskeyView.swift
fedemkr, line 1: ⛏️ Add docs to private functions and accessibility identifiers where applicable.

TestHarnessShared/UI/Autofill/CreatePasskey/CreatePasskeyProcessor.swift
```diff
@@ -0,0 +1,221 @@
...
+        performRegistration = { _, _, _ in throw PasskeyRegistrationError.notAvailable }
+        super.init(state: CreatePasskeyState())
+
+        performRegistration = { [weak self] rpId, userName, displayName in
+            try await self?.performPasskeyRegistration(
+                rpId: rpId,
+                userName: userName,
+                displayName: displayName,
+            )
+        }
```
fedemkr: 🤔 Why is `performRegistration` needed at all? Can't you call `performPasskeyRegistration(...)` function directly in `registerPasskey`?

fedemkr (dismissed):
⚠️ I see the passkey registration flow is triggered here so a provider can save it but no passkey information is saved for the Test Harness app (like e.g. the challenge). How is it suppose to check it's working if it doesn't save the information to use the passkey afterwards?
Additionally, if we want this to work as close as a real production app as possible, the `challenge` should actually come from a server so perhaps we should discuss this internally on which approach we want to follow for this.

TestHarnessShared/UI/Autofill/CreatePasskey/CreatePasskeyProcessor.swift
morganzellers-bw: Yes, I like that idea better too!

TestHarnessShared/UI/Autofill/CreatePasskey/CreatePasskeyProcessor.swift
morganzellers-bw: I was using it to set the closure from the tests, but injecting it will be better I think.

TestHarnessShared/UI/Autofill/CreatePasskey/CreatePasskeyProcessor.swift
morganzellers-bw: I ended up adding `CreatePasskeyCoordinator` that has the `presentationAnchor` injected into it and holds onto the continuation instead of using `objc_setAssociatedObject`.

TestHarnessShared/UI/Platform/Root/ScenarioPicker/ScenarioPickerState.swift
```diff
@@ -33,7 +33,7 @@ struct ScenarioPickerState: Equatable {
             ScenarioItem(id: "createAccountForm", title: Localizations.createAccountForm, route: .createAccountForm),
             ScenarioItem(id: "simpleLoginForm", title: Localizations.simpleLoginForm, route: .simpleLoginForm),
             ScenarioItem(id: "totpAutofillForm", title: Localizations.totpAutofillForm, route: .totpAutofillForm),
-            ScenarioItem(id: "passkeyAutofill", title: Localizations.passkeyAutofill, route: nil),
+            ScenarioItem(id: "registerPasskey", title: Localizations.registerPasskey, route: .registerPasskey),
```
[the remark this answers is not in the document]
morganzellers-bw: Fixed in commit [36a1738](https://github.com/bitwarden/ios/pull/2745/commits/36a173820c3ffa78f77aba8d856f76f2f8189317)

PR #2747, opened 2026-06-03
[PM-37920] feat: Hide storage cost row when value is zero or absent

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37920

## 📔 Objective

Storage cost row was always rendered in the subscription cost breakdown, even when the amount was zero. This PR adds a `showStorageCost` computed property that returns `true` only when the storage cost is greater than zero, and conditionally renders the row accordingly — matching the web behaviour described in the ticket.

## 📸 Screenshots
| Before | After |
|--------|--------|
|  |  |

### DIFF
--- a/BitwardenShared/UI/Billing/PremiumPlan/PremiumPlanState.swift
+++ b/BitwardenShared/UI/Billing/PremiumPlan/PremiumPlanState.swift
@@ -142,6 +142,11 @@ struct PremiumPlanState: Equatable {
         !discount.isEmpty
     }
 
+    /// Whether the storage cost row should be shown (only when cost is greater than zero).
+    var showStorageCost: Bool {
+        (subscription?.storageCost ?? 0) > 0
+    }
+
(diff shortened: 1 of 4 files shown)


BitwardenShared/UI/Billing/PremiumPlan/PremiumPlanState.swift
```diff
@@ -142,6 +142,11 @@ struct PremiumPlanState: Equatable {
         !discount.isEmpty
     }
 
+    /// Whether the storage cost row should be shown (only when cost is greater than zero).
```
SaintPatrck, line 145: ⚠️ In the [client logic](https://github.com/bitwarden/clients/blob/034a8de76f5ad7e31f4a43b9d175ca0fc17bdd64/libs/pricing/src/components/cart-summary/cart-summary.component.html#L106-L135), storage cost is only hidden when the response property is `null`. If it's present and the cost is `$0.00`, the row should still display.

PR #2748, opened 2026-06-03
[PM-37886] refactor: Update Key Connector vault unlock to use SDK's keyConnectorUrl method

matt-livefront:
## 🎟️ Tracking

[PM-37886](https://bitwarden.atlassian.net/browse/PM-37886)

## 📔 Objective

Migrates the Key Connector vault unlock to let the SDK handle Key Connector communication, removing the app's direct `GET /user-keys` network call.

#### Changes

- `AuthRepository.unlockVaultWithKeyConnectorKey`: switches from `InitUserCryptoMethod.keyConnector(masterKey:userKey:)` (which required a prior `getMasterKeyFromKeyConnector` API call) to `InitUserCryptoMethod.keyConnectorUrl(url:keyConnectorKeyWrappedUserKey:)`, which delegates the Key Connector fetch to the SDK.
- Removes `getMasterKeyFromKeyConnector` from `KeyConnectorService` and `KeyConnectorAPIService` protocols and implementations.
- Deletes `KeyConnectorUserKeyRequest` and `KeyConnectorUserKeyResponseModel`.

[PM-37886]: https://bitwarden.atlassian.net/browse/PM-37886?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


fedemkr (dismissed):

BitwardenShared/UI/Auth/Login/SingleSignOn/SingleSignOnProcessor.swift
```diff
@@ -235,8 +234,10 @@ extension SingleSignOnProcessor: SingleSignOnFlowDelegate {
                             keyConnectorURL: keyConnectorUrl,
                             orgIdentifier: state.identifierText,
                         )
+                        coordinator.hideLoadingOverlay()
                         await coordinator.handleEvent(.didCompleteAuth)
                     } catch StateServiceError.noAccountCryptographicState {
+                        coordinator.hideLoadingOverlay()
```
KatherineInCode, line 240: 🎨  To one of the things Claude generated for this, is there a way we could test to make sure the overlay is hidden here?

🤔 Would the hiding of the overlay be something we could `defer`?

KatherineInCode, line 240: 🎨  To one of the things Claude generated for this, is there a way we could test to make sure the overlay is hidden here?

🤔 Would the hiding of the overlay be something we could `defer`?

matt-livefront: > 🎨 To one of the things Claude generated for this, is there a way we could test to make sure the overlay is hidden here?

Yep, I updated the tests for this.
 
> 🤔 Would the hiding of the overlay be something we could `defer`?

That would be ideal. At one point we ran into a problem where defer was causing the loading overlay to not hide when transitioning from the auth flow to the vault, which is why they are manually hidden in some places like this before completing auth. I'd like to clean this up, but I'd prefer to do it separately. I created a ticket: https://bitwarden.atlassian.net/browse/PM-38748.

fedemkr (approved):

PR #2749, opened 2026-06-03
[PM-29571] feat: Save password credentials without user interaction

fedemkr:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-29571

## 📔 Objective

Implements the iOS 26.2 silent password save flow (`performWithoutUserInteractionIfPossible(savePasswordRequest:)`). When the system calls this entry point, the extension unlocks the vault with the neverlock key and saves a new login cipher directly — no UI shown.

Note: `Info.plist` declares the `SupportsSavePasswordCredentials` capability required for the system to invoke this flow.

## 📸 Screenshots

https://github.com/user-attachments/assets/cace1116-1242-4042-84cf-5fde78f36216


BitwardenAutoFillExtension/CredentialProviderViewController.swift
```diff
@@ -312,6 +312,43 @@ extension CredentialProviderViewController {
...
+                with: DefaultCredentialProviderContext(
+                    .savePasswordWithoutUserInteraction(savePasswordRequest),
+                ),
+            )
+            await savePassword(savePasswordRequest: savePasswordRequest)
+        }
+    }
+
+    @available(iOSApplicationExtension 26.2, *)
+    private func savePassword(savePasswordRequest: ASSavePasswordRequest) async {
```
matt-livefront, line 334: ⛏️ Add docs?

BitwardenShared/Core/Vault/Extensions/CipherView+Extensions.swift
```diff
@@ -17,4 +18,65 @@ extension CipherView {
...
+            folderId: nil,
+            collectionIds: [],
+            key: nil,
+            name: name ?? URL(string: uri)?.host ?? uri,
+            notes: nil,
+            type: .login,
+            login: BitwardenSdk.LoginView(
+                username: username,
+                password: password,
+                passwordRevisionDate: creationDate,
```
matt-livefront: 🤔 Should this be set on initial creation? It looks like we normally don't set it until after you update the initial password.

fedemkr: Yes, you're right. Made the fix, nice catch!

matt-livefront (approved):

PR #2730, opened 2026-05-29
[PM-38152] feat: Add Passport to vault list and search

SaintPatrck:
## 🎟️ Tracking

- [PM-38152](https://bitwarden.atlassian.net/browse/PM-38152)

## 📔 Objective

Surface synced **Passport** ciphers in the vault list and search, gated behind the
`pm-32009-new-item-types` feature flag — mirroring the Driver's License integration (#2728). With the
flag ON a "Passports" group appears in the Types section and passport ciphers list, search, and offer
overflow actions; with the flag OFF they are hidden but still persist.

- Adds `.passport` to `CipherType.newItemTypesGatedCases` so the feature-flag filter applies automatically.
- `VaultListGroup.passport` (group filtering, name/title); id-card icon for the row and item; no add-button for the group.
- Type-section passport group row (gated) + `countPerCipherType`; `belongsToGroup` matching for list/search.
- Overflow ("more options") menu: **Copy passport number** action (View / Edit / Archive come automatically).

Search flows through the SDK-computed `cipherView.subtitle`, the same path Driver's License uses.

Add/Edit and View passport screens remain intentionally stubbed (`EmptyView()`) pending **PM-38153** / **PM-38154**.

## 📸 Screenshots

| Figma | Actual |
|--------|--------|
|  |  | 

> [!IMPORTANT]
> Figma currently displays the incorrect "Current icons". Icons will be updated as part of a separate ticket.

[PM-38152]: https://bitwarden.atlassian.net/browse/PM-38152?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2754, opened 2026-06-04
llm: Add alphabetical enum case ordering rule to iOS skill

SaintPatrck:
## 🎟️ Tracking

#2730

## 📔 Objective

Codifies a convention in the `implementing-ios-code` skill: when adding a case to an existing enum, insert it at its alphabetical position in **every `switch` over that enum** (across all layers) rather than appending it — matching the surrounding order.

This came out of #2730, where the new `.passport` case was appended instead of slotted in alphabetically across three switch statements. Documenting the rule so future work follows it.


withinfocus (approved):

SaintPatrck:
Thanks @withinfocus

PR #2746, opened 2026-06-02
[PM-38149] feat: Add Driver's License Add/Edit screens

SaintPatrck:
## 🎟️ Tracking

- Jira: [PM-38149](https://bitwarden.atlassian.net/browse/PM-38149)

## 📔 Objective

Adds the Driver's License Add/Edit screen and wires it through the shared `AddEditItem` flow, so users can create and edit Driver's License items (gated behind the `newItemTypes` feature flag).

- New `AddEditDriversLicenseItem` file-set (state, action, view) covering all 11 fields; the license number is a hidden field with a reveal toggle.
- Completes the create and edit state↔SDK round-trip, including the model→SDK mapping.
- Exposes Driver's License in the add-item type picker, gated behind the `newItemTypes` flag.

Date of birth, issue date, and expiration date are read-only for now, shown as long localized dates (e.g. "August 10, 2026"); editable date pickers will follow in separate work.

## 📸 Screenshots

| Figma | Actual |
|--------|--------|
|  |   | 

[PM-38149]: https://bitwarden.atlassian.net/browse/PM-38149?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2755, opened 2026-06-04
llm: Add review-driven iOS skill conventions

SaintPatrck:
## 🎟️ Tracking

Conventions surfaced during review of #2746.

> [!NOTE]
> Consolidates #2761 (alphabetization rules) into this PR so the `implementing-ios-code`, `testing-ios-code`, and `reviewing-changes` skill guidance stays consistent and contradiction-free in one place.

## 📔 Objective

Codify recurring iOS review feedback as durable skill conventions so future work follows them without re-litigation.

**`implementing-ios-code`:**
- Use `#Preview` macros for new views (don't add new `PreviewProvider` structs); snapshot coverage comes from a test that instantiates the view directly.
- Member ordering: alphabetize members within a `// MARK:` section and function/initializer parameters (defaults, variadic, then trailing-closure last). Carve-outs: UI objects follow visual layout order; protocol conformance ordering is not enforced.
- Data models: alphabetize stored properties and initializer parameters; insert at the alphabetical position and update the synthesized initializer and call sites.
- Prefer `static` members over file-scope globals.
- Prefer typed over stringly-typed values (with a deferred-SDK-contract exception).

**`testing-ios-code`:**
- Reusable SDK/`CipherView` fixtures live in the shared `BitwardenSdk+<Area>Fixtures.swift` file for the model's area, not in individual test files.
- Split large test files into per-feature Swift Testing extensions.
- Snapshot tests: two valid patterns (instantiate the view directly vs. iterate `PreviewProvider._allPreviews`); when a view could go either way, use `AskUserQuestion` to let the user choose.

**`reviewing-changes`:**
- Alphabetization checklist updated to the canonical, source-linked member/parameter rules, with the UI-visual-layout carve-out.


.claude/skills/testing-ios-code/SKILL.md
```diff
@@ -142,3 +142,9 @@ New protocols need mocks:
 3. Or just build — Sourcery runs automatically in pre-build phase
 
 See `references/mock-generation.md` for full details.
+
+## Conventions
+
+- **Fixtures live in the shared fixtures file** — reusable SDK-model and `CipherView` fixtures belong in `BitwardenShared/UI/Vault/PreviewContent/BitwardenSdk+VaultFixtures.swift` (e.g. `CipherView.cardFixture()`), not co-located inside an individual test file. Add new reusable fixtures there.
```
KatherineInCode: ⛏️ Except when they belong in the `AuthenticatorShared` version of that file

.claude/skills/testing-ios-code/SKILL.md
```diff
@@ -142,3 +142,9 @@ New protocols need mocks:
...
+- **Split large test files per feature** — when a test file grows unwieldy, break it into `<Type>Tests+<Feature>.swift` extensions (e.g. `AddEditItemProcessorTests+DriversLicense.swift`), and author the new file in Swift Testing.
+- **Snapshot tests instantiate the view directly** — build the view with a `MockProcessor`/`Store` in `setUp`, then `assertSnapshot(of: subject.navStackWrapped, as:)` per state (all prefixed `disabletest_`). Do not iterate `PreviewProvider._allPreviews`; new views use `#Preview` macros, which the snapshot harness does not consume. See `AddEditSendItemView+SnapshotTests.swift` for the pattern.
```
KatherineInCode: 🤔 There are cases when iterating over all the previews in the file is actually the better way to go, and there's no way to do it over all the macros. It really is a human-discretion sort of thing which path to go.

.claude/skills/testing-ios-code/SKILL.md
```diff
@@ -142,3 +142,9 @@ New protocols need mocks:
 3. Or just build — Sourcery runs automatically in pre-build phase
 
 See `references/mock-generation.md` for full details.
+
+## Conventions
+
+- **Fixtures live in the shared fixtures file** — reusable SDK-model and `CipherView` fixtures belong in `BitwardenShared/UI/Vault/PreviewContent/BitwardenSdk+VaultFixtures.swift` (e.g. `CipherView.cardFixture()`), not co-located inside an individual test file. Add new reusable fixtures there.
+- **Split large test files per feature** — when a test file grows unwieldy, break it into `<Type>Tests+<Feature>.swift` extensions (e.g. `AddEditItemProcessorTests+DriversLicense.swift`), and author the new file in Swift Testing.
```
fedemkr, line 149: 🤔 Should we explicitly explain "unwieldy" to something like number of new lines / the file becoming > than N lines?

.claude/skills/testing-ios-code/SKILL.md
SaintPatrck: ❓ Gotcha. Would you prefer we leave this out, in that case? 

💭 Alternatively we could have Claude ask the user when it encounters this fork in the road.

.claude/skills/testing-ios-code/SKILL.md
fedemkr: ℹ️ Just in case, there are other fixture files like `BitwardenSdk+AuthFixtures` where we put SDK model fixtures. So it depends on the area that the model belongs to.
🤔 Should we further explain the convention to take that into consideration?

.claude/skills/testing-ios-code/SKILL.md
SaintPatrck: We can, if you have a threshold in mind. Or we can let this ride as-is for now, and specify if/when we see Claude having trouble deciding what "unwieldy" is. 

I usually lean more towards less specificity/verbosity unless it's warranted, only because every token has monetary cost. If we are more verbose to start with we have a harder time gauging if we're over-explaining. Conversely, if we let Claude use its best judgement, we'll find our pretty quickly if we need to elaborate more, thus justifying extra cost.

fedemkr: Sounds good leaving as is.

.claude/skills/testing-ios-code/SKILL.md
SaintPatrck: Updated prose. Let me know what you think.

.claude/skills/testing-ios-code/SKILL.md
KatherineInCode: I think it's a place where we need input by a person, so probably having it ask makes sense.

.claude/skills/testing-ios-code/SKILL.md
KatherineInCode: Looks reasonable to me

.claude/skills/testing-ios-code/SKILL.md
SaintPatrck: ✅ Updated — the snapshot bullet now presents both patterns and instructs Claude to ask the user when a view could go either way (direct instantiation vs. iterating `PreviewProvider._allPreviews`).

KatherineInCode (approved):

SaintPatrck:
Thanks @fedemkr and @KatherineInCode

PR #2756, opened 2026-06-04
chore: Include new item type data in cipher requests

SaintPatrck:
## Tracking

N/A - Internal tech-debt post SDK update.

## 📔 Objective

Completes the deferred wiring of the new item types — bank account, driver's license, and passport — into `CipherRequestModel` now that their SDK dependencies are in place. These types were already handled on the read path but were missing from the add/update request body, so their data wasn't persisted to the server. Maps all three (add and update share the same request path) and adds request-body regression tests.

### DIFF
--- a/BitwardenShared/Core/Vault/Models/Request/CipherRequestModel.swift
+++ b/BitwardenShared/Core/Vault/Models/Request/CipherRequestModel.swift
@@ -64,11 +64,11 @@ struct CipherRequestModel: JSONRequestBody {
     /// The organization identifier for the cipher.
     let organizationID: String?
 
-    /// The password history for this cipher.
-    let passwordHistory: [CipherPasswordHistoryModel]?
-
     /// Passport data if the cipher is a passport.
     let passport: CipherPassportModel?
+
+    /// The password history for this cipher.
+    let passwordHistory: [CipherPasswordHistoryModel]?
 
     /// Whether the user needs to be re-prompted for their master password prior to autofilling the
     /// cipher's password.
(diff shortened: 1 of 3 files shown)


BitwardenShared/Core/Vault/Models/Request/CipherRequestModel.swift
```diff
@@ -61,6 +67,9 @@ struct CipherRequestModel: JSONRequestBody {
     /// The password history for this cipher.
     let passwordHistory: [CipherPasswordHistoryModel]?
 
+    /// Passport data if the cipher is a passport.
+    let passport: CipherPassportModel?
```
KatherineInCode, line 68: ⛏️ Alphabetize

BitwardenShared/Core/Vault/Models/Request/CipherRequestModel.swift
```diff
@@ -104,6 +115,7 @@ extension CipherRequestModel {
             notes: cipher.notes,
             organizationID: cipher.organizationId,
             passwordHistory: cipher.passwordHistory?.map(CipherPasswordHistoryModel.init),
+            passport: cipher.passport.map(CipherPassportModel.init),
```
KatherineInCode, line 117: ⛏️ Alphabetize (should happen as the synthesized initializer is updated once the properties are alphabetized)

BitwardenShared/Core/Vault/Models/Request/CipherRequestModelTests.swift
```diff
@@ -0,0 +1,133 @@
+import BitwardenSdk
+import XCTest
+
+@testable import BitwardenShared
+
+class CipherRequestModelTests: BitwardenTestCase {
```
KatherineInCode: 🎨 We prefer new tests to be Swift Testing

🤔 Would it be possible to do subsuites for the different types? Mostly just thinking about possible architecture sorts of things. May not actually be worth doing here.

KatherineInCode (commented):
Just some small things

BitwardenShared/Core/Vault/Models/Request/CipherRequestModelTests.swift
SaintPatrck: Subsuites sounds reasonable. Android does something similar in places with nested test sub-classes. I agree, that's scope for another PR though.

fedemkr (commented):
Looks good, but it seems to be missing Tracking section and the Jira ticket ID in the title

SaintPatrck:
@fedemkr there isn't a ticket to track this. I just happened to notice it was overlooked after the SDK update applied.

fedemkr (approved):

PR #2758, opened 2026-06-05
[PM-38150] feat: Add Driver's License view screen

SaintPatrck:
## 🎟️ Tracking

- [PM-38150](https://bitwarden.atlassian.net/browse/PM-38150)

## 📔 Objective

Adds the read-only detail view for Driver's License items. Renders the **License details** section: first/middle/last name (with copy), license number (masked with a reveal toggle + copy), date of birth / issue date / expiration date (formatted long dates), issuing country / state / authority, and license class. Empty fields are hidden. Wires the view into the per-type dispatch in `ViewItemDetailsView`.

## 📸 Screenshots

| Figma | Actual |
|--------|--------|
|  |  | 

[PM-38150]: https://bitwarden.atlassian.net/browse/PM-38150?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenShared/UI/Vault/VaultItem/ViewItem/ViewDriversLicenseItem/ViewDriversLicenseItemView.swift
```diff
@@ -0,0 +1,249 @@
...
+
+/// A view for displaying the contents of a driver's license item.
+struct ViewDriversLicenseItemView: View {
+    // MARK: Properties
+
+    /// The `Store` for this view.
+    @ObservedObject var store: Store<DriversLicenseItemState, ViewItemAction, Void>
+
+    var body: some View {
+        if !store.state.isLicenseDetailsSectionEmpty {
```
fedemkr, line 16: ♻️ Not to change int his PR, but @KatherineInCode @matt-livefront do you remember why we took this approach of checking if the specific cipher type details are empty inside the cipher type item view instead of doing it in `ViewItemDetailsView` wrapping each specific cipher type item view with the empty check?

BitwardenShared/UI/Vault/VaultItem/ViewItem/ViewDriversLicenseItem/ViewDriversLicenseItemView.swift
```diff
@@ -0,0 +1,249 @@
...
+
+                    licenseClassItem
+                }
+            }
+        }
+    }
+
+    // MARK: Private Views
+
+    @ViewBuilder private var firstNameItem: some View {
```
fedemkr, line 49: ⛏️ Add docs as well as the other view builder properties.

KatherineInCode (dismissed):

BitwardenShared/UI/Vault/VaultItem/ViewItem/ViewDriversLicenseItem/ViewDriversLicenseItemView.swift
```diff
@@ -0,0 +1,249 @@
...
+                    .styleGuide(.body)
+                    .foregroundColor(SharedAsset.Colors.textPrimary.swiftUIColor)
+                    .accessibilityIdentifier("DriversLicenseNumberEntry")
+            } accessoryContent: {
+                PasswordVisibilityButton(
+                    accessibilityIdentifier: "ShowDriversLicenseNumberButton",
+                    isPasswordVisible: isVisible,
+                ) {
+                    store.send(.driversLicenseItemAction(.toggleLicenseNumberVisibilityChanged(!isVisible)))
+                }
```
fedemkr, line 112: 🤔 I know we're doing this on this way on Card, but I feel Identity/Login handles this better by not having to pass `isVisible` value here. I mean IMO the view should only send the processor `.toggleLicenseNumberVisibilityChanged` without parameters and then the processor check the current state and invert the visibility of the field.
I feel like doing it this way removes some logic from the View and it goes to the processor where it should live.
@bitwarden/team-ios @matt-livefront what do you think? Also I think we should have a tech-debt to update the Card approach as well.

BitwardenShared/UI/Vault/VaultItem/ViewItem/ViewDriversLicenseItem/ViewDriversLicenseItemView.swift
matt-livefront: I don't really mind it here, creating the view is cheap and it keeps the empty check encapsulated within this view. If you move the check into the parent view, that view would need to reach into the child state to do the empty check.

BitwardenShared/UI/Vault/VaultItem/ViewItem/ViewDriversLicenseItem/ViewDriversLicenseItemView.swift
matt-livefront: From a unidirectional data flow perspective, it would be better to not have the view do the toggling and dictate the new value to the processor. We're not very consistent on this though, and SwiftUI sometimes fights us. If you use a binding, the binding sends the new value as part of the action, which ends up being easier to use than to ignore it. I like updating this and the card cases though, it would end up being simpler than manually toggling it.

BitwardenShared/UI/Vault/VaultItem/ViewItem/ViewDriversLicenseItem/ViewDriversLicenseItemView.swift
fedemkr: Oh that makes sense, thanks Matt!

fedemkr (approved):

PR #2760, opened 2026-06-05
[PM-38442] feat: Add Fill-Assist endpoint and models

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38442

## 📔 Objective

Adds the networking layer and data models needed to fetch Fill-Assist targeting rules from the [map-the-web](https://github.com/bitwarden/map-the-web) repository.

**API & Models**
- `FillAssistAPIService` — fetches `forms.v0.json` via a dedicated `mapTheWebService` HTTP service
- `FormsMapResponseModel` and supporting types (`FormsMapHostEntry`, `FormsMapPathnameEntry`, `FormsMapContent`, `FormsMapSelector`) — decode the forms map JSON, including a custom `init(from:)` that silently excludes null-valued host entries

**Environment URL plumbing**
- `fillAssistRules: String?` added to `EnvironmentServerConfigResponseModel` and `EnvironmentServerConfig` so the server can provide a custom rules URL
- `fillAssistRulesUrl: URL?` added to `EnvironmentURLData` (persisted per account) with a `with(fillAssistRulesUrl:)` copy helper
- `fillAssistRulesURL: URL` added to `EnvironmentURLs`, `EnvironmentService` protocol, and both `DefaultEnvironmentService` implementations (PM and Authenticator), defaulting to the GitHub releases URL
- `DefaultStateService.setServerConfig` stamps the server-provided `fillAssistRules` URL into the active account's `EnvironmentURLData` so it is available the next time `loadURLsForActiveAccount()` runs


BitwardenKit/Core/Platform/Models/Domain/EnvironmentURLs.swift
```diff
@@ -131,6 +137,7 @@ public extension EnvironmentURLs {
             identityURL = environmentURLData.identity ?? URL(string: "https://identity.bitwarden.com")!
             webVaultURL = environmentURLData.webVault ?? URL(string: "https://vault.bitwarden.com")!
         }
+        fillAssistRulesURL = URL(string: "https://github.com/bitwarden/map-the-web/releases/latest/download")!
```
andrebispo5: ❓ **QUESTION**: `fillAssistRulesURL` is always set to the hardcoded GitHub URL — the new `EnvironmentServerConfig.fillAssistRules` field is decoded but never consulted here, so the server-config override described in the PR description does not yet take effect. Is the override intentionally deferred to a follow-up, or should this read from the active server config (and `EnvironmentURLData` for self-hosted) before falling back to the GitHub default?

<details>
<summary>Details</summary>

The PR description states: "Includes `fillAssistRules` in `EnvironmentServerConfigResponseModel` so the server can override the base URL." However, `EnvironmentURLs.init(environmentURLData:)` sets `fillAssistRulesURL` unconditionally to `https://github.com/bitwarden/map-the-web/releases/latest/download`, and `DefaultEnvironmentService.fillAssistRulesURL` returns that value as-is.

As a result, even when the server returns a non-nil `fillAssistRules`, the `mapTheWebService` in `APIService` (whose `baseURLGetter` reads `environmentService.fillAssistRulesURL`) still points at the hardcoded GitHub URL. If the override behavior is intended to land in a later PR, a brief note in the PR description (or a `// TODO: PM-XXXXX` here) would help reviewers understand the scope boundary.
</details>

BitwardenShared/Core/Autofill/Models/FormsMapResponseModel.swift
```diff
@@ -0,0 +1,87 @@
...
+// MARK: - FormsMapResponseModel
+
+/// The response model for the Forms Map from the map-the-web repository.
+///
+struct FormsMapResponseModel: Equatable, JSONResponse {
+    /// The schema version of the map.
+    let schemaVersion: String
+
+    /// The host entries keyed by hostname. Null-valued entries are excluded during decoding.
+    let hosts: [String: FormsMapHostEntry]
```
andrebispo5, line 19: ❓ **QUESTION**: The doc comment claims "Null-valued entries are excluded during decoding," but no custom decoder is implemented for `FormsMapResponseModel`, so any `null` value in the `hosts` JSON object will cause `Dictionary` decoding to throw `DecodingError.typeMismatch` against `FormsMapHostEntry`. Does the upstream `forms.v0.json` ever emit null host entries, and if so, should `hosts` use a custom `init(from:)` that filters them out (or should the comment be removed)?

<details>
<summary>Details</summary>

Swift's synthesized `Codable` conformance for `[String: FormsMapHostEntry]` does not skip null values — it requires every value to decode into the non-optional element type. If the map-the-web schema may include entries like `"foo.com": null`, decoding the entire response will fail. Either:

- Implement a custom `init(from:)` for `FormsMapResponseModel` that decodes `hosts` as `[String: FormsMapHostEntry?]` and filters `nil` values, or
- Remove the comment if the schema is guaranteed not to produce nulls.

A quick test case using fixture JSON with a `null` host entry would lock the behavior down either way.
</details>

BitwardenShared/Core/Platform/Services/API/APIService.swift
```diff
@@ -22,6 +22,9 @@ class APIService {
     /// The API service used for user identity requests.
     let identityService: HTTPService
 
+    /// The API service used for map-the-web requests.
+    let mapTheWebService: HTTPService
```
matt-livefront: ❓ Everything else in this PR uses "fill assist" as the naming convention / prefix. Does it make sense to match that here (e.g. `fillAssistService`) for consistency?

BitwardenKit/Core/Platform/Models/Domain/EnvironmentURLData.swift
```diff
@@ -150,6 +156,26 @@ public extension EnvironmentURLData {
         return url?.host
     }
 
+    /// Returns a copy of this `EnvironmentURLData` with the given Fill-Assist rules URL applied.
+    ///
+    /// - Parameter fillAssistRulesUrl: The Fill-Assist rules URL to set.
+    /// - Returns: A new `EnvironmentURLData` with `fillAssistRulesUrl` updated.
+    ///
+    func with(fillAssistRulesUrl: URL?) -> EnvironmentURLData {
```
matt-livefront: 🤔 If the fill assist URL needs to be updated / mutable, I wonder if we should just make it a `var`?

BitwardenShared/Core/Autofill/Services/API/FillAssistAPIServiceTests.swift
```diff
@@ -0,0 +1,51 @@
...
+        stateService = MockStateService()
+        let accountTokenProvider = MockAccountTokenProvider()
+        accountTokenProvider.getTokenReturnValue = "ACCESS_TOKEN"
+        subject = APIService(
+            accountTokenProvider: accountTokenProvider,
+            activeAccountStateProvider: activeAccountStateProvider,
+            client: client,
+            stateService: stateService,
+        )
+    }
```
matt-livefront: 🎨 There's an APIService convenience initializer in the tests which makes it so you don't have to provide all of these properties:


```suggestion
    var client: MockHTTPClient!
    var subject: FillAssistAPIService!

    // MARK: Initialization

    init() {
        client = MockHTTPClient()
        subject = APIService(client: client)
    }
```

BitwardenShared/Core/Platform/Services/API/APIService.swift
andrebispo5: Fill Assist is the name of the feature client side but the repo and origin of data is the _Map The Web_ repo which has zero references to Fill Assist. 👍  https://github.com/bitwarden/map-the-web

BitwardenKit/Core/Platform/Models/Domain/EnvironmentURLData.swift
andrebispo5: Well it would be easier, I'll make the change and add a comment to why this specific field needs to be var.

BitwardenShared/Core/Platform/Services/API/APIService.swift
matt-livefront: I get that, but the base URL for the service is `fillAssistRulesUrl`. Using `mapTheWeb` for this single HTTP service seems inconsistent. Unless `mapTheWebService` is going to be used for things other than fill assist, I think it'd be better to just be consistent with the "fill assist" naming convention, which makes things easier to follow and search for in the codebase.

andrebispo5: makes sense I'm all for making things easier to find in the codebase and the name being different makes it harder than it should be 👍

matt-livefront (approved):

PR #2761, opened 2026-06-05
llm: Extend iOS alphabetization rules to properties and parameters

SaintPatrck:
## 🎟️ Tracking

#2756

## 📔 Objective

Extends the alphabetization conventions in the `reviewing-changes` and `implementing-ios-code` skills, which previously only covered enum cases, tests, and static/computed members. Per the [Swift style guide](https://contributing.bitwarden.com/contributing/code-style/swift#alphabetization), all `// MARK:`-section members and function/initializer parameters are alphabetized (default-value, variadic, and trailing-closure parameters last).

This came out of #2756, where stored properties and an initializer parameter were appended rather than slotted in alphabetically. Documenting the rule so future work follows it.

### DIFF
--- a/.claude/skills/implementing-ios-code/SKILL.md
+++ b/.claude/skills/implementing-ios-code/SKILL.md
@@ -27,6 +27,7 @@ Implement from the bottom up:
 
 **Data Models** (if needed)
 - Request/Response types in `Core/<Domain>/Models/Request/` and `Response/`
+  - Alphabetize stored properties and initializer parameters (parameters with default values go last). When you add one to an existing type, insert it at its alphabetical position and update the synthesized initializer and every call site to match — don't append.
 - Enum types in `Core/<Domain>/Models/Enum/`
   - Keep cases alphabetical. When you add a case to an existing enum, insert it at its alphabetical position in **every `switch` over that enum** (across all layers) rather than appending it — match the surrounding order.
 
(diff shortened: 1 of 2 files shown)


.claude/skills/implementing-ios-code/SKILL.md
```diff
@@ -27,6 +27,7 @@ Implement from the bottom up:
 
 **Data Models** (if needed)
 - Request/Response types in `Core/<Domain>/Models/Request/` and `Response/`
+  - Alphabetize stored properties and initializer parameters (parameters with default values go last). When you add one to an existing type, insert it at its alphabetical position and update the synthesized initializer and every call site to match — don't append.
```
KatherineInCode, line 30: 🤔 Do we also want to include the "visual layout order" provision here, or is just in the review skill enough?

SaintPatrck: Good call — yes, the implementing skill needs it too. We are consolidating these skill changes into #2755 so the implementing/reviewing/testing guidance stays consistent in one place. As part of that, the implementing skill's member-ordering convention now carries the same carve-out: **UI objects (views, view modifiers) follow visual layout order, not alphabetical** (alongside the protocol-conformance-not-enforced note). Closing this PR in favor of #2755.

SaintPatrck:
Consolidated into #2755 so the alphabetization rules land alongside the related skill conventions and stay contradiction-free in one place. Kat's visual-layout-order question is addressed there — the implementing skill now carries the same carve-out. The single content commit (fb9c92395) was cherry-picked into #2755.

.claude/skills/implementing-ios-code/SKILL.md
SaintPatrck: I think that makes sense.

Side note: I'm closing this to incorporate the changes into #2755.

PR #2764, opened 2026-06-05
[PM-37886] refactor: Pass keyConnectorKeyWrappedUserKey through LoginUnlockMethod instead of persisting in state

matt-livefront:
## 🎟️ Tracking

[PM-37886](https://bitwarden.atlassian.net/browse/PM-37886)

## 📔 Objective

When a user logs in with key connector, their `keyConnectorKeyWrappedUserKey` is being persisted to the device as `encryptedUserKey`. For key connector, once the user is logged in and the vault is unlocked, this key is no longer used. As such, it doesn't need to be persisted.

Additional context: https://github.com/bitwarden/ios/pull/2677#discussion_r3282105331

This updates the key connector flows to pass the `keyConnectorKeyWrappedUserKey` from the login response through `LoginUnlockMethod.keyConnector`, which can then be used to unlock the vault in `AuthRepository.unlockVaultWithKeyConnectorKey()`.

- `LoginUnlockMethod.keyConnector` now carries `encryptedUserKey: String?` populated from `response.key` in `AuthService.unlockMethod(for:)`, eliminating the need to read it back from state during unlock.
- `AuthRepository.unlockVaultWithKeyConnectorKey` accepts a non-optional `encryptedUserKey: String` parameter directly, removing the redundant `getAccountEncryptionKeys` state read.
- `KeyConnectorService.convertNewUserToKeyConnector` no longer persists `encryptedUserKey` to state (both v1 and v2 paths); the `cryptographicState` (private key) is still persisted as required for future operations.
- `SingleSignOnProcessor` now uses an explicit `guard let encryptedUserKey` to trigger the key connector migration dialog for new users, replacing the implicit `noAccountCryptographicState` error path.

[PM-37886]: https://bitwarden.atlassian.net/browse/PM-37886?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/UI/Auth/Login/TwoFactorAuth/TwoFactorAuthProcessor.swift
+++ b/BitwardenShared/UI/Auth/Login/TwoFactorAuth/TwoFactorAuthProcessor.swift
@@ -245,11 +245,15 @@ final class TwoFactorAuthProcessor: StateProcessor<TwoFactorAuthState, TwoFactor
                     ),
                 )
                 coordinator.navigate(to: .dismiss)
-            case let .keyConnector(keyConnectorUrl):
+            case let .keyConnector(keyConnectorKeyWrappedUserKey, keyConnectorUrl):
                 guard let orgIdentifier = state.orgIdentifier else {
                     throw TwoFactorAuthError.missingOrgIdentifier
                 }
+                guard let keyConnectorKeyWrappedUserKey else {
+                    throw StateServiceError.noEncUserKey
+                }
                 try await services.authRepository.unlockVaultWithKeyConnectorKey(
+                    keyConnectorKeyWrappedUserKey: keyConnectorKeyWrappedUserKey,
                     keyConnectorURL: keyConnectorUrl,
                     orgIdentifier: orgIdentifier,
                 )
--- a/BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
+++ b/BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
@@ -267,10 +267,15 @@ protocol AuthRepository: AnyObject {
     /// Attempts to unlock the user's vault with the user's Key Connector key.
     ///
     /// - Parameters:
-    ///   - keyConnectorUrl: The URL to the Key Connector API.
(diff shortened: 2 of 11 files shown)


BitwardenShared/UI/Auth/Login/TwoFactorAuth/TwoFactorAuthProcessor.swift
```diff
@@ -245,11 +245,15 @@ final class TwoFactorAuthProcessor: StateProcessor<TwoFactorAuthState, TwoFactor
...
                 )
                 coordinator.navigate(to: .dismiss)
-            case let .keyConnector(keyConnectorUrl):
+            case let .keyConnector(keyConnectorKeyWrappedUserKey, keyConnectorUrl):
                 guard let orgIdentifier = state.orgIdentifier else {
                     throw TwoFactorAuthError.missingOrgIdentifier
                 }
+                guard let keyConnectorKeyWrappedUserKey else {
+                    throw StateServiceError.noEncUserKey
+                }
```
fedemkr, line 254: 🤔 Could any of the other unlock mechanism end up throwing this error? If so, not sure if we could add something here to indicate it's the key connector key and also to indicate that happens on this path. As I think it won't appear on the stack trace on the logs.

fedemkr (commented):
Looks good, just one small ⛏️

BitwardenShared/UI/Auth/Login/TwoFactorAuth/TwoFactorAuthProcessor.swift
matt-livefront: Good question, this has been refactored in the next PR (https://github.com/bitwarden/ios/pull/2780), which updates the error to `TwoFactorAuthError.missingKeyConnectorKey`:

https://github.com/bitwarden/ios/blob/28699bd27e832801f5f71e154c577c6f0c9cb475/BitwardenShared/UI/Auth/Login/TwoFactorAuth/TwoFactorAuthProcessor.swift#L255-L256

fedemkr (approved):

PR #2769, opened 2026-06-09
[PM-38145] feat: Add View Bank Account screen

SaintPatrck:
## 🎟️ Tracking

- [PM-38145](https://bitwarden.atlassian.net/browse/PM-38145)

## 📔 Objective

Adds the read-only **View Bank Account** detail screen for the Password Manager.

- Renders the **Account details** section; empty fields are hidden, and the whole section is hidden when the item has no bank account details.
- Copy buttons on account number, routing number, SWIFT code, IBAN, and PIN.
- Account number, PIN, and IBAN are masked by default with reveal toggles.
- Bank name, name on account, account type, branch number, and bank contact phone display as plain text.

## 📸 Screenshots

| Figma | Actual |
|--------|--------|
|  |  | 

[PM-38145]: https://bitwarden.atlassian.net/browse/PM-38145?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/UI/Vault/VaultItem/ViewItem/ViewBankAccountItem/ViewBankAccountItemView.swift
+++ b/BitwardenShared/UI/Vault/VaultItem/ViewItem/ViewBankAccountItem/ViewBankAccountItemView.swift
@@ -1,0 +1,302 @@
+import BitwardenKit
+import BitwardenResources
+import BitwardenSdk
+import SwiftUI
+
+// MARK: - ViewBankAccountItemView
+
+/// A view for displaying the contents of a bank account item.
+struct ViewBankAccountItemView: View {
+    // MARK: Properties
+
+    /// The `Store` for this view.
+    @ObservedObject var store: Store<BankAccountItemState, ViewItemAction, Void>
+
+    var body: some View {
+        if !store.state.isBankAccountDetailsSectionEmpty {
+            SectionView(Localizations.accountDetails, contentSpacing: 8) {
+                ContentBlock {
+                    textField(
+                        title: Localizations.bankName,
+                        value: store.state.bankName,
+                        valueAccessibilityIdentifier: "BankAccountBankNameEntry",
+                    )
+
+                    textField(
+                        title: Localizations.nameOnAccount,
+                        value: store.state.nameOnAccount,
+                        valueAccessibilityIdentifier: "BankAccountNameOnAccountEntry",
+                    )
+
+                    accountTypeItem
+
+                    accountNumberItem
+
+                    copyableTextField(
+                        title: Localizations.routingNumber,
+                        value: store.state.routingNumber,
+                        valueAccessibilityIdentifier: "BankAccountRoutingNumberEntry",
+                        copyButtonAccessibilityIdentifier: "BankAccountCopyRoutingNumberButton",
+                        copyField: .routingNumber,
+                    )
+
+                    textField(
+                        title: Localizations.branchNumber,
+                        value: store.state.branchNumber,
+                        valueAccessibilityIdentifier: "BankAccountBranchNumberEntry",
+                    )
+
+                    pinItem
+
+                    copyableTextField(
+                        title: Localizations.swiftCode,
+                        value: store.state.swiftCode,
+                        valueAccessibilityIdentifier: "BankAccountSwiftCodeEntry",
+                        copyButtonAccessibilityIdentifier: "BankAccountCopySwiftCodeButton",
+                        copyField: .swiftCode,
+                    )
+
+                    ibanItem
+
+                    textField(
+                        title: Localizations.bankContactPhone,
+                        value: store.state.bankContactPhone,
+                        valueAccessibilityIdentifier: "BankAccountContactPhoneEntry",
+                    )
+                }
+            }
+        }
+    }
+
+    // MARK: Private Views
+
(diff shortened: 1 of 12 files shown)


BitwardenShared/UI/Vault/VaultItem/ViewItem/ViewBankAccountItem/ViewBankAccountItemView.swift
```diff
@@ -0,0 +1,308 @@
...
+            .accessibilityElement(children: .contain)
+        }
+    }
+}
+
+// MARK: - BankAccountItemState
+
+extension BankAccountItemState {
+    /// Whether the bank account details section has no values to display.
+    var isBankAccountDetailsSectionEmpty: Bool {
```
matt-livefront: ❓ Should this be moved into the state file?

BitwardenShared/UI/Vault/VaultItem/ViewItem/ViewItemAction.swift
```diff
@@ -62,6 +65,30 @@ enum ViewItemAction: Equatable, Sendable {
 /// The text fields within the `ViewItemView` that can be copied.
 ///
 enum CopyableField {
+    /// The bank account number field.
+    case accountNumber
+
+    /// The bank contact phone field.
+    case bankContactPhone
```
matt-livefront, line 74: 🤔 Does it make sense to separate these fields by type, maybe just with a `// MARK: Bank Account Fields` comment?

BitwardenShared/UI/Vault/VaultItem/ViewItem/ViewBankAccountItem/ViewBankAccountItemView.swift
```diff
@@ -0,0 +1,308 @@
...
+                }
+            }
+        }
+    }
+
+    // MARK: Private Views
+
+    /// The account type field, displaying the selected type's localized name, hidden when no type is selected.
+    @ViewBuilder private var accountTypeItem: some View {
+        if case let .custom(type) = store.state.accountType {
```
fedemkr: ⛏️ There's a computed property `customValue` to get the value directly:
```suggestion
        if let type = store.state.accountType.customValue {
```

BitwardenShared/UI/Vault/VaultItem/ViewItem/ViewBankAccountItem/ViewBankAccountItemView.swift
```diff
@@ -0,0 +1,308 @@
...
+                PasswordText(password: accountNumber, isPasswordVisible: isVisible)
+                    .styleGuide(.body)
+                    .foregroundColor(SharedAsset.Colors.textPrimary.swiftUIColor)
+                    .accessibilityIdentifier("BankAccountNumberEntry")
+            } accessoryContent: {
+                PasswordVisibilityButton(
+                    accessibilityIdentifier: "ShowBankAccountNumberButton",
+                    isPasswordVisible: isVisible,
+                ) {
+                    store.send(.bankAccountItemAction(.toggleAccountNumberVisibilityChanged(!isVisible)))
```
fedemkr, line 106: 🎨  If possible, I would like us to move to the approach used in `ViewSSHKeyItemView` where we don't pass the final result of the state and just fire the action of toggling the visibility leaving the actual logic to the processor. This would align better with the architecture of the Unidirectional data flow.
Just in case, this is not blocking the PR and we can just tackle this when we refactor the other places where this approach is currently being used.
Same applies to IBAN and PIN item views.

fedemkr (commented):
Looks good, have some minor suggestions but they shouldn't block the PR.
I think with Matt suggested changes this would be good to approve.

BitwardenShared/UI/Vault/VaultItem/ViewItem/ViewBankAccountItem/ViewBankAccountItemView.swift
SaintPatrck: Relocated in 72f023b6a

BitwardenShared/UI/Vault/VaultItem/ViewItem/ViewItemAction.swift
SaintPatrck: Makes sense. Added markers in 72f023b6a

BitwardenShared/UI/Vault/VaultItem/ViewItem/ViewBankAccountItem/ViewBankAccountItemView.swift
SaintPatrck: Applied suggestion in 72f023b6a

BitwardenShared/UI/Vault/VaultItem/ViewItem/ViewBankAccountItem/ViewBankAccountItemView.swift
SaintPatrck: Logged https://bitwarden.atlassian.net/browse/PM-39242 to track this refactor.

matt-livefront (approved):

PR #2771, opened 2026-06-09
[PM-38515] feat: Add expand/collapse section headers to Password Manager vault

SaintPatrck:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38515

## 📔 Objective

Password Manager vault list sections can now be expanded and collapsed by tapping their headers. Each section's collapsed state is persisted per account and restored across app launches.

### Expanded

| Figma | Actual |
|--------|--------|
|  |  | 

### Collapsed

| Figma | Actual |
|--------|--------|
|  |  |

### DIFF
--- a/BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessor.swift
+++ b/BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessor.swift
@@ -178,6 +178,8 @@ final class VaultListProcessor: StateProcessor<
             state.searchText = newValue
         case let .searchVaultFilterChanged(newValue):
             state.searchVaultFilterType = newValue
+        case let .sectionExpandToggled(sectionId, isExpanded):
+            setSectionExpanded(sectionId: sectionId, isExpanded: isExpanded)
         case .showImportLogins:
             coordinator.navigate(to: .importLogins)
         case let .toastShown(newValue):
@@ -238,7 +240,8 @@ extension VaultListProcessor {
 
         state.shouldShowArchiveOnboardingActionCard = await services.stateService.shouldDoArchiveOnboarding()
 
-        state.shouldShowUpgradedToPremiumActionCard = await services.billingService.shouldShowUpgradedToPremiumActionCard()
+        state.shouldShowUpgradedToPremiumActionCard = await services.billingService
+            .shouldShowUpgradedToPremiumActionCard()
 
         let isBannerDismissed = await services.stateService.isPremiumUpgradeBannerDismissed()
         guard !isBannerDismissed else {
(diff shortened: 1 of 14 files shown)


BitwardenShared/UI/Vault/Vault/VaultList/VaultListView.swift
```diff
@@ -302,8 +302,14 @@ private struct SearchableVaultListView: View { // swiftlint:disable:this type_bo
 
             vaultFilterRow
 
-            ForEach(sections) { section in
-                VaultListSectionView(section: section) { item in
+            ForEach(sections.filter { !$0.items.isEmpty }) { section in
```
fedemkr: 🤔 I don't see that before we were filtering out sections without items. Is that something new for this? Also we're doing a similar filter on `VaultListSectionsBuilder` by not adding sections that don't have any items.

SaintPatrck: @fedemkr Good catch. Looks like https://bitwarden.atlassian.net/browse/PM-37931 snuck into this PR. If there's no objection, I can add a screenshot of empty groups being omitted and leave this change here. Thoughts?

fedemkr: I think it'd be better to have that be done in `VaultListSectionsBuilder -> addTypesSection()` checking `preparedData.countPerCipherType` to whether append the type item to the vault section. Therefore we have all logic there and reduce the in-view logic which in some flows improves performance as well.

SaintPatrck: I misread your original comment. Not enough caffeine yet. 😆 

You're correct, we should not be hiding the entire sections like this here. Removed in ff34c43.

fedemkr: Great thanks Patrick! Just to double check, are you going to implement PM-37931 in a different PR? If so, changes here look good 🚀 .

SaintPatrck: Yes, PM-37931 will be a separate PR.

BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessor.swift
```diff
@@ -656,9 +659,36 @@ extension VaultListProcessor {
...
+                services.errorReporter.log(error: error)
+            }
+        }
+    }
+
     /// Streams the user's vault list.
     private func streamVaultList() async {
         do {
+            let collapsedSectionIds = await (try? services.stateService.getCollapsedVaultListSectionIds()) ?? []
+            state.collapsedSectionIds = Set(collapsedSectionIds)
```
fedemkr, line 691: 🎨 I think we should log the error if `getCollapsedVaultListSectionIds` throws instead of ignoring it.
🤔 Also, I think we may want to move this to `appeared` as this stream [gets called](https://github.com/bitwarden/ios/blob/main/BitwardenShared/UI/Vault/Vault/VaultList/VaultListView.swift#L438) every time the `state.vaultFilterType` changes and as far as I can see that doesn't affect the collapsed vault list. However, maybe it introduces glitches on the vault load animation so it would need to be tested as streaming the vault and loading the collapsed vault list section will be done in parallel.
Thoughts? cc: @KatherineInCode @matt-livefront

SaintPatrck: We can certainly log the error. I'm not sure what value it would provide though. The only failure point is "no active user", which is not reconcilable and would already be caught and logged by other processes. IMO, logging this error would just add undue noise in logs.

The purpose for calling it here is to guarantee the sections are in the correct state when first drawn. If we move it to `appear` we risk having sections appear expanded and then snap closed.

fedemkr: Alright, makes sense. I'll approve

fedemkr (approved):

SaintPatrck:
Thanks @fedemkr

PR #2773, opened 2026-06-09
[PM-38808] Support optional encrypted Cipher.name

shane-melton:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38808

## 📔 Objective

The SDK is making the encrypted `Cipher.name` field optional (`String` → `String?`) as part of the blob encryption migration ([sdk-internal#1122](https://github.com/bitwarden/sdk-internal/pull/1122)) — blob-encrypted ciphers carry the name inside the sealed `data` blob rather than as a top-level field.

The SDK → network conversions `CipherRequestModel.init(cipher:)` and `CipherDetailsResponseModel.init(cipher:)` assign into non-optional `name` fields, so the now-nullable `name` no longer compiles. These fall back to an empty string when the name is absent.

The decrypted `CipherView` / `CipherListView` `name` fields are unchanged, so all UI is unaffected.

> ⚠️ `?? ""` emits a non-optional-operand warning until the bumped SDK lands — this PR is intended to merge alongside the SDK version bump.

### DIFF
--- a/BitwardenShared/Core/Vault/Extensions/BitwardenSdk+Vault.swift
+++ b/BitwardenShared/Core/Vault/Extensions/BitwardenSdk+Vault.swift
@@ -132,7 +132,7 @@ extension CipherDetailsResponseModel {
             identity: cipher.identity.map(CipherIdentityModel.init),
             key: cipher.key,
             login: cipher.login.map(CipherLoginModel.init),
-            name: cipher.name,
+            name: cipher.name ?? "",
             notes: cipher.notes,
             organizationId: cipher.organizationId,
             organizationUseTotp: cipher.organizationUseTotp,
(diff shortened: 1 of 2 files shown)


KatherineInCode:
🤔 Under what circumstances will the user see a blank name because of the nil-coalescing to an empty string?

shane-melton:
> 🤔 Under what circumstances will the user see a blank name because of the nil-coalescing to an empty string?

The encrypted name will only come back null for [blob encrypted ciphers](https://github.com/bitwarden/sdk-internal/pull/1122). In such cases the only way to access the decrypted name property is via decrypting the blob and access the resulting `CipherView.Name` property.

LRNcardozoWDF:
This has been addressed in [this PR](https://github.com/bitwarden/ios/pull/2825).

PR #2778, opened 2026-06-10
[PM-27140] feat: Add File Share view to Test Harness app.

morganzellers-bw:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-27140

## 📔 Objective

Adds a view to the Test Harness app where the Bitwarden Share feature can be tested.

## 📸 Screenshots

<details>
<summary>Expand</summary>

Simulator Screenshot - iPhone 17 Pro - 2026-06-10 at 12 23 55

</details>


TestHarnessShared/UI/Tools/FileShare/FileShareState.swift
```diff
@@ -0,0 +1,30 @@
+import Foundation
+
+/// The state for the file share test screen.
+///
+struct FileShareState: Equatable {
+    // MARK: Static Properties
+
+    /// The raw bytes written to the sample PDF file.
+    static let sampleFileData = Data("%PDF-1.4\n%%EOF\n".utf8)
```
fedemkr: 🎨 I think we could have a PDF file with some more data in it so we can check in the Bitwarden app that it got shared correctly by being able to opening it and see its contents. If I'm not mistaken this would be like an empty PDF file, right? So perhaps adding some text would help verifying after sharing.

TestHarnessShared/UI/Tools/FileShare/FileShareView.swift
fedemkr, line 1: 🎨 I think we should be adding the `accessibilityIdentifier` as we do in the main apps as well in case QA wants to perform some automation on the TestHarness as well.

fedemkr (commented):
Looks good, some minor improvements. Btw really like the approach of generating the files to share 🚀

TestHarnessShared/UI/Tools/FileShare/FileShareView.swift
morganzellers-bw: Added them for the share view and the previously created views 👍

TestHarnessShared/UI/Tools/FileShare/FileShareState.swift
morganzellers-bw: Updated to have the PDF contain "Hello World" & updated the UI message as well indicating this.

fedemkr (approved):

PR #2779, opened 2026-06-11
[PM-31718] fix: Guard against double-invoked ASCredentialIdentityStore completion handlers in AutoFill

matt-livefront:
## 🎟️ Tracking

[PM-31718](https://bitwarden.atlassian.net/browse/PM-31718)

## 📔 Objective

Attempts to fix a low-volume `EXC_BREAKPOINT` crash in the AutoFill credential provider seen on iOS 15, 16, and 17. The crash occurs on the `com.apple.NSXPCConnection.user.com.apple.SafariFoundation.CredentialProviderExtensionHelper` thread, in a compiler-generated async-bridging thunk inside `BitwardenShared` — specifically when updating the system credential identity store.

**Root cause:** We call `ASCredentialIdentityStore` APIs via their Swift `async` bridges. That bridge wraps a `CheckedContinuation` that must be resumed exactly once. On iOS 15–17, the system can invoke the underlying completion handler **more than once** over XPC (e.g. an interrupted/redelivered reply), and the second resume trips the fatal "continuation resumed more than once" trap. The completion-bridged APIs in play are `replaceCredentialIdentities(with:)` (iOS <17 path) and `removeAllCredentialIdentities()` (all OS versions — this is the iOS 17 crash path). It's a race, not a deterministic failure, which matches the low crash volume.

**Fix:** Introduces a `SafeCredentialIdentityStore` adapter conforming to `CredentialIdentityStore` that wraps `ASCredentialIdentityStore`. The completion-handler-based APIs now route through a single-resume continuation helper that resumes the continuation at most once and ignores any further callbacks. The iOS 17+ APIs that take `[ASCredentialIdentity]` are `async`-only (no completion variant to guard), so they forward unchanged.

Using a distinct adapter type, rather than redeclaring the async bridge on `ASCredentialIdentityStore`, keeps the inner calls bound to the concrete completion-handler variants, with no same-name async overload in scope to accidentally recurse into.

**Changes:**
- Add `SafeCredentialIdentityStore` (adapter) + `SingleResumeGuard` (NSLock-backed one-shot flag).
- Remove the empty `extension ASCredentialIdentityStore: CredentialIdentityStore {}`.
- Default the three call sites (`AuthService`, `ReviewPromptService`, `AutofillCredentialService`)
  to `SafeCredentialIdentityStore()` instead of `ASCredentialIdentityStore.shared`.
- Annotate the wrapper with `@available(iOS, deprecated: 18.0)` plus a removal note: the bug
  doesn't occur on iOS 18+, so once iOS 17 support is dropped the deprecation warning will flag
  the type for deletion and restoration of the one-line direct conformance.
- Add `SafeCredentialIdentityStoreTests` covering single, multiple synchronous, and concurrent
  (100×) callbacks, plus error and error-then-success cases.

No behavior change for the happy path (single callback → single resume); the guard only affects the duplicate-callback case that previously crashed.

[PM-31718]: https://bitwarden.atlassian.net/browse/PM-31718?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/Core/Autofill/Services/SafeCredentialIdentityStore.swift
+++ b/BitwardenShared/Core/Autofill/Services/SafeCredentialIdentityStore.swift
@@ -25,9 +25,8 @@ import AuthenticationServices
 ///   extension ASCredentialIdentityStore: CredentialIdentityStore {}
 ///   ```
 ///
-///   Then change the three call sites that default to `SafeCredentialIdentityStore()` back to
-///   `ASCredentialIdentityStore.shared` (`AuthService`, `ReviewPromptService`,
-///   `AutofillCredentialService`).
+///   Then change any call sites that default to `SafeCredentialIdentityStore()` back to
+///   `ASCredentialIdentityStore.shared`.
 @available(iOS, deprecated: 18.0, message: "Obsolete on iOS 18+; delete and restore the direct conformance (see note).")
 class SafeCredentialIdentityStore: CredentialIdentityStore {
     // MARK: Properties
--- a/BitwardenShared/Core/Autofill/Services/SafeCredentialIdentityStore.swift
+++ b/BitwardenShared/Core/Autofill/Services/SafeCredentialIdentityStore.swift
@@ -1,0 +1,134 @@
+import AuthenticationServices
+
+// MARK: - SafeCredentialIdentityStore
+
+/// A `CredentialIdentityStore` that wraps `ASCredentialIdentityStore` and guards against the
+/// system invoking a completion handler more than once.
+///
+/// Several `ASCredentialIdentityStore` completion-handler APIs can deliver their reply over XPC
+/// more than once (observed as a crash on iOS 15–17). The compiler-generated `async` bridge for
+/// those APIs resumes its continuation on every callback, and resuming a continuation more than
+/// once is a fatal error (`EXC_BREAKPOINT`). This wrapper calls the completion-handler variants
+/// directly and resumes the continuation at most once, ignoring any further callbacks.
+///
+/// Only the completion-handler-based APIs (`removeAllCredentialIdentities()` and
+/// `replaceCredentialIdentities(with:)`) can be guarded this way. The iOS 17+ APIs that take
+/// `[ASCredentialIdentity]` are `async`-only with no completion-handler variant, so they forward
+/// directly to the system's `async` methods.
+///
+/// - Note: The double-callback bug does not occur on iOS 18+. Once the minimum deployment target
+///   reaches iOS 18 (i.e. iOS 17 support is dropped), the `@available` annotation below will start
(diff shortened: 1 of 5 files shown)


BitwardenShared/Core/Autofill/Services/SafeCredentialIdentityStore.swift
```diff
@@ -0,0 +1,134 @@
...
+///   emitting deprecation warnings at the call sites. At that point this entire type can be deleted
+///   and the conformance restored to the original one-liner:
+///
+///   ```swift
+///   extension ASCredentialIdentityStore: CredentialIdentityStore {}
+///   ```
+///
+///   Then change the three call sites that default to `SafeCredentialIdentityStore()` back to
+///   `ASCredentialIdentityStore.shared` (`AuthService`, `ReviewPromptService`,
+///   `AutofillCredentialService`).
```
fedemkr: ⛏️ I think we can remove this so we don't need to maintain it if we add/remove more calls to this class.

BitwardenShared/Core/Autofill/Services/SafeCredentialIdentityStoreTests.swift
```diff
@@ -0,0 +1,67 @@
...
+    /// `withSingleResumedContinuation(_:)` does not crash when the completion handler is invoked
+    /// concurrently from multiple threads, resuming the continuation only once.
+    @Test
+    func withSingleResumedContinuation_concurrentCallbacks_resumesOnce() async throws {
+        try await SafeCredentialIdentityStore.withSingleResumedContinuation { completion in
+            DispatchQueue.concurrentPerform(iterations: 100) { _ in
+                completion(true, nil)
+            }
+        }
+    }
```
fedemkr, line 43: 🤔 Shouldn't this also test something for the output that it was called exactly once? Or are you inferring that because it's not crashing/throwing an error?

fedemkr (commented):
Nice catch, great fix 🎉 . Just some minor improvements.

BitwardenShared/Core/Autofill/Services/SafeCredentialIdentityStoreTests.swift
matt-livefront: Good question, I don't think there's a clean way to assert that it's called exactly once directly. Exactly once is enforced by the checked continuation, if it's never resumed, the test will hang / timeout, and if it's resumed more than once, it will crash. If you take out the `claimResume()` call, the tests will crash. So it is tested, just implicitly, through that mechanism.

fedemkr: Perfect!

fedemkr (approved):

PR #2780, opened 2026-06-11
[PM-37888] refactor: Read masterKeyWrappedUserKey from masterPasswordUnlock

matt-livefront:
## 🎟️ Tracking

[PM-37888](https://bitwarden.atlassian.net/browse/PM-37888)

## 📔 Objective

Updates instances of `AccountEncryptionKeys.encryptedUserKey` to instead use `UserDecryptionOptions.masterPasswordUnlock?.masterKeyEncryptedUserKey` since these are all called for users with a master password. I've left `AccountEncryptionKeys.encryptedUserKey` for now, to be cleaned up in a separate PR.

- Switch `unlockVaultFromLoginWithDevice`, `validatePassword`, and `KeyConnectorService.migrateUser` to read the master-key-encrypted user key from `UserDecryptionOptions.masterPasswordUnlock?.masterKeyEncryptedUserKey` instead of `AccountEncryptionKeys.encryptedUserKey`
- Add `setAccountMasterPasswordUnlock` call in the V1 JIT and TDE paths of `setMasterPassword` (these paths previously never populated `masterPasswordUnlock`, so subsequent reads would return nil)
- Add `setAccountMasterPasswordUnlock` call in `updateMasterPassword` so the canonical source stays current after a password change
- Set `encryptedUserKey: nil` on all master-password-related `setAccountEncryptionKeys` calls — the value now lives exclusively in `masterPasswordUnlock`

[PM-37888]: https://bitwarden.atlassian.net/browse/PM-37888?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
```diff
@@ -995,9 +995,16 @@ extension DefaultAuthRepository: AuthRepository {
         )
 
         try await accountAPIService.setPassword(requestModel)
+        try await stateService.setAccountMasterPasswordUnlock(
+            MasterPasswordUnlockResponseModel(
+                kdf: account.kdf,
+                masterKeyEncryptedUserKey: requestUserKey,
+                salt: account.profile.email,
+            ),
```
fedemkr, line 1002: 🤔 From what I can see the `MasterPasswordUnlockResponseModel` it's always initialized with the account's KDF and Email properties. Would it make sense to have an init taking an Account directly instead of being split, so we centralize the KDF and email there?

fedemkr (commented):
Looks good, just one possible improvement.

BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
matt-livefront: Yep, good idea!

fedemkr (approved):

PR #2782, opened 2026-06-12
[PM-29570] feat: Implement password generation with UI in AutoFill extension

fedemkr:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-29570

## 📔 Objective

Implements `prepareInterface(for:)` in `CredentialProviderViewController` to handle `ASGeneratePasswordsRequest` with user interaction (iOS 26.2+).

When the OS triggers this entry point, the extension:

- Initialises the app with a generatePasswordCredential(_:userInteraction: true) context (handled by prepareInterface(for:generatePasswordsRequest:)). A performWithoutUserInteraction(generatePasswordsRequest:) path also exists for the no-interaction case, using the generator with the SDK directly.
- After auth, AppCoordinator handles the .generatePasswordCredential app route by calling showGeneratePasswordCredential(), which creates a fresh GeneratorCoordinator (navigated directly to .generator(staticType: .password)) and presents it — bypassing VaultCoordinator entirely.
- A GeneratePasswordExtensionDelegate bridges GeneratorCoordinatorDelegate callbacks to CredentialProviderExtensionDelegate. It maps GeneratorType + password string to ASGeneratedPassword.Kind: .passphrase for passphrases, and a heuristic for passwords — .strong if the value contains any non-alphanumeric character, .alphanumeric otherwise. (A TODO PM-29569 tracks replacing the heuristic once the SDK exposes a mapping API.)
- On completion, CredentialProviderExtensionDelegate.completeGeneratePasswordRequest(kind:password:) calls extensionContext.completeGeneratePasswordRequest(results:) with the user-selected password and derived kind

> [!WARNING]  
> Info.plist values has not been updated as we need to wait for #2743 to be merged to have https://github.com/bitwarden/sdk-internal/pull/1102 included to get the appropriate generation rules. Currently the password generation rules are not the ones that come from the OS. The SDK incorporation and update of the `Info.plist` will be done in a future PR.

## 📸 Screenshots

https://github.com/user-attachments/assets/f6bc9785-0d93-44ea-8eb6-78145cba4f8a


BitwardenShared/UI/Autofill/Utilities/GeneratePasswordExtensionDelegateTests.swift
```diff
@@ -0,0 +1,88 @@
+import AuthenticationServices
+import XCTest
```
matt-livefront, line 2: ⛏️ Want to convert this to Swift Testing?

BitwardenShared/UI/Platform/Application/AppProcessor.swift
```diff
@@ -256,6 +256,31 @@ public class AppProcessor {
...
     }
 
+    /// Generates a password for a credential provider generate-password request without showing UI.
+    ///
+    /// - Parameter request: The generate-password request containing developer-specified rules.
+    /// - Returns: The generated password string.
+    ///
+    @available(iOSApplicationExtension 26.2, *)
+    public func generatePasswordCredential(request: any GeneratePasswordRequestProxy) async throws -> String {
+        try await unlockVaultWithNeverlockKey()
```
matt-livefront: 🤔 Does the vault need to be unlocked to generate a password? Or is this needed for something else?

BitwardenShared/UI/Vault/Vault/VaultRoute.swift
```diff
@@ -53,6 +53,9 @@ public enum VaultRoute: Equatable, Hashable {
     /// A route to the flight recorder settings in the settings tab.
     case flightRecorderSettings
 
+    /// A route to the generate password screen in the autofill extension.
+    case generatePassword
```
matt-livefront: 🤔 If this is something that wouldn't ever be used standalone, and only used with the extension, I wonder if we should name it to be specific to the extension?

BitwardenShared/UI/Vault/Vault/VaultCoordinator.swift
```diff
@@ -349,6 +356,31 @@ final class VaultCoordinator: Coordinator, HasStackNavigator { // swiftlint:disa
...
+        generatePasswordExtensionDelegate = delegate
+        // Use the vault coordinator's own stack navigator so the generator view is set as the
+        // root via `replace` rather than presented modally. At auth-completion time the stack
+        // navigator is not yet in the window hierarchy, so `present` would fail.
+        let coordinator = module.makeGeneratorCoordinator(
+            delegate: delegate,
+            stackNavigator: stackNavigator,
+        ).asAnyCoordinator()
+        coordinator.start()
+        coordinator.navigate(to: .generator(staticType: .password))
```
matt-livefront: 🤔 Given that we're essentially bypassing the vault here and just showing the generator, would it make sense to add a generator route to `AppRoute` and have `AppCoordinator` handle displaying the generator? Is that more or less complex than using the vault coordinator?

BitwardenShared/UI/Autofill/Utilities/GeneratePasswordExtensionDelegateTests.swift
fedemkr: As far as I know I can't, as we would require `@availability` check at the suite level (given `var subject: GeneratePasswordExtensionDelegate!`) and that's not possible for the time being. See https://github.com/swiftlang/swift-testing/issues/608

BitwardenShared/UI/Vault/Vault/VaultRoute.swift
fedemkr: I think it could potentially be used with `AppIntents` if we were to integrate the UI somehow there. Also I believe that given that it's in the `VaultRoute` it doesn't matter whether that is for the extension or for whatever else, it just tells what's the route. I can see that I may need to remove the comment of "autofill extension" on the comment though. Would that be fine or would you still like me to rename it?

BitwardenShared/UI/Platform/Application/AppProcessor.swift
fedemkr: I've asked Product because if we need to save the password in the password generator history, then we need to unlock the vault. Otherwise it wouldn't be necessary.

BitwardenShared/UI/Vault/Vault/VaultRoute.swift
matt-livefront: I guess there was two things that prompted my initial question. 1) the comment "in the autofill extension", which could be removed. But 2) `showGeneratePassword()` requires `appExtensionDelegate` to be set, which kind of limits it to being used in the extension, at least how it's set up today.

See my other [comment](https://github.com/bitwarden/ios/pull/2782#discussion_r3416220971) though, if this route just displays the generator as the root view, instead of showing it on/over the vault list, I feel like it would be better to promote this to `AppRoute`, since that would better match the view hierarchy of how it's displayed.

BitwardenShared/UI/Vault/Vault/VaultCoordinatorTests.swift
```diff
@@ -284,6 +284,14 @@ class VaultCoordinatorTests: BitwardenTestCase { // swiftlint:disable:this type_
         XCTAssertEqual(view.store.state.group, .identity)
     }
 
+    /// `navigate(to:)` with `.flightRecorderSettings` notifies the delegate to switch to the about
+    /// screen in the settings tab.
+    @MainActor
+    func test_navigateTo_flightRecorderSettings() throws {
```
matt-livefront, line 282: ⚠️ I think this is duplicated test, leading to the build failure.

BitwardenResources/Localizations/az.lproj/Localizable.strings
```diff
@@ -128,7 +128,7 @@
 "ItemUpdated" = "Element saxlanıldı";
 "Submitting" = "Təqdim olunur…";
 "Syncing" = "Sinxronlaşdırılır…";
-"SyncingLogins" = "Girişlər sinxronlaşdırılır...";
+"SyncingLogins" = "Girişlər sinxronlaşdırılır…";
```
matt-livefront: ❓ Any idea where these changes came from? These will just be overwritten with the next Crowdin import, right?

fedemkr: hmmm I thought I got rid of that before I pushed sorry. Yeah the ellipsis script run on non-english files and fixed the ellipsis to the correct format there which shouldn't do. I have the fix for the script but I was going to do that in a different PR. Thanks for catching this!

matt-livefront (approved):

PR #2796, opened 2026-06-16
[PM-20012] fix: Incorrect endpoint saved

LRNcardozoWDF:
## 🎟️ Tracking

[PM-20012](https://bitwarden.atlassian.net/browse/PM-20012)

## 📔 Objective

We are setting the pre-auth environment URLs when using the autofill. That data is shared with the main app, so if we use the autofill in middle of performing login we will have the wrong environment URLs for the account.

[PM-20012]: https://bitwarden.atlassian.net/browse/PM-20012?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenShared/Core/Platform/Services/EnvironmentService.swift
```diff
@@ -76,7 +80,6 @@ class DefaultEnvironmentService: EnvironmentService {
             urls = .defaultUS
         }
 
-        await setPreAuthURLs(urls: managedSettingsURLs ?? urls)
```
matt-livefront, line 79: 🤔 Are there any downsides to removing this? I think this currently is used so that the pre-auth URLs are updated when you switch the active account (the environment URLs when adding a new account are always set to the last active account). Assuming that's still a requirement, we should account for that before removing this.

LRNcardozoWDF: You may be right, I didn't test that flow. I'll test that, meanwhile this PR will be in draft. Thank you for bringing this to my attention.

BitwardenShared/Core/Auth/Services/AuthService.swift
```diff
@@ -984,8 +984,15 @@ class DefaultAuthService: AuthService { // swiftlint:disable:this type_body_leng
...
-            let urls = await stateService.getPreAuthEnvironmentURLs()
+            // Create the account. Prefer the environment URLs snapshotted for this email when the
+            // login flow started, since the global pre-auth URLs may have since been overwritten
+            // by an unrelated active-account sync (e.g. triggered by the AutoFill extension).
+            let urls: EnvironmentURLData? = if let accountCreationURLs = await stateService
+                .getAccountCreationEnvironmentURLs(email: email) {
+                accountCreationURLs
+            } else {
+                await stateService.getPreAuthEnvironmentURLs()
+            }
```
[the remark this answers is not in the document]
fedemkr: Agree with this could you check @LRNcardozoWDF ?

PR #2798, opened 2026-06-16
[PM-33861] Display organization notification banner on vault list

matt-livefront:
## 🎟️ Tracking

[PM-33861](https://bitwarden.atlassian.net/browse/PM-33861)

## 📔 Objective

Add support for the `organizationUserNotification` organization policy and surface it as a dismissible banner (action card) on the vault list.

Changes:
- Add the `pm-31948-org-user-notification-banner` feature flag.
- Add the `organizationUserNotification` `PolicyType` and its associated `PolicyOptionType` keys (`header`, `description`, `buttonText`, `showAfterEveryLogin`). 
- Add `OrganizationUserNotificationBannerData` and `PolicyService.getOrganizationUserNotificationBannerData()`, which (behind the feature flag) returns the banner data for the earliest-revision policy applying to the user, or `nil` when no policy applies or the policy has no `description`.
- Render the banner as an `ActionCard` on the vault list (empty and populated states). When the policy supplies both a header and button text, an acknowledgement button is shown; otherwise a standard X dismiss button is shown. Dismissing clears the banner.
- Extract the vault-list action cards into a dedicated `extension` and share a`policyWithEarliestRevisionDate` helper in `PolicyService`.

Currently this just dismisses the banner in memory. There's additional logic for persisting the dismissal of the banner coming in a future PR.

## 📸 Screenshots

| Description Only | Header + Description | Header, Description, Custom Button |
|--------|--------|--------|
| Screenshot 2026-06-12 at 11 39 53 AM | Screenshot 2026-06-12 at 11 33 24 AM | Screenshot 2026-06-12 at 11 38 56 AM | 

[PM-33861]: https://bitwarden.atlassian.net/browse/PM-33861?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenShared/Core/Vault/Services/PolicyService.swift
```diff
@@ -306,6 +314,15 @@ actor DefaultPolicyService: PolicyService {
...
     }
+
+    /// Returns the policy with the earliest revision date from the given list, or `nil` if the list is empty.
+    ///
+    /// - Parameter policies: The list of policies to search.
+    /// - Returns: The policy with the earliest revision date.
+    ///
+    private func policyWithEarliestRevisionDate(from policies: [Policy]) -> Policy? {
+        policies.min { ($0.revisionDate ?? .distantFuture) < ($1.revisionDate ?? .distantFuture) }
+    }
```
fedemkr, line 325: ❓ I know this comes from the organization logic, but I'm a bit hesitant as this is getting the policy that has been updated least recently, i.e. the oldest one. Maybe it's correct as is, but I would think the most recently updated policy should prevail instead as it would have the latest changes.
Would this be to display the cards ordered from earliest to latest? Would the shown card policy be removed in a future PR?

BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessor.swift
```diff
@@ -328,6 +332,13 @@ extension VaultListProcessor {
         await services.flightRecorder.setFlightRecorderBannerDismissed()
     }
 
+    /// Dismisses the organization user notification banner.
+    ///
+    private func dismissOrganizationBanner() async {
+        // TODO: PM-33861 Persist banner dismissal data
```
fedemkr, line 340: ⛏️ This is referencing the same ticket as this PR. Will this be added as part of this PR?

BitwardenShared/Core/Vault/Services/PolicyService.swift
matt-livefront: In practice, there can only be a single notification banner. This policy requires the "Single organization" policy, so you can't be a member of multiple organizations and you can only specify a single banner policy. Also, once the dismissal logic is implemented, which is based on on revision date, if there were multiple, dismissing an older one would then cause the newer one to be shown.

BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessor.swift
matt-livefront: I'll have a follow-up PR for this work, as there was already a lot here. The dismissal logic was easy to break off from this work.

fedemkr: Roger, I'll approve, although the TODO should reference the ticket in which the work will be done, but I guess it doesn't matter much here as it will be worked on soon.

fedemkr (approved):

PR #2806, opened 2026-06-18
[PM-39234] feat: Add manifest endpoint and models

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39234

## 📔 Objective

Adds `FillAssistManifestRequest`, `FillAssistManifestResponseModel`, and a `getManifest()` method to `FillAssistAPIService` to fetch `manifest.json` from the map-the-web CDN. The manifest describes the available versioned map artifacts and their filenames, enabling downstream consumers to dynamically select the correct forms file (e.g. `forms.v1.json`) rather than hardcoding it.

**Depends on #2760** (PM-38442 — Add Fill-Assist endpoint and models), which establishes `mapTheWebService` and the base URL plumbing this PR builds upon.


matt-livefront (dismissed):

BitwardenShared/Core/Autofill/Models/FillAssistManifestResponseModel.swift
```diff
@@ -0,0 +1,38 @@
...
+// MARK: - FillAssistManifestEntryModel
+
+/// Describes a single versioned artifact entry within the Fill-Assist manifest.
+///
+struct FillAssistManifestEntryModel: Codable, Equatable {
+    /// The content identifier for the artifact in `sha256:<hex>` format.
+    let cid: String
+
+    /// Whether this major version has entered its end-of-life support window.
+    let deprecated: Bool?
```
fedemkr: 🤔 Shouldn't this non-optional and have `@DefaultFalse`? Not sure I understand what would be the usage of this value as `nil`.

fedemkr (commented):
Looks good, just one minor ⛏️

BitwardenShared/Core/Autofill/Models/FillAssistManifestResponseModel.swift
andrebispo5: yes that makes sense. when it's deprecated is comes as `"deprecated":true`

fedemkr (approved):

PR #2810, opened 2026-06-19
[PM-38443] feat: Add fill-assist domain models, CSS parser, and dynamic filename API

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38443

## 📔 Objective

Updates `getFormsMap` to accept a dynamic `filename` parameter driven by the manifest. Adds `FillAssistCachedData`, `FillAssistHostRules`, and `FillAssistFieldAttributes` domain models for persisting parsed fill-assist rules, and `FillAssistSelectorParser` which extracts HTML element attributes from CSS selectors (excluding shadow DOM and class-only selectors).

Switches `FillAssistAPIService` from `send()` to `download(filename:)` — GitHub Releases uses multi-hop 302 redirects and `CertificateHTTPClient` deliberately blocks 302s for SSO, so a dedicated `download(filename:)` method on `HTTPService` using `URLSession` download tasks (which follow all redirects) is used instead. Also fixes `FormsMapHostEntry` to silently exclude null-valued pathname entries, which the live forms file contains.

This is PR 1/3 for PM-38443. PR 2/3 adds storage, PR 3/3 adds the repository and sync integration.


morganzellers-bw (dismissed):

BitwardenShared/Core/Autofill/Models/FormsMapResponseModel.swift
```diff
@@ -36,11 +36,29 @@ struct FormsMapResponseModel: Equatable, JSONResponse {
...
     let pathnames: [String: FormsMapPathnameEntry]?
+
+    // MARK: Codable
+
+    init(from decoder: any Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        forms = try container.decodeIfPresent([FormsMapContent].self, forKey: .forms)
+        pathnames = try container.decodeIfPresent([String: FormsMapPathnameEntry?].self, forKey: .pathnames)?
+            .compactMapValues { $0 }
+    }
```
fedemkr: 🤔 The custom init is only to perform `.compactMapValues` or do you need it for anything else? I'm asking because normally specifying the Coding keys should be enough.
And if you only need it to compact values then we could add something like:
```swift
@propertyWrapper
struct CompactDecodable<T: Decodable>: Decodable {
    var wrappedValue: [String: T]?

    init(wrappedValue: [String: T]? = nil) {
        self.wrappedValue = wrappedValue
    }

    init(from decoder: Decoder) throws {
        let container = try decoder.singleValueContainer()
        if container.decodeNil() {
            wrappedValue = nil
        } else {
            let nullable = try container.decode([String: T?].self)
            wrappedValue = nullable.compactMapValues { $0 }
        }
    }
}

// usage

@CompactDecodable let pathnames: [String: FormsMapPathnameEntry]?
```

Having that we can reuse it where we need to without adding the init(from decoder:). Also I'd add something alike for arrays:

```swift
@propertyWrapper
struct CompactDecodable<T: Decodable>: Decodable {
    var wrappedValue: [T]

    init(from decoder: Decoder) throws {
        let container = try decoder.singleValueContainer()
        let nullable = try container.decode([T?].self)
        wrappedValue = nullable.compactMap { $0 }
    }
}
```

What do you think? cc: @bitwarden/team-ios

BitwardenShared/Core/Autofill/Services/API/FillAssistAPIServiceTests.swift
fedemkr, line 1: ⚠️ I see this got deleted, but where are the tests now?

BitwardenShared/Core/Autofill/Services/API/FillAssistAPIService.swift
```diff
@@ -21,11 +22,15 @@ protocol FillAssistAPIService { // sourcery: AutoMockable
...
-        try await fillAssistService.send(FormsMapRequest())
+    func getFormsMap(filename: String) async throws -> FormsMapResponseModel {
+        let fileURL = try await fillAssistService.download(filename: filename)
+        let data = try Data(contentsOf: fileURL)
+        return try JSONDecoder.pascalOrSnakeCaseDecoder.decode(FormsMapResponseModel.self, from: data)
     }
 
     func getManifest() async throws -> FillAssistManifestResponseModel {
-        try await fillAssistService.send(FillAssistManifestRequest())
+        let fileURL = try await fillAssistService.download(filename: "manifest.json")
```
fedemkr: ⛏️ Move to Constants file.

BitwardenShared/Core/Autofill/Utilities/FillAssistSelectorParser.swift
```diff
@@ -0,0 +1,90 @@
...
+        let fullRange = NSRange(location: 0, length: nsSelector.length)
+        attributeRegex?.enumerateMatches(in: selector, range: fullRange) { match, _, _ in
+            guard let match,
+                  let attrRange = Range(match.range(at: 1), in: selector),
+                  let valRange = Range(match.range(at: 2), in: selector)
+            else { return }
+            let attr = String(selector[attrRange])
+            let val = String(selector[valRange])
+            switch attr {
+            case "id": if id == nil { id = val }
```
fedemkr: ⛏️ This can be simplified as:

```suggestion
            case "id": id = id ?? val
```

BitwardenShared/Core/Autofill/Utilities/FillAssistSelectorParser.swift
```diff
@@ -0,0 +1,90 @@
...
+    // MARK: Private
+
+    /// Extracts the leading tag name from a selector (e.g. `"input"` from `"input#user"`).
+    ///
+    private static func extractTagName(from selector: String) -> String? {
+        let stopChars = CharacterSet(charactersIn: "#[.:")
+        let trimmed = selector.trimmingCharacters(in: .whitespaces)
+        guard let stopIndex = trimmed.unicodeScalars.firstIndex(where: { stopChars.contains($0) }) else {
+            // No stop character — entire string is the tag (if non-empty and not a class selector).
+            return trimmed.isEmpty || trimmed.first == "." ? nil : trimmed
```
fedemkr, line 85: ⛏️ I may be wrong but I think there's no test for this case

BitwardenShared/Core/Platform/Extensions/CompactDecodable.swift
fedemkr, line 1: 🎨 I think we should have tests for this with an example object and some json decoding to see that this behaves as it's supposed to as we do with `@DefaultFalse` property wrapper.

fedemkr (commented):
Looks good, just some tests missing.

fedemkr (dismissed):

fedemkr (approved):

PR #2812, opened 2026-06-19
[PM-38443] feat: Add FillAssistRepository and sync integration

andrebispo5:
## 🗒️ Note

To avoid further conflicts, this PR contains the following stacked approved PRs:
https://github.com/bitwarden/ios/pull/2876
https://github.com/bitwarden/ios/pull/2882

## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38443

## 📔 Objective

Adds `FillAssistRepository` (protocol + `DefaultFillAssistRepository`) which orchestrates the full fill-assist sync pipeline: feature flag check, 6-hour interval guard, manifest fetch, cid/sourceUrl cache comparison, forms download, schema version validation, CSS selector parsing, and per-account storage. Wires the repository into `SyncService` (non-blocking) and `ServiceContainer`.

Also fixes `APIService.fillAssistService` to use a plain `HTTPService(client: URLSession.shared)` — the standard `CertificateHTTPClient` deliberately blocks 302 redirects for SSO, which broke the GitHub Releases CDN redirect chain.


BitwardenShared/Core/Autofill/Repositories/FillAssistRepository.swift
```diff
@@ -0,0 +1,210 @@
...
+
+        // 6. Download forms file.
+        let formsMap = try await fillAssistAPIService.getFormsMap(filename: entry.filename)
+
+        // 7. Validate schema major version; update timestamp and skip if unsupported.
+        let schemaMajor = formsMap.schemaVersion.split(separator: ".").first.map(String.init) ?? ""
+        guard schemaMajor == Constants.fillAssistExpectedSchemaMajor else {
+            appSettingsStore.setFillAssistLastFetchTimestamp(Date(), userId: userId)
+            return
+        }
```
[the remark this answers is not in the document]
andrebispo5: start version is 1.0.0

BitwardenShared/Core/Autofill/Repositories/FillAssistRepository.swift
```diff
@@ -0,0 +1,179 @@
...
+    /// Returns the cached fill-assist rules for a given hostname, or `nil` if unavailable.
+    ///
+    /// - Parameter hostname: The hostname to look up.
+    /// - Returns: The cached `FillAssistHostRules`, or `nil`.
+    ///
+    func fillAssistRules(for hostname: String) async -> FillAssistHostRules?
+
+    /// Clears all cached fill-assist data for the active account.
+    ///
+    func clearFillAssistRules() async throws
```
fedemkr: ⛏️ Given that this is the "Fill Assist" repository, I think we can remove the use of "FillAssist" in the function names as it's a bit repetitive IMO.

BitwardenShared/Core/Autofill/Repositories/FillAssistRepository.swift
```diff
@@ -0,0 +1,179 @@
...
+        } catch {
+            // Sync failures are non-fatal — existing cache remains available.
+            errorReporter.log(error: error)
+        }
+    }
+
+    func fillAssistRules(for hostname: String) async -> FillAssistHostRules? {
+        guard let userId = try? await stateService.getActiveAccountId() else { return nil }
+        return appSettingsStore.fillAssistCachedData(userId: userId)?.rules[hostname]
+    }
```
fedemkr: 🤔 Do we always need to resolve for _exact_ hostname or should we apply any heuristics here?

BitwardenShared/Core/Autofill/Repositories/FillAssistRepository.swift
```diff
@@ -0,0 +1,179 @@
...
+    ///
+    private func performSync() async throws {
+        // 1. Feature flag guard.
+        guard await configService.getFeatureFlag(.fillAssistTargetingRules) else { return }
+
+        // 2. Time-interval guard — skip if last fetch was less than fillAssistUpdateInterval ago.
+        let sourceUrl = environmentService.fillAssistRulesURL
+        let userId = try await stateService.getActiveAccountId()
+        let lastFetch = appSettingsStore.fillAssistLastFetchTimestamp(userId: userId)
+        if let lastFetch, Date().timeIntervalSince(lastFetch) < Constants.fillAssistUpdateInterval {
```
fedemkr: 🎨 Use `TimeProvider` instead of `Date()` so you can better mock the date on the tests.
This applies to all `Date()` in this file.

BitwardenShared/Core/Autofill/Repositories/FillAssistRepository.swift
```diff
@@ -0,0 +1,179 @@
...
+        let userId = try await stateService.getActiveAccountId()
+        appSettingsStore.setFillAssistCachedData(nil, userId: userId)
+    }
+
+    // MARK: Private
+
+    /// Runs the full sync pipeline if sync conditions are met
+    ///
+    private func performSync() async throws {
+        // 1. Feature flag guard.
```
fedemkr: ⛏️ I don't think all of these comments are necessary. This and some others are self-explanatory by the code.

BitwardenShared/Core/Autofill/Repositories/FillAssistRepository.swift
```diff
@@ -0,0 +1,179 @@
...
+            appSettingsStore.setFillAssistLastFetchTimestamp(Date(), userId: userId)
+            return
+        }
+
+        // 6. Download forms file.
+        let formsMap = try await fillAssistAPIService.getFormsMap(filename: entry.filename)
+
+        // 7. Validate schema major version; update timestamp and skip if unsupported.
+        let schemaMajor = formsMap.schemaVersion.split(separator: ".").first.map(String.init) ?? ""
+        guard schemaMajor == Constants.fillAssistExpectedSchemaMajor else {
```
fedemkr: 🤔 Could we reuse `ServerVersion` here or is the format different?

BitwardenShared/Core/Autofill/Repositories/FillAssistRepository.swift
```diff
@@ -0,0 +1,179 @@
...
+
+    /// Builds a `[hostname: FillAssistHostRules]` dictionary by pooling all non-null field
+    /// definitions across each host's top-level forms and every pathname entry.
+    ///
+    private func buildRules(from formsMap: FormsMapResponseModel) -> [String: FillAssistHostRules] {
+        var result = [String: FillAssistHostRules]()
+        for (hostname, hostEntry) in formsMap.hosts {
+            var pooled = [String: [FillAssistFieldAttributes]]()
+            let allForms = (hostEntry.forms ?? [])
+                + (hostEntry.pathnames?.values.compactMap(\.self).flatMap(\.forms) ?? [])
```
fedemkr: ⛏️ I'd move this to a computed get-only property in `FormsMapHostEntry` so the logic here is simpler.
You can use `.flatMap { $0?.forms ?? [] } ?? [])` instead of `.compactMap(\.self).flatMap(\.forms) ?? [])`, but I'll leave that up to you on which one is better to read.

BitwardenShared/Core/Autofill/Repositories/FillAssistRepository.swift
```diff
@@ -0,0 +1,179 @@
...
+    /// Converts a `FormsMapSelector` into zero or more `FillAssistFieldAttributes`.
+    ///
+    private func parseSelector(_ selector: FormsMapSelector) -> [FillAssistFieldAttributes] {
+        switch selector {
+        case let .single(css):
+            FillAssistSelectorParser.parse(css).map { [$0] } ?? []
+        case let .sequence(parts):
+            parts.compactMap { FillAssistSelectorParser.parse($0) }
+        }
+    }
```
fedemkr: 🎨 I think this function can be directly in `FormsMapSelector`.
🤔 Additionally, I think we don't have tests for both `.single` and `.sequence` kind of selectors or to test the part of `?? []`.

BitwardenShared/Core/Autofill/Repositories/FillAssistRepositoryTests.swift
```diff
@@ -0,0 +1,284 @@
...
+import BitwardenKitMocks
+import XCTest
+
+@testable import BitwardenShared
+@testable import BitwardenSharedMocks
+
+// MARK: - FillAssistRepositoryTests
+
+@MainActor
+class FillAssistRepositoryTests: BitwardenTestCase {
```
fedemkr: 🎨 Can we use Swift Testing here instead?

BitwardenShared/Core/Platform/Services/API/APIService.swift
```diff
@@ -113,8 +113,12 @@ class APIService {
...
             baseURLGetter: { environmentService.identityURL },
         )
-        fillAssistService = httpServiceBuilder.makeService(
+        // Fill-Assist requests target an external CDN (GitHub Releases) that uses multi-hop 302
+        // redirects. CertificateHTTPClient deliberately blocks 302s for SSO, so we construct
+        // a plain HTTPService with URLSession.shared which follows all redirects.
+        fillAssistService = HTTPService(
             baseURLGetter: { environmentService.fillAssistRulesURL },
+            client: URLSession.shared,
         )
```
fedemkr, line 123: 🤔 Wouldn't that case go through [here](https://github.com/bitwarden/ios/blob/main/BitwardenShared/Core/Platform/Services/CertificateHTTPClient.swift#L61), performing the default handling of the request?

BitwardenShared/Core/Vault/Services/SyncService.swift
```diff
@@ -474,6 +479,9 @@ extension DefaultSyncService {
             )
         }
 
+        // Sync fill-assist rules alongside vault data. Failures must not block vault sync.
```
fedemkr: ⛏️ I'd remove "Failures must not block vault sync." as the function doesn't throw any errors so I think it's a direct assumption.

BitwardenShared/Core/Vault/Services/SyncService.swift
```diff
@@ -474,6 +479,9 @@ extension DefaultSyncService {
...
+        await fillAssistRepository.syncFillAssistRules()
```
fedemkr: ⚠️ Add tests for this.

BitwardenShared/Core/Autofill/Repositories/FillAssistRepositoryTests.swift
```diff
@@ -0,0 +1,284 @@
...
+                    "forms": [{ "category": "account-login", "fields": { "username": ["input#user"] } }]
+                }
+            }
+        }
+        """
+        return try JSONDecoder.pascalOrSnakeCaseDecoder.decode(
+            FormsMapResponseModel.self,
+            from: Data(json.utf8),
+        )
+    }
```
fedemkr: 🤔 Can't we directly initialize the structs instead of using JSON decoding?

BitwardenShared/Core/Autofill/Repositories/FillAssistRepository.swift
```diff
@@ -0,0 +1,179 @@
...
+        appSettingsStore.setFillAssistCachedData(data, userId: userId)
+        appSettingsStore.setFillAssistLastFetchTimestamp(Date(), userId: userId)
+    }
+
+    /// Builds a `[hostname: FillAssistHostRules]` dictionary by pooling all non-null field
+    /// definitions across each host's top-level forms and every pathname entry.
+    ///
+    private func buildRules(from formsMap: FormsMapResponseModel) -> [String: FillAssistHostRules] {
+        var result = [String: FillAssistHostRules]()
+        for (hostname, hostEntry) in formsMap.hosts {
```
fedemkr, line 239: 🎨 I see tests are only covering when there is only 1 host. But what about 0 or multiple hosts?

BitwardenShared/Core/Autofill/Repositories/FillAssistRepository.swift
```diff
@@ -0,0 +1,179 @@
...
+                + (hostEntry.pathnames?.values.compactMap(\.self).flatMap(\.forms) ?? [])
+            for form in allForms {
+                for (fieldKey, selectors) in form.fields {
+                    let attrs = selectors.flatMap { parseSelector($0) }
+                    pooled[fieldKey, default: []].append(contentsOf: attrs)
+                }
+            }
+            if !pooled.isEmpty {
+                result[hostname] = FillAssistHostRules(fields: pooled)
+            }
```
fedemkr, line 251: 🎨 I think a test is missing when `pooled` is actually empty.

BitwardenShared/Core/Autofill/Repositories/FillAssistRepository.swift
andrebispo5: it's always the host name.

BitwardenShared/Core/Autofill/Repositories/FillAssistRepository.swift
andrebispo5: yup I've already removed in a PR upstream 👍

BitwardenShared/Core/Autofill/Repositories/FillAssistRepository.swift
andrebispo5: > ServerVersion doesn't expose a .major property at all — the version component parsing is entirely private (versionComponents() is internal). The only public API is Comparable, so you'd have to write a range check like version >= ServerVersion("1.0.0") && version < ServerVersion("2.0.0") using force-unwrapped literals, which is more complex and less readable than extracting the major component directly.

KatherineInCode: 🤔 Would it make sense to actually create a `SemanticVersion` struct of some sort (and probably typealias ServerVersion 🤔 ) that also has a way of easily saying "is this within this major version" and such?

fedemkr: @andrebispo5 you could just expose a `major` variable there and use the internal `versionComponents()` to get it. However, perhaps it's a bit confusing to use `ServerVersion` on this given that it wouldn't be a server version. Let's leave it as is for now unless/until there is another place where we need to use it.

BitwardenShared/Core/Autofill/Models/FormsMapResponseModel.swift
```diff
@@ -21,6 +21,13 @@ struct FormsMapResponseModel: Equatable, JSONResponse {
     /// The schema version of the map.
     let schemaVersion: String
 
+    // MARK: Initialization
+
+    init(hosts: [String: FormsMapHostEntry], schemaVersion: String) {
```
fedemkr, line 32: ⛏️ Add docs.

BitwardenShared/Core/Autofill/Models/FormsMapResponseModel.swift
```diff
@@ -43,6 +50,18 @@ struct FormsMapHostEntry: Codable, Equatable {
...
     @CompactDecodable var pathnames: [String: FormsMapPathnameEntry]?
+
+    /// All form descriptions for this host, pooling top-level and pathname-specific entries.
+    var allForms: [FormsMapContent] {
+        (forms ?? []) + (pathnames?.values.flatMap(\.forms) ?? [])
+    }
+
+    // MARK: Initialization
+
+    init(forms: [FormsMapContent]?, pathnames: [String: FormsMapPathnameEntry]? = nil) {
```
fedemkr, line 73: ⛏️ Add docs

BitwardenShared/Core/Autofill/Repositories/FillAssistRepositoryTests.swift
```diff
@@ -0,0 +1,353 @@
...
+        return FormsMapResponseModel(
+            hosts: [
+                "example.com": FormsMapHostEntry(
+                    forms: [loginContent],
+                    pathnames: ["/login": FormsMapPathnameEntry(forms: [pathnameContent])],
+                ),
+            ],
+            schemaVersion: "1.0.0",
+        )
+    }
```
fedemkr: ⛏️ Add function docs.

BitwardenShared/Core/Autofill/Repositories/FillAssistRepository.swift
```diff
@@ -0,0 +1,176 @@
...
+
+        let sourceUrl = environmentService.fillAssistRulesURL
+        let userId = try await stateService.getActiveAccountId()
+        let cached = appSettingsStore.fillAssistCachedData(userId: userId)
+        let lastFetch = appSettingsStore.fillAssistLastFetchTimestamp(userId: userId)
+        if let lastFetch,
+           cached != nil,
+           timeProvider.presentTime.timeIntervalSince(lastFetch) < Constants.FillAssist.updateInterval {
+            return
+        }
```
fedemkr, line 141: 🤔 I believe we should consider checking `forceSync` here to bypass this check. Otherwise users will need to wait up to 6 hours when a change is done in map-the-web.
I think it can be done in a different PR so not to keep adding stuff to this one.

fedemkr (dismissed):
Looks good, approving as `forceSync` should be handled in a different PR.

fedemkr (dismissed):

BitwardenShared/Core/Platform/Services/StateService.swift
```diff
@@ -1912,8 +1948,6 @@ actor DefaultStateService: StateService, ActiveAccountStateProvider, ConfigState
         appSettingsStore.setDefaultUriMatchType(nil, userId: knownUserId)
         appSettingsStore.setDisableAutoTotpCopy(nil, userId: knownUserId)
         appSettingsStore.setEncryptedUserKey(key: nil, userId: knownUserId)
-        appSettingsStore.setFillAssistCachedData(nil, userId: knownUserId)
-        appSettingsStore.setFillAssistLastFetchTimestamp(nil, userId: knownUserId)
         appSettingsStore.setHasPerformedSyncAfterLogin(nil, userId: knownUserId)
```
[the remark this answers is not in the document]
andrebispo5: I'm calling Clear rules, which clears that cache, here: https://github.com/bitwarden/ios/pull/2812/changes/BASE..49ad11ab3c1f39a8f024ef68796e26136df376be#diff-cb8846480d62aaf308bd0e3a3cc1f018f98ec0031d052faf961fe8e90982fc1bL868

BitwardenShared/Core/Platform/Services/StateService.swift
```diff
@@ -1912,8 +1948,6 @@ actor DefaultStateService: StateService, ActiveAccountStateProvider, ConfigState
...
         appSettingsStore.setLastSyncTime(nil, userId: knownUserId)
```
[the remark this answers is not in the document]
andrebispo5: I'm calling Clear rules, which clears that cache, here: https://github.com/bitwarden/ios/pull/2812/changes/BASE..49ad11ab3c1f39a8f024ef68796e26136df376be#diff-cb8846480d62aaf308bd0e3a3cc1f018f98ec0031d052faf961fe8e90982fc1bL868

fedemkr (approved):

PR #2814, opened 2026-06-19
[PM-28349] feat: Support https:// redirects for SSO and Duo auth connectors

matt-livefront:
## 🎟️ Tracking

[PM-28349](https://bitwarden.atlassian.net/browse/PM-28349)

## 📔 Objective

Our mobile authentication connectors currently complete the `ASWebAuthenticationSession` flow at the end of the SSO and Duo flows using a custom `bitwarden://` URL scheme. This adds support for switching this to a HTTPS callback URL for **Cloud** users (US, EU, and Internal regions)

What changed:
- **Region-gated callback selection (no feature flag).** The callback descriptor is now resolved centrally in `AuthService` from `EnvironmentService.region`. Cloud regions use an `https://{apexHost}/{sso,duo}-callback` callback at the region's fixed apex host (US→`bitwarden.com`, EU→`bitwarden.eu`, Internal→`bitwarden.pw`); self-hosted continues to use `bitwarden://`.
- **New `RegionType.internal`** for Bitwarden's `bitwarden.pw` (QA/Internal) environments. It is hidden from the region picker and detected automatically by `bitwarden.pw` host suffix (`EnvironmentService.region` / `EnvironmentURLData.region`). The apex host comes from `RegionType.authCallbackHost`.
- **`ASWebAuthenticationSession.Callback.https(host:path:)`** is used to intercept the `https://` redirect for SSO and Duo. This API is iOS 17.4+, and the deployment target is iOS 15.0, so it is `@available`-gated: Cloud users on iOS < 17.4 transparently fall back to the existing `bitwarden://` callback.
- SSO's OAuth `redirect_uri` follows the same region/availability resolution so the value sent to the identity provider always matches the callback the app intercepts.
- Session construction for the SSO, Duo, and WebAuthn-self-hosted flows is consolidated behind a single `AuthService` factory (the coordinator retains presentation/start). The now-unused `callbackUrlScheme` parameter is removed from `AuthRoute.singleSignOn`.

Scope notes:
- **WebAuthn is intentionally excluded.** On iOS, Cloud WebAuthn 2FA already uses the native passkey API (`ASAuthorizationController`), not a connector callback URL; only the self-hosted connector uses `webauthn-callback`, and self-hosted remains on `bitwarden://`.

[PM-28349]: https://bitwarden.atlassian.net/browse/PM-28349?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/Core/Auth/Services/AuthService.swift
+++ b/BitwardenShared/Core/Auth/Services/AuthService.swift
@@ -55,6 +55,53 @@ enum AuthError: Error {
     case unableToResendNewDeviceOtp
 }
 
+// MARK: - AuthWebSessionCallback
+
+/// How an `ASWebAuthenticationSession` should match the redirect that ends an auth connector flow.
+///
+enum AuthWebSessionCallback: Equatable {
+    /// Matches a custom URL scheme redirect (e.g. `bitwarden://sso-callback`).
+    case customScheme(String)
+
+    /// Matches an HTTPS redirect (e.g. `https://vault.bitwarden.com/sso-callback`). Requires iOS 17.4+.
+    case https(host: String, path: String)
+}
+
+// MARK: - AuthWebSessionCallbackKind
+
+/// The auth connector flow that an `ASWebAuthenticationSession` is being created for. Determines the
+/// callback path and whether the HTTPS callback is eligible for Cloud users.
+///
+enum AuthWebSessionCallbackKind {
+    /// The Duo two-factor connector.
+    case duo
+
+    /// The single sign-on (SSO) connector.
+    case singleSignOn
+
+    /// The self-hosted WebAuthn connector.
+    case webAuthnSelfHosted
+
+    /// The callback path component, without a leading slash (e.g. `sso-callback`).
+    var callbackPath: String {
+        switch self {
+        case .duo: "duo-callback"
+        case .singleSignOn: "sso-callback"
+        case .webAuthnSelfHosted: "webauthn-callback"
+        }
+    }
+
+    /// Whether this connector supports the HTTPS callback for Cloud users. WebAuthn is excluded
+    /// because Cloud WebAuthn uses the native passkey API rather than a connector
+    /// callback URL; the self-hosted WebAuthn connector continues to use the custom scheme.
+    var supportsHttpsCallback: Bool {
+        switch self {
+        case .duo, .singleSignOn: true
+        case .webAuthnSelfHosted: false
+        }
+    }
+}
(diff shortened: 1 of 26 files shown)


BitwardenShared/Core/Auth/Services/AuthService.swift
```diff
@@ -780,17 +853,53 @@ class DefaultAuthService: AuthService { // swiftlint:disable:this type_body_leng
...
+    /// - Parameter kind: The connector flow the redirect is for.
+    /// - Returns: e.g. `https://vault.bitwarden.com/sso-callback` for Cloud/iOS 17.4+, otherwise
+    ///   `bitwarden://sso-callback`.
+    ///
+    private func callbackUrl(for kind: AuthWebSessionCallbackKind) -> String {
+        switch resolveCallback(for: kind) {
+        case let .customScheme(scheme):
+            "\(scheme)://\(kind.callbackPath)"
+        case let .https(host, path):
+            "https://\(host)\(path)"
```
fedemkr, line 899: 🤔 Would it be safer to use `URLComponents` here or is it not worth it to have all the sanitization and safe guards it provides?
```
case let .https(host, path):
    var components = URLComponents()
    components.scheme = "https"
    components.host = host
    components.path = path
    return components.url?.absoluteString ?? ""
```

fedemkr (commented):
Looks good, really like the `RegionType` simplification for base URLs.
Just a small question and it seems that Codecov has caught 24 lines not being tested in `AuthService` although it's the `switch resolveCallback(for: callbackKind)` block which I'm not sure if we can test it correctly.

BitwardenShared/Core/Auth/Services/AuthService.swift
matt-livefront: I'm not sure that it's worth it here. Both host and path come from hardcoded string literals in the app. And the URL that it creates isn't used directly, it's passed as a query parameter value for `redirect_uri`. So if any percent encoding was required, it would be handled there.

matt-livefront:
> it seems that Codecov has caught 24 lines not being tested in `AuthService` although it's the `switch resolveCallback(for: callbackKind)` block which I'm not sure if we can test it correctly.

Correct, there's no way to get at the URL or parameters used to create the `ASWebAuthenticationSession` unfortunately.

fedemkr (approved):

PR #2816, opened 2026-06-19
[PM-33861] feat: Add dismissal tracking for organization banners

matt-livefront:
## 🎟️ Tracking

[PM-33861](https://bitwarden.atlassian.net/browse/PM-33861)

## 📔 Objective

Adds dismissal tracking for the organization user notification banner so a dismissed banner stays dismissed instead of reappearing on every vault-list load. Builds on #2798, which introduced the banner display.

A dismissal is persisted per user and keyed to the policy's `revisionDate`. The banner reappears only when:
- the organization publishes an updated banner (a different `revisionDate`), or
- the user logs back in, subject to the banner's `showAfterEveryLogin` setting and the logout type:
  - hard (user-initiated) logout always clears the record, so the banner reappears, and
  - soft logout (e.g. vault-timeout) clears it only when `showAfterEveryLogin` is `true`.

This PR also completes [PM-39144](https://bitwarden.atlassian.net/browse/PM-39144) by adding the SDK `PolicyType` mapping for `organizationUserNotification` (removing the placeholder `nil` mapping).

[PM-33861]: https://bitwarden.atlassian.net/browse/PM-33861?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PM-39144]: https://bitwarden.atlassian.net/browse/PM-39144?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenShared/Core/Platform/Services/StateService.swift
```diff
@@ -291,6 +291,15 @@ protocol StateService: AnyObject, BillingStateService {
     ///
     func getNotificationsLastRegistrationDate(userId: String?) async throws -> Date?
 
+    /// Gets the organization user notification banner dismissal record for a user ID.
+    ///
+    /// - Parameter userId: The user ID of the account. Defaults to the active account if `nil`.
+    /// - Returns: The organization user notification banner dismissal record, or `nil` if not dismissed.
+    ///
+    func getOrganizationUserNotificationBannerDismissal(
```
KatherineInCode, line 299: 🤔 I find myself wondering if there's a way to carve this out to a more-specialized `*StateService` like we've been doing, but I'm not sure what the right grouping would be, particularly.

BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessor.swift
```diff
@@ -334,11 +334,30 @@ extension VaultListProcessor {
...
     ///
-    private func dismissOrganizationBanner() async {
-        // TODO: PM-33861 Persist banner dismissal data
+    /// - Parameter fromActionButton: Whether the dismissal came from tapping the banner's action
+    ///   button rather than the dismiss button.
+    ///
+    private func dismissOrganizationBanner(fromActionButton: Bool) async {
+        guard let data = state.organizationUserNotificationBannerData else { return }
         state.organizationUserNotificationBannerData = nil
+        if fromActionButton {
```
KatherineInCode, line 351: 🤔 It's a little strange to me that we collect an event if the user dismissed the banner by action button, but not if they dismissed it by dismiss button. (Though if I'm reading the ticket correctly, if there *is* an action button, there shouldn't also be a dismiss button? 🤔 )

matt-livefront: That is the requirement as I understand it. I believe some organizations want an explicit "I acknowledge" button which logs an event so this can be used to comply with security requirements if needed. But this only applies specifically to the action button so dismissal via the close button doesn't have the same need.

BitwardenShared/Core/Platform/Services/StateService.swift
matt-livefront: Good question, I don't really see anything else that we could extract with this into a more-specialized service. We could potentially do something like `OrganizationNotificationStateService`, but I'm not sure if that's worth creating if we only have this get/set method within it 🤔.

KatherineInCode: Yeah, it's the eternal problem with segregating `StateService` interfaces out, figuring out what the boundaries should be and when we should segregate. I don't have a good answer 🤔

matt-livefront: In the absence of a clear grouping, I think I'd like to keep it as-is for now and reevaluate as we carve out more specialized services.

KatherineInCode (approved):

PR #2823, opened 2026-06-22
[PM-37992] fix: Show "subscription needs attention" card for past-due users

KatherineInCode:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37992

## 📔 Objective

Past-due and update-payment users were seeing no action card on the vault list. The "Upgrade to Premium" card is correctly hidden for these users (the server keeps `premium = true` during the grace period), but a "Your subscription needs attention" card with a "View plan" button was missing entirely.

**Card behavior**
- Shown for personal Premium subscriptions in `pastDue` or `updatePayment` status
- Tapping "View plan" switches to the Settings tab and pushes the plan screen (matching the existing flight recorder settings pattern — no new modal)
- Dismissible: no — it reflects an actionable subscription state the user should address

**Subscription visibility caching**
Rather than making a live `getSubscription()` API call on every vault appearance, the card visibility is now persisted in `AppSettingsStore` per user — following the exact same pattern as `shouldShowUpgradedToPremiumActionCard()`.

The cache is refreshed at three points:
1. `AppProcessor.onFetchSyncSucceeded()` — every vault sync (awaited inline so the cache is populated before `VaultListProcessor.appeared()` reads it)
2. `PremiumPlanProcessor.loadPremiumPlan()` — when the user views the plan screen, passing the already-fetched subscription to avoid a redundant API call
3. `BillingService.premiumStatusChanged()` — on push notifications, before the existing guard that excludes premium-intact users

**BillingStateService cleanup**
As part of this work, billing card state was consolidated into a dedicated `BillingStateService` protocol:
- Card visibility get/set methods (`getSubscriptionAttentionCardVisible`, `setSubscriptionAttentionCardVisible`, `getUpgradedToPremiumActionCardVisible`, `setUpgradedToPremiumActionCardVisible`) moved from `StateService` to `BillingStateService`
- `DefaultBillingService` now takes two narrowly-typed state dependencies: `billingStateService: BillingStateService` for card visibility operations, and `stateService: StateService` for premium-status queries (`doesActiveAccountHavePremium`, `doesActiveAccountHavePremiumPersonally`) which remain on `StateService` where the rest of the app uses them
- The `DefaultStateService: BillingStateService` conformance lives in `StateService.swift` alongside the other protocol conformances, giving it full access to private members like `dataStore` and `getActiveAccountUserId()`

## 📸 Screenshots

Subscription needs attention card on vault list

Plan screen shown after tapping View plan


BitwardenShared/Core/Platform/Services/StateServiceTests.swift
```diff
@@ -1948,21 +1948,21 @@ class StateServiceTests: BitwardenTestCase { // swiftlint:disable:this type_body
...
-    /// `getUpgradedToPremiumActionCardVisible()` returns `false` and logs an error when there is no active account.
-    func test_getUpgradedToPremiumActionCardVisible_noActiveAccount() async {
-        let isVisible = await subject.getUpgradedToPremiumActionCardVisible()
-        XCTAssertFalse(isVisible)
-        XCTAssertEqual(errorReporter.errors as? [StateServiceError], [.noActiveAccount])
+    /// `getUpgradedToPremiumActionCardVisible()` throws when there is no active account.
+    func test_getUpgradedToPremiumActionCardVisible_noActiveAccount() async throws {
+        await assertAsyncThrows(error: StateServiceError.noActiveAccount) {
+            _ = try await self.subject.getUpgradedToPremiumActionCardVisible()
+        }
```
matt-livefront: ⛏️ Should these tests be moved into `StateService+BillingStateServiceTests.swift`?

KatherineInCode: Good call

matt-livefront (approved):

PR #2830, opened 2026-06-25
[PM-38444] feat: Use FillAssist cached rules to guide DOM field targeting in Action Extension

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-38444

## 📔 Objective

When the Action Extension autofills credentials, use FillAssist cached rules to prepend CSS-selector-derived field targets (`id`/`name` attributes) for username and password in the fill script. This improves targeting accuracy on sites where standard DOM heuristics fail to identify the correct fields.

Also fixes the host lookup to use the registered domain (eTLD+1) rather than the raw `URL.host`, so rules stored under `datacamp.com` are found correctly when the page URL contains a subdomain like `www.datacamp.com`.

### DIFF
--- a/BitwardenShared/UI/Vault/Helpers/AutofillHelper.swift
+++ b/BitwardenShared/UI/Vault/Helpers/AutofillHelper.swift
@@ -110,19 +110,16 @@ class AutofillHelper {
         guard await services.configService.getFeatureFlag(.fillAssistTargetingRules) else { return [] }
         guard let uri,
               let url = URL(string: uri),
-              let host = url.host else { return [] }
-        let lookupHost = DomainName.parseBaseDomain(url: url) ?? host
+              let lookupHost = url.domain else { return [] }
         guard let rules = await services.fillAssistRepository.fillAssistRules(for: lookupHost) else { return [] }
 
         var result: [(String, String)] = []
 
-        if let attrs = rules.fields["username"]?.first,
-           let selector = attrs.id ?? attrs.name {
+        if let selector = rules.fields["username"]?.lazy.compactMap({ $0.id ?? $0.name }).first {
             result.append((selector, username))
         }
 
-        if let attrs = rules.fields["password"]?.first,
-           let selector = attrs.id ?? attrs.name {
+        if let selector = rules.fields["password"]?.lazy.compactMap({ $0.id ?? $0.name }).first {
             result.append((selector, password))
         }
 
--- a/BitwardenShared/UI/Vault/Helpers/AutofillHelper.swift
+++ b/BitwardenShared/UI/Vault/Helpers/AutofillHelper.swift
@@ -111,7 +111,7 @@ class AutofillHelper {
         guard let uri,
               let url = URL(string: uri),
               let lookupHost = url.domain else { return [] }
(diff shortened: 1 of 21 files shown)


BitwardenShared/UI/Vault/Helpers/AutofillHelper.swift
```diff
@@ -94,6 +97,35 @@ class AutofillHelper {
...
+
+        var result: [(String, String)] = []
+
+        if let selector = rules.fields["username"]?.lazy.compactMap({ $0.id ?? $0.name }).first {
+            result.append((selector, username))
+        }
+
+        if let selector = rules.fields["password"]?.lazy.compactMap({ $0.id ?? $0.name }).first {
+            result.append((selector, password))
+        }
```
fedemkr: ⛏️ I think you don't need `lazy` here or why was it added?
🎨 Tests are missing for the different cases of what fields have:
- Has "username" but not "password"
- Has "password" but not "username"
- Has other fields named differently than "username" or "password"

BitwardenShared/UI/Vault/Helpers/AutofillHelper.swift
```diff
@@ -126,10 +158,20 @@ class AutofillHelper {
...
+
+            let assistFields = await fillAssistFields(
+                for: appExtensionDelegate?.uri,
+                username: username,
+                password: password,
+            )
+
+            let combinedFields: [(String, String)]? = assistFields.isEmpty && cipherFields.isEmpty
+                ? nil
+                : assistFields + cipherFields
```
fedemkr: 🎨 You can directly put here `let combinedFields = assistFields + cipherFields` and then when you need to pass it in the parameter you can use `combinedFields.nilIfEmpty`

BitwardenShared/UI/Vault/Helpers/AutofillHelperTests.swift
```diff
@@ -466,4 +477,140 @@ class AutofillHelperTests: BitwardenTestCase { // swiftlint:disable:this type_bo
...
+
+        await subject.handleCipherForAutofill(cipherListView: .fixture(id: "1")) { _ in }
+
+        // Rules should be looked up under the registered domain, not the full host.
+        XCTAssertEqual(fillAssistRepository.fillAssistRulesReceivedHostname, "example.com")
+        let fields = try XCTUnwrap(appExtensionDelegate.didCompleteAutofillRequestFields)
+        XCTAssertEqual(fields[0].0, "login-email")
+        XCTAssertEqual(fields[0].1, "user@bitwarden.com")
+        XCTAssertEqual(fields[1].0, "login-pwd")
+        XCTAssertEqual(fields[1].1, "PASSWORD")
```
fedemkr: 🎨 I would change the mock `didCompleteAutofillRequestFields` to have names on the tuple.
`var didCompleteAutofillRequestFields: [(selector: String, value: String)]?`
So then you don't need to use `.0` or `.1` and you can directly say `.selector` or `.value`.

fedemkr (approved):

andrebispo5:
Merging to cascading PR, test that failed is from a timeout.

PR #2845, opened 2026-06-30
[PM-39358] fix: Reject partial card scans via Luhn and brand-length validation

fedemkr:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39358

## 📔 Objective

Partial OCR card scans could produce digit sequences that matched the regex in `CardTextParser` and were auto-filled into the card-number field despite being incomplete (e.g. the card was not fully in frame). This PR adds two validation layers inside `CardTextParser.extractCardNumber(from:)`:

1. **Luhn mod-10 checksum** — rejects numbers that fail the standard Luhn algorithm, which rules out nearly all truncated card numbers.
2. **Brand-specific digit-length enforcement** — after inferring the brand from leading digits, confirms the digit count matches the expected PAN length for that network (e.g. Amex = 15, Visa = 13/16/19, Diners = 14).

`CardComponent.Brand` is extended with a `validDigitLengths: Set<Int>` property that drives the length check.

No UI changes; scanner behaviour is unchanged from the user's perspective — the form simply never populates from an invalid number.


BitwardenShared/Core/Vault/Services/CardTextParser.swift
```diff
@@ -167,4 +168,23 @@ final class DefaultCardTextParser: CardTextParser {
...
+            }
+            sum += digit
+        }
+        return sum.isMultiple(of: 10)
+    }
+
+    /// Returns `true` if `digits` has a length valid for the brand detected from its leading digits.
+    private func hasValidLength(_ digits: String) -> Bool {
+        CardComponent.Brand.detect(from: digits).validDigitLengths.contains(digits.count)
+    }
```
matt-livefront: ⛏️ Can you alphabetize these?

matt-livefront (dismissed):

matt-livefront (approved):

PR #2847, opened 2026-06-30
[PM-39361] fix: Restart card scanner after camera interruption or reopen

fedemkr:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39361

## 📔 Objective

Fixes a blank camera screen when opening the card scanner after the Add Card view was closed while the scanner was still open, or after the device locks and unlocks while the app is on the Add Card screen.

**Changes in `CardScannerView.swift`:**
- Replace `try? startScanning()` with explicit error handling; on failure a new
  `onScannerUnavailable` callback is fired (dispatched after the current SwiftUI
  update pass to avoid re-entrant state mutations).
- Implement `dataScanner(_:becameUnavailableWithError:)` in the `Coordinator` to
  forward runtime camera-interruption events through the same callback.
- Add `@Environment(\.scenePhase)` observation in `CardScannerWrapperView`; when
  the scene becomes `.active` while scanning is supposed to be running, a
  stop-then-restart cycle is triggered.
- Retry up to 2 times with a 300 ms delay before auto-dismissing the sheet, so
  the user is never left with a permanently blank screen.


BitwardenShared/UI/Vault/VaultItem/AddEditItem/AddEditCardItem/CardScannerView.swift
```diff
@@ -114,7 +151,11 @@ struct CardScannerView: UIViewControllerRepresentable {
 
     func updateUIViewController(_ uiViewController: DataScannerViewController, context: Context) {
         if isScanning {
-            try? uiViewController.startScanning()
+            do {
+                try uiViewController.startScanning()
+            } catch {
+                DispatchQueue.main.async { context.coordinator.onScannerUnavailable?() }
```
matt-livefront, line 172: ❓ Is the dispatch here just to avoid updating the view in the middle of a view update?

BitwardenShared/UI/Vault/VaultItem/AddEditItem/AddEditCardItem/CardScannerView.swift
```diff
@@ -51,6 +62,28 @@ struct CardScannerWrapperView: View {
             }
         }
         .navigationViewStyle(.stack)
+        .onChange(of: scenePhase) { newPhase in
+            if newPhase == .active, isScanning {
+                restartScanning()
```
matt-livefront: 🤔 Does the change of `scenePhase` trigger for you? I think the scene phase might only work if you're using SwiftUI navigation, not UIKit. You might need to observe the notification center notifications instead.

BitwardenShared/UI/Vault/VaultItem/AddEditItem/AddEditCardItem/CardScannerView.swift
fedemkr: Indeed.

BitwardenShared/UI/Vault/VaultItem/AddEditItem/AddEditCardItem/CardScannerView.swift
```diff
@@ -62,9 +62,12 @@ struct CardScannerWrapperView: View {
             }
         }
         .navigationViewStyle(.stack)
-        .onChange(of: scenePhase) { newPhase in
-            if newPhase == .active, isScanning {
-                restartScanning()
+        .onReceive(NotificationCenter.default.publisher(for: UIApplication.willEnterForegroundNotification)) { _ in
```
fedemkr, line 65: @matt-livefront I wasn't sure whether this would be a good idea to put it directly here. It seems more like SwiftUI but less testable so if needed I can move it to the processor and do the logic somehow there.

matt-livefront: I think it's ok given that we would have used scene phase here and it's for a minor and self-contained use.

matt-livefront (dismissed):

matt-livefront (approved):

PR #2848, opened 2026-07-01
[PM-39812] fix: Apply SSO cookie request handlers to HTTPService download functions

fedemkr:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39812

## 📔 Objective

The `download(from:)` and `download(filename:)` methods in `HTTPService` bypassed
the request handler pipeline entirely, meaning `SSOCookieVendorRequestHandler` never
injected load-balancer cookies into download requests. On servers configured with SSO
cookie vendor bootstrap, this caused attachment downloads to throw errors.

**Fix:** Both download methods now call `applyRequestHandlers(to:)` before passing the
request to the underlying client, mirroring what `send(_:)` already did. A failable
`HTTPRequest.init?(from: URLRequest)` bridges the `URLRequest`-based download API into
the `HTTPRequest`-based handler pipeline.

> [!NOTE]  
> The response handler was not added here as that handles `HTTPResponse` instead of `URL` on the download functions. We could potentially add handling for that as well but requires more complex changes. So for now the request handler has been added which adds the cookie if it's available as we think it's an edge case that just at the moment of downloading an attachment the cookie expires.


matt-livefront (commented):
These changes look good, looks like there's one test failure though.

matt-livefront (approved):

PR #2849, opened 2026-07-01
[PM-38944] fix: Hide storage cost row when user has no additional storage

KatherineInCode:
## 🎟️ Tracking

[PM-38944](https://bitwarden.atlassian.net/browse/PM-38944) — Storage line item is displayed when a user doesn't have additional storage

Also resolves [PM-38945](https://bitwarden.atlassian.net/browse/PM-38945) — Storage line item is displayed when there is additional storage at zero cost (same root cause and fix).

## 📔 Objective

The storage cost line item in the Premium plan view was always rendered, displaying "$0.00" for users without additional storage. This matched neither web nor Android behavior, where the row is hidden when `additionalStorage` is absent from the API response.

Adds a `showStorageCost` computed property to `PremiumPlanState` following the existing `showDiscount` pattern, and gates the storage `billingRow` on it. The property returns `true` only when `storageCost > 0`. Since the server sends `additionalStorage: null` for both "no storage purchased" and "zero-cost storage" scenarios, this single check covers both PM-38944 and PM-38945.

## 📸 Screenshots

| No additional storage | With additional storage ($4.00) |
|---|---|
| Plan view without storage cost row | Plan view with storage cost row showing $4.00 |

[PM-38944]: https://bitwarden.atlassian.net/browse/PM-38944?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PM-38945]: https://bitwarden.atlassian.net/browse/PM-38945?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/UI/Billing/PremiumPlan/PremiumPlanState.swift
+++ b/BitwardenShared/UI/Billing/PremiumPlan/PremiumPlanState.swift
@@ -142,6 +142,11 @@ struct PremiumPlanState: Equatable {
         !discount.isEmpty
     }
 
+    /// Whether the storage cost row should be shown.
+    var showStorageCost: Bool {
+        (subscription?.storageCost ?? 0) > 0
+    }
+
     /// The storage cost label (e.g. "$4.00" or "$0.00").
     var storageCostLabel: String {
         guard let subscription else { return "" }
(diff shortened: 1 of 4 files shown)


andrebispo5:
<del>Just to confirm, is this the correct behaviour now? because I had a similar PR up but it turned out that the line should always be displayed. 🤔<del>

<del>PR that I closed https://github.com/bitwarden/ios/pull/2747<del>

Edit:After checking, it looks like this is the correct behaviour 👍

andrebispo5 (approved):

PR #2859, opened 2026-07-06
llm: Add Step 7 verify/lint/format section to implementing-ios-code skill

KatherineInCode:
## 📔 Objective

Adds a "Step 7: Verify — Lint, Format, and Regenerate" section to the `implementing-ios-code` skill. This makes the post-implementation verification step explicit and actionable:

- SwiftGen regeneration command if new localization strings were added
- `swiftlint`, `swiftformat --lint --lenient`, and `typos` commands
- Guidance on when to suppress vs. fix SwiftLint violations (`line_length`, `file_length`, `type_body_length`)

Without this step, implementations were sometimes handed back with lint or format violations, or with missing SwiftGen output after adding localization strings — requiring a follow-up round of corrections. Making verification explicit and placing it before the handoff point means violations get caught and fixed in the same session, not discovered during preflight or review.

### DIFF
--- a/.claude/skills/implementing-ios-code/SKILL.md
+++ b/.claude/skills/implementing-ios-code/SKILL.md
@@ -86,6 +86,30 @@ All new public types and methods require DocC (`///`) documentation.
 Exceptions: protocol property/function implementations (docs live in the protocol), mock classes.
 Use pragma marks to organize code. `// MARK: -` is used to denote different objects in the same file; `// MARK:` is used to denote different sections within an object.
 
+## Step 7: Verify — Lint, Format, and Regenerate
+
+Run these checks after all code and tests are written, **before** handing back. Fix every violation found — do not leave warnings for the commit phase.
+
+**If new localization strings were added**, regenerate first so SwiftGen picks them up:
+```bash
+mint run swiftgen config run --config swiftgen-bwr.yml   # BitwardenResources (most common)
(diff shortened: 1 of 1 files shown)


SaintPatrck (dismissed):

.claude/skills/implementing-ios-code/SKILL.md
```diff
@@ -86,6 +86,30 @@ All new public types and methods require DocC (`///`) documentation.
 Exceptions: protocol property/function implementations (docs live in the protocol), mock classes.
 Use pragma marks to organize code. `// MARK: -` is used to denote different objects in the same file; `// MARK:` is used to denote different sections within an object.
 
+## Step 7: Verify — Lint, Format, and Regenerate
```
fedemkr, line 89: 🤔 Should we call the same scripts that we call in the [pre-commit](https://github.com/bitwarden/ios/blob/main/Scripts/pre-commit#L20) hook instead? or combine script calls with some Claude prompts? so we can centralize at least a part of this.

KatherineInCode: I have it now just literally calling the pre-commit hooks, which hopefully centralizes things a little better 🤔

andrebispo5 (approved):
Looks good now with the pre-commit script usage 👍

SaintPatrck (approved):

fedemkr (approved):

PR #2867, opened 2026-07-08
[PM-37992] fix: Show subscription attention card in empty vault and suppress conflicting CTAs

KatherineInCode:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37992

## 📔 Objective

QA follow-up on #2823. Two failures were found:

1. **Card missing for empty/near-empty vaults** — `subscriptionNeedsAttentionActionCard` was only rendered inside `vaultContents(with:)`, which only displays when the vault has items. Users with 0–4 vault items saw `emptyVault` instead, where the card was absent. Fixed by adding the card to `emptyVault`.

2. **Subscription attention card and archive onboarding card appear simultaneously** — When a past-due user had an undismissed archive onboarding CTA, both cards showed at once. Fixed by suppressing `archiveOnboardingActionCard` (and `importLoginsActionCard` in the empty vault) when the subscription attention card is visible.

Also converts `VaultListView` previews from `PreviewProvider` to `#Preview` macros, and adds previews for the two failing states.


BitwardenShared/UI/Vault/Vault/VaultList/VaultListView+SnapshotTests.swift
```diff
@@ -255,14 +253,75 @@ class VaultListViewTests: BitwardenTestCase {
...
+    }
+
+    @MainActor
+    func disabletest_snapshot_profileSwitcherMultiAccount() {
+        processor.state.loadingState = .data([])
+        processor.state.profileSwitcherState = .dualAccounts
+        assertSnapshots(of: subject, as: [.defaultPortrait, .defaultPortraitDark])
+    }
+
+    // The following tests have no direct preview counterpart.
```
matt-livefront: ⛏️ Is this still needed?

BitwardenShared/UI/Vault/Vault/VaultList/VaultListStateTests.swift
```diff
@@ -1,40 +1,76 @@
...
     }
 
-    override func tearDown() {
-        super.tearDown()
+    // MARK: Tests
 
-        subject = nil
+    /// `activeActionCard` returns the highest-priority active card across all 32 flag combinations,
+    /// or `nil` when no flags are set.
+    @Test(arguments: allCardSubsets)
```
matt-livefront: 👍 Nice!

BitwardenShared/UI/Vault/Vault/VaultList/VaultListView+SnapshotTests.swift
KatherineInCode: No, removed.

matt-livefront (dismissed):

BitwardenShared/UI/Vault/Vault/VaultList/VaultListState.swift
```diff
@@ -72,6 +84,24 @@ struct VaultListState: Equatable {
...
+    /// The import logins card is additionally gated on the vault being empty, since it is only
+    /// relevant to users who have not yet added any items.
+    var activeActionCard: VaultListActionCard? {
+        if shouldShowUpgradedToPremiumActionCard { return .upgradedToPremium }
+        if shouldShowPremiumUpgradeActionCard { return .upgradeNeeded }
+        if shouldShowSubscriptionAttentionCard { return .subscriptionNeedsAttention }
+        if shouldShowArchiveOnboardingActionCard { return .introducingArchive }
+        if shouldShowImportLoginsActionCard, isVaultEmpty { return .importItems }
+        return nil
+    }
```
fedemkr, line 92: 🤔 Instead of this what do you think of doing something like:
```swift
enum VaultListActionCard: CaseIterable, Equatable {
    case upgradedToPremium
    case upgradeNeeded
    case subscriptionNeedsAttention
    case introducingArchive
    case importItems
}

var activeActionCard: VaultListActionCard? {
    VaultListActionCard.allCases.first { shouldShow($0) }
}

private func shouldShow(_ card: VaultListActionCard) -> Bool {
    switch card {
    case .upgradedToPremium: shouldShowUpgradedToPremiumActionCard
    case .upgradeNeeded: shouldShowPremiumUpgradeActionCard
    case .subscriptionNeedsAttention: shouldShowSubscriptionAttentionCard
    case .introducingArchive: shouldShowArchiveOnboardingActionCard
    case .importItems: shouldShowImportLoginsActionCard && isVaultEmpty
    }
}
```
`CaseIterable` makes the priorities just reordering the enum cases which is very easy to maintain. Additionally, `shouldShow` uses a `switch` instead of `activeActionCard`'s `if` statements, so when/if a new card is added to the enum it will force to update the `shouldShow` function because the compiler forces exhaustiveness, whereas on the `if` it will silently pass.

KatherineInCode: Good thought, though it breaks the alphabetization rule. Still seems worth it, though; I'll make the change, and include a doc comment about the priority order.

BitwardenShared/UI/Vault/Vault/VaultList/VaultListStateTests.swift
```diff
@@ -1,40 +1,78 @@
...
+import Testing
 
 @testable import BitwardenShared
 @testable import BitwardenSharedMocks
 
-class VaultListStateTests: BitwardenTestCase {
+struct VaultListStateTests {
+    // MARK: Static Properties
+
+    /// All 32 subsets of `VaultListActionCard` cases, used for exhaustive priority testing.
```
fedemkr: ⛏️ Very minor but I'd not explicitly say the number of combinations as we'd need to maintain that should the card cases increase/decrease. I'd just say:
```suggestion
    /// All subsets of `VaultListActionCard` cases, used for exhaustive priority testing.
```

BitwardenShared/UI/Vault/Vault/VaultList/VaultListStateTests.swift
```diff
@@ -1,40 +1,78 @@
...
         subject = VaultListState()
     }
 
-    override func tearDown() {
-        super.tearDown()
+    // MARK: Tests
 
-        subject = nil
+    /// `activeActionCard` returns the highest-priority active card across all 32 flag combinations,
+    /// or `nil` when no flags are set. Tests use an empty vault so the import logins card is eligible.
```
fedemkr: Idem.
```suggestion
    /// `activeActionCard` returns the highest-priority active card across all flag combinations,
    /// or `nil` when no flags are set. Tests use an empty vault so the import logins card is eligible.
```

fedemkr (dismissed):
Looks good, just put 2 really minor ⛏️ but approved 🚀

BitwardenShared/UI/Vault/Vault/VaultList/VaultListStateTests.swift
KatherineInCode: Makes sense to me

BitwardenShared/UI/Vault/Vault/VaultList/VaultListView.swift
```diff
@@ -82,7 +82,7 @@ private struct SearchableVaultListView: View {
             Group {
                 organizationBannerActionCard
 
-                importLoginsActionCard
+                actionCard
```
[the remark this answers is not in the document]
KatherineInCode: It makes sense to also avoid showing some cards in an empty vault. I'll make a change.

matt-livefront (approved):

PR #2868, opened 2026-07-08
[PM-40133] feat: Add Send Controls policy support and AnyCodable array/dictionary decoding

matt-livefront:
## 🎟️ Tracking

[PM-40133](https://bitwarden.atlassian.net/browse/PM-40133)

## 📔 Objective

Adds the new Send Controls policy and feature flag. This also fixes a JSON decoding error for the policy since we didn't previously have `AnyCodable` support for array types.

- Adds `PolicyType.sendControls` (raw value `21`) and completes its bidirectional SDK mapping in `PolicyService+SdkMapping.swift`.
- Adds the `pm-31885-send-controls` feature flag to the `FeatureFlag` catalog (no client-side gating call site yet — this pass is plumbing only).
- Fixes a decode error and closes the whole class of it: `AnyCodable` gains array (`case array([AnyCodable])`) and object/dictionary (`case dictionary([String: AnyCodable])`) support, so it now covers every JSON value kind. A policy `data` field containing a JSON array (e.g. Send Controls' `allowedSendTypes`) or a nested object no longer throws `DecodingError.dataCorruptedError`.

[PM-40133]: https://bitwarden.atlassian.net/browse/PM-40133?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenKit/Core/Platform/Utilities/AnyCodable.swift
+++ b/BitwardenKit/Core/Platform/Utilities/AnyCodable.swift
@@ -3,8 +3,14 @@
 /// A custom codable type that can be used to encode/decode a variety of primitive types.
 ///
 public enum AnyCodable: Codable, Equatable, Sendable {
+    /// The wrapped value is an array of `AnyCodable` values.
+    case array([AnyCodable])
+
     /// The wrapped value is a bool value.
     case bool(Bool)
+
+    /// The wrapped value is a keyed dictionary of `AnyCodable` values.
+    case dictionary([String: AnyCodable])
 
     /// The wrapped value is a double value.
     case double(Double)
@@ -34,6 +40,10 @@ public enum AnyCodable: Codable, Equatable, Sendable {
             self = .null
         } else if let stringValue = try? container.decode(String.self) {
             self = .string(stringValue)
+        } else if let arrayValue = try? container.decode([AnyCodable].self) {
+            self = .array(arrayValue)
+        } else if let dictionaryValue = try? container.decode([String: AnyCodable].self) {
+            self = .dictionary(dictionaryValue)
         } else {
             throw DecodingError.dataCorruptedError(
                 in: container,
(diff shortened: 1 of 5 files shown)


BitwardenKit/Core/Platform/Utilities/AnyCodable.swift
```diff
@@ -3,9 +3,15 @@
...
 public enum AnyCodable: Codable, Equatable, Sendable {
+    /// The wrapped value is an array of `AnyCodable` values.
+    case array([AnyCodable])
+
     /// The wrapped value is a bool value.
     case bool(Bool)
 
+    /// The wrapped value is a keyed dictionary of `AnyCodable` values.
+    case dictionary([String: AnyCodable])
+
```
fedemkr, line 14: 🤔 Do we need to consider the possibility of array/dictionary values being `nil`? E.g. 
`let json = #"{"something": {"a": 1, "bOptional": null}}"#`
I guess it would be more likely in a dictionary's key/value to have a potential `nil` value, on arrays it seems a bit far-fetched but raising it just in case.

fedemkr (commented):
Looks good, just one question.

BitwardenKit/Core/Platform/Utilities/AnyCodable.swift
matt-livefront: Good question, that is handled by `AnyCodable.null`: https://github.com/bitwarden/ios/blob/5d839bd80d94d5d6e2db93d3c994cdaa4af3e3f6/BitwardenKit/Core/Platform/Utilities/AnyCodable.swift#L21-L22


```
(lldb) po dictionary
▿ 1 element
  ▿ 0 : 2 elements
    - key : "something"
    ▿ value : AnyCodable
      ▿ dictionary : 2 elements
        ▿ 0 : 2 elements
          - key : "a"
          ▿ value : AnyCodable
            - int : 1
        ▿ 1 : 2 elements
          - key : "bOptional"
          - value : BitwardenKit.AnyCodable.null
```

fedemkr: oh excellent! thanks

fedemkr (approved):

PR #2869, opened 2026-07-08
[PM-39884] feat: Hide Gov region from start registration region picker

fedemkr:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39884

## 📔 Objective

FedRAMP does not allow self-registration, so the Gov region
(`bitwarden-gov.com`) is now hidden from the region selector on the
start registration screen.


vvolkgang (dismissed):
:shipit: Test is failing, a unit test needs to be updated. 

Small note about the description:

> Additionally, the processor subscribes to
configPublisher and dismisses with an alert if disableUserRegistration
becomes true after a region change — mirroring the pattern used in
LandingProcessor.

I saw that this was implemented in https://github.com/bitwarden/ios/pull/2911 so we can remove it here!

vvolkgang (dismissed):
:shipit:

morganzellers-bw (approved):

PR #2870, opened 2026-07-09
[PM-33981] feat: Add device management models and API layer

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33981

## 📔 Objective

Adds the complete data and API layer for the device management feature — no UI changes in this PR.

- Promotes `DeviceType` from an `Int` typealias to a proper `enum DeviceType: Int` in BitwardenKit, with a safe `Decodable` fallback to `.unknownBrowser` for unknown server values
- Adds `DeviceTypeCategory` and `DeviceType` display extensions (`category`, `platform`, `displayName`, `pendingRequestMatchKey`) in BitwardenShared Core
- Adds `DeviceActivityStatus` enum mapping server timestamps to activity buckets using calendar-day comparison
- Adds `DeviceListItem` domain model, `DeviceResponse`/`DevicesListResponse` API models, `CurrentDeviceRequest`/`DevicesListRequest`, and `DeviceAPIService` with `getCurrentDevice(appId:)` and `getDevices()` endpoints
- Promotes `deviceAPIService` to a proper `ServiceContainer` init parameter (consistent with all other services); `withMocks()` defaults it to the shared `apiService` mock (overridable with an explicit `MockDeviceAPIService()`), matching how other `APIService`-extension-based services are wired
- Adds unit tests for all new types and request/response models; adds JSON fixtures

## 📸 Screenshots

No UI changes in this PR.


BitwardenShared/Core/Auth/Models/Domain/DeviceListItem.swift
```diff
@@ -0,0 +1,108 @@
...
+        deviceType = device.type
+        displayName = device.type.displayName
+        firstLogin = device.creationDate
+        id = device.id
+        identifier = device.identifier
+        isCurrentSession = false
+        isTrusted = device.isTrusted
+        lastActivityDate = device.lastActivityDate
+        pendingRequest = nil
+    }
```
matt-livefront: 🎨 If you move the custom initializer `init(device:timeProvider)` into an extension on this type, then you'll get the top initializer synthesized for free.

BitwardenShared/Core/Auth/Models/Enum/DeviceActivityStatus.swift
```diff
@@ -0,0 +1,93 @@
...
+
+        switch daysDifference {
+        case 0:
+            self = .today
+        case 1 ... 7:
+            self = .thisWeek
+        case 8 ... 14:
+            self = .lastWeek
+        case 15 ... 30:
+            self = .thisMonth
```
matt-livefront: 🤔 Shouldn't "this month" only be the case if the month components are the same? If it's July 15, and the device was last active June 20 (25 days ago), this would return `.thisMonth`, but that's actually last month.

BitwardenShared/Core/Platform/Services/ServiceContainer.swift
```diff
@@ -1274,10 +1281,6 @@ extension ServiceContainer {
         apiService
     }
 
-    var deviceAPIService: DeviceAPIService {
-        apiService
-    }
```
matt-livefront, line 1301: ❓ I'm curious why change this? Do we need to store a copy of the `APIService` as `DeviceAPIService` in the service container instead of just using this computed variable?

BitwardenShared/Core/Auth/Models/Enum/DeviceTypeCategory.swift
matt-livefront, line 1: ⛏️ Should this be named `DeviceTypeCategory.swift` to match the enum?

BitwardenKit/Core/Platform/Models/Enum/DeviceType.swift
```diff
@@ -0,0 +1,47 @@
...
+    /// for values not yet recognized by the client.
+    public init(_ rawValue: Int) {
+        self = DeviceType(rawValue: rawValue) ?? .unknownBrowser
+    }
+
+    public init(from decoder: any Decoder) throws {
+        let container = try decoder.singleValueContainer()
+        let rawValue = try container.decode(Int.self)
+        self = DeviceType(rawValue: rawValue) ?? .unknownBrowser
+    }
```
matt-livefront: 🎨 Instead of implementing a custom initializer to provide a default unknown value, you can use the `DefaultValue` property wrapper.

DeviceType would need to conform to `DefaultValueProvider`:
```
// MARK: - DefaultValueProvider

extension DeviceType: DefaultValueProvider {
    public static var defaultValue: DeviceType { .unknownBrowser }
}
```

And then in DeviceResponse, the type can be wrapped with the property wrapper:

```
    @DefaultValue var type: DeviceType
```

BitwardenShared/Core/Platform/Services/ServiceContainer.swift
andrebispo5: > It's needed as a stored property, not revertible to a computed passthrough — checked the git history on this branch and it was a deliberate change. Earlier revisions had a test-only mutable `deviceAPIServiceOverride` var with a computed `deviceAPIService` property (`deviceAPIServiceOverride ?? apiService`), which is the anti-pattern this commit removed. Promoting it to a proper `init` parameter lets `ServiceContainer.withMocks(deviceAPIService:)` inject a real `MockDeviceAPIService` through the standard DI path (consistent with `authService`, `configService`, and most other services in the container), instead of a mutable side-door override. Reverting to a computed `{ apiService }` passthrough would mean bringing that mutable override var back to keep it testable — happy to do that if you'd prefer it, but I think the stored-property version is the better outcome here.

BitwardenShared/Core/Auth/Models/Enum/DeviceActivityStatus.swift
```diff
@@ -0,0 +1,94 @@
...
+            self = .today
+        case 1 ... 7:
+            self = .thisWeek
+        case 8 ... 14:
+            self = .lastWeek
+        default:
+            // Beyond the last two weeks, only classify as `.thisMonth` if the date actually
+            // falls within the current calendar month; a day count alone can't tell June 20
+            // apart from July 20 relative to a July 15 "now".
+            self = calendar.isDate(date, equalTo: now, toGranularity: .month) ? .thisMonth : .overThirtyDaysAgo
```
matt-livefront: 🤔 I still feel like this isn't quite right. Now if it's July 15, and the last activity was June 20, it would return `.overThirtyDaysAgo` even though it's less than 30 days. Is it worth adding an "X days ago" label to handle the case where it isn't this month but it also isn't over 30 days ago?

andrebispo5: I just noticed that this changed to match other clients, fixing it now.

BitwardenResources/Localizations/en.lproj/Localizable.strings
```diff
@@ -1169,3 +1169,18 @@
...
+"PastSevenDays" = "Past 7 days";
+"PastFourteenDays" = "Past 14 days";
+"PastThirtyDays" = "Past 30 days";
+"OverThirtyDaysAgo" = "Over 30 days ago";
+"Unknown" = "Unknown";
+/* Device display name. %1$@ = category (e.g. "Mobile"), %2$@ = platform (e.g. "iOS"). */
+"DeviceDisplayName" = "%1$@ - %2$@";
+"Mobile" = "Mobile";
+"BrowserExtension" = "Browser extension";
+"WebVaultDeviceType" = "Web vault";
```
fedemkr: ⛏️ This should just be `"WebVault"`

BitwardenResources/Localizations/en.lproj/Localizable.strings
```diff
@@ -1169,3 +1169,18 @@
...
 "MoneyMarket" = "Money market";
 "Savings" = "Savings";
+"Today" = "Today";
+"PastSevenDays" = "Past 7 days";
+"PastFourteenDays" = "Past 14 days";
+"PastThirtyDays" = "Past 30 days";
+"OverThirtyDaysAgo" = "Over 30 days ago";
+"Unknown" = "Unknown";
+/* Device display name. %1$@ = category (e.g. "Mobile"), %2$@ = platform (e.g. "iOS"). */
+"DeviceDisplayName" = "%1$@ - %2$@";
```
fedemkr: ⛏️ I would use `DeviceDisplayNameXHyphenY` to explicitly indicate that it has parameters and also to avoid any potential conflicts if in the future we actually have the string `Device display name` which seems like a possibility and would match this key.

BitwardenShared/Core/Auth/Models/Domain/DeviceListItem.swift
```diff
@@ -0,0 +1,74 @@
...
+    /// The display name of the device.
+    let displayName: String
+
+    /// The date of the first login on this device.
+    let firstLogin: Date
+
+    /// Whether the device has a pending login request.
+    ///
+    /// Computed from `pendingRequest` — `true` when a pending request is present.
+    var hasPendingRequest: Bool { pendingRequest != nil }
```
fedemkr, line 24: ⚠️ Docs do not reflect the implementation, docs says `true` and implementation `!= nil`. Also I don't think docs should be giving that information, it would only have to say the first line IMO.

BitwardenShared/Core/Auth/Models/Enum/DeviceActivityStatus.swift
```diff
@@ -0,0 +1,95 @@
...
+
+        // `.day` is always non-nil when explicitly requested via `dateComponents([.day]:)`;
+        // the guard is defensive.
+        guard let daysDifference = calendar.dateComponents([.day], from: startOfDate, to: startOfToday).day else {
+            self = .unknown
+            return
+        }
+
+        // Bucket boundaries mirror the Android app's day-count logic exactly, so both platforms
+        // show identical "last active" labels for the same date.
```
fedemkr: ⛏️ Which additional value adds this comment? I mean, Last active labels should be the same no matter the platform so is there any special treatment here that differs from `clients`?

BitwardenShared/Core/Auth/Models/Enum/DeviceActivityStatus.swift
```diff
@@ -0,0 +1,95 @@
...
+        let calendar = Calendar.current
+        let startOfDate = calendar.startOfDay(for: date)
+        let startOfToday = calendar.startOfDay(for: now)
+
+        // `.day` is always non-nil when explicitly requested via `dateComponents([.day]:)`;
+        // the guard is defensive.
+        guard let daysDifference = calendar.dateComponents([.day], from: startOfDate, to: startOfToday).day else {
+            self = .unknown
+            return
+        }
```
fedemkr, line 78: 🤔 If `date` was produced/stored assuming UTC day boundaries (e.g. a server-issued timestamp) but the device is in a different time zone, `startOfDay` can shift which "day" a moment falls on, causing off-by-one results near midnight UTC vs. local midnight.
Can you check how `date` is created?

BitwardenShared/Core/Auth/Models/Enum/DeviceActivityStatus.swift
fedemkr: The same happens with "This week" and "Last week"

BitwardenShared/Core/Auth/Models/Enum/DeviceTypeCategory.swift
```diff
@@ -0,0 +1,140 @@
...
+        case .vivaldiBrowser, .vivaldiExtension:
+            "Vivaldi"
+        case .safariBrowser, .safariExtension:
+            "Safari"
+        case .ieBrowser:
+            "Internet Explorer"
+        case .duckDuckGoBrowser:
+            "DuckDuckGo"
+        case .unknownBrowser:
+            ""
```
fedemkr: ⛏️ Can we put this alongside with `case .sdk, .server`? AS far as I can see they have the same value.

BitwardenShared/Core/Auth/Models/Enum/DeviceActivityStatus.swift
andrebispo5: it's Claude doing its thing and I didn't noticed. This logic mimic all platforms now. I can also update the enums to now reflect the string localization they refer to.

BitwardenResources/Localizations/en.lproj/Localizable.strings
andrebispo5: There is already a key named WebVault but it's value is "Bitwarden Web Vault". I'll change both.

BitwardenShared/Core/Auth/Models/Enum/DeviceActivityStatus.swift
andrebispo5: Claude special: 

> Checked how `date` is produced: `DeviceResponse.lastActivityDate` decodes via `JSONDecoder.defaultDecoder` (BitwardenKit/Core/Platform/Services/API/Extensions/JSONDecoder+Bitwarden.swift), which parses the server's ISO8601 timestamp (with explicit UTC offset, via `ISO8601DateFormatter`) into a `Date`. A `Date` is just an absolute instant, not a timezone-tagged wall-clock value, so there's no 'UTC day boundary' baked into it. `Calendar.current.startOfDay(for:)` then buckets that instant using the device's own local timezone — exactly what we want for 'how many days ago, from this user's perspective.' There isn't a mismatch here: if the server sent a UTC-offset timestamp and we localize with `Calendar.current` (local timezone), that's the correct pairing. The off-by-one you're describing would only happen if `date` were a naively-truncated UTC midnight reinterpreted as local, which isn't the case here.

BitwardenResources/Localizations/en.lproj/Localizable.strings
```diff
@@ -138,7 +138,7 @@
 "UnlockWithOpticID" = "Unlock with Optic ID";
 "UnlockWithPIN" = "Unlock with PIN code";
 "VerificationCode" = "Verification code";
-"WebVault" = "Bitwarden web vault";
+"BitwardenWebVault" = "Bitwarden web vault";
```
fedemkr: ⚠️ @andrebispo5 we need to be careful when doing this kind of things as we might be losing all translations for this key or worse using the already in-place translations for this key in the new one so you might be getting "Caja fuerte web de Bitwarden" in Spanish for `WebVault` now that you changed it, but in the other localization files it still the same old key. 
We might need to just use `WebVault2` for the new key until we come up with a flow to safely do the renaming of keys without risking wrong translations.

fedemkr (changes requested):

BitwardenResources/Localizations/en.lproj/Localizable.strings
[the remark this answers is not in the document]
andrebispo5: Fixed by reverting `WebVault` back to its original key/meaning ("Bitwarden web vault") instead of renaming it to `BitwardenWebVault" — so no cross-locale updates are needed and no existing translations are lost. The new device-type "Web vault" string now lives at its own `WebVault2` key, which starts fresh in every locale like any newly added string.

andrebispo5: Yup that makes sense, thank you. I've reverted the change and added the WebVault2 key for the new value.

fedemkr (approved):

PR #2876, opened 2026-07-10
[PM-39851] feat: Add SHA-256 integrity check for cached Fill Assist rules

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-39851

## 📔 Objective

Adds a SHA-256 integrity check for the Fill Assist rules cached on device. When rules are fetched and saved, a fingerprint is computed and stored in the Keychain; when the cached rules are read (during autofill or the next sync), the fingerprint is recomputed and compared. On mismatch (or on any inconsistent absent/present combination), the tampered cache and its fingerprint are cleared, and the normal sync flow redownloads and recreates the fingerprint.

Uses a sorted-keys `JSONEncoder` for fingerprint computation specifically to keep the hash deterministic regardless of `Dictionary` iteration order (the cached rules contain nested dictionaries), verified by a dedicated round-trip regression test.

This is not a reintroduction of the SHA-256 check removed in a prior commit — that check hashed CDN-delivered content against the CDN's own untrusted `cid` (circular trust, no real security benefit). This fingerprint is computed and verified entirely on-device, to detect local tampering of the cached data (e.g. jailbreak/local-tamper), independent of CDN trust.


BitwardenShared/Core/Autofill/Repositories/FillAssistRepository.swift
```diff
@@ -150,6 +167,63 @@ class DefaultFillAssistRepository: FillAssistRepository {
...
+                try await keychainRepository.setUserAuthKey(
+                    for: .fillAssistRulesFingerprint(userId: userId),
+                    value: newFingerprint,
+                )
+            } catch {
+                errorReporter.log(error: error)
+            }
+        } else {
+            errorReporter.log(error: FillAssistFingerprintError.encodingFailed)
+        }
```
fedemkr: 🎨 Instead of `if let ... = try? ...` and handling the error in the `else` you should be using `do... catch` and log the error thrown.

BitwardenShared/Core/Autofill/Repositories/FillAssistRepository.swift
```diff
@@ -150,6 +167,63 @@ class DefaultFillAssistRepository: FillAssistRepository {
...
+        guard let storedFingerprint,
+              let computed = try? fingerprint(for: cached),
+              storedFingerprint == computed
+        else {
+            appSettingsStore.setFillAssistCachedData(nil, userId: userId)
+            try? await keychainRepository.deleteUserAuthKey(for: .fillAssistRulesFingerprint(userId: userId))
+            return nil
+        }
+
+        return cached
```
fedemkr, line 218: 🤔 I see you're using all `try?`. Shouldn't we log all the errors or why are they all being ignored?

BitwardenShared/Core/Autofill/Repositories/FillAssistRepository.swift
```diff
@@ -150,6 +167,63 @@ class DefaultFillAssistRepository: FillAssistRepository {
...
+    /// concurrent use across calls on a shared instance.
+    ///
+    /// - Parameter data: The cached data to fingerprint.
+    /// - Returns: A lowercase hexadecimal SHA-256 digest of the encoded data.
+    ///
+    private func fingerprint(for data: FillAssistCachedData) throws -> String {
+        let encoder = JSONEncoder()
+        encoder.outputFormatting = [.sortedKeys]
+        return try encoder.encode(data).generatedHash(using: SHA256.self)
     }
```
fedemkr: 🎨 I'd move this to a different object which can be called from here. Therefore you can properly mock it on the tests instead of having to duplicate the logic in the fixtures. The only difference would be to have an encodable data as parameter so it can be more flexible and used in other places if necessary:
```swift
func fingerprint(for data: Encodable) throws -> String
```

BitwardenShared/Core/Autofill/Repositories/FillAssistRepositoryTests.Fixtures.swift
```diff
@@ -1,12 +1,29 @@
...
+        return try encoder.encode(data).generatedHash(using: SHA256.self)
+    }
+
+    /// Pre-populates `appSettingsStore` with `data` and stores its matching fingerprint in
+    /// `keychainRepository`, so the cache will pass integrity verification when read.
+    func cacheVerifiedData(_ data: FillAssistCachedData, userId: String = "1") throws {
+        appSettingsStore.fillAssistCachedDataByUserId[userId] = data
+        keychainRepository.getUserAuthKeyValueReturnValue = try fingerprint(for: data)
+    }
+
```
fedemkr: 🎨 Remove when SUT fingerprint function is moved to a different object which can be mocked instead of duplicating logic here. Adjust to use the mock instead.

fedemkr (changes requested):

fedemkr (approved):

PR #2871, opened 2026-07-09
[PM-33981] feat: Add device management UI and Account Security integration

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-33981

## 📔 Objective

⛓️ Depends on #2870 (PM-33981 Core models and API layer) — merge that PR first. 

Adds the complete device management UI on top of the Core layer from the preceding PR.

- Adds `PillBadgeStyle.info` (blue palette) to `PillBadgeView` for the Current Session badge
- Adds the `DeviceManagementView` (list with pull-to-refresh and empty state), `DeviceRow` (pill badge at top, Trusted label, Recently active and First login info rows, full-row tap), and `DeviceManagementProcessor` (parallel device + pending-request loading, calendar-day activity buckets, platform-based request matching via `DeviceType.platform` — `pendingRequestMatchKey` was reverted since the server sends the same `requestDeviceType` for browser and extension variants)
- Wires `.deviceManagement` route in `SettingsRoute` and `SettingsCoordinator`
- Adds `manageDevices` feature flag; gates a "Devices" row in the Account Security Other section behind it; reorders the section: Account fingerprint phrase → Two-step login → Devices → Lock now → Log out → Delete account
- Adds `DeviceManagementProcessorTests` (Swift Testing) and `DeviceManagementViewTests` (ViewInspector); adds `AccountSecurityProcessorTests` coverage for the new action and flag load

## 📸 Screenshots
image
image
image

PR #2423, opened 2026-03-09
wip: Implement DeviceAuthKeyService using SDK

iinuwa:
## 📔 Objective

Uses https://github.com/bitwarden/sdk-internal/pull/816 to implement the DeviceAuthKeyService.

PR #2878, opened 2026-07-11
[PM-40262] Complete Device Auth Key core support

2001Y:
## Summary

- complete the iOS Device Auth Key service against the current Bitwarden SDK
- keep Account Security UI changes out of this PR to avoid overlap with #2871
- preserve account isolation and fail closed for mismatched credentials

## Current state

This draft currently contains the first TDD specification only. The expected RED is `DeviceAuthKeyError.notImplemented`; implementation commits will follow after that behavior is confirmed in CI.

This continues the core-service direction explored in #2423 without cherry-picking its conflicting branch.

## Test plan

- [ ] RED: missing-record assertion test fails on the current stub
- [ ] GREEN: targeted `DeviceAuthKeyServiceTests` pass
- [ ] related keychain/autofill tests pass
- [ ] `BitwardenShared` generic simulator build passes

### DIFF
--- a/BitwardenShared/Core/Auth/Services/DeviceAuthKeyServiceTests.swift
+++ b/BitwardenShared/Core/Auth/Services/DeviceAuthKeyServiceTests.swift
@@ -40,23 +40,24 @@ final class DeviceAuthKeyServiceTests: BitwardenTestCase {
 
     // MARK: Tests - assertDeviceAuthKey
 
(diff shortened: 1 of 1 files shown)


2001Y:
Closing this premature test-only draft while coordinating with the existing Device Auth Key work in #2423. Local implementation and validation will continue before any new public PR.

PR #2899, opened 2026-07-21
[PM-40305] fix: Prevent login-with-device vault key wrapped to a substituted public key

matt-livefront:
## 🎟️ Tracking

[PM-40305](https://bitwarden.atlassian.net/browse/PM-40305)

## 📔 Objective

Fixes an issue where approving a login-with-device request could wrap the vault key to a public key the user never verified.

- On Approve, `LoginRequestProcessor.reloadData()` re-fetched the login request from the server and silently overwrote `state.request` (including `publicKey`) after only checking expiry/answered status — the fingerprint the user had just verified was never re-checked against the key the vault key was ultimately wrapped to.
- A malicious or compromised server could return the genuine device's key while the user verifies the fingerprint, then swap in an attacker key on the approve-time reload, causing the vault key to be wrapped to the attacker's key instead.
- `reloadData()` no longer writes the reloaded request back into `state.request` at all — it only uses the fresh fetch to check `isExpired`/`isAnswered`. The vault key is now always wrapped to the public key from the request as originally loaded and verified, never to whatever key a later reload returns. This also fixes a related gap where `answerRequest` would still call `answerLoginRequest` after `reloadData()` had shown an expired/answered alert.

[PM-40305]: https://bitwarden.atlassian.net/browse/PM-40305?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/UI/Platform/LoginRequest/LoginRequestProcessor.swift
+++ b/BitwardenShared/UI/Platform/LoginRequest/LoginRequestProcessor.swift
@@ -99,7 +99,11 @@ final class LoginRequestProcessor: StateProcessor<LoginRequestState, LoginReques
             coordinator.showLoadingOverlay(title: Localizations.loading)
 
             // First reload the request to ensure it hasn't expired or been answered already.
-            await reloadData()
+            // If the reload blocked the update, an alert has already been shown; don't
+            // answer the request.
+            guard await reloadData() else {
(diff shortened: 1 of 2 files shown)


vvolkgang (approved):
:shipit:

PR #2903, opened 2026-07-22
[PM-40767] fix: Cancel extension when dismissing save-password flow

fedemkr:
## 🎟️ Tracking

PM-40767

## 📔 Objective

When the AutoFill extension presents `AddEditItemView` for the save-password flow (iOS 26.2+) and the user taps Cancel, the extension was showing a blank screen instead of closing.

`AddEditItemProcessor.handleDismiss()` was calling `coordinator.navigate(to: .dismiss())`, which dismissed only the `VaultItemCoordinator`'s navigation controller without calling `extensionContext.cancelRequest()`. The fix casts `appExtensionDelegate` to `CredentialProviderExtensionDelegate` and, when `isSavingPasswordCredential` is true, calls `didCancel()` to close the extension properly.


BitwardenShared/UI/Vault/VaultItem/AddEditItem/AddEditItemProcessor.swift
matt-livefront: 🎨 As this grows, I wonder if this would better read as multiple `if` statements rather than a guard? 

Also, I noticed that when in the action extension, we handle both completing the request and cancelling in this method. But for the autofill extension, we handle completing the request in `addItem()` and cancelling it here. Does it make sense to standardize? Having both complete and dismiss handled together would help to consolidate the logic.

fedemkr: Totally agree Matt, I've made the changes, thank you!

matt-livefront (approved):

PR #2920, opened 2026-07-28
[PM-41028] feat: Restrict View Send screen for policy non-compliant Sends

matt-livefront:
## 🎟️ Tracking

[PM-41028](https://bitwarden.atlassian.net/browse/PM-41028)

## 📔 Objective

- Add `SendPolicyOptions.isSendTypeRestricted(for:)` to determine whether an existing Send violates the active Send Controls policy (restricted type takes priority over restricted access type/deletion date/hide-email).
- View Send screen now hides the send-link (copy/share) section and the floating edit button, and shows an `ActionCard` "Organization policy restriction" banner for non-compliant Sends, with copy varying by violation: restricted text type, restricted file type, or restricted fields (with a "Make a copy" action).
- File Sends never offer "Make a copy" (an existing attachment can't be carried into a new Send), regardless of which policy field is actually violated.
- "Make a copy" opens a new Send pre-filled from the original Send's data, adjusted to comply with the active policy via the existing `AddEditSendItemProcessor` policy-enforcement logic.

## 📸 Screenshots

| Send Text Restriction |
|--------|
| send-text-restriction | 

| Send File Restriction |
|--------|
| send-file-restriction | 

| Send Restricted Field |
|--------|
| send-restricted-field | 

[PM-41028]: https://bitwarden.atlassian.net/browse/PM-41028?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/UI/Tools/Send/SendItem/AddEditSendItem/AddEditSendItemState.swift
+++ b/BitwardenShared/UI/Tools/Send/SendItem/AddEditSendItem/AddEditSendItemState.swift
@@ -257,6 +257,36 @@ extension AddEditSendItemState {
         )
     }
 
+    /// Creates a new `AddEditSendItemState` for creating a policy-compliant copy of a restricted
+    /// `sendView`. Used when the user taps "Make a copy" from the View Send screen's restriction
+    /// banner; unlike `init(sendView:)`, this omits the id, access id, and key so that saving
+    /// creates a brand-new Send rather than updating the original, and defaults to `.add` mode.
+    /// Policy-enforced fields (access type, deletion date) are applied afterward by the existing
+    /// `loadData()` policy-override logic, the same as for any other new Send.
+    ///
+    /// - Parameter sendView: The restricted `SendView` to copy details from.
+    ///
+    init(copyingFrom sendView: SendView) {
+        let accessType: SendAccessType = if sendView.hasPassword {
+            .anyoneWithPassword
+        } else {
+            SendAccessType(authType: sendView.authType)
+        }
+
+        self.init(
+            accessType: accessType,
+            isHideMyEmailOn: sendView.hideEmail,
+            isHideTextByDefaultOn: sendView.text?.hidden ?? false,
+            maximumAccessCount: sendView.maxAccessCount.map(Int.init) ?? 0,
+            mode: .add,
+            name: sendView.name,
+            notes: sendView.notes ?? "",
+            recipientEmails: sendView.emails.isEmpty ? [""] : sendView.emails,
+            text: sendView.text?.text ?? "",
+            type: SendType(sendType: sendView.type),
+        )
+    }
+
     /// Returns a `SendView` based on the properties of the `AddEditSendItemState`.
     ///
     func newSendView() -> SendView {
--- a/BitwardenShared/UI/Tools/Send/SendItem/AddEditSendItem/AddEditSendItemState.swift
+++ b/BitwardenShared/UI/Tools/Send/SendItem/AddEditSendItem/AddEditSendItemState.swift
@@ -261,6 +261,36 @@ extension AddEditSendItemState {
         )
     }
 
+    /// Creates a new `AddEditSendItemState` for creating a policy-compliant copy of a restricted
+    /// `sendView`. Used when the user taps "Make a copy" from the View Send screen's restriction
+    /// banner; unlike `init(sendView:)`, this omits the id, access id, and key so that saving
+    /// creates a brand-new Send rather than updating the original, and defaults to `.add` mode.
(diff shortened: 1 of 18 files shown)


BitwardenShared/UI/Tools/Send/SendItem/AddEditSendItem/AddEditSendItemState.swift
```diff
@@ -257,6 +257,36 @@ extension AddEditSendItemState {
...
+
+        self.init(
+            accessType: accessType,
+            isHideMyEmailOn: sendView.hideEmail,
+            isHideTextByDefaultOn: sendView.text?.hidden ?? false,
+            maximumAccessCount: sendView.maxAccessCount.map(Int.init) ?? 0,
+            mode: .add,
+            name: sendView.name,
+            notes: sendView.notes ?? "",
+            recipientEmails: sendView.emails.isEmpty ? [""] : sendView.emails,
```
fedemkr, line 288: 🤔 Is it needed to add an empty string as one of the emails recipient when the `sendView.emails` is empty? If that's actually necessary, then a test is missing for that case.

BitwardenShared/UI/Tools/Send/SendItem/ViewSendItem/ViewSendItemStateTests.swift
```diff
@@ -21,6 +21,44 @@ class ViewSendItemStateTests: BitwardenTestCase {
...
+
+    /// A disabled Text Send whose type isn't restricted by policy offers the "Make a copy" action.
+    func test_isDisabled_disabledTextSend_typeNotRestricted() {
+        let subject = ViewSendItemState(sendView: .fixture(type: .text, disabled: true))
+
+        XCTAssertTrue(subject.isDisabled)
+        XCTAssertEqual(subject.restrictionBannerMessage, Localizations.toEditThisSendMakeACopyDescriptionLong)
+        XCTAssertTrue(subject.canMakeCopy)
+    }
+
```
fedemkr: 🤔 This is a different approach as how we normally do tests. It's a bit confusing to have the test saying `isDisabled` when that's not the actual focus of the test and at least you would just need 2 instead of 4 test functions to test that particular property.
I understand this is to save doing a lot of tests to test the functions separately but wanted to chat a bit on whether we'd like to deviate from our standard into this kind of several-in-one tests which saves several lines of code. cc: @bitwarden/team-ios 
If so, perhaps we'd need to update a bit the naming convention for these cases and/or the properties to be tested in the test functions' docs.

fedemkr (commented):
Looks good, just some minor comments.

BitwardenShared/UI/Tools/Send/SendItem/AddEditSendItem/AddEditSendItemState.swift
matt-livefront: It's needed to provide an empty text field for the user to add an email. I'll add a test.

BitwardenShared/UI/Tools/Send/SendItem/ViewSendItem/ViewSendItemState.swift
```diff
@@ -24,16 +27,48 @@ struct ViewSendItemState: Equatable {
...
+
     /// The send's share URL without a scheme for displaying in the UI.
     var displayShareURL: String? {
         shareURL?.withoutScheme
     }
 
+    /// Whether this Send has been disabled.
+    var isDisabled: Bool {
+        sendView.disabled
+    }
```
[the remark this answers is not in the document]
matt-livefront: Disabled sends shouldn't have been editable previously, so this is fixing an existing bug and doesn't need to be behind the feature flag.

BitwardenShared/UI/Tools/Send/SendItem/ViewSendItem/ViewSendItemStateTests.swift
matt-livefront: Good point. It seemed worthwhile to combine these tests given that these properties are all tightly coupled. The common theme here is testing the properties related to disabled sends due to type restrictions. What if we rename them to `test_restrictionState_xyz`?

fedemkr: Yeah I agree on that name

fedemkr (commented):
Looks good, only renaming the tests left ⛏️

fedemkr (dismissed):

fedemkr (approved):

PR #2925, opened 2026-08-01
[PM-41325] feat: Add custom header support for self-hosted environments

jalenfran:
## 🎟️ Tracking

Community feature request: [Allow custom header for the Bitwarden app (Android, iOS)](https://community.bitwarden.com/t/allow-custom-header-for-the-bitwarden-app-android-ios/65289)

## 📔 Objective

Allow users to configure custom HTTP headers on the pre-auth Self-Hosted environment screen,
applied to every request sent to the environment's hosts. This supports self-hosted deployments
behind reverse proxies that gate access on a header — most commonly Cloudflare Access
(`CF-Access-Client-Id` / `CF-Access-Client-Secret`) — as a lighter-weight alternative to mTLS.

The implementation deliberately mirrors the existing mTLS client-certificate feature:

- **Storage**: header name/value pairs are credentials, so they are stored in the iOS Keychain
  (`AfterFirstUnlockThisDeviceOnly`). Only an opaque UUID identifier is persisted in
  `EnvironmentURLData`, exactly like the certificate fingerprint. The identifier is shared across
  accounts pointing at the same environment, with reference-counted Keychain cleanup.
- **Injection**: a new `CustomHeadersRequestHandler` runs in the `HTTPServiceBuilder` chain.
  It is **host-scoped**: headers are only attached when the request host matches one of the
  environment's hosts, so credentials are never sent to third parties (e.g. HIBP).
- **Icons**: `CipherIconImageLoader` bypasses `HTTPService`, so it attaches the headers itself —
  the same dual-path treatment mTLS required.
- **UI**: a "Custom headers" section on the Self-Hosted screen with add/remove name–value rows;
  values are masked with a reveal toggle like other credential fields.

The second commit fixes three pre-existing credential-hygiene patterns shared with mTLS,
found while auditing the feature:

- Switching from a self-hosted environment to a built-in region now removes Keychain
  credentials (client certificate identity and custom headers) whose pre-auth references were
  dropped, instead of orphaning them.
- `deleteAccount` now cleans up the account's client certificate and custom headers,
  matching `logout`.
- `OSLogHTTPLogger` now redacts the values of headers not on a known-safe allow-list, so
  authorization tokens, cookies, and user-configured custom headers no longer reach the
  unified log verbatim.

## 📸 Screenshots

Custom headers section on the Self-Hosted environment screen:

| Value revealed | Value masked (default) |
| :---: | :---: |
| Self-hosted settings with a custom header value shown in plain text | Self-hosted settings with the custom header value masked |


hugoduraes:
This would be an awesome addition! 🚀

PR #2930, opened 2026-08-03
[PM-16841] fix: Save active account last-active time before adding new account

fedemkr:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-16841

## 📔 Objective

When a user added a second account (Account B) via the Account Switcher, Account A was incorrectly locked — regardless of vault timeout settings — the next time the Account Switcher was opened.

**Root cause:** `ProfileSwitcherHandler.handleProfileSwitcherEffect(.addAccountPressed)` navigated to the add-account flow without first saving the current account's `lastActiveTime`. When `checkSessionTimeouts` later ran (triggered by opening the Account Switcher), a nil `lastActiveTime` was treated as a timed-out session, causing Account A to be locked prematurely.

This contrasts with the `switchAccountRedirect` path (switching to an *existing* account), which correctly calls `vaultTimeoutService.setLastActiveTime(userId:)` before switching.

**Fix:** Added `setLastActiveAccountTime()` to `AuthRepository`, which records the current timestamp for the active account. The `.addAccountPressed` handler now calls this method (with do-catch error logging) before navigating, mirroring the existing `switchAccountRedirect` pattern.


BitwardenShared/Core/Auth/Repositories/TestHelpers/MockAuthRepository.swift
```diff
@@ -75,6 +75,8 @@ class MockAuthRepository: AuthRepository { // swiftlint:disable:this type_body_l
     var setMasterPasswordResult: Result<Void, Error> = .success(())
     var setPinsRequirePasswordAfterRestart: Bool?
     var setPinsResult: Result<Void, Error> = .success(())
+    var setLastActiveAccountTimeCalled = false
```
andrebispo5: ⛏️ alphabetise

BitwardenShared/Core/Auth/Repositories/TestHelpers/MockAuthRepository.swift
```diff
@@ -362,6 +364,13 @@ class MockAuthRepository: AuthRepository { // swiftlint:disable:this type_body_l
         try setMasterPasswordResult.get()
     }
 
+    func setLastActiveAccountTime() async throws {
```
andrebispo5: ⛏️ same alphabetical ordering.

BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
```diff
@@ -239,6 +239,11 @@ protocol AuthRepository: AnyObject {
         resetPasswordAutoEnroll: Bool,
     ) async throws
 
+    /// Records the current timestamp as the active account's last-active time.
+    /// Call this before navigating away from the active account (e.g., to add a new account).
```
andrebispo5: This last part of the doc is very specific and could be removed.

BitwardenShared/Core/Auth/Repositories/AuthRepository.swift
```diff
@@ -239,6 +239,11 @@ protocol AuthRepository: AnyObject {
...
+    ///
+    func setLastActiveAccountTime() async throws
```
andrebispo5: ⛏️ same alphabetical ordering.

andrebispo5 (changes requested):
Just small stuff to take a second look at. 👍

andrebispo5 (approved):
Awesome work! 🎉

PR #2938, opened 2026-08-06
[PM-41520] fix: Gracefully handle unknown Send types

matt-livefront:
## 🎟️ Tracking

[PM-41520](https://bitwarden.atlassian.net/browse/PM-41520)

## 📔 Objective

- Prevent a full Send sync failure when the server returns a Send `type` this app version doesn't recognize, by decoding unrecognized types to a new `.unknown` `SendType` case instead of throwing during JSON decoding (via `@DefaultValue`).
- `SendService.replaceSends(_:userId:)` now drops any Send that maps to `.unknown` during sync, logging a non-fatal error via `ErrorReporter` instead of persisting it with an incorrect type or failing the whole sync.
- `Send(sendResponseModel:)` and `AddEditSendItemState.newSendView()` now throw `DataMappingError.invalidData` for `.unknown` types rather than force-mapping them, since the SDK's `BitwardenSdk.SendType` has no equivalent case.
- Added exhaustive-switch handling for `.unknown` in `SendListItem`, `SendListItemRowView`, `AddEditSendItemState`, and `AddEditSendItemView` so the UI compiles and degrades gracefully (empty labels/icons fall back sensibly) rather than crashing, even though `.unknown` is never offered in the type picker.

[PM-41520]: https://bitwarden.atlassian.net/browse/PM-41520?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


andrebispo5:
Can we have screenshots on how the screen is going to handle the unknown types?

matt-livefront:
> Can we have screenshots on how the screen is going to handle the unknown types?

@andrebispo5  Unknown types are dropped before being saved to the database, so there's no change in the existing screens. You won't see the unknown types

andrebispo5:
> > Can we have screenshots on how the screen is going to handle the unknown types?
> 
> @andrebispo5 Unknown types are dropped before being saved to the database, so there's no change in the existing screens. You won't see the unknown types

@matt-livefront I asked because in the PR description it is meantioned that the UI adapts :
"Added exhaustive-switch handling for .unknown in SendListItem, SendListItemRowView, AddEditSendItemState, and AddEditSendItemView so the UI compiles and degrades gracefully (empty labels/icons fall back sensibly) rather than crashing, even though .unknown is never offered in the type picker."

BitwardenShared/Core/Tools/Services/SendServiceTests.swift
```diff
@@ -1,3 +1,4 @@
+import BitwardenKitMocks
```
KatherineInCode, line 1: 🤔 Would it be worth converting this to Swift Testing, do you think? It might not be, given the bulk of the file, but just thinking from a campsite-rule perspective

BitwardenShared/UI/Tools/Send/SendItem/AddEditSendItem/AddEditSendItemStateTests.swift
```diff
@@ -100,7 +100,7 @@ class AddEditSendItemStateTests: BitwardenTestCase { // swiftlint:disable:this t
         )
     }
 
-    func test_newSendView_text() {
+    func test_newSendView_text() throws {
```
KatherineInCode, line 103: 🤔 Similarly to previous comment, does it make sense to convert this to Swift Testing? Probably not in this case, since so little is changing, but worth a thought.

KatherineInCode (commented):
Looks good, just some thoughts on testing—if you'd rather not do that at this time for whatever reason (expediency or otherwise), that's fine. I don't have a strong opinion on Claude's finding.

BitwardenShared/Core/Tools/Services/SendService.swift
```diff
@@ -197,21 +203,33 @@ extension DefaultSendService {
...
+                return try Send(sendResponseModel: model)
+            } catch {
+                errorReporter.log(error: BitwardenError.generalError(
+                    type: "SendService: Dropped Send During Sync",
+                    message: "A send couldn't be converted and was excluded from the synced list of sends.",
+                    error: error,
+                ))
+                return nil
+            }
+        }
```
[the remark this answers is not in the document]
fedemkr: Yes, totally agree on this and not sure if even like that we're going to have potentially a lot of errors reported in Crashlytics. But well maybe we can think about it in a different PR not to block the release with this.

BitwardenShared/Core/Tools/Extensions/TestHelpers/BitwardenSdk+ToolsFixtures.swift
```diff
@@ -25,14 +25,17 @@ extension Send {
         text: SendTextModel? = nil,
         type: BitwardenShared.SendType = .text,
     ) -> Send {
-        self.init(
+        guard let sdkType = SendType(type: type) else {
+            preconditionFailure("The SDK's `SendType` has no case for `.unknown`; can't build a fixture with it.")
```
fedemkr, line 29: 🤔 I don't think we have another option without a lot of refactoring, but if this happens the whole test suite will crash, right?

matt-livefront:
> > > Can we have screenshots on how the screen is going to handle the unknown types?
> > 
> > 
> > @andrebispo5 Unknown types are dropped before being saved to the database, so there's no change in the existing screens. You won't see the unknown types
> 
> @matt-livefront I asked because in the PR description it is meantioned that the UI adapts : "Added exhaustive-switch handling for .unknown in SendListItem, SendListItemRowView, AddEditSendItemState, and AddEditSendItemView so the UI compiles and degrades gracefully (empty labels/icons fall back sensibly) rather than crashing, even though .unknown is never offered in the type picker."

@andrebispo5 Yeah, unfortunately if we add an unknown case, it still needs to be handled by the UI in the above cases, even if we never expect those values to reach the UI.

BitwardenShared/Core/Tools/Services/SendService.swift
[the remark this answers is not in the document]
matt-livefront: I added the aggregated error. Hopefully the work to add support for any new types will occur soon, but even so once per sync doesn't seem too egregious for something that we'll end up fixing.

matt-livefront:
> Looks good, just some thoughts on testing—if you'd rather not do that at this time for whatever reason (expediency or otherwise), that's fine. I don't have a strong opinion on Claude's finding.

@KatherineInCode  I think I'd like to wait in order to get this change in. I'm not a huge fan of converting a whole file if we're just changing/adding a single test as it makes it harder to see what actually changed. I'd rather do that separately, but maybe that's just my preference. I could do it as a follow-up though.

fedemkr (approved):
Looks good, just a small question but approved.

BitwardenShared/Core/Tools/Extensions/TestHelpers/BitwardenSdk+ToolsFixtures.swift
matt-livefront: Correct, but you would catch this during development at the time of creating the fixture.

KatherineInCode (approved):

PR #2940, opened 2026-08-06
[PM-37877] feat: Implement SDK KM state bridge

matt-livefront:
## 🎟️ Tracking

[PM-37877](https://bitwarden.atlassian.net/browse/PM-37877)

## 📔 Objective

Implements the SDK's `StateBridgeForeignImpl` interface on iOS and wires it to real mobile state, so the SDK can read/write the user key, PIN envelopes, encrypted PIN, V2 upgrade token, account cryptographic state, and master password unlock data without mobile passing it explicitly on every call.

- Adds `SdkStateBridge`, a per-user type conforming to `StateBridgeForeignImpl`, registered via `client.kmStateBridge().registerBridgeImpl(bridgeImpl:)` in `DefaultClientService.configureNewClient(_:for:)` for every user-scoped `Client` (covers the main app + AutoFill/Action/Share extensions + watchOS, which all share this one path).
- Adds a `kmStateBridge()` passthrough to `BitwardenSdkClient`/`Client`, matching the existing `auth()`/`crypto()`/`platform()` pattern.
- Adds a new `SdkStateBridgeStateService` protocol (mirroring `AutofillStateService`/`ServerCommunicationConfigStateService`) that narrows `SdkStateBridge`'s dependency down from the full `StateService` to just what it needs: `accountCryptographicState`, `v2UpgradeToken`, `encryptedPin`, `persistentPinEnvelope`/`ephemeralPinEnvelope` (the latter backed by the existing in-memory `accountVolatileData` slot), `setAccountMasterPasswordUnlock`/`getAccountMasterPasswordUnlock`, and `clearAccountMasterPasswordUnlockData`.
- Adds `v2UpgradeToken(userId:)`/`setV2UpgradeToken(_:userId:)` storage to `AppSettingsStore`.
- The raw `userKey` slot is in-memory only on the bridge instance and is never persisted (Keychain-only guardrail for unwrapped key material).

[PM-37877]: https://bitwarden.atlassian.net/browse/PM-37877?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenShared/Core/Platform/Services/StateService.swift
```diff
@@ -2729,3 +2727,78 @@ extension DefaultStateService: AutofillStateService {
         appSettingsStore.setLastRequestToTurnOnCredentialProvider(date)
     }
 }
+
+// MARK: SdkStateBridgeStateService
+
+extension DefaultStateService: SdkStateBridgeStateService {
```
quexten, line 2733: Disclaimer: I don't know the standards for swift. Question (non-blocking): Is there a reason we have another layer of indirection here rather than making the SdkStateBridge impl directly perform these?

BitwardenShared/Core/Platform/Services/StateService.swift
```diff
@@ -2729,3 +2727,78 @@ extension DefaultStateService: AutofillStateService {
...
+    func setEphemeralPinEnvelope(_ envelope: String?, userId: String) async {
+        accountVolatileData[userId, default: AccountVolatileData()].pinProtectedUserKey = envelope
+    }
+
+    // MARK: Master Password Unlock Data
+
+    func clearAccountMasterPasswordUnlockData(userId: String) async {
+        do {
+            try updateAccountProfile(userId: userId) { profile in
+                profile.userDecryptionOptions?.masterPasswordUnlock = nil
```
quexten, line 2765: Non-blocking: User decryption options should really be separate from synced unlock options. At least, they are on `clients`.

quexten (dismissed):

BitwardenShared/Core/Platform/Services/StateService.swift
matt-livefront: @quexten Just to confirm, are you suggesting that we store `masterPasswordUnlock` as it's own thing rather than nested within `profile.userDecryptionOptions`? I think we did it this way because the data is bundled together in the identity token response, but we can split it up if that'd be better.

BitwardenShared/Core/Platform/Services/StateService.swift
matt-livefront: This mirrors a pattern we already use throughout the app, anything that needs to read/write persisted state goes through a `*StateService` protocol implemented by `DefaultStateService` (e.g. `AutofillStateService`, `BillingStateService`, `UserSessionStateService`), rather than components reaching into `AppSettingsStore`/Keychain directly. `SdkStateBridgeStateService` extends that same convention to the new SDK bridge. It makes the bridge easier to test/mock and allows for reuse of any of the existing `StateService` methods.

BitwardenShared/Core/Platform/Services/ServiceContainer.swift
```diff
@@ -1164,6 +1165,7 @@ public class ServiceContainer: Services { // swiftlint:disable:this type_body_le
                 serverCommunicationConfigStateService: stateService,
                 stateService: stateService,
             ),
+            sdkStateBridgeStateService: stateService,
```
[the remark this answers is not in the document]
matt-livefront: Yes, this needs to be shared.

BitwardenShared/Core/Platform/Services/StateService.swift
quexten: Sorry took me a while to get back to this. Yes, that seems simpler, but don't feel it's blocking for this PR or anything.

quexten (dismissed):

quexten (approved):

fedemkr (approved):

PR #2952, opened 2026-08-10
[PM-41799] fix: Hide Premium upgrade path for self-hosted and internal QA accounts

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-41799

## 📔 Objective

Self-hosted accounts (and Bitwarden's own internal QA `*.bitwarden.pw` environments) could still see and use the in-app Premium upgrade flow when they shouldn't. There were two separate issues:

1. `BillingService.isSelfHosted()` only checked for `RegionType.selfHosted`, so it missed `RegionType.internal` (any `*.bitwarden.pw` host). That meant on our own internal QA environments the self-hosted restriction never kicked in, even with the debug override flag off.
2. The vault list's "Upgrade to Premium" banner wasn't checking self-hosted status at all before deciding whether to show itself, so self-hosted accounts could see and tap into it.

This fixes both: `isSelfHosted()` now also matches the internal QA region, and the vault list only shows the upgrade banner when the account isn't self-hosted. The `debug-disable-self-host-premium-check` flag still works as before for QA testing.

## 📸 Screenshots 
Before:

https://github.com/user-attachments/assets/6c330ced-aa04-4878-82f6-a28b59c8de02

After:

https://github.com/user-attachments/assets/850861fb-824f-4cd5-880e-c743e114f665


KatherineInCode (approved):

andrebispo5:
@KatherineInCode Thank you! 🎉

PR #2991, opened 2026-08-25
[PM-41957] fix: Show New folder created toast on all folder creation paths

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-41957

## 📔 Objective

Creating a folder only told you it worked in one of the three places you can do it.

- Settings > Folders showed the "New folder created." toast. The item editor's folder picker and the vault list's add-folder button showed nothing.
- The item editor selected the new folder but never set a toast.
- The vault list wasn't passing itself as the delegate, so nothing was listening.
- The vault list also wiped the toast on every refresh. It cleared it to dismiss the "This is taking longer than expected" message, and since the vault list watches folders, creating one triggered a refresh that killed the toast a moment after it appeared. Now only that one toast gets dismissed, which also fixes the item deleted/archived/restored toasts that were losing the same race.
- Fixed the swapped `folderDeleted`/`folderEdited` doc comments on the delegate protocol while I was in there.

## 📸 Screenshots

https://github.com/user-attachments/assets/a1e47f14-f5c1-4016-915d-5273b28378e5

### DIFF
--- a/BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessor.swift
+++ b/BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessor.swift
@@ -38,6 +38,14 @@ final class VaultListProcessor: StateProcessor<
         & HasTimeProvider
         & HasVaultRepository
 
+    // MARK: Static Properties
+
+    /// The toast shown while the vault is taking an unusually long time to load. A new instance is
+    /// returned on each access so that the view animates between successive toasts.
+    private static var slowLoadingToast: Toast {
+        Toast(title: Localizations.thisIsTakingLongerThanExpected, mode: .manualDismiss)
+    }
+
     // MARK: Private Properties
 
     /// The `Coordinator` that handles navigation.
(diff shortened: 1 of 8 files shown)


BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessor.swift
```diff
@@ -813,6 +832,22 @@ extension VaultListProcessor {
     }
 }
 
+// MARK: - AddEditFolderDelegate
+
+extension VaultListProcessor: AddEditFolderDelegate {
+    func folderAdded(_: FolderView) {
+        state.toast = Toast(title: Localizations.folderCreated)
```
matt-livefront, line 839: 🤔 One thing I noticed is that if this toast is displayed while the "This is taking longer than expected" toast is visible, the "New folder created" toast is never dismissed.

https://github.com/user-attachments/assets/335695e0-7149-4433-8326-a245189ea9ac

This might be an existing bug introduced by this: https://github.com/bitwarden/ios/pull/2827/changes#diff-68b12b1d47e8d01df3d910e70d76c1ba459cafe2c142501b36a2febad412be45. Any idea if we can fix this, revert that change, or switch the `task` to `onChange(of:)`?

andrebispo5: Good catch, thanks. The root cause was in `ToastView` rather than here:

- The `.task` running the dismiss timer had no `id`, and `.id(toast.id)` sits further in on the `VStack`, so swapping `state.toast` for a different toast never restarted it.
- The task that did run started while the manual dismiss toast was up, hit the `guard mode == .automaticDismiss` and returned, so the "New folder created" toast ended up with no timer at all.
- Changed it to `.task(id: toast.id)` in 87458dd5e. That also fixes a second case where one automatic toast replacing another inherited the first's leftover time and could disappear almost immediately.

Verified in the simulator with both toasts in sequence.

andrebispo5: https://github.com/user-attachments/assets/f99d561a-a400-4e10-ad8b-80f92c5a7903

matt-livefront (approved):

PR #2992, opened 2026-08-25
[PM-42462] fix: Show None as the vault item picker placeholder

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-42462

## 📔 Objective

- The card brand picker showed `--Select--` as its placeholder. Per the design discussion on this PR, the placeholder is now `None`.
- Card expiration month, identity title, and bank account type used the same `--Select--` format, so I fixed all four. Fixing only the brand would leave two placeholder styles on the same screen.
- Uses the existing `None` localization, so no new string was added.
- Tidied up the doc comments on those four so they match. A couple said "default state title for title type" even on the card brand.
- Added a test per type so the dashes don't come back.

## 📸 Screenshots

| Before | After |
|--------|--------|
|  |  |

### DIFF
--- a/BitwardenShared/Core/Vault/Models/Enum/BankAccountTypeTests.swift
+++ b/BitwardenShared/Core/Vault/Models/Enum/BankAccountTypeTests.swift
@@ -1,3 +1,4 @@
+import BitwardenResources
 import Foundation
 import Testing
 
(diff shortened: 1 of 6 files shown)


fedemkr (commented):
Changes look good, although I left a comment on the Figma design as I'm not sure if leaving plain "Select" here is the best option as it could be a bit confusing IMO.

andrebispo5:
> Changes look good, although I left a comment on the Figma design as I'm not sure if leaving plain "Select" here is the best option as it could be a bit confusing IMO.

Totally agree on this one, having the Select option there makes it a bit confusing.

fedemkr (approved):

PR #2993, opened 2026-08-26
[PM-42025] fix: Show a toast after a vault item is saved

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-42025

## 📔 Objective

- Saving a license showed no "License saved" toast. Turns out no item type showed one; QA just happened to catch it on the license.
- Nothing was wired up for it. `itemAdded()` also fires when you cancel out of the editor, so the toast couldn't just hang off the existing delegate methods.
- Added an `itemSaved(type:)` delegate call that only fires after the item actually saves, plus a per-type title so each type gets its own wording.
- The vault list, group list, and item view screens all show it now.
- Found a second problem while testing: the item view rebuilds its whole state whenever the cipher changes, so saving wiped its own toast. That's fixed too, which also steadies the existing "Item saved" toast on that screen.

## 📸 Screenshots

# Before 

https://github.com/user-attachments/assets/5fbcbe6d-2367-4748-80b8-dd4af8d1909e

# After

https://github.com/user-attachments/assets/8629dbbc-31d1-407d-b3c0-d6cda9fdd33a

### DIFF
--- a/BitwardenShared/Core/Vault/Models/Enum/CipherType.swift
+++ b/BitwardenShared/Core/Vault/Models/Enum/CipherType.swift
@@ -111,4 +111,18 @@ extension CipherType {
             [.text, .hidden, .boolean]
         }
     }
+
+    /// The title of the toast shown after an item of this type is saved.
+    var savedToastTitle: String {
+        switch self {
+        case .bankAccount: Localizations.bankAccountSaved
+        case .card: Localizations.cardSaved
+        case .driversLicense: Localizations.licenseSaved
+        case .identity: Localizations.identitySaved
+        case .login: Localizations.loginSaved
+        case .passport: Localizations.passportSaved
+        case .secureNote: Localizations.secureNoteSaved
+        case .sshKey: Localizations.sshKeySaved
+        }
+    }
 }
--- a/BitwardenShared/UI/Vault/Helpers/VaultItemMoreOptionsHelper.swift
+++ b/BitwardenShared/UI/Vault/Helpers/VaultItemMoreOptionsHelper.swift
@@ -52,6 +52,10 @@ class DefaultVaultItemMoreOptionsHelper: VaultItemMoreOptionsHelper {
     /// The services used by this helper.
     private var services: Services
 
+    /// The closure used to display a toast, stored when navigating to the edit item screen so
+    /// that a confirmation toast can be displayed once the item has been saved.
+    private var handleDisplayToastAfterEdit: ((Toast) -> Void)?
+
     // MARK: Initialization
 
     /// Initialize a `VaultItemMoreOptionsHelper`.
@@ -227,6 +231,7 @@ class DefaultVaultItemMoreOptionsHelper: VaultItemMoreOptionsHelper {
             }
         case let .edit(cipherView):
             await masterPasswordRepromptHelper.repromptForMasterPasswordIfNeeded(cipherView: cipherView) {
+                self.handleDisplayToastAfterEdit = handleDisplayToast
                 self.coordinator.navigate(to: .editItem(cipherView), context: self)
             }
         case let .launch(url):
@@ -270,5 +275,20 @@ class DefaultVaultItemMoreOptionsHelper: VaultItemMoreOptionsHelper {
             await coordinator.showErrorAlert(error: error)
             services.errorReporter.log(error: error)
         }
+    }
+}
+
+// MARK: - CipherItemOperationDelegate
+
+extension DefaultVaultItemMoreOptionsHelper: CipherItemOperationDelegate {
+    func itemDismissed() -> Bool {
+        handleDisplayToastAfterEdit = nil
+        return true
+    }
+
+    func itemUpdated(type: CipherType) -> Bool {
+        handleDisplayToastAfterEdit?(Toast(title: type.savedToastTitle))
+        handleDisplayToastAfterEdit = nil
+        return true
     }
 }
--- a/BitwardenShared/UI/Vault/Helpers/VaultItemMoreOptionsHelper.swift
+++ b/BitwardenShared/UI/Vault/Helpers/VaultItemMoreOptionsHelper.swift
@@ -13,12 +13,15 @@ protocol VaultItemMoreOptionsHelper {
     ///
     /// - Parameters
     ///   - item: The selected item to show the options for.
+    ///   - delegate: The delegate notified of operations performed on the item from within the
+    ///     add/edit item view, used when the edit option is selected.
     ///   - handleDisplayToast: A closure called to handle displaying a toast.
     ///   - handleNavigateToPremiumUpgrade: A closure called to navigate to the Premium upgrade flow.
     ///   - handleOpenURL: A closure called to open a URL.
     ///
     func showMoreOptionsAlert(
         for item: VaultListItem,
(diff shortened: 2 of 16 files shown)


BitwardenShared/UI/Vault/VaultItem/AddEditItem/AddEditItemProcessor.swift
```diff
@@ -26,6 +26,12 @@ protocol CipherItemOperationDelegate: AnyObject {
     /// Called when the cipher item has been successfully restored.
     func itemRestored()
 
+    /// Called when the cipher item has been successfully saved, whether it was added or updated.
+    ///
+    /// - Parameter type: The type of the cipher item that was saved.
+    ///
+    func itemSaved(type: CipherType)
```
matt-livefront: 🤔 This delegate already has both an `itemAdded()` and `itemUpdated()` method. Would it be better to wire your changes into those existing methods?

BitwardenShared/Core/Vault/Models/Enum/CipherType.swift
```diff
@@ -111,4 +111,18 @@ extension CipherType {
             [.text, .hidden, .boolean]
         }
     }
+
+    /// The title of the toast shown after an item of this type is saved.
+    var savedToastTitle: String {
+        switch self {
+        case .bankAccount: Localizations.bankAccountSaved
```
matt-livefront, line 118: 🤔 All of our other toasts are more generic, like "Item deleted", "Item moved to archive", "Item restored". Should this be similar for consistency? Or are we intentionally deviating from that here?

andrebispo5: This is a new requirement: 
https://bitwarden.atlassian.net/browse/PM-42025
https://bitwarden.atlassian.net/browse/PM-41935

BitwardenShared/UI/Vault/VaultItem/AddEditItem/AddEditItemProcessor.swift
andrebispo5: itemAdded() also fires when you cancel out of the editor, so the toast couldn't just hang off the existing delegate methods.

matt-livefront: That seems unintuitive and likely a bug. We shouldn't be calling `itemAdded()` if you didn't actually add an item.

It looks like the main reason we use `itemAdded()` currently is for when you scan an OTP QR code. That shows the `VaultItemSelectionView` and then you can navigate to `AddEditItemView`. After adding an item, the goal was to dismiss both sheets and return you back to the vault. However, that seems broken due to other changes in the coordinator. These navigation flows are complicated and brittle 😢 .

I'd ultimately be in favor of using `itemAdded()` and `itemUpdated()` here rather than adding another `itemSaved()` method, which feels redundant, and also fixing it so `itemAdded()` is only called when you actually add the item. I can add a ticket to fix the `VaultItemSelection` dismissal, but I don't think we'd be any worse off than today by making it so `itemAdded()` is only called when the item is actually added.

andrebispo5: My main concern was the brittle aspect of this navigation and that it looked like a house of cards. Given the very narrow scope of this ticket, I was hesitant to add structural changes to it. But fixing the `itemAdded()` seems crucial.

BitwardenShared/UI/Vault/Helpers/VaultItemMoreOptionsHelper.swift
```diff
@@ -272,3 +277,18 @@ class DefaultVaultItemMoreOptionsHelper: VaultItemMoreOptionsHelper {
...
+
+extension DefaultVaultItemMoreOptionsHelper: CipherItemOperationDelegate {
+    func itemDismissed() -> Bool {
+        handleDisplayToastAfterEdit = nil
+        return true
+    }
+
+    func itemUpdated(type: CipherType) -> Bool {
+        handleDisplayToastAfterEdit?(Toast(title: type.savedToastTitle))
+        handleDisplayToastAfterEdit = nil
```
matt-livefront: 🤔 Instead of storing `handleDisplayToastAfterEdit` and using that for displaying the toast, would it work to use the coordinator method (`coordinator.showToast(type.savedToastTitle)`) to display the toast?

BitwardenShared/UI/Vault/Helpers/VaultItemMoreOptionsHelper.swift
```diff
@@ -227,6 +231,7 @@ class DefaultVaultItemMoreOptionsHelper: VaultItemMoreOptionsHelper {
             }
         case let .edit(cipherView):
             await masterPasswordRepromptHelper.repromptForMasterPasswordIfNeeded(cipherView: cipherView) {
+                self.handleDisplayToastAfterEdit = handleDisplayToast
                 self.coordinator.navigate(to: .editItem(cipherView), context: self)
```
matt-livefront: 🤔 Instead of passing `self` as the context/delegate, would it make sense to add `CipherItemOperationDelegate` as a parameter to this function so the calling processor could be used to display the toast directly without needing to persist the "display toast" closure?

BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessor.swift
```diff
@@ -859,6 +859,10 @@ extension VaultListProcessor: CipherItemOperationDelegate {
         state.toast = Toast(title: Localizations.itemDeleted)
     }
 
+    func itemSaved(type: CipherType) {
```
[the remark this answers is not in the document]
andrebispo5: Good catch, that gap was real. Fixed:

- `.edit` now passes a delegate through to the add/edit screen
- `VaultListProcessor` and `VaultGroupProcessor` show the toast via `itemUpdated(type:)`

All three paths behave the same now.

BitwardenShared/UI/Vault/VaultItem/AddEditItem/AddEditItemProcessor.swift
andrebispo5: Agreed, went with your suggestion:

- Dropped `itemSaved(type:)`
- `itemAdded(type:)` / `itemUpdated(type:)` now carry the type
- Added `itemDismissed()` so cancelling no longer fires `itemAdded()`

BitwardenShared/UI/Vault/Helpers/VaultItemMoreOptionsHelper.swift
andrebispo5: Makes sense, done in e341a1604:

- `showMoreOptionsAlert` takes the delegate now, `.edit` passes it as the context
- Dropped the stored toast closure and the helper's `CipherItemOperationDelegate` conformance
- `VaultListProcessor` / `VaultGroupProcessor` were missing `itemUpdated(type:)`, added

One exception: `VaultItemSelectionProcessor` can't pass `self`, since its conformance dismisses on save for the OTP key flow. It passes a small `CipherSavedToastDelegate` instead, so a plain edit from the menu still just shows the toast.

matt-livefront (approved):

PR #2995, opened 2026-08-26
[PM-31117] feat: Add SDK-managed PIN unlock

matt-livefront:
## 🎟️ Tracking

[PM-31117](https://bitwarden.atlassian.net/browse/PM-31117)

## 📔 Objective

Adds support for using the SDK for PIN management.

- Adds the `pm-31059-sdk-pin-unlock` feature flag and exposes the SDK's `userCryptoManagement().pinSettings()` client for SDK-managed PIN state.
- With the flag ON, PIN unlock and validation route through the SDK-managed path (`InitUserCryptoMethod.pinState`, `PinSettingsClient.validatePin`) instead of the legacy envelope/key parameters the app previously read and passed in itself.
- PIN enrollment (`setPins`) and removal now unconditionally dual-write to both the legacy iOS-managed state and the new SDK-managed state, so the flag can be toggled in either direction without leaving a stale or missing PIN unlock method.

[PM-31117]: https://bitwarden.atlassian.net/browse/PM-31117?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


quexten (dismissed):
Sweet! After we roll out and remove the feature flag, that should drop a fair bit of logic from mobile's codebase :)

PR #2999, opened 2026-08-27
llm: Exclude claude-code-action's restored config paths from the SDK-update push guard

SaintPatrck:
## 🎟️ Tracking

No tracking ticket — a correctness fix to the SDK-update-evaluation workflow's push guard, found while investigating a failed run.

## 📔 Objective

The push guard treated any working-tree diff as an uncommitted Claude edit and failed the job, but **claude-code-action** restores `.claude`, `.mcp.json`, `CLAUDE.md`, and a few other sensitive paths from `origin/main` before running Claude, since the PR head is untrusted — and the long-lived `sdlc/sdk-update` branch had gone stale relative to `main` for those paths. Both guards now exclude the exact restore list `claude-code-action` pulls from base, since this workflow never has a legitimate reason to touch those paths itself.

### DIFF
--- a/.github/workflows/sdlc-sdk-update-evaluate.yml
+++ b/.github/workflows/sdlc-sdk-update-evaluate.yml
@@ -278,7 +278,7 @@ jobs:
           # The second alternation is the same restore list claude-code-action pulls from
           # origin/main above — this job never has a legitimate reason to commit changes there either.
           CHANGED=$(git diff --name-only "origin/$_BRANCH_NAME" HEAD)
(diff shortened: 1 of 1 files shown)


fedemkr (approved):
Looks good, should we also exclude the `LICENSE.txt` file as well?

SaintPatrck:
> Looks good, should we also exclude the `LICENSE.txt` file as well?

`LICENSE.txt` is not touched by `claude-code-action`, so it should not be an issue. If it surfaces, we can make an exclusion rule for it.

PR #3000, opened 2026-08-27
[PM-39767] feat: Persist Premium upgrade pending state and resolve it on any sync

KatherineInCode:
## 🎟️ Tracking

[PM-39767](https://bitwarden.atlassian.net/browse/PM-39767)

## 📔 Objective

Replaces #2916. Persists Premium upgrade pending/failure state per account, and resolves it on any sync — not just the one that started the checkout. Fixes QA finding #3 ("Delayed Sync").

Checkout-success and the background sync watcher now share `resolvePendingUpgrade(userId:syncFailed:)`, scoped to an explicit `userId`. No lock needed (unlike #2916): an account switch mid-flight can't corrupt another account's state, and double-processing the same sync is harmless.

Also fixes a gap left by #2873's CTA fix: none of the other eight upgrade entry points checked for an already-pending upgrade before starting a second one. `PremiumUpgradeHelper.startInAppPremiumUpgrade()` — the shared choke point all nine go through — now checks once for everyone.

**Trade-offs:**
- `lastAttemptFailed` can briefly be wrong after a specific compound sync failure. Left alone — nothing reads it yet.
- A dropped-webhook upgrade that never resolves has no escape hatch to the web fallback, since every entry point now defers to the pending alert. Revisit with the "Sync Unsuccessful" PR.

Celebration sheet and "Sync Unsuccessful" alert are later PRs in this stack; the latter's design is still open.

[PM-39767]: https://bitwarden.atlassian.net/browse/PM-39767?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenShared/Core/Billing/Services/BillingService.swift
```diff
@@ -228,13 +270,66 @@ class DefaultBillingService: BillingService {
...
+
+    func premiumUpgradePendingStatePublisher() -> AnyPublisher<PremiumUpgradePendingState, Never> {
+        // `premiumUpgradePendingStateSubject.send(_:)` is called from whichever background
+        // `Task` (`start()`'s account/sync subscribers) happens to be resolving it — never
+        // guaranteed to be the main thread. Pinned here so a consumer driving `@Published` UI
+        // state from this publisher doesn't need to hop to main itself.
+        premiumUpgradePendingStateSubject
+            .receive(on: DispatchQueue.main)
+            .eraseToAnyPublisher()
+    }
```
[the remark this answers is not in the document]
KatherineInCode: This will be consumed in later PRs on the stack. I'll look into the lack of test coverage, though.

BitwardenShared/Core/Billing/Services/BillingService+BillingStateTests.swift
```diff
@@ -0,0 +1,559 @@
...
+/// Resolves a `nil` `userId` to `stateService`'s active account.
+func resolvedUserId(_ userId: String?, stateService: MockStateService) throws -> String {
+    if let userId {
+        return userId
+    }
+    guard let activeAccount = stateService.activeAccount else {
+        throw StateServiceError.noActiveAccount
+    }
+    return activeAccount.profile.userId
+}
```
fedemkr: 🤔 Why does this need to be a global function? Can't the caller just use `func getAccountIdOrActiveId(userId: String?) async throws -> String` which already exists in `MockStateService`?

BitwardenShared/Core/Billing/Services/BillingService.swift
```diff
@@ -84,15 +103,24 @@ protocol BillingService: AnyObject { // sourcery: AutoMockable
     /// - Returns: Whether the action card should be shown.
     ///
     func shouldShowUpgradedToPremiumActionCard() async -> Bool
+
+    /// Starts observing sync completions to resolve any pending Premium upgrade. Should be
+    /// called once, for the lifetime of the app.
+    ///
+    func start() async
```
fedemkr: 🤔 I was wondering whether this is too generic for a function name. I mean "Start" for a Billing service can mean a lot of things and even though we have the DocC, perhaps it could be more straightforward to call it `startObservingSyncCompletions` or something more descriptive of what it's starting.
What do you think?

BitwardenShared/Core/Billing/Models/Domain/PremiumUpgradePendingState.swift
```diff
@@ -0,0 +1,13 @@
...
+///
+struct PremiumUpgradePendingState: Equatable, Sendable {
+    // MARK: Properties
+
+    /// Whether a Premium upgrade is currently pending.
+    let isPending: Bool
+
+    /// Whether the last sync attempt for the pending Premium upgrade failed.
+    let lastAttemptFailed: Bool
+}
```
fedemkr, line 19: 🤔 I feel the naming could be a bit contradictory as the struct is named `PremiumUpgradePendingState` and internally it could be `isPending = false` so the "pending" struct is not actually "pending". What do you think about renaming it to just `PremiumUpgradeState` or something like that?

BitwardenShared/Core/Billing/Services/BillingService.swift
```diff
@@ -228,13 +269,66 @@ class DefaultBillingService: BillingService {
...
+    }
+
+    func premiumUpgradePendingStatePublisher() -> AnyPublisher<PremiumUpgradePendingState, Never> {
+        // `premiumUpgradePendingStateSubject.send(_:)` is called from whichever background
+        // `Task` (`start()`'s account/sync subscribers) happens to be resolving it — never
+        // guaranteed to be the main thread. Pinned here so a consumer driving `@Published` UI
+        // state from this publisher doesn't need to hop to main itself.
+        premiumUpgradePendingStateSubject
+            .receive(on: DispatchQueue.main)
+            .eraseToAnyPublisher()
```
fedemkr, line 301: 🤔 I'm not sure if the service would need to enforce dispatching to main queue. What's the benefit over the caller deciding which queue it should be dispatched to? In fact, `premiumCheckoutStatusPublisher` is currently deferring that to the caller, check `PremiumUpgradeHelper` for example.
However if you think this has some other benefits I'm not seeing, we should update the other ones to follow the same thing as well if the behavior is consistent.

BitwardenShared/Core/Billing/Services/BillingService.swift
```diff
@@ -292,4 +386,144 @@ class DefaultBillingService: BillingService {
             return false
         }
     }
+
+    func start() async {
+        guard !started else { return }
+        started = true
+
+        activeAccountSubscriber = Task {
```
fedemkr, line 394: ⚠️  `activeAccountSubscriber`'s `Task` captures `self` strongly and is never cancelled, creating a retain cycle (`self` → `activeAccountSubscriber` → closure → `self`).
Same happens with `syncCompletionSubscriber`.
Given that `DefaultBillingService` is a Singleton then it won't matter for production but it will on tests. Every unit test that constructs its own `DefaultBillingService` leaks that instance (and its mocks) for the rest of the test run so memory won't be freed.

BitwardenShared/Core/Billing/Services/BillingService.swift
```diff
@@ -292,4 +386,144 @@ class DefaultBillingService: BillingService {
...
+    ///
+    private func isEligibleForPremiumUpgradePath() async -> Bool {
+        await refreshSubscriptionAttentionCard(subscription: nil)
+        guard await !isSelfHosted(),
+              await configService.getFeatureFlag(.premiumUpgradePath)
+        else {
+            return false
+        }
+        return true
+    }
```
fedemkr, line 430: 🎨 I see this guard check being done in several places in this service. 
Could we move `await refreshSubscriptionAttentionCard(subscription: nil)` to `reconcileCheckoutSuccess` as it's the only place where it's being used and use this function without it so it can use the guard check in all places accordingly?

BitwardenShared/Core/Billing/Services/BillingService.swift
```diff
@@ -292,4 +386,144 @@ class DefaultBillingService: BillingService {
...
+
+    /// Subscribes to the active account's sync completions and resolves a pending Premium
+    /// upgrade after each genuinely new sync.
+    ///
+    /// - Parameter userId: The user ID of the account this subscriber was started for.
+    ///
+    private func reconcileOnEachNewSync(userId: String) async {
+        let publisher: AnyPublisher<Date?, Never>
```
fedemkr, line 438: 🎨 Use a more descriptive name like `lastSyncTimePublisher` as it's easier to follow in code.

BitwardenShared/Core/Billing/Services/BillingService.swift
```diff
@@ -292,4 +386,144 @@ class DefaultBillingService: BillingService {
...
+        }
+
+        // Snapshot the last-known sync time directly, rather than relying on whichever value
+        // the publisher happens to deliver first: its `CurrentValueSubject` backing replays the
+        // existing cached value immediately on subscribe (not evidence that a sync just
+        // happened), and a *different* account's sync can also re-emit this shared store
+        // without this account's own value having changed.
+        var lastSeenDate: Date?
+        do {
+            lastSeenDate = try await stateService.getLastSyncTime()
```
fedemkr, line 453: 🤔 Why is this named `lastSeenDate` instead of `lastSyncTime`?

BitwardenShared/Core/Billing/Services/BillingService.swift
```diff
@@ -292,4 +386,144 @@ class DefaultBillingService: BillingService {
...
+        } catch {
+            errorReporter.log(error: error)
+        }
+
+        for await date in publisher.values {
+            guard date != lastSeenDate else { continue }
+            lastSeenDate = date
+            _ = await resolvePendingUpgrade(userId: userId, syncFailed: false)
+            await refreshPremiumUpgradePendingStateSubject()
+        }
```
fedemkr, line 463: 🎨 I think you can wrap the whole code in a `do...catch` logging the error which would be cleaner than catching each time on-site and having to declare and set variables separately.

fedemkr (commented):
> lastAttemptFailed can briefly be wrong after a specific compound sync failure. Left alone — nothing reads it yet.

🤔 I think we should at least have a note somewhere like in the model to state that the value may be wrong on compound sync failure. Therefore, if at some point it starts to be read the dev can know about that behavior and fix/improve if necessary. I believe that is talking about the model because in `resolvePendingUpgrade` it's reading `billingStateService.getPremiumUpgradeLastSyncAttemptFailed`.

_-- Reviewed partially, will continue reviewing later --_

BitwardenShared/Core/Billing/Services/BillingService+BillingStateTests.swift
KatherineInCode: Because I completely forgot the other one existed 😂 good catch

BitwardenShared/Core/Billing/Services/BillingService.swift
KatherineInCode: I'd patterned it after the one in [AuthenticatorSyncService](https://github.com/bitwarden/ios/blob/main/BitwardenShared/Core/Platform/Services/AuthenticatorSyncService.swift#L15), but upon reflection, I agree—that one is more clear from the class name, and this isn't. Might be worth circling back on the `AuthenticatorSyncService` one eventually to make it slightly more descriptive. I'll update this one for now, though.

BitwardenShared/Core/Billing/Models/Domain/PremiumUpgradePendingState.swift
KatherineInCode: Agreed, the naming around this could use some work, though I think it's probably important to keep "pending" in the name of the struct, since it's an outgrowth of that "is there a Premium upgrade pending" boolean extended to keep track of failed sync attempts. I'll poke at it to try to come up with a better name 🤔

BitwardenShared/Core/Billing/Services/BillingService.swift
KatherineInCode: Good call.

PR #3002, opened 2026-08-27
[PM-42029] fix: Match the add item menu order to the designs

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-42029

## 📔 Objective

- The add item menu listed the types in the wrong order: Identity and Secure note came before Bank account, License and Passport.
- Order now matches the designs: Login, Card, Bank account, Identity, License, Passport, Secure note, then Folder at the bottom.
- The list gets reversed on its way to the menu, since iOS draws it bottom-up when it opens above the button. So `canCreateCases` reads the same way you see it on screen.
- Also reversed the two state defaults and the view's default parameter. They were feeding the menu an unreversed list, so tapping the button before the vault list finished loading showed everything upside-down.

## 📸 Screenshots 

| Before | After |
|--------|--------|
|  |  |

### DIFF
--- a/BitwardenShared/Core/Vault/Repositories/TestHelpers/MockVaultRepository.swift
+++ b/BitwardenShared/Core/Vault/Repositories/TestHelpers/MockVaultRepository.swift
@@ -75,7 +75,7 @@ class MockVaultRepository: VaultRepository { // swiftlint:disable:this type_body
 
     var getTOTPKeyIfAllowedToCopyResult: Result<String?, Error> = .success(nil)
 
-    var getItemTypesUserCanCreateResult: [BitwardenShared.CipherType] = CipherType.canCreateCases
+    var getItemTypesUserCanCreateResult: [BitwardenShared.CipherType] = CipherType.canCreateCases.reversed()
 
     var hasMinimumCipherCountResult: Result<Bool, Error> = .success(false)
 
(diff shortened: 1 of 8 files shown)


BitwardenShared/Core/Vault/Repositories/TestHelpers/MockVaultRepository.swift
```diff
@@ -75,7 +75,7 @@ class MockVaultRepository: VaultRepository { // swiftlint:disable:this type_body
 
     var getTOTPKeyIfAllowedToCopyResult: Result<String?, Error> = .success(nil)
 
-    var getItemTypesUserCanCreateResult: [BitwardenShared.CipherType] = CipherType.canCreateCases
```
KatherineInCode, line 78: 🤔 Are all these `reversed()` calls even necessary? It seems like the issue was purely in the order in `CipherType`?

andrebispo5: Fair question. The `CipherType` reorder is what fixes the reported bug, but the defaults still need the reversal:

- `appeared()` awaits `refreshVault()` before `loadItemTypesUserCanCreate()`, and the FAB overlay sits outside the `LoadingView`, so the button is tappable for the whole sync.
- During that window the menu renders off the default. Unreversed, it comes out upside-down (Secure note at the top, Login at the bottom).
- The repository hands back `canCreateCases.reversed()`, so the defaults have to match or they disagree with what replaces them.

PR #3012, opened 2026-09-02
[PM-42022] fix: Drivers License accessibility fixes

morganzellers-bw:
## 🎟️ Tracking

- [PM-42022](https://bitwarden.atlassian.net/browse/PM-42022) - View screen show/hide toggle has a generic button announcement
- [PM-42020](https://bitwarden.atlassian.net/browse/PM-42020) - Add & Edit Screen: show/hide toggle name announcement is not custom by line item
- [PM-42027](https://bitwarden.atlassian.net/browse/PM-42027) - Entered characters are not announced individually
- [PM-42021](https://bitwarden.atlassian.net/browse/PM-42021) - MP Re-prompt information icon is not announced and can't be accessed

## 📔 Objective

Adds a batch of VoiceOver accessibility fixes under the Drivers License item type, mirroring the same batch of fixes already implemented for Bank Account in #2985:

- View/Add-Edit screen toggles now announce which field they control instead of generic "Password"/"Toggle Visibility" wording
- The license number spells out its characters individually to VoiceOver when visible, instead of being read as a whole number/word
- The Master Password Re-prompt info icon is independently announced and reachable instead of being swallowed by the toggle row

The overflow ("more options") menu fix (PM-42026) is stacked on top in #3013, since it grew into a larger shared-component rework and made sense to track as its own PR.

## 📸 Screenshots

[PM-42022]: https://bitwarden.atlassian.net/browse/PM-42022?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PM-42020]: https://bitwarden.atlassian.net/browse/PM-42020?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PM-42027]: https://bitwarden.atlassian.net/browse/PM-42027?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[...]

PR #2985, opened 2026-08-20
[pm-41938] fix: Bank Account accessibility fixes

morganzellers-bw:
## 🎟️ Tracking

- [PM-41938](https://bitwarden.atlassian.net/browse/PM-41938) - View Bank Account screen show/hide toggle has a generic button announcement
- [PM-41939](https://bitwarden.atlassian.net/browse/PM-41939) - Add & Edit Screen: show/hide toggle name announcement is not custom by line item
- [PM-41940](https://bitwarden.atlassian.net/browse/PM-41940) - Entered characters are not announced individually
- [PM-41941](https://bitwarden.atlassian.net/browse/PM-41941) - Overflow menu is not announced and can't be accessed
- [PM-41942](https://bitwarden.atlassian.net/browse/PM-41942) - MP Re-prompt information icon is not announced and can't be accessed

## 📔 Objective

Adds a batch of accessibility bug fixes under the Bank Account item type

## 📸 Screenshots

<details>
<summary>Click to reveal</summary>

#### PM-41938

https://github.com/user-attachments/assets/fd5170a8-f23d-4b8d-a5cb-ba74c758c197

#### PM-41939

https://github.com/user-attachments/assets/fb024d79-b186-4696-96f2-a4181d4de1f0

#### PM-41940

https://github.com/user-attachments/assets/5e3e6440-d6e2-47f0-aa67-c282caac8d04

#### PM-41941

https://github.com/user-attachments/assets/16feae53-a94d-46ef-ad74-ae6e062720a9

#### PM-41942

https://github.com/user-attachments/assets/7578fac0-4883-4b0b-bedc-788e8d66c360

</details>

[PM-41938]: https://bitwarden.atlassian.net/browse/PM-41938?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PM-41939]: https://bitwarden.atlassian.net/browse/PM-41939?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[...]

PR #3013, opened 2026-09-02
[PM-42026] fix:  Overflow menu is not announced and can't be accessed via VoiceOver

morganzellers-bw:
## 🎟️ Tracking

- [PM-42026](https://bitwarden.atlassian.net/browse/PM-42026) - Overflow menu is not announced and can't be accessed

## 📔 Objective

Stacked on top of #3012. Fixes the vault list row's overflow ("more options") menu being unreachable to VoiceOver, mirroring the identical fix shipped for Bank Account in #2985 (commit `7e137f3eb`).

The button was nested inside another Button at every call site (the row itself is the label of an outer Button used to open the item). SwiftUI always flattens a Button's whole label subtree into a single VoiceOver element, so the nested button's own accessibility label and activation were swallowed by the outer button — a simpler fix exposing "More options" as a custom rotor action would only make it reachable via the actions rotor, not as its own swipeable, independently selectable stop.

Instead, moves "open item" tap handling inside `VaultListItemRowView` itself, behind a new `.pressed` effect, so the primary open action and the secondary options/copy buttons become siblings instead of nested buttons — mirroring the pattern already used by `SendListItemRowView`. All 4 call sites (`VaultListView`, `VaultGroupView`, `VaultItemSelectionView`, `VaultAutofillListView`) drop their outer `Button` wrapper and forward `.pressed` through the `mapEffect` closure they already use for `.morePressed`. `itemPressed` moves from a `VaultListAction`/`VaultGroupAction` to a `VaultListEffect`/`VaultGroupEffect` to route through the same mechanism. Because a real `Button`/`AsyncButton` is preserved (just relocated), there's no loss of native press-highlight feedback.

Since this reworks the shared `VaultListItemRowView`, it benefits every cipher type, not just Drivers License.

Local review (`code-review-local`) flagged two non-blocking items, both also present unchanged in the identical Bank Account implementation (#2985) — left as-is here to stay consistent with that PR rather than diverge:
- `VaultAutofillListView`'s `mapEffect` now completes an autofill request on `.morePressed` instead of being a no-op; currently unreachable since `CipherOptionsButton` is always hidden when `isFromExtension: true`.
- `VaultListItemRowView`'s row-level `accessibilityIdentifier` moved from the outer combined element to a `.contain` container; the tap target is now identified via `"VaultListItemRowButton"` instead.

## 📸 Screenshots

N/A — behavior matches the already-reviewed Bank Account implementation in #2985; VoiceOver announcements should be confirmed on-device.

[PM-42026]: https://bitwarden.atlassian.net/browse/PM-42026?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenKit/Core/Platform/Extensions/String.swift
+++ b/BitwardenKit/Core/Platform/Extensions/String.swift
@@ -115,6 +115,16 @@ public extension String {
         return result
     }
 
+    /// Returns a copy of the string with each character separated by a space so that VoiceOver
+    /// announces the characters individually instead of interpreting the string as a whole word
+    /// or number.
+    ///
+    /// - Returns: A copy of the string with each character separated by a space.
+    ///
+    func spellingOutCharacters() -> String {
+        map { String($0) }.joined(separator: " ")
+    }
+
     /// Validates whether the string is a valid email address.
     ///
     /// - Parameter useStrictValidation: If `true`, validates against a regex pattern requiring
--- a/BitwardenShared/Core/Vault/Extensions/CipherListView+Extensions.swift
+++ b/BitwardenShared/Core/Vault/Extensions/CipherListView+Extensions.swift
@@ -1,5 +1,6 @@
 import BitwardenResources
 import BitwardenSdk
+import Foundation
 
 extension CipherListView {
     // MARK: Properties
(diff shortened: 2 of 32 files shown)


matt-livefront:
> Follow-up to the overflow-menu fix in #3012 (stacked on top of that branch). A prior fix exposed the vault list row's "more options" button as a custom rotor action because it was nested inside the outer button used to open the item, and SwiftUI flattens a Button's whole label subtree into a single VoiceOver element — that only makes the button reachable via the actions rotor, not as its own swipeable, independently selectable stop.

Is this the desired behavior? When navigating the list via VoiceOver, this now doubles the numbers of swipes needed to navigate through the list. You also lose context for which item the "more options" button belongs to, if you aren't able to visually see which row the button belongs to (is it the item you previously swiped from or the next item?). I thought that exposing these actions as accessibility actions was a better VoiceOver experience. I think it could be improved by exposing each available action as an accessibility action, rather than just "more options", but I think because it isn't a static list that's probably a bit more complicated than what we have today.

morganzellers-bw:
> > Follow-up to the overflow-menu fix in #3012 (stacked on top of that branch). A prior fix exposed the vault list row's "more options" button as a custom rotor action because it was nested inside the outer button used to open the item, and SwiftUI flattens a Button's whole label subtree into a single VoiceOver element — that only makes the button reachable via the actions rotor, not as its own swipeable, independently selectable stop.
> 
> Is this the desired behavior? When navigating the list via VoiceOver, this now doubles the numbers of swipes needed to navigate through the list. You also lose context for which item the "more options" button belongs to, if you aren't able to visually see which row the button belongs to (is it the item you previously swiped from or the next item?). I thought that exposing these actions as accessibility actions was a better VoiceOver experience. I think it could be improved by exposing each available action as an accessibility action, rather than just "more options", but I think because it isn't a static list that's probably a bit more complicated than what we have today.

Those are great points, I agree this may be a less user-friendly experience. I'll take a look at improving it with individual accessibility actions and see what it looks/feels like.

morganzellers-bw:
Superseded by #3025, which carries forward this PR's more complete overflow-menu VoiceOver fix (including the double-swipe regression fix) as a shared, cipher-type-agnostic PR in the combined stack. Closing in favor of that.

PR #3015, opened 2026-09-02
[PM-37884] fix: Show the scan card button when editing a card

andrebispo5:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-37884

## 📔 Objective

The "Scan card" button never showed up when editing an existing card, only when adding a new one.

- The edit screen listens for cipher updates, and every update rebuilt the card state straight from the cipher
- That rebuild threw away `cardScannerEnabled`, which the feature flag had just set, so the button disappeared before anyone could see it
- Adding a card was fine because that flow has no cipher to stream yet
- The card state now carries the scanner values over when it's rebuilt, so the button sticks around (and an incoming sync no longer closes the scanner sheet mid-scan)
- Added tests on both `CipherItemState` and `AddEditItemProcessor` that fail without the fix

## 📸 Screenshots 

| Before | After |
|--------|--------|
|  |  |

### DIFF
--- a/BitwardenShared/UI/Vault/VaultItem/AddEditItem/AddEditCardItem/CardItemState.swift
+++ b/BitwardenShared/UI/Vault/VaultItem/AddEditItem/AddEditCardItem/CardItemState.swift
@@ -56,6 +56,21 @@ extension CardItemState {
             number: cardNumber.nilIfEmpty,
         )
     }
+
+    /// Returns a copy of the state with the card scanner's state copied over from another state.
+    ///
+    /// The card scanner's state is driven by the feature flag and the UI rather than by the cipher,
+    /// so it needs to be carried over whenever the state is rebuilt from an updated cipher.
+    ///
+    /// - Parameter other: The state to copy the card scanner's state from.
+    /// - Returns: A copy of the state with the card scanner's state applied.
+    ///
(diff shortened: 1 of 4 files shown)


matt-livefront (approved):

speedy1804:
@andrebispo5 @matt-livefront 
Hi, sorry to bother you, but I have a question:
The initial pull requests for the "camera scan" feature were merged quite a while ago, yet the feature still isn't available in the Bitwarden app (the button is missing). Is this intentional—meaning the feature flag simply hasn't been disabled yet—or is it a bug?
I’d really appreciate an answer.
Regards, Speedy 😃

matt-livefront:
> @andrebispo5 @matt-livefront Hi, sorry to bother you, but I have a question: The initial pull requests for the "camera scan" feature were merged quite a while ago, yet the feature still isn't available in the Bitwarden app (the button is missing). Is this intentional—meaning the feature flag simply hasn't been disabled yet—or is it a bug? I’d really appreciate an answer. Regards, Speedy 😃

@speedy1804 Yes, this feature is still behind a feature flag which is currently off. I'm not sure if we're waiting on anything else other than this fix to enable it, but I will check.

speedy1804:
@matt-livefront Thank you so much for your reply and the clarity!

PR #3022, opened 2026-09-08
[PM-43016] fix: Load item types before vault refresh and reload on sync

fedemkr:
## 🎟️ Tracking

https://bitwarden.atlassian.net/browse/PM-43016

## 📔 Objective

The item types a user can create (used to build the "add item" button/menu on the vault list and vault group screens) were loaded only once, after the rest of the `.appeared` flow had already run, and defaulted to all cases until that load completed. That caused two problems:

- The menu could briefly show item types the user isn't actually allowed to create (or omit ones they are) while the vault was syncing, since the load happened after the sync-triggered refresh rather than before it.
- Feature-flag or policy changes picked up by a later sync were never reflected without leaving and reopening the screen, since the load only ran once at `.appeared`.

This PR:

- Loads item types before refreshing the vault on `.appeared`, and changes the default to `[]` so the menu doesn't render stale/incorrect types while unloaded.
- Reloads item types whenever a sync completes, via a new `syncCompletePublisher()` subscription in both `VaultListProcessor` and `VaultGroupProcessor`.
- Guards against overlapping loads racing each other with a generation token, so a slower, older call can't overwrite a newer result.

## 📸 Screenshots

Not applicable — no visual/UI change, only timing/correctness of an existing menu's contents.


BitwardenShared/UI/Vault/Vault/VaultGroup/VaultGroupProcessor.swift
```diff
@@ -60,6 +65,9 @@ final class VaultGroupProcessor: StateProcessor<// swiftlint:disable:this type_b
     /// An object to manage TOTP code expirations and batch refresh calls for search results.
     private var searchTotpExpirationManager: TOTPExpirationManager?
 
+    /// A task to handle the sync-complete stream.
+    private var syncCompleteStreamTask: Task<Void, Never>?
```
matt-livefront: 🎨 Would it make sense to turn this into an `Effect` that the view performs when it appears? That would allow the view to own the task and handle cancellation for you. And it's more consistent with how we do this elsewhere as well.

BitwardenShared/UI/Vault/Vault/VaultGroup/VaultGroupProcessorTests.swift
```diff
@@ -253,6 +257,54 @@ class VaultGroupProcessorTests: BitwardenTestCase { // swiftlint:disable:this ty
...
+        waitFor(subject.state.itemTypesUserCanCreate == [.card])
+
+        // The older, first call resolving afterwards must not overwrite the newer result.
+        vaultRepository.getItemTypesUserCanCreateContinuations[0].resume(returning: [.login])
+
+        // Give the stale result a chance to (wrongly) apply, then confirm it didn't.
+        let deadline = Date(timeIntervalSinceNow: 0.25)
+        while Date() < deadline {
+            RunLoop.current.run(mode: .default, before: Date(timeIntervalSinceNow: 0.05))
+        }
```
matt-livefront: 🤔 Is calling the continuations synchronous? Is this extra delay necessary?

BitwardenShared/UI/Vault/Vault/VaultGroup/VaultGroupProcessor.swift
fedemkr: Nice catch, changed 🚀

BitwardenShared/UI/Vault/Vault/VaultList/VaultListProcessorTests.swift
```diff
@@ -501,6 +501,78 @@ class VaultListProcessorTests: BitwardenTestCase { // swiftlint:disable:this typ
...
+    }
+
+    /// The processor reloads the item types the user can create whenever a sync completes, so
+    /// feature-flag or policy changes picked up by the sync are reflected without needing the
+    /// screen to reappear.
+    @MainActor
+    func test_streamSyncComplete_reloadsItemTypesUserCanCreate() {
+        vaultRepository.getItemTypesUserCanCreateResult = [.card]
+
+        syncService.syncCompleteSubject.send(())
```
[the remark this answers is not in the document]
fedemkr: Fixed

matt-livefront (approved):

PR #3024, opened 2026-09-08
[pm-32809] fix: Add shared VoiceOver a11y infra for toggles and spelled-out values

morganzellers-bw:
## 🎟️ Tracking

- [PM-41942](https://bitwarden.atlassian.net/browse/PM-41942) — MP Re-prompt information icon is not announced and can't be accessed
- [PM-42021](https://bitwarden.atlassian.net/browse/PM-42021) — MP Re-prompt information icon is not announced and can't be accessed (Drivers License)

## 📔 Objective

Consolidates the shared VoiceOver accessibility work for Bank Account and Drivers License item types.

Adds the shared mechanisms:

- `PasswordVisibilityButton` now applies its `accessibilityLabel`
- `BitwardenTextField` gains an optional `passwordVisibilityFieldName` parameter so the show/hide toggle can announce which field it affects (e.g. "PIN is visible, tap to hide.") instead of generic "Password" wording. Defaults to `nil`, preserving today's behavior for every existing field.
- `PasswordText` gains an opt-in `spellOutAccessibilityValue` flag (default `false`) that spells out characters individually to VoiceOver via `speechSpellsOutCharacters`, with an overridden `accessibilityLabel` to avoid a double announcement.
- `BitwardenTextValueField` gains the same opt-in `spellOutAccessibilityValue` flag for non-password-style value fields.
- Fixes the Master Password Re-prompt info icon being swallowed by its parent toggle's combined VoiceOver element (same root cause independently fixed in both #2985 and #3012 as PM-41942/PM-42021) by moving it into `BitwardenToggle`'s accessory slot.
- Also fixes the same defect for the Appearance screen's "Show website icons" info button, found and fixed alongside the Bank Account work.

**Note on scope:** an intermediate commit on #2985 had `PasswordText.spellOutAccessibilityValue` defaulting to `true`, which would have silently turned on character-by-character spelling for every existing password-style field app-wide (login passwords, card numbers, SSH keys, passport numbers, etc.), not just the intended Bank Account/Drivers License fields. This PR restores the opt-in default of `false` to match the original design intent.

> We probably want to make a ticket to handle doing this elsewhere ^^

## 📸 Screenshots

https://github.com/user-attachments/assets/c01c7fae-54e7-42a9-ae90-c4a3050c7e33

[...]

PR #3025, opened 2026-09-08
[PM-42026] fix: Expose the vault list overflow menu to VoiceOver

morganzellers-bw:
## 🎟️ Tracking

- [PM-41941](https://bitwarden.atlassian.net/browse/PM-41941) — Overflow menu is not announced and can't be accessed (Bank Account)
- [PM-42026](https://bitwarden.atlassian.net/browse/PM-42026) — Overflow menu is not announced and can't be accessed (Drivers License)

## 📔 Objective

The row's "more options" button (and, for TOTP rows, the "copy code" button) was nested inside another `Button` at every call site, so SwiftUI flattened the whole row into a single VoiceOver element and swallowed the nested button's own label and activation.

This PR:
- Moves "open item" tap handling inside `VaultListItemRowView` itself behind a new `.pressed` effect, so the primary open action and secondary buttons become siblings instead of nested `Button`s, mirroring the pattern already used by `SendListItemRowView`.
- Exposes each applicable overflow menu action (View, Edit, per-type Copy actions, Launch, Archive/Unarchive) as an individually named VoiceOver accessibility action, rather than a single generic "More options" stop — this is the fix for the double-swipe regression found during review of the first approach.
- Updates all 4 call sites (`VaultListView`, `VaultGroupView`, `VaultItemSelectionView`, `VaultAutofillListView`) and their processors accordingly.

Base: #3024 (shared VoiceOver accessibility infrastructure).

#### PM-41941

https://github.com/user-attachments/assets/7dfa30d3-a9c0-4c8d-9464-90f588248b3d

[PM-41941]: https://bitwarden.atlassian.net/browse/PM-41941?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PM-42026]: https://bitwarden.atlassian.net/browse/PM-42026?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenShared/Core/Vault/Extensions/CipherListView+Extensions.swift
```diff
@@ -31,8 +32,75 @@ extension CipherListView {
...
+        var kinds: [MoreOptionsActionKind] = [.view]
+        if deletedDate == nil { kinds.append(.edit) }
+
+        switch type {
+        case .card:
+            if copyableFields.contains(.cardNumber) { kinds.append(.copyCardNumber) }
+            if copyableFields.contains(.cardSecurityCode) { kinds.append(.copySecurityCode) }
+        case let .login(loginListView):
+            if copyableFields.contains(.loginUsername) { kinds.append(.copyUsername) }
+            if copyableFields.contains(.loginPassword) { kinds.append(.copyPassword) }
```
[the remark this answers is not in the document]
morganzellers-bw: Fixed in [380c03b](https://github.com/bitwarden/ios/pull/3025/commits/380c03b09bdd2106b80c45a5ff3105ebe1a99dbe)

BitwardenShared/UI/Vault/Extensions/Alert+Vault.swift
```diff
@@ -667,6 +667,237 @@ extension Alert {
...
+            return .archive(cipherView: cipherView)
+        case .copyAccountNumber:
+            guard let accountNumber = cipherView.bankAccount?.accountNumber else { return nil }
+            return .copy(
+                toast: Localizations.accountNumber,
+                value: accountNumber,
+                requiresMasterPasswordReprompt: true,
+                logEvent: nil,
+                cipherId: nil,
+            )
```
matt-livefront, line 785: 🎨 Building these actions and deciding which options are available is very similar to the `Alert.moreOptions()` method. Is it possible to update the alert method to use this new functionality? Then this logic would exist in one place rather than duplicating it.

BitwardenShared/UI/Vault/Extensions/Alert+Vault.swift
```diff
@@ -667,6 +667,237 @@ extension Alert {
...
+/// The kind of an individual more-options action for a vault list item, independent of the
+/// concrete cipher values needed to actually perform it.
+///
+/// Used both to decide which options are available (from a `CipherListView`, synchronously, for
+/// VoiceOver custom accessibility actions on the vault list row) and to build the more-options
+/// action sheet's `AlertAction`s (from a full `CipherView`, in `Alert.moreOptions`). If you
+/// add/remove a case here, also update `CipherListView.applicableMoreOptionsActionKinds` and
+/// `Alert.moreOptions` so the two stay in sync.
+///
+enum MoreOptionsActionKind: CaseIterable, Equatable {
```
matt-livefront, line 681: ⛏️ This should probably get it's own file rather than being lumped in here. And I would consider if it makes sense to move the `applicableMoreOptionsActionKinds` extension method into this new file as well so all of the logic for which additional options are available for a cipher exist in a single place.

BitwardenShared/Core/Vault/Extensions/CipherListView+Extensions.swift
```diff
@@ -31,8 +32,75 @@ extension CipherListView {
...
+
     // MARK: Methods
 
+    /// Computes, synchronously and without decrypting anything beyond what's already in this
+    /// list view, the set of more-options actions applicable to this cipher. Used to populate the
+    /// vault list row's VoiceOver custom accessibility actions without a `fetchCipher` round trip.
+    ///
+    /// Scope note: intentionally matches `Alert.moreOptions`'s scope exactly (e.g. no Identity
+    /// fields, no bank IBAN/SWIFT/PIN/branch number), even though `copyableFields` exposes more
+    /// than the sheet currently surfaces.
```
matt-livefront, line 53: ❓ Are we missing functionality from `Alert.moreOptions`? Should we be exposing an action for anything in the `copyableFields` array?

PR #3031, opened 2026-09-09
[PM-43234] fix: Hide Edit option for non-compliant Sends in the send list menu

matt-livefront:
## 🎟️ Tracking

[PM-43234](https://bitwarden.atlassian.net/browse/PM-43234)

## 📔 Objective

- The Send List's "..." item-options menu only suppressed Edit when Sends were disabled org-wide (`SendPolicyOptions.isSendDisabled`), but never checked the per-item `sendView.disabled` flag that marks an individual Send as non-compliant — the same flag already used to hide the Edit FAB on the View Send screen.
- Closes the remaining gap: a non-compliant Send could still be edited via the row's "..." menu even though the FAB path was already blocked.
- Gated the Edit button in `SendListItemRowView.optionsMenu(for:)` on `!sendView.disabled`. No other menu items (View/Copy/Share/RemovePassword/Delete) or the global `isSendDisabled` behavior changed.

## 📸 Screenshots

| Before | After |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/fc595936-9122-45b6-85d0-a22506a7896a" /> | <video src="https://github.com/user-attachments/assets/6a56cb48-9b87-49b7-9176-783f18619bd9" /> | 

[PM-43234]: https://bitwarden.atlassian.net/browse/PM-43234?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/BitwardenShared/UI/Tools/Send/Send/SendList/SendListItemRowView+ViewInspectorTests.swift
+++ b/BitwardenShared/UI/Tools/Send/Send/SendList/SendListItemRowView+ViewInspectorTests.swift
@@ -1,0 +1,81 @@
+// swiftlint:disable:this file_name
+import BitwardenKit
+import BitwardenKitMocks
+import BitwardenResources
+@preconcurrency import BitwardenSdk
+import ViewInspector
+import XCTest
+
+@testable import BitwardenShared
+
+// MARK: - SendListItemRowViewTests
+
+class SendListItemRowViewTests: BitwardenTestCase {
+    // MARK: Properties
+
+    var processor: MockProcessor<SendListItemRowState, SendListItemRowAction, SendListItemRowEffect>!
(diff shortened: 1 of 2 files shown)


fedemkr (approved):
Looks good, although is it correct that the only action hidden is "Edit" for non-compliant particular sends? Feels odd that something non-compliant can be shared so raising it just  in case.

matt-livefront:
> Looks good, although is it correct that the only action hidden is "Edit" for non-compliant particular sends? Feels odd that something non-compliant can be shared so raising it just in case.

Good callout! Sounds like we'll handle this in a separate ticket/PR.

PR #3044, opened 2026-09-15
[PM-27189] fix: Use master password unlock salt instead of email across the app

matt-livefront:
## 🎟️ Tracking

[PM-27189](https://bitwarden.atlassian.net/browse/PM-27189)

## 📔 Objective

- Sources the salt and KDF params used for master-password hashing/derivation (`AuthService.hashPassword`, `AuthRepository.updateMasterPassword`) from `MasterPasswordUnlockData` instead of the account's email/KDF settings. This is necessary to handle the case where the salt could be changed to something other than email after the SDK starts generating the salt.
- Removes the now-redundant post-login hash-caching step (`AuthService.saveMasterPasswordHash`) — every login flow already recomputes and caches the same hash via `unlockVaultWithPassword`.
- `ChangeKdfService.updateKdfToMinimums` updated the account's KDF settings but never refreshed the cached `MasterPasswordUnlockData` to match, so it now does both together.

[PM-27189]: https://bitwarden.atlassian.net/browse/PM-27189?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BitwardenShared/Core/Auth/Services/AuthService.swift
```diff
@@ -659,9 +662,6 @@ class DefaultAuthService: AuthService { // swiftlint:disable:this type_body_leng
             masterPassword: masterPassword,
         )
 
-        // Save the master password hash.
-        try await saveMasterPasswordHash(password: masterPassword)
```
matt-livefront, line 663: ℹ️ Hashing the master password here (and for two factor login) is redundant since it's also hashed and saved when the vault is unlocked after login.

https://github.com/bitwarden/ios/blob/f92c62c43eb193a738c2b28318adff63578169e6/BitwardenShared/Core/Auth/Repositories/AuthRepository.swift#L1137-L1141

